PR #11, opened 2026-04-21
fix(codex): single-quote binary path in per-project config.toml writer

suzuke:
## Summary

Same escape-interpretation bug as PR #5 (the `[projects."..."]` trust-entry writer, commit 09a66c8) — different writer, different file, second place to miss it.

`configure_codex` writes `<project>/.codex/config.toml` with:

```toml
[mcp_servers.agend-terminal]
command = "{binary_path}"
...
AGEND_HOME = "{home}"
```

On Windows `binary_path()` is `C:\...\agend-terminal.exe` and `home()` is `C:\Users\...\.agend` — both contain `\U`, which a TOML basic string interprets as a unicode escape. Codex then refuses to load its own config with `too few unicode value digits` and falls back to `C:\WINDOWS` as cwd.

Fix: extract `toml_string_value(s)` helper that emits a single-quoted literal (no escape interpretation) when the value doesn't contain a single quote, and falls back to an escaped basic string otherwise. Apply to both `command =` and `AGEND_HOME =`.

## Test plan

- [x] `toml_string_value_uses_literal_for_windows_paths`
- [x] `toml_string_value_falls_back_to_escaped_double_quoted_for_apostrophes`
- [x] `configure_codex_writes_parseable_toml_for_windows_style_binary_path` — round-trips the written file through `toml::from_str` and asserts the `command` field equals `binary_path()` verbatim.
- [x] All 19 pre-existing codex tests pass.
- [ ] Manual: restart daemon with new binary, confirm `~/.agend/workspace/codex/.codex/config.toml` parses cleanly and codex launches without "UNC paths are not supported" → `C:\WINDOWS` fallback.

## Not in scope

- `configure_codex` still uses a literal `contains("[mcp_servers.agend-terminal]")` check to skip rewriting — which means an already-broken per-project config survives the upgrade and must be deleted by hand. A future change could parse-then-regenerate on `toml::from_str` failure; deferred to keep this PR focused.

### DIFF
--- a/src/mcp_config.rs
+++ b/src/mcp_config.rs
@@ -64,6 +64,24 @@ fn mcp_server_entry() -> serde_json::Value {
     })
 }
 
+/// Render a string value as a TOML string literal, picking the form that
+/// survives escape parsing.
+///
+/// Single-quoted literal strings (`'foo\bar'`) don't interpret escapes, which
+/// is what we want for Windows paths — a double-quoted `"C:\Users\..."` makes
+/// codex reject its own config.toml with "too few unicode value digits" when
+/// the path happens to contain `\U`, `\n`, `\t`, etc. Fall back to the
+/// backslash-escaped double-quoted form only if the value itself contains a
+/// single quote, which a single-line literal cannot represent.
+fn toml_string_value(s: &str) -> String {
+    if s.contains('\'') {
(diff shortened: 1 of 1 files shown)


suzuke:
Superseded by #12 which removes codex_trust_directory entirely. The per-project TOML single-quoting fix is folded into that refactor.

PR #19, opened 2026-04-22
ci(release): Linux x86_64 AppImage with tray

suzuke:
## Summary
- New `appimage` job in `release.yml` that bundles GTK3 + appindicator runtime deps into a self-contained `.AppImage` so Linux users can run the tray without `cargo install --features tray`.
- Job uses `linuxdeploy` + `linuxdeploy-plugin-gtk` (continuous release), `APPIMAGE_EXTRACT_AND_RUN=1` to bypass missing FUSE on GitHub runners, and `imagemagick` to resize the 1024×1024 master icon to the 256×256 required by the AppImage spec + FreeDesktop hicolor standard.
- `Exec=agend-terminal tray` in the `.desktop` — AppImage convention is that double-click launches resident mode.
- `needs: [build, appimage]` so the release job waits for both; the AppImage is added to the release files list and `SHA256SUMS`.

## Why x86_64 only
aarch64 Linux is a rare desktop target; deferring until user signal justifies the extra matrix cell.

## Why not make the `linux-gnu` tarball ship `--features tray`
tray-icon dyn-links `libgtk-3.so` + `libayatana-appindicator3.so` at runtime. A raw tarball on a headless box would fail to load. AppImage solves the exact same problem by bundling those libs; tarball users still get the portable CLI without tray, and `cargo install --features tray` remains the path for Linux folks who want tray from source.

## Test plan
- [ ] Tag a throwaway prerelease (e.g. `v0.3.2-rc1`) and verify `agend-terminal-x86_64.AppImage` appears in the release with a matching `SHA256SUMS` line.
- [ ] Download the AppImage on a GTK3-capable desktop, `chmod +x`, run — tray icon appears.
- [ ] Confirm the `.desktop` `Exec=agend-terminal tray` invokes the resident mode when the AppImage is double-clicked or integrated via `appimaged`.

Closes out item #6 from `docs/PLAN-tray-resident.md`.


suzuke:
Pushed `9196aac` to address the review:

**Done this round (commit 9196aac):**
- **Supply-chain pin**: linuxdeploy `1-alpha-20251107-1` (`c20cd71e…`) + plugin-gtk commit `3b67a1d` (`b0f4cbc6…`), both gated by `sha256sum -c -`. No more `continuous` / `master` URLs.
- **workflow_dispatch**: manual pre-tag validation. Release job now `if: startsWith(github.ref, 'refs/tags/')` so dispatched runs upload artifacts for inspection but don't touch the release page.
- **.desktop Categories**: dropped `TerminalEmulator`, left `Utility;Development;`.
- **desktop-file-validate**: explicit gate before AppDir staging.

**Deferred to follow-ups:**
- **`Exec=agend-terminal tray` vs AppRun default**: reviewer is right that linuxdeploy-plugin-gtk's AppRun may not honor the `.desktop` `Exec=` on direct double-click — worth confirming empirically (first dispatched run will tell us) and documenting in README if AppRun bypasses Exec.
- **`convert` → `magick`**: `magick` isn't in ubuntu-latest's default ImageMagick 6 package name; will switch when runners flip to IM7.
- **Artifact naming (`-x86_64` vs `-<triple>`)**: AppImage convention is `-x86_64`; leaving as-is for discoverability.

I'd suggest merging this round and then cutting `v0.3.2-rc1` (or running the workflow via dispatch first) to empirically verify the double-click-Exec question — that's the one open unknown.

Issue #29, opened 2026-04-22
Task #9 Option C post-merge polish (umbrella)


suzuke:
## Scope

Umbrella tracking 3 non-blocking follow-ups identified during PR #20–#24 review / merge. Root of changes is Task #9 Option C (ops/mcp 雙軌收斂) + Task #12 (start_instance relocation).

## Linked issues

- #26 — Stale ops:: doc references (tests/mcp_characterization.rs + src/telegram.rs) — suggested assignee: at-dev-2
- #27 — Add test coverage for `from_command("kiro")` alias branch — suggested assignee: at-dev-2
- #28 — Future-proof dual-track drift detector against raw strings / extern fns — suggested assignee: at-dev-3

## Rationale for umbrella

- Issues 粒度細，各自可獨立 PR + review
- Umbrella 提供 reviewer 整體 context（三者都是 Task #9 Option C 的 cleanup 尾聲）
- 關閉條件：3 個子 issue 全 close

## 歷史 PR references

- PR #18 (Task #9 Option C Commit 1+2 + caller swap)
- PR #20 (`validate_branch` 3rd copy consolidation)
- PR #21 (21 dead CLI-wrapper fn + `#![allow(dead_code)]` removal)
- PR #22 (Task #12 — `start_instance` inline + delete `src/ops.rs`)
- PR #23 (dual-track drift CI lint)
- PR #24 (TeamCreated positive-pin via RecordingNotifier)

## LESSONS reference

`recovery/LESSONS-04-21.md` 有相關 conventions adopted 條目：
- Dual-track code structural rule
- Migration SOP — 4-point review checklist
- Post-migration residue hard-resolve heuristic
- Patch-id auto-drop for shared commits across branches

## Non-goal

本 umbrella 不擴大 scope。任何超出上述 3 子 issue 的 cleanup（例如 module re-organization、新 feature）應開獨立 issue 不併進來。

suzuke:
All 3 children resolved:

- #26 closed by PR #30 (@ 02:03:26)
- #27 closed by PR #30 (@ 02:03:26)
- #28 closed by PR #31 (@ 02:16:29)

Main @ post-umbrella sanity: 625 passed / 0 failed (was 622, +3 from PR #31 new tests).

Task #9 Option C post-merge polish complete.

PR #23, opened 2026-04-22
feat(ci): add dual-track fn drift detector

suzuke:
## Summary

Task #9 Option C follow-up #4 — CI lint 擋 future ops/mcp 雙軌 fn drift。

## Why

LESSONS-04-21 記載 `cleanup_working_dir` 在 `ops.rs` vs `mcp/handlers.rs` 靜默 drift 8 天（2026-04-14 → 04-22），原因是兩邊各有一份 function 且沒有機制保證同步。Option C 主 PR 把 7 個 utility 收斂到 `agent_ops` 解決當前情況，但沒防止未來重現（若有人 re-introduce dual-track fn，一樣會 silently drift）。

## 變更

- `tests/no_dual_track_drift.rs`（新 306 LOC）
  - `extract_top_level_fns()`：brace-balanced body extraction
  - `compare_fns_from_sources()`：純 helper 方便 unit test
  - Test 1: `no_dual_track_fn_drift_between_ops_and_mcp_handlers` — 同名 fn body 若不同則 fail
  - Test 2: `handles_missing_ops_rs_gracefully` — `ops.rs` 被刪（Task #12 之後）此 test 仍 pass

## Robustness

- `.unwrap_or_default()` 處理 missing file（Task #12 會刪 `src/ops.rs`，此 lint 不應 block 那個 PR）
- Test 寫為「source-agnostic」—helper 吃 file content string 而非硬 coded path，easier unit test

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test --release` — 625 passed / 0 failed / 2 ignored
- [x] 手動驗：artificially inject 2 同名 fn with different body → test fail（pre-push check）
- [x] 手動驗：刪除 `src/ops.rs` → test pass（graceful handle 生效）

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #30, opened 2026-04-22
chore: Task #9 Option C post-merge polish (closes #26, closes #27)

suzuke:
## Summary

Resolves the two non-blocking items identified during PR #22 review,
consolidated as the first batch of umbrella issue #29 (Task #9 Option C
post-merge polish). Pure cleanup — no behavioral change.

- **closes #26** — update stale `ops::` doc references to current
  canonical paths. Three call sites referred to `ops::delete_instance`
  which was removed by PR #21 (`2db67fc`) + PR #22 (`8a7240d`):
  - `tests/mcp_characterization.rs:373` test fn renamed from
    `delete_instance_succeeds_via_ops_when_daemon_down` →
    `delete_instance_succeeds_via_mcp_handler_when_daemon_down`
  - Same test's inline comment + assertion message now describe the MCP
    handler's disk-fallback path (`fleet.yaml` removal +
    `cleanup_working_dir`) instead of the removed function
  - `src/telegram.rs:353` doc comment on `cleanup_deleted_topic` now
    cites the MCP `delete_instance` handler's disk mutation as the
    source of truth

- **closes #27** — restore `Backend::from_command("kiro")` alias test
  coverage. The `backend_resolves_to_preset_command` test dropped by
  PR #22 was only partially duplicated by `from_command_detection`:
  the `claude` roundtrip was covered, but the **bare `"kiro"` alias**
  (without the `-cli` suffix — handled by `src/backend.rs:314`'s
  `basename == "kiro"` branch) lost its pin. A single assertion folded
  into the existing `from_command_detection` matrix restores the
  roundtrip `"kiro" → Backend::KiroCli → preset().command == "kiro-cli"`.

## Scope

- 3 files touched, +16 / −4 lines
- No production logic modified (doc-comment text only in `telegram.rs`
  and a test-assert addition in `backend.rs`)
- Test count unchanged: 622 passed (rename preserves count, added
  assertion is inside an existing test fn)

## Test plan

- [x] `cargo fmt --check` — clean
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo test --release` — 622 passed / 0 failed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #31, opened 2026-04-22
test(ci): harden dual-track drift detector — raw strings + extern ABI (closes #28)

### LINKED ISSUE #28: Future-proof dual-track drift detector against raw strings and extern fns
> Part of umbrella: #29

## Background

PR #23 (`tests/no_dual_track_drift.rs`) 建立 ops/mcp 雙軌 fn drift 偵測。at-dev-1 在 PR #23 review 提出 2 類 future-proofing gap（均非當前 blocker，0 current instances）：

### Gap 1: Raw string literals in top-level fn bodies

目前 `skip_string` helper 只認 `"..."`（含 `\"` escape）+ `'...'`，**不認 `r#"..."#`** 等 raw strings。若未來有人在 top-level fn body 放 raw string 含 unbalanced `{` 或 `}`，brace counter 會 mis-track fn boundary。

### Gap 2: `extern "C" fn` prefix

`extract_top_level_fns` 只認 `pub fn` / `fn` 開頭，不認 `extern "C" fn` / `pub extern fn`。目前 codebase 無此 pattern，但 FFI 擴充時可能 hit。

## Proposed fix（2 選 1）

**Option A**: 擴充 `skip_string` 認 `r#"..."#`
- Pros: direct fix，其他 Rust parser 也是這條路
- Cons: 要處理 `r##`、`r###` 等任意 `#` 數量

**Option B**: helper 自我 assert「頂層 fn body 無 raw string」+ fail loudly if violated
- Pros: 簡單，明確 invariant
- Cons: 若未來真要 raw string 仍需解（deferred cost）

## Acceptance

- 選一個方向實作
- 加 test case inject raw string fn → 確保 detector 行為正確（fail-loudly 或 correctly-parse）

Suggested assignee: at-dev-3（原作者，`skip_string` / brace-balance 最熟）。at-dev-1 flexible backup。

suzuke:
## Summary

Follow-up to PR #23 (Task #9 Option C epilogue lint). Closes **#28** (dual-track lint future-proofing — raw strings + extern fn). Part of umbrella **#29** (Task #9 Option C post-merge polish).

Addresses two parser gaps identified in at-dev-1's PR #23 review. Both were flagged as non-blockers with no current offenders; this lands the robustness cleanly before they become active drift.

### 1. Raw string literals (`r"..."` / `r#"..."#`) in top-level fn bodies

`skip_string` uses bare `"` as delimiter, incompatible with raw-string termination (`"#`). A raw string containing `{` or `}` could silently corrupt brace counting — dropping the fn from the drift map (silent miss) or extracting a wrong body (silent drift false-result).

**Fix — Option B (fail-loud, not full raw-string support):**
- New `body_contains_raw_string_token` scans each extracted body with the same string/char/comment skippers as `match_balanced_brace`, panics on raw-string prefix at a token boundary.
- Scoped to *extracted* top-level fn bodies (Refinement 1), not whole-file — so raw strings inside `mod tests {...}` / impl blocks / etc. that the column-0 parser never enters do **not** false-fail. `src/mcp/handlers.rs` L787/L833 (both inside `mod tests`) cleanly skipped.
- Panic message points at the fix path: either move the raw string out, or upgrade `skip_string` and remove this guard.

### 2. `extern "C" fn` / `extern "Rust" fn` / any quoted-ABI variant

Existing qualifier stripper only matched bare `extern ` — any `extern "ABI" fn` at column 0 was silently dropped. Particularly dangerous since FFI exports are where drift between two files causes the most damage.

**Fix:** new `strip_extern_maybe_abi` replaces the bare-`extern` prefix strip. Consumes `extern ` plus the optional quoted ABI clause.

## Defense-in-depth: panic on silent drop

If `find_body_brace` or `match_balanced_brace` returns None *after* `detect_fn_name` already matched, the detector now panics with a parser-limitation hint. Covers the pathological case where raw-string miscount corrupts brace counting so badly that no closing `}` is found at all (guard #1 wouldn't fire because the extracted body range may be truncated). Prior behavior: silently advance line-by-line.

## Tests (3 new)

| Test | Purpose |
|------|---------|
| `raw_string_in_top_level_fn_fails_loudly` | `#[should_panic(expected = "raw string literal")]` — pins the fail-loud contract. |
| `raw_string_in_indented_block_is_ignored` | Scope pin: raw strings inside `mod tests {...}` do NOT trigger. Prevents regression on `src/mcp/handlers.rs` L787/L833. |
| `extern_abi_fn_detected` | `extern "C" fn` + `extern "Rust" fn` both register; divergent bodies detected correctly. |

## Scope discipline

- Raw-string support itself is **not** implemented (Option A deferred). Guard docstring points future authors at the upgrade path if a top-level fn ever needs a raw string.
- Refinement 1 (body-scope scan) picked over whole-file scan specifically to avoid false-fail on `src/mcp/handlers.rs` tests.
- Follows Task #9 Option C's "closed loop, don't over-engineer" basis.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test --release` — **625 passed / 0 failed** across 10 test binaries
- [x] `tests/no_dual_track_drift.rs`: **5/5 pass** (2 existing + 3 new), including `#[should_panic]` verification

### DIFF
--- a/tests/no_dual_track_drift.rs
+++ b/tests/no_dual_track_drift.rs
@@ -16,12 +16,38 @@
 //! Scope is intentionally narrow — only these two files. Macro-generated
 //! fn and non-top-level definitions (inside `mod tests`, `impl`, etc.) are
 //! deliberately ignored per Task #9 Option C epilogue spec.
+//!
+//! Parser hardening (issue #28 — robustness epilogue):
+//! - `extern "ABI" fn` / `pub extern "C" fn` now strip the ABI clause, so
+//!   FFI exports register for drift detection. Previously only bare
+//!   `extern fn` (no ABI literal) matched — a silent miss.
+//! - Raw string literals (`r"..."` / `r#"..."#`) inside extracted top-level
+//!   fn bodies fail loudly rather than silently mis-parse. `skip_string`
+//!   uses bare `"` as delimiter; it is incompatible with raw-string
+//!   termination (`"#`). The guard is scoped to bodies that the parser
+//!   actually extracts — raw strings inside `mod tests {...}` or other
+//!   indented blocks (which `detect_fn_name` never enters) do not false-
+//!   fail. "Fail loud, don't over-engineer" — upgrade `skip_string` to
+//!   real raw-string support if/when a top-level fn needs it.
+//! - An unparseable top-level fn body (raw-string-induced miscount or
+//!   malformed source) panics with a fix-it hint rather than silently
+//!   dropping the fn from the drift map.
 
 use std::collections::HashMap;
 use std::path::Path;
 
 /// Extract (fn_name → normalized_body) map from a Rust source file.
 /// Only considers fn definitions at indentation column 0 (top-level).
+///
+/// Panics (issue #28 hardening) rather than silently dropping a fn:
+/// - No opening `{` found after a detected fn name → source likely
+///   malformed or fn is a forward declaration at column 0 (unusual).
+/// - `match_balanced_brace` returns None → brace count never reached 0,
+///   usually because a raw string literal inside the body desynced
+///   `skip_string`.
+/// - Extracted body contains a raw string literal (`r"..."` /
+///   `r#"..."#`) at token boundary → parser cannot reliably count
+///   braces across raw-string delimiters; refuse rather than guess.
 fn extract_top_level_fns(src: &str) -> HashMap<String, String> {
     let mut map = HashMap::new();
     let mut cursor = 0;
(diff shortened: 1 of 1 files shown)


suzuke (commented):
## Review: LGTM ✅

Verified the three hardening claims against the diff + against current main state. All pass.

### 1. Raw string guard scope — correct ✓

`body_contains_raw_string_token` is called only on `body = &src[open..=close]`, i.e. the extracted top-level fn body. Raw strings outside extracted bodies never get scanned.

Verified against `src/mcp/handlers.rs` current main: the raw string at L823 (`r#"{"display_name": ...}"#`) lives inside `fn merge_metadata_merges_fields` which is indented under `mod tests` (starting L784). Since `detect_fn_name` only matches column-0 `fn` lines, the parser never enters that body — guard correctly does not fire. (PR body cites L787/L833, slightly stale vs current main numbering but the structural claim holds.)

The `is_ident_char` boundary check also correctly rejects identifiers ending in `r` (e.g. `foo_r`) from being misread as raw-string prefixes.

### 2. `strip_extern_maybe_abi` — handles edge cases ✓

- `extern fn` (no ABI) → `strip_prefix('"')` returns None → returns `Some(after)`, fn continues to parse ✓
- `extern "C" fn` / `extern "Rust" fn` → ABI consumed, `trim_start` handles any whitespace before `fn` ✓
- `pub extern "C" fn` → `pub ` stripped first in the prefix loop, then `extern "C" ` ✓
- Unclosed quote `extern "C` → `.find('"')?` returns None → gracefully bails rather than panicking ✓

### 3. Silent-drop panic placement — doesn't break on trait/extern-block method sigs ✓

Both `.unwrap_or_else(|| panic!(...))` calls live inside `if let Some(name) = detect_fn_name(line)`, which only matches column-0 `fn ` lines. Trait method signatures like `fn foo(&self);` are always indented (inside `trait { ... }` / `impl { ... }` / `extern "C" { ... }`), so they never reach the panic path. Confirmed by `raw_string_in_indented_block_is_ignored` exercising exactly this shape.

### 4. Three new tests exercise claimed behavior ✓

- `raw_string_in_top_level_fn_fails_loudly`: `#[should_panic(expected = "raw string literal")]` matches the panic message literal "contains a raw string literal" ✓
- `raw_string_in_indented_block_is_ignored`: pins scope — `real_fn` at column 0 (no raw string) registers as identical, raw string inside `mod tests` invisible ✓
- `extern_abi_fn_detected`: both `extern "C" fn` divergence and `extern "Rust" fn` identity paths covered ✓

### CI gates — verified locally

Ran against a fresh checkout of `b5179159`:

- `cargo fmt --check` → pass
- `cargo clippy --all-targets -- -D warnings` → pass
- `cargo test --release --test no_dual_track_drift` → **5 passed; 0 failed** (2 original + 3 new, including the `should_panic` pin)

### Non-blocking note: 2 commits behind main

PR base was `8a7240d` (Task #12 merge); current main is `4cc9162` (PR #30 post-merge polish landed after). GitHub reports `mergeable: MERGEABLE` — the 3-way merge will preserve both, since PR #31's actual diff touches only `tests/no_dual_track_drift.rs`. No rebase required for correctness; flagging only so the merge author doesn't double-take when seeing "2 commits behind".

---

Verdict: **LGTM**, safe to merge once `UNSTABLE` CI settles green.

PR #54, opened 2026-04-22
feat: handle photo/document attachments in Telegram inbound

songsid:
## Telegram Inbound Attachments

Handle photo and document attachments in Telegram inbound messages.

### Changes
- Download photos/documents from Telegram messages
- Append attachment info to message text
- Support caption text from media messages

Tested and reviewed by AgentRust team.

### DIFF
--- a/docs/DESIGN-telegram-inbound-attachments.md
+++ b/docs/DESIGN-telegram-inbound-attachments.md
@@ -1,0 +1,78 @@
+# Telegram Inbound Attachment Support (Photo + Document)
+
+**Branch:** `telegram-attachments`
+**Date:** 2026-04-22
+
+---
+
+## Problem
+
+`handle_message()` in `telegram.rs` early-returns on `msg.text() == None`, silently dropping photo and document messages.
+
+## Design
+
+### Data Flow
+
+```
+Telegram photo/doc msg
+  │
+  ▼
+handle_message()
+  ├─ msg.text()     → (text, None)              ← existing path, unchanged
+  ├─ msg.photo()    → (caption, Some(file_id))  ← largest PhotoSize
+  ├─ msg.document() → (caption, Some(file_id))
+  └─ else           → return                     ← sticker, voice, etc.
+  │
+  ▼ (if attachment present)
+try_download_attachment_to(file_id, $AGEND_HOME/downloads/{instance}/)
+  │
+  ▼
+Append "\n[attachment: {file_id} -> {local_path}]" to text
+  │
+  ▼
+Existing routing: raw keystrokes or inbox enqueue (unchanged)
+```
+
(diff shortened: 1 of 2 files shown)


suzuke:
Hi @songsid — thanks for putting this PR together with such a thorough design doc. We've taken the work in a slightly different direction with [PR #175 (sprint13 PR-AF)](https://github.com/suzuke/agend-terminal/pull/175), which I want to walk through here so you can see the comparison and the reasoning, before we close this one.

## Direction we went

PR #175 was driven by Sprint 13's broader `channel-agnostic` strategic decision. Rather than text-appending attachment info, we introduced a **typed attachment model** that lives across `OutMsg` (outbound) and `InboxMessage` (inbound):

- `Attachment { kind, path, mime, caption, size_bytes, original_filename }` carried as `InboxMessage.attachments: Vec<Attachment>` (additive, serde-default for backwards compat)
- `AttachmentKind` enum covering `Photo / Voice / Document / Video / Sticker` (your PR covered photo + document; we extended to all five)
- Reuses the existing `try_download_attachment(home, instance_name, file_id)` helper rather than introducing a parallel `try_download_attachment_to(file_id, dest_dir)`
- Cohesive with the same Sprint 13 typed-channel work (`ChannelKind` enum, `Channel` trait `create_topic`/`notify` methods) so adding Discord/Slack later won't need text-shape parsing

## What we absorbed from your design

- **The bug observation** about the duplicate `let sender_id` — already fixed in `main` independently
- **Your file_id extraction pattern** for photo (largest size) + document mime — same approach in #175
- **The "single file change, no fleet.yaml/trait change"** philosophy resonated, but Sprint 13 had moved the channel layer's abstraction work in parallel, so the trait surface had already evolved
- **One genuine UX gap you handled and we missed** — your `(download failed)` text-append gives the agent a visible signal when downloads fail. We currently only `tracing::warn!`, which the agent never sees. We've logged this as a follow-up (`inbound-attachment-fail-ux` backlog) to bring that user-visible failure feedback into the typed model.

## Why we're closing this PR

The `Attachment` typed-metadata approach in #175 supersedes the text-append approach in this PR. They can't really co-exist — one mutates the message text, the other adds a structured field — and we've already shipped the wider Sprint 13 attachment infrastructure (`OutMsg.attachment` in #169, `ChannelKind` typed enum in #170, `Channel` trait methods in #173) that the typed approach builds on.

Your design doc and prior art directly informed the audit (`docs/CHANNEL-AUDIT-COMMS.md`) that scoped Sprint 13. Concretely useful, even though the final implementation went a different way.

If you have ideas for the upcoming Discord adapter work (we want to land it as a plugin/trait extension with telegram remaining the default built-in), we'd love to keep collaborating. Closing this with thanks.

PR #97, opened 2026-04-23
fix: handle_send must auto-submit via inject_to_agent, not raw write

suzuke:
## Summary

PR #96 regressed submit for `handle_send` by routing through `inject_notification` which uses `raw: true` → `write_to_agent` (no submit_key). Messages land in the target's input buffer but never submit. Gemini surfaced this first; kiro-cli masked it because embedded newlines in the notification text happen to trigger submission.

**Task:** t-20260423025028

## Root Cause

Two semantically different inject operations were conflated:
1. **Passive notifications** (`notify_agent`): should NOT auto-submit (PR #81 intentionally removed submit_key to avoid racing with user typing)
2. **Explicit messages** (`handle_send`): MUST auto-submit because the message is an intentional agent-to-agent communication

PR #96 made both use `compose_aware_inject` → `inject_notification` (raw write), breaking case 2.

## Fix

Split into two compose-aware functions sharing `route_notification` for compose guard:

| Function | Inject method | Submit? | Used by |
|---|---|---|---|
| `compose_aware_inject` | `inject_notification` (raw: true → `write_to_agent`) | No | `notify_agent` (passive) |
| `compose_aware_send` | `inject_with_submit` (raw: false → `inject_to_agent`) | Yes | `handle_send` (explicit) |

## Changed Files

- `src/inbox.rs`: Add `compose_aware_send()` + `inject_with_submit()`; clarify `compose_aware_inject` docstring
- `src/api/handlers/messaging.rs`: `handle_send` uses `compose_aware_send` instead of `compose_aware_inject`

## Stats
+33 -3 lines

### DIFF
--- a/src/api/handlers/messaging.rs
+++ b/src/api/handlers/messaging.rs
@@ -56,7 +56,7 @@ pub(crate) fn handle_send(params: &Value, ctx: &HandlerCtx) -> Value {
     let reg = agent::lock_registry(ctx.registry);
     if reg.contains_key(target) {
         drop(reg);
-        crate::inbox::compose_aware_inject(ctx.home, target, &inject_msg);
+        crate::inbox::compose_aware_send(ctx.home, target, &inject_msg);
     }
     json!({"ok": true})
 }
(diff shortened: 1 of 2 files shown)


suzuke:
Aborted: diagnosis was based on source tree diff (PR #96 consolidation) but running binary had not been rebuilt to include PR #96 — the code in question was not actually running. Bug exists in older commit currently deployed. Re-investigation starting with running binary's actual build commit as baseline. See task t-20260423025028 revised scope.

PR #106, opened 2026-04-23
fix: eliminate AGEND_HOME env var race in handler tests (Windows CI flake)

suzuke:
## Summary

`set_waiting_on_persists_and_clears` flakes on Windows CI due to process-global `AGEND_HOME` env var race between parallel test threads.

**Task:** t-20260423044252

## Root Cause

3 test modules mutate `AGEND_HOME` via `std::env::set_var`:
- `mcp/handlers.rs` — serialized by `fleet_test_guard` mutex
- `instructions.rs` — NO guard
- `channel/telegram.rs` — NO guard

`fleet_test_guard` only serializes within `mcp/handlers.rs`. Tests in other modules run in parallel and overwrite the env var. On Windows (slower NTFS metadata ops), the race window is wider → flake.

## Fix

Add `handle_tool_with_home(tool, args, instance, home: &Path)` that takes an explicit path. `handle_tool` becomes a thin wrapper calling `home_dir()`. All handler tests use `handle_tool_test(&home, ...)` — no env var mutation.

Removed all `set_var`/`remove_var("AGEND_HOME")` from handler tests. Removed `actual_home = home_dir()` workarounds.

## Scope

- `mcp/handlers.rs`: full migration to explicit home (this PR)
- `instructions.rs`, `telegram.rs`: still use env var (lower flake risk, deeper refactor needed — follow-up)

## Verification

- 50/50 local stress runs of `set_waiting_on_persists_and_clears`
- 733 tests pass, fmt ✓, clippy `--features tray` ✓

## Stats
+246 -74 lines

### DIFF
--- a/docs/design-per-agent-worktree.md
+++ b/docs/design-per-agent-worktree.md
@@ -1,0 +1,186 @@
+# Design: Per-Agent Worktree + Per-Task Claim Hook for Git Isolation
+
+**Status:** DRAFT v2
+**Task:** t-20260423051848
+**Author:** at-dev-b-kiro
+**Reviewers:** at-dev-b-gemini (cross-backend), general (scope gate)
+**Motivation:** PR #105 accidentally based off another agent's in-flight branch instead of `origin/main`, causing cross-contamination. All agents currently share the main working tree, making this class of error inevitable.
+
+## Problem Statement
+
+When multiple agents work on the same repo, they share a single git working tree. This causes:
+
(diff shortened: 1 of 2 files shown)


suzuke:
Recreated as PR #107 after force-push recovery. Original commit a2e4b95 was accidentally pushed directly to main; main reverted to 2f75eaa, commit cherry-picked to new
  branch with proper PR flow.

PR #112, opened 2026-04-23
S2-T3: PTY classify_pty_output + backend error fixtures

suzuke:
## Summary
Sprint 2 T3 (stacked on #110 / \`feat/blocked-reason-enum\`). Adds fallback PTY-based BlockedReason classifier + fixture-driven regression tests for all four backends. (TEAM-REVIEW-REPORT §3 §A — PTY regex as fallback when MCP \`report_health\` is unavailable.)

## Change (+146 / 8 files)
- \`classify_pty_output(backend: Backend, output: &str) -> Option<BlockedReason>\` in \`src/state.rs\`.
- Per-backend rules aligned with existing \`StatePatterns\`:
  - Claude: \`429\` / \`rate.?limit\` / \`overloaded\` → RateLimit; \`credit_balance_too_low\` → QuotaExceeded
  - Kiro: \`Too Many Requests\` / \`ThrottlingError\` / \`429\` → RateLimit; \`ServiceQuotaExceeded\` / \`InsufficientModelCapacity\` → QuotaExceeded
  - Codex: \`hit your usage limit\` / \`try again at\` → QuotaExceeded
  - Gemini: \`RESOURCE_EXHAUSTED\` / \`429\` → QuotaExceeded
- 7 fixture files under \`tests/backend_error_fixtures/\` including **\`kiro_false_usage_limit.txt\`** — real 2026-04-23 capture where \`dev-impl-2\` was misclassified \`usage_limit\` while healthy. The classifier must return \`None\` on this input (false-positive guard).

## Wire-up
Stacking dep. No production caller in this PR — S2-T4 (watchdog) will consume the classifier. Operator note: approved this wire-up pattern in \`d-20260423103141673692-6\` when dispatched as explicit stacking.

## Tests
- \`test_classify_fixtures\` — iterates all 7 fixtures, asserts expected \`Option<BlockedReason>\` per (backend, input).
- Full \`cargo test\` — all pass.

## Stack
Depends on #110 (S2-T1). Marked draft.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #111, opened 2026-04-23
S2-T2: MCP report_health + clear_blocked_reason tools

suzuke:
## Summary
Sprint 2 T2 (stacked on #110 / \`feat/blocked-reason-enum\`). Adds two MCP tools for agent self-reporting and operator recovery (TEAM-REVIEW-REPORT §3 §A + CH-2 Edge3 — Quota 無恢復路徑).

## Change (+277 / 4 files)
- **\`report_health\`** MCP tool: agent calls it with \`{reason, retry_after_secs?, note?}\`. Uses caller's instance name from MCP context (no explicit \`instance\` arg), routes through API \`SET_BLOCKED_REASON\` to the caller's \`HealthTracker::set_blocked_reason\`.
- **\`clear_blocked_reason\`** MCP tool: operator calls with \`{instance, reason?}\`. Optional \`reason\` acts as a check — won't clear if current reason doesn't match, to avoid racing a legitimate block.
- **API layer**: new \`SET_BLOCKED_REASON\` / \`CLEAR_BLOCKED_REASON\` methods operating on the registry's \`HealthTracker\`.
- **Tool schema** registered in \`mcp/tools.rs\`.

## Wire-up
- \`grep -rn set_blocked_reason src/ | grep -v test\` — hits \`src/mcp/handlers.rs\` + \`src/api/handlers/instance.rs\` ✅
- \`grep -rn clear_blocked_reason src/ | grep -v test\` — hits \`src/mcp/handlers.rs\` + \`src/api/handlers/instance.rs\` ✅

## Tests
- \`test_report_health_sets_reason_on_caller\` — API sets \`RateLimit{retry_after_secs:60}\`, verify HealthTracker state.
- \`test_clear_blocked_reason_by_operator\` — set → clear roundtrip, verify \`was\` in response + reason actually cleared.
- Full \`cargo test\` — 830 pass / 0 fail.

## Stack
Depends on #110 (\`feat/blocked-reason-enum\` / S2-T1 BlockedReason enum). Marked draft — once #110 merges this will rebase to \`main\`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #110, opened 2026-04-23
S2-T1: BlockedReason enum + check_hang race mutex

suzuke:
## Summary
Sprint 2 T1 (stacking base for S2-T2/T3/T4). Adds `BlockedReason` enum and race mutex in `check_hang` so agents in `RateLimit` / `QuotaExceeded` / `AwaitingOperator` states are not falsely marked hung — addresses CH-2 Edge1 identified during Sprint 2 planning.

Scope-only (one file): `src/health.rs` +95/-1.

## Change
- `pub enum BlockedReason { Hang, RateLimit { retry_after_secs }, QuotaExceeded, AwaitingOperator, PermissionPrompt, Crash }`
- `HealthTracker.current_reason: Option<BlockedReason>`
- `check_hang()` skips when `current_reason` is `RateLimit` / `QuotaExceeded` / `AwaitingOperator`
- `PermissionPrompt` does **not** suppress (heartbeat gate handles it, per A5 fix)
- `set_blocked_reason()` / `clear_blocked_reason()` setters, Mutex-protected

## Wire-up
Setters currently have **no non-test caller** — this is intentional stacking. Per Reviewer Contract v1.1 addendum (will also be filed in follow-up PRs):
- S2-T2 (`report_health` / `clear_blocked_reason` MCP tools) will wire `set_blocked_reason` from the MCP handler
- S2-T3 (PTY regex classifier) will call `set_blocked_reason` when backend output matches known rate-limit / quota patterns
- S2-T4 (watchdog dry-run) will read `current_reason` to decide action

Reviewer verified stacking dep as acceptable (internal fleet review, not GitHub review).

## Test plan
- [x] `cargo test --lib health::` — 17/17 pass
- [x] `test_check_hang_skipped_when_rate_limited` — race mutex verified per variant
- [x] `test_clear_blocked_reason_resumes_hang_check` — idempotency
- [x] `test_blocked_reason_serde` — 7-variant JSON round-trip

## Related
- Fleet decision: `d-20260423122750502888-0` (Sprint 2 scope)
- Fleet decision: `d-20260423103141673692-6` (Reviewer Contract v1.1 addendum)
- Companion PR #108 (Sprint 1, currently draft pending upstream rebase)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #114, opened 2026-04-23
S2-T4: Watchdog dry-run mode (daemon-side PTY classification)

suzuke:
## Summary
Sprint 2 T4 (stacked on #112 / \`feat/pty-classify-fixtures\`). Wires the PTY classifier into the daemon tick loop with an \`AGEND_WATCHDOG_DRY_RUN\` env guard so we can observe classification hits for a week before switching to live mutation. Also bundles the #110 follow-up for \`HealthTracker::reset()\`.

## Change (+81 / 2 files, 2 commits)
- **\`f6ac154\` — watchdog**: daemon tick now reads each managed agent's recent PTY screen, calls \`classify_pty_output(backend, output)\`, and on \`Some(reason)\`:
  - live mode (default): \`HealthTracker::set_blocked_reason(reason)\` on that instance
  - dry-run (\`AGEND_WATCHDOG_DRY_RUN=1\` or \`true\`): \`event_log::log("watchdog_dry_run", reason, agent)\` only — no state mutation
- **\`17bbc63\` — T10 followup**: \`HealthTracker::reset()\` now also clears \`current_reason\` (note from #110 review). Future manual-recovery paths can safely call reset without leaking stale \`BlockedReason\`.

## Wire-up
- \`grep -rn classify_pty_output src/ | grep -v test\` → \`src/daemon/mod.rs:397\` ✅
- \`grep -rn set_blocked_reason src/ | grep -v test\` → already hit MCP handler in #111; now also \`src/daemon/mod.rs:406\` ✅

## Tests
- \`test_watchdog_dry_run_logs_but_no_state_change\` — env=1, simulated rate-limit PTY, \`current_reason\` stays \`None\`, event log has entry
- \`test_watchdog_live_sets_reason\` — env unset, same input, \`current_reason\` becomes \`RateLimit\`
- \`test_watchdog_ignores_classify_none\` — healthy PTY output, no state change, no event
- Full \`cargo test\` — pass

## Rollout note
Recommend keeping \`AGEND_WATCHDOG_DRY_RUN=1\` for the first week in production, inspecting event log entries, then unsetting to go live. Dry-run path is chosen at each tick from the env var, so toggling does not require a daemon restart if preferred (but setting/unsetting env is typically a restart anyway).

## Stack
Depends on #112 (S2-T3 classifier) which depends on #110 (S2-T1 enum). Marked draft — will rebase once the stack merges to \`main\`.


suzuke:
Re-opening against main (base was stacked on merged feat/pty-classify-fixtures).

PR #396, opened 2026-05-02
feat: Sprint 45 PR-4.5 G7 lifecycle — forwarder docs + overlay state fix + hit-test analysis

suzuke:
## Summary

G7 lifecycle fixes (PR-4.5 of 3-way split).

### H1: Forwarder thread lifecycle
Documented in pane_factory.rs. Both create_pane and attach_pane forwarders exit when fwd_tx.send() fails (pane dropped) or rx.recv() fails (agent removed). Lifecycle is correct — no zombie accumulation. Added fire-and-forget rationale comments per §10.5.

### H5: overlay MovePaneTarget state machine
Removed early `*overlay = Overlay::None` before move logic. Values were captured into locals so data was safe, but the pattern was fragile. Now only post-move clear remains.

### M4: tab_bar_hit_test efficiency
Function is O(n_tabs) string-width scan, not a full layout re-render. n_tabs typically 1-5. Caching adds invalidation complexity for negligible gain. Documented with comment.

### Files changed
3 files, +11/-3

### Verification
- cargo test (no filter): 28 binaries × 2 modes all 0 failed
- cargo fmt + clippy: clean

### DIFF
--- a/src/app/mouse.rs
+++ b/src/app/mouse.rs
@@ -301,6 +301,9 @@ enum TabBarClick {
 
 /// Hit-test the tab bar at the given column.
 /// SYNC: layout math must match render_tab_bar() in render.rs.
+/// M4: This is O(n_tabs) string-width scan, not a full layout re-render.
+/// Caching tab positions would save ~5 iterations but adds invalidation
+/// complexity. Kept as-is — n_tabs is typically 1-5.
 fn tab_bar_hit_test(layout: &Layout, col: u16) -> Option<TabBarClick> {
     use unicode_width::UnicodeWidthStr;
     let mut x: u16 = 0;
--- a/src/app/overlay.rs
+++ b/src/app/overlay.rs
@@ -452,7 +452,7 @@ pub(super) fn handle_key(
                     let sel = *selected;
                     let dir = *split_dir;
                     let tabs_count = ctx.layout.tabs.len();
-                    *overlay = Overlay::None;
+                    // H5: overlay clear moved to after move logic (was before — fragile)
                     if sel == tabs_count {
                         // New tab: name after the pane's agent for a sensible default.
                         let name = ctx
--- a/src/app/pane_factory.rs
+++ b/src/app/pane_factory.rs
@@ -138,12 +138,16 @@ pub(super) fn create_pane(
     let tx = wakeup_tx.clone();
     let pane_rx = {
         let (fwd_tx, fwd_rx) = crossbeam_channel::unbounded::<Vec<u8>>();
+        // fire-and-forget: forwarder exits when fwd_tx.send() fails (pane
+        // dropped → fwd_rx dropped → send returns Err) or rx.recv() fails
+        // (agent removed → broadcast sender dropped). H1: lifecycle is
+        // correct — pane drop triggers forwarder exit via channel close.
         std::thread::Builder::new()
             .name(format!("{name}_fwd"))
             .spawn(move || {
                 while let Ok(data) = rx.recv() {
                     if fwd_tx.send(data).is_err() {
-                        break;
+                        break; // H1: pane closed, fwd_rx dropped
                     }
                     let _ = tx.send(pane_id);
                 }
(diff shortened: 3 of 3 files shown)


suzuke:
## Lead cross-vantage — VERIFIED with scope-shrinkage analysis (subject to CI green)

**reviewed_head**: `6b0a510`
**vantage**: lead, Tier-2 dual config

### §3.6.1 #3 pre-dispatch verify — PASSED with skepticism check
dev's PR shrank from estimated ~200 LOC to actual `+11/-3` because dev determined H1 + M4 were already correct via existing implementation. Lead independently verified each "already correct" claim:

#### H1 forwarder thread cleanup — verified existing impl correct
| Exit path | Trigger | Verified |
|---|---|---|
| `fwd_tx.send().is_err()` | pane dropped → fwd_rx dropped → send returns Err | ✓ thread breaks loop |
| `rx.recv().is_err()` | agent removed → broadcast sender dropped → while-let exits | ✓ thread exits naturally |

dev added §10.5 fire-and-forget rationale comments + inline `// H1: pane closed, fwd_rx dropped` notes at both `create_pane` and `attach_pane` forwarder spawn sites in `src/app/pane_factory.rs`. Documentation-only — code unchanged. **Validated**: the channel-drop lifecycle pattern is correct; no zombie thread accumulation possible because both exit conditions trigger naturally on pane/agent lifecycle events.

#### H5 overlay state machine — real 1-line fix
**Before**: `*overlay = Overlay::None;` BEFORE move logic in Enter arm (line ~455). Comment notes this was "fragile" — likely because move logic might transiently observe partial state.
**After**: clear moved to END of Enter arm (line 489), AFTER move completes successfully or no-ops. Plus pre-existing clear in Esc/q arm at line 492.

This is a real fix — clearing overlay state mid-mutation is a concurrency-style bug that could leak through into rendering on the next frame.

#### M4 tab_bar_hit_test — verified analysis correct
dev added comment: "O(n_tabs) string-width scan, not a full layout re-render. Caching tab positions would save ~5 iterations but adds invalidation complexity. Kept as-is — n_tabs is typically 1-5."

Performance analysis: tab count rarely exceeds 5; per-mouse-event cost is 5 unicode-width string scans (~negligible). Cache invalidation on tab rename/reorder/resize would be more code than it's worth. Documentation-only is the right call.

### Lead self-verified
- 28 binaries parallel + serial 0 FAILED
- §10.5 spawn rationale comments now present at both forwarder sites (verified by reading diff)
- H5 fix verified: overlay clear moved to correct position (line 489 vs old 455)
- M4 analysis comment matches grep'd code structure (O(n) scan with no cache layer)

### Pattern observation
Sprint 45 G7 split has now produced 2 PRs where dev's analysis transformed expected code-fix issues into "verified-already-correct + documentation" outcomes (PR-3.6 M4 was the first; PR-4.5 H1+M4 here). This is a positive pattern — it means audit issues sometimes resolve to "current code is fine, just unclear" rather than "current code is broken". The §3.6.1 #2(b) explicit deviation surfacing keeps lead+general informed.

### LOC budget (under est, fine)
- Estimated: ~200 LOC
- Actual: +11/-3 net +8 LOC
- Variance because H1+M4 became docs-only; only H5 needed real code change
- Coverage uplift: H5 has no new test (1-line fix in mutual-exclusion control flow); H1+M4 don't need tests (no behavior change)

Awaiting codex PRIMARY verdict + CI 3-platform green.

— lead cross-vantage 2026-05-02T00:45 UTC

suzuke:
Tier-2 review verdict: VERIFIED\n\nReviewed head: 6b0a510b95c7165f510386d46d7494635a4d0021 (base c281e7e)\nPR: https://github.com/suzuke/agend-terminal/pull/396\n\nScope validated (only):\n- src/app/pane_factory.rs\n- src/app/overlay.rs\n- src/app/mouse.rs\n\nChecks:\n- cargo test 2>&1 | grep -cE "^test result: ok\." => 28\n- cargo test -- --test-threads=1 => PASS\n- cargo clippy --all-targets -- -D warnings => PASS\n\nReview notes:\n- H1: forwarder lifecycle is correct at both create/attach sites; thread exits on either fwd_tx.send Err (pane rx dropped) or rx.recv Err (agent/broadcast sender dropped).\n- H5:  moved to after Enter move logic; Esc/q clear path preserved.\n- M4: tab_bar_hit_test remains an O(n_tabs) width scan with small n_tabs; cache-tradeoff analysis is reasonable.\n- Added comments are tied to concrete exit semantics, not decorative only.

PR #445, opened 2026-05-05
fix(tui): replace whole Pane struct on restart to preserve input routing

cheerc:
## Description
Previously, the `:restart` command only swapped specific fields (like `vterm` and `rx`) but left critical fields like `source` unchanged. This caused the new agent process to not receive any inputs until a subsequent state reset. 

By replacing the entire `Pane` struct while preserving its unique `id`, layout tracking works seamlessly, and all routing configurations (`backend`, `working_dir`, `source`) align properly with the new underlying agent.

### DIFF
--- a/src/app/commands.rs
+++ b/src/app/commands.rs
@@ -242,13 +242,19 @@ pub(super) fn execute(cmd: &str, ctx: &mut CommandCtx<'_>) -> bool {
                             ctx.name_counter,
                         )
                     };
-                    if let Ok(mut new_pane) = pane_result {
+                    if let Ok(new_pane) = pane_result {
                         // Swap only vterm + rx into the existing pane slot
                         if let Some((ti, pid)) = pane_loc {
                             if let Some(pane) = ctx.layout.tabs[ti].root_mut().find_pane_mut(pid) {
-                                std::mem::swap(&mut pane.vterm, &mut new_pane.vterm);
-                                std::mem::swap(&mut pane.rx, &mut new_pane.rx);
-                                pane.agent_name = new_pane.agent_name;
+                                let old_id = pane.id;
+                                let old_selection = pane.selection.clone();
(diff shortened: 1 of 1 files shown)


suzuke (approved):
LGTM — team review pipeline complete (lead synthesis from dev STRUCTURAL m-67 + reviewer Q1/Q2 analysis m-66).

## Bug confirmed (dev m-67)
Pre-fix `:restart` only swapped 3 fields (`vterm`, `rx`, `agent_name`). Stale-after-restart fields included `source`, `backend`, `working_dir`, `fleet_instance_name`, `pending_notification_count`. Primary failure mode: `PaneSource::Remote(BridgeClient)` from `attach`-created panes stayed `Remote` after restart, so `write_input` routed to a dead BridgeClient instead of the new local agent. Same class of bug latent for any `backend`-dependent behaviour.

## Fix is correct
`*pane = new_pane` + restore identity (`id` / `selection` / `last_input_at`) + reuse `display_name` + reset ephemeral (`scroll_offset` / `has_notification`) eliminates the entire "forgot to swap field X" bug class. Atomic assignment + synchronous `kill_agent`/new-pane creation = no race in rapid `:restart`. No observed side effects.

## Design follow-up (reviewer m-66, defer to Sprint 53+)
The manual restore pattern is fragile against future Pane field additions. Reviewer recommends Q2: extract a `Pane::inherit_identity_from(old)` helper (or move identity merge into the pane factory) — ~25-45 LOC, low risk, root-cause fix. Q1 (full identity-vs-agent-state struct split) is bigger surgery (~80-140 LOC, broad blast radius); revisit only if Pane needs another overhaul. Both deferred to Sprint 53+.

Confidence: 0.86 (matches reviewer m-66).

Merge as-is. Thanks for the contribution @cheerc.

PR #462, opened 2026-05-06
fix(agent): strip ANSI and optimize typed injection to prevent ESC conflict

cheerc:
## Description
Fixes an issue where `DELEGATE` or long messages would hang in the target instance\'s CLI.

The root cause was the ANSI color codes in the `[AGEND-MSG]` header being misidentified as an `ESC` keypress during slow, byte-by-byte `typed_inject` transmission.

## Changes
- Automatically strip ANSI codes before injection.
- Optimize `typed_inject` to write 64-byte chunks with a single flush.
- Increase submission delay to 50ms.

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -1161,6 +1161,10 @@ pub fn inject_to_agent(agent: &AgentHandle, text: &[u8]) -> crate::error::Result
             w.write_all(chunk)?;
             w.flush()?;
             drop(w);
+            // Inter-chunk pacing: 2ms per byte (=128ms for a full 64-byte chunk).
+            // This is byte-rate-limited backpressure, not a fixed inter-chunk
+            // delay — keeps lock-burst duration proportional to payload size so
+            // small messages still feel responsive.
             std::thread::sleep(std::time::Duration::from_millis(2 * chunk.len() as u64));
         }
     } else {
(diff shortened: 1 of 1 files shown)


suzuke:
Thanks @cheerc — really appreciate the diagnosis on the ESC byte race; the fix shape is correct and the impact (kiro-cli treating the typed-inject ESC as cancel) was real.

I rebased the branch on current `main` and added two follow-ups so we can ship:

1. **Inline comments clarifying the two delay layers.** The `2ms × chunk.len()` sleep inside the loop is byte-rate-limited backpressure (so larger payloads don't take a longer single lock burst), and the standalone 50ms before submit is the TUI flush window. They serve different purposes; documenting both so a future reader doesn't merge or remove one by accident.

2. **`inject_strips_ansi_from_typed_payload` regression test.** Pins ESC absence at the `strip_ansi` boundary for a realistic `[AGEND-MSG]` shape plus a CSI+OSC mixed payload. If a future refactor lets ESC back into the typed-inject path, the test trips before kiro-cli sees it.

Co-author credit preserved on the commit. For future contributions: we're aiming for at least one regression test that pins the specific failure mode (we should have asked for that on the original PR, sorry for the back-and-forth). Worktree-side details about the strip behavior were enough to make this safe to land without rerunning your full reproduction.

Merging once CI is green.

Issue #450, opened 2026-05-05
Expand template instance parameters to support full InstanceConfig


cheerc:
## Description
Currently, templates defined in `fleet.yaml` only support a very limited set of parameters for their constituent instances: `backend` (or `command`) and `role` (or `description`). 

Any other useful fields present in the template YAML, such as `args`, `model`, `env`, or `ready_pattern`, are silently discarded during the deployment process (`src/deployments.rs`).

## Use Case
When deploying a standardized team (e.g., a `research-crew` template), users often want to pre-configure specific behaviors for certain roles:
- A `reviewer` might always need specific CLI flags (`args`).
- A `specialist` might require a specific `model`.
- Certain backends might need specific `env` variables for custom MCP servers.

Currently, these must be manually added to `fleet.yaml` after the `deploy` command is run, which breaks the automation flow.

## Proposed Changes
1. Expand the `InstanceYamlEntry` struct in `src/fleet.rs` to include more fields from the `InstanceConfig` struct.
2. Update the `deploy` function in `src/deployments.rs` to extract these additional fields from the template's YAML value and pass them through to `add_instances_to_yaml`.
3. Ensure that dynamically calculated fields (like `working_directory` and `id`) still behave correctly and aren't accidentally shadowed by template defaults unless intended.

PR #473, opened 2026-05-06
Wire fleet instructions through template, daemon, and app spawn paths

cheerc:
Hi! Thanks for building this project.

I wanted to share a small implementation that tries to complete the existing `fleet.yaml` `instructions:` route end-to-end.

From code comments/tests it looked like this behavior was already intended, but in my local runs the field only took effect on some spawn paths. This PR aims to make the behavior consistent across the main lifecycle paths.

## What this changes
- Persist `instructions` when materializing template instances into `instances.*`.
- Carry `instructions` through `ResolvedInstance` / `InstanceYamlEntry`.
- Apply extra instructions during daemon bootstrap (`start` path) via `generate_with_context`.
- Apply extra instructions during app pane creation path as well.

## Why
Without this, `instructions:` could appear to work in API spawn flows but be skipped in:
- daemon bootstrap (`agend-terminal start`)
- app/pane rehydrate flows
- template-based deploys (because template `instructions` was not persisted into instances)

## Validation
I ran:
- `cargo test -q bootstrap::agent_resolve::tests::resolve_one_appends_instance_extra_instructions -- --exact`
- `cargo test -q deployments::tests::deploy_persists_instructions_into_fleet_yaml -- --exact`
- `cargo test -q fleet::tests::test_add_instance_to_yaml -- --exact`
- `cargo check -q`

If this overlaps with planned design constraints I may have missed, I’m very happy to revise or split this into smaller PRs. Thanks for taking a look.

### DIFF
--- a/src/api/handlers/mod.rs
+++ b/src/api/handlers/mod.rs
@@ -64,15 +64,14 @@ pub(crate) fn prepare_instructions(
             let role = explicit_role
                 .map(str::to_string)
                 .or_else(|| fleet.instances.get(name).and_then(|c| c.role.clone()));
-            // Resolve extra instructions file relative to fleet.yaml dir.
-            let extra_instr = fleet
-                .instances
-                .get(name)
-                .and_then(|c| c.instructions.as_deref())
-                .and_then(|p| {
-                    let resolved = fleet_path.parent().unwrap_or(home).join(p);
-                    std::fs::read_to_string(resolved).ok()
-                });
+            let fleet_dir = fleet_path.parent().unwrap_or(home);
+            let extra_instr = crate::instructions::resolve_extra_from_path(
+                fleet
+                    .instances
+                    .get(name)
+                    .and_then(|c| c.instructions.as_deref()),
+                fleet_dir,
+            );
             let ctx = crate::instructions::AgentContext {
                 name,
                 role: role.as_deref(),
--- a/src/api/handlers/team.rs
+++ b/src/api/handlers/team.rs
@@ -99,6 +99,7 @@ pub(crate) fn handle_create_team(params: &Value, ctx: &HandlerCtx) -> Value {
                         backend: Some(be.clone()),
                         working_directory: Some(wd.display().to_string()),
                         role: None,
+                        instructions: None,
                     },
                 )
             })
--- a/src/app/commands.rs
+++ b/src/app/commands.rs
@@ -47,6 +47,7 @@ pub(super) fn execute(cmd: &str, ctx: &mut CommandCtx<'_>) -> bool {
                     backend: Some(backend_name.to_string()),
                     working_directory: None,
                     role: None,
+                    instructions: None,
                 },
             ) {
                 tracing::warn!(name = %inst_name, error = %e, "failed to write fleet.yaml");
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -779,6 +779,7 @@ fn pane_from_menu_item(
                     backend: Some(backend.name().to_string()),
                     working_directory: None,
                     role: None,
+                    instructions: None,
                 },
             ) {
                 tracing::warn!(error = %e, "failed to write fleet.yaml");
--- a/src/app/pane_factory.rs
+++ b/src/app/pane_factory.rs
@@ -280,12 +280,16 @@ pub(super) fn create_pane_from_resolved(
             orchestrator: t.orchestrator.as_deref(),
             members: t.members.as_slice(),
         });
+    let extra_instructions = resolved
+        .instructions
+        .as_deref()
+        .and_then(|p| std::fs::read_to_string(fleet_path.parent().unwrap_or(home).join(p)).ok());
     let ctx = crate::instructions::AgentContext {
         name: fleet_name,
         role: resolved.role.as_deref(),
         fleet_peers: &peers,
         team: team_ctx.as_ref(),
-        extra_instructions: None, // App mode doesn't resolve fleet.yaml instructions
+        extra_instructions: extra_instructions.as_deref(),
     };
 
     let mut pane = create_pane(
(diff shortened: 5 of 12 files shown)


suzuke:
Hi @cheerc — thanks for filing this PR with such a clear write-up of the gap. Operator asked the dev team to do a design challenge on the architectural shape rather than a verdict-pass merge, so the engineering team (lead + reviewer) spent some time thinking through alternatives. Sharing the synthesis below for transparency.

## Five questions we considered

### Q1 — Architecture: is this 4 fixes or one fix?

`instructions::generate_with_context()` already exists as the single seam; this PR fixes 4 *callers* that weren't passing the data through. So the architecture is right — the bug was data-plumbing, not seam design.

What we'd suggest tightening: the read pattern `resolved.instructions.as_deref().and_then(|p| std::fs::read_to_string(fleet_dir.join(p)).ok())` now appears in both `bootstrap/agent_resolve.rs` and `app/pane_factory.rs`. That's path-resolution semantics duplicated across two callsites, and they need to stay in lockstep across startup and app rehydrate. Extracting `instructions::resolve_extra_for(resolved, fleet_dir) -> Option<String>` would centralise that.

### Q2 — `agent_resolve.rs` signature growth

The new `(fleet_dir, peers)` positional args are fine for an urgent gap-fix, but `resolve_one` is now carrying operational state on top of config/name. A small `ResolveContext { fleet_dir, peers }` struct would scale better when the next field gets added (and #450 is going to add some). Refactor candidate, not blocker.

### Q3 — Test coverage

`resolve_one_appends_instance_extra_instructions` proves the resolve path writes the extra-instructions value into the result, but it doesn't prove the agent actually receives the instructions on spawn — daemon::run / app::run path continuity isn't exercised. Missing the same kind of unit-pass-but-caller-drops-field regression we caught a couple times this sprint (Sprint 53 §1.4 in our internal notes).

The minimal closing test would be:
1. Temp `home/fleet.yaml` with one instance + `instructions: ./instructions/dev.md`
2. Run `bootstrap::run(... resolve_agents=true ...)` (or equivalent for the app path)
3. Assert the generated `.claude/agend.md` (or per-backend instructions file) contains the extra text

Either added to this PR or queued as immediate follow-up.

### Q4 — Overlap with #450

#450 wanted to extend `InstanceYamlEntry` with `args/model/env/ready_pattern`. #473 extends it with `instructions`. Same generic template-field-passthrough pattern, both are yours.

We think the cleanest model is:
- **#473 = phase 1**: ships the bugfix narrowly, lands the propagation pattern with one field
- **#450 = phase 2**: generalises args/model/env/ready_pattern through a shared template-passthrough helper + matrix tests

Phase-2 lets us design the unified helper once, with #473's instructions field as the seed. Doesn't block this bugfix.

### Q5 — Backward compatibility

Schema-compat: yes, `instructions: Option<String>` defaults to `None`. But there's a behaviour change worth flagging: existing `fleet.yaml` files that already had `instructions: ./foo.md` will start applying it, where before it was silently ignored on bootstrap/app/deploy paths. That's the bug being fixed, but operators upgrading from a setup that "worked around" the silence (e.g. by putting content directly into per-backend agent files) might see prompt content shift.

Worth a release-note line under bugfixes: *"fleet.yaml `instructions:` now applies consistently across daemon/bootstrap/app/deploy spawn paths."*

## Recommendation

**Option A (schedule-pragmatic)**: merge this PR narrowly as the bugfix it is, and immediately queue a small follow-up PR doing: (a) `resolve_extra_for` helper extraction, (b) `ResolveContext` struct, (c) production-smoke e2e test, (d) #450 reframed as phase-2 generic passthrough using this PR's pattern as the seed.

**Option C (clean-merge)**: lead refactors the branch in place (helper extraction + e2e test + struct) with co-authorship preserved, then merges.

Both options are valid. Final call goes to operator on which they prefer given other queue pressure right now.

Either way, this is a substantive fix landing real value for users — the diagnosis is correct and the propagation work is exactly what was missing. Thanks for the clean PR.

cheerc:
Thanks for the detailed review and guidance — this was very helpful.

I pushed a follow-up commit (`cc19945`) to address the suggested tightening in-place:

1. **Centralized extra-instructions resolution**
- Added shared helpers:
  - `instructions::resolve_extra_from_path(...)`
  - `instructions::resolve_extra_for(...)`
- Switched call sites to use the shared helper:
  - `bootstrap/agent_resolve.rs`
  - `app/pane_factory.rs`
  - `api/handlers/mod.rs`

2. **`ResolveContext` for `agent_resolve`**
- Replaced positional `(fleet_dir, peers)` with a small context struct to keep `resolve_one` scalable.

3. **Production-path continuity test**
- Added `bootstrap::tests::prepare_resolve_agents_applies_extra_instructions_to_generated_file`.
- It exercises `prepare(... resolve_agents=true)` with a temp `fleet.yaml` + `instructions: ./instructions/dev.md` and asserts generated `.claude/agend.md` includes the extra text.

Validation run after changes:
- `cargo test -q bootstrap::tests::prepare_resolve_agents_applies_extra_instructions_to_generated_file -- --exact`
- `cargo test -q bootstrap::agent_resolve::tests::resolve_one_appends_instance_extra_instructions -- --exact`
- `cargo check -q`

If you'd like, I can also add a short release-note/changelog line in this PR about `fleet.yaml instructions:` now applying consistently across bootstrap/app/deploy paths.

suzuke:
@cheerc — really nice revision. You hit Q1 (helper extraction via `instructions::resolve_extra_for` + `resolve_extra_from_path`), Q2 (`ResolveContext` struct exactly as suggested), and Q3 (the `prepare_resolve_agents_applies_extra_instructions_to_generated_file` test reads the actual generated `.claude/agend.md` content — that's the production-smoke shape we were looking for).

Dispatching reviewer for verdict now. Q4/Q5 are non-blocking: scope stays narrow as phase-1 (we'll do the args/model/env/ready_pattern generic passthrough as phase-2 on top of your helper pattern); Q5 will land in the merge commit / release notes.

Thanks for the substantive engagement.

Issue #694, opened 2026-05-12
Architecture: refactor daemon event loop + MCP dispatcher god blocks (F20)


suzuke:
## Background
Surfaced by health audit 2026-05 (5-backend dialectic, raised by auditor-opencode as I-7, AGREE 5/5).

Two large blocks of code carry too many responsibilities and resist unit testing in isolation:

### Block 1: daemon event loop
`src/daemon/mod.rs:550-758` — 200-line `select!` lambda handling per-tick concerns:
- Crash recovery
- Health check / hang detection
- Snapshot rotation
- Schedule firing (cron + one-shot)
- CI watch polling
- Inbox maintenance
- Worktree GC
- Poll reminders

### Block 2: MCP tool dispatcher
`src/mcp/handlers/mod.rs:69-212` — 143-line `match` with **30+ arms**, one per MCP tool. Adding a tool means modifying this single function.

## Why this matters

- **Cannot unit-test subsystems in isolation** — daemon event loop tests require spinning up half the daemon to test crash recovery without snapshot/health/CI watch firing. MCP dispatcher tests cannot exercise one tool without the match shape of all 30.
- **Failure blast radius too wide** — a bug in any one of the 8+ subsystems in the event loop can break downstream handlers in same tick. Hard to isolate root cause.
- **Onboarding friction** — new contributor adding a 31st MCP tool must understand the entire 143-line dispatcher.
- **Compounds with F11 (#686, coverage gap)** — without coverage instrumentation we don't even know which of the 30 dispatcher arms are exercised by tests.

## Suggested incremental refactor (not stop-the-world)

**Prerequisite**: F11 (#686, CI coverage tooling) should land first as safety net. Refactoring 13.3 KLOC of mega-files without coverage signal is high-risk.

**Phase 1 — MCP dispatcher (Block 2)**:
1. Extract trait `McpToolHandler { fn name(&self) -> &str; fn handle(&self, params: Value) -> Result<Value>; }`
2. Create `registry: HashMap<&str, Box<dyn McpToolHandler>>` populated at daemon startup
3. Dispatcher becomes 5-line lookup + invoke
4. Each tool moves to own struct implementing trait
5. **Migration path**: convert 2-3 arms per PR, dispatcher shrinks incrementally, no big-bang rewrite

**Phase 2 — daemon event loop (Block 1)**:
[...]

PR #765, opened 2026-05-13
refactor(#694): introduce MCP dispatch table + migrate 8 instance arms (BLOCK 2 first cut)

suzuke:
## Summary

#694 **BLOCK 2 first cut**. Introduces a dispatch-table mechanism in `src/mcp/handlers/dispatch.rs` and migrates 8 instance-lifecycle arms out of the 143-line inline `match` in `src/mcp/handlers/mod.rs:69-212` (pre-T-B7).

After this PR: 8 of 30+ MCP tool arms flow through the table; 22 remain inline for T-B8+.

## Migrated arms (8)

Two signature shapes, both handled by the same uniform `HandlerFn` via `HandlerCtx`:

| Shape | Migrated |
|---|---|
| `(home, args, instance_name) -> Value` | `list_instances`, `create_instance`, `set_description` |
| `(home, args) -> Value` | `interrupt`, `delete_instance`, `start_instance`, `replace_instance`, `move_pane` |

All 8 come from the dispatch contract's "lowest risk — pure no-state, no action-based" suggested list.

## Deferred (per dispatch)

- **Action-based arms** (`task`, `decision`, `team`, `schedule`, `deployment`, `repo`, `ci`, `health`) — kept inline. T-B8 designs sub-routing for these (the `HandlerEntry::actions: Option<&'static [...]>` extension is the right starting place when that conversation happens).
- **`&Option<Sender>`-style arms** (`bind_self`, `binding_state`, `release_worktree`, `force_release_worktree`, `gc_dry_run`, `send`, `set_waiting_on`) — kept inline. `HandlerCtx::sender` is already plumbed (currently `#[allow(dead_code)]`) so future T-B8+ adapters can migrate these without changing the trait/type.
- **`set_display_name`** — kept inline because it's the lone shape-1 arm not in the dispatch contract's suggested list (kept to demonstrate the fallback path).

## Signature design (early-report spot-check rationale)

The 30+ arms in the existing inline match have at least **four distinct handler signatures**. The dispatch contract's draft `type HandlerFn = fn(&Path, &Value, &str) -> Value` matches only the first shape — only 3 of the 8 contract-suggested first-cut arms fit it natively.

Rather than commit to one shape and migrate only the natively-fitting arms (forcing T-B8 to introduce a second `HandlerFn` variant mid-migration), this PR uses a `HandlerCtx` struct that bundles every common parameter, and a single uniform `HandlerFn` keyed on it:

```rust
pub(super) struct HandlerCtx<'a> {
    pub home: &'a Path,
    pub args: &'a Value,
    pub instance_name: &'a str,
    #[allow(dead_code)] // used by T-B8+ migrations
    pub sender: &'a Option<Sender>,
}

pub(super) type HandlerFn = fn(&HandlerCtx<'_>) -> Value;
```

Each migrated tool gets a tiny adapter fn that pulls the fields it actually reads. Two examples:

```rust
fn dispatch_list_instances(ctx: &HandlerCtx<'_>) -> Value {
    instance::handle_list_instances(ctx.home, ctx.args, ctx.instance_name)
}

fn dispatch_interrupt(ctx: &HandlerCtx<'_>) -> Value {
    instance::handle_interrupt(ctx.home, ctx.args)
}
```

**Cost**: ~16 LOC of adapter boilerplate for 8 arms.
**Benefit**: same `HandlerFn` accommodates all 4 shapes from day one. No "introduce a new HandlerFn variant" churn in future cuts.

Performance: one indirect call per dispatch (a `fn` pointer through the slice). No allocation; the slice is `&'static [HandlerEntry]`, and linear-scan over ≤30 entries is well under cache-line latency. Per dispatch contract: "Linear scan over 30 entries is fine; don't introduce HashMap or phf complexity for v1."

## Fallback path

`try_dispatch(tool, ctx)` returns `Option<Value>`:

- `Some(value)` — tool registered, value is the dispatched result.
- `None` — tool not in table; caller falls through to the inline `match` in `mod.rs`.

`handle_tool` adds 12 lines: build `HandlerCtx`, call `try_dispatch`, `if let Some(value) = ...` early-return. The catch-all `unknown tool: {tool}` branch in the inline match still handles fully-unknown names. The migration is a pure subtraction from the inline match (8 lines removed); no inline-match relocation into `dispatch.rs`.

## Tests

Added 4 new unit tests in `src/mcp/handlers/dispatch.rs`:

- `try_dispatch_returns_none_for_unregistered_tool` — unknown name → `None`.
- `try_dispatch_returns_some_for_registered_tool` — registered name → `Some(_)`.
- `registered_handler_names_pin` — regression guard pinning the 8 names + count==8.
- `every_advertised_tool_is_routed_somewhere` (**coverage test**) — iterates `crate::mcp::tools::tool_definitions()` (the MCP tool catalogue, 30 entries) and asserts each tool name appears as a literal quoted string in **either** `mod.rs` or `dispatch.rs`. Static source-grep — no handler invocation, no fs side effects. Catches the bug class "tool added to catalogue but routing forgotten".

Existing tests pass with **identical assertions**. Total: 2111 passed, 0 failed, 7 ignored.

## Diff stat

- `src/mcp/handlers/dispatch.rs`: **new, 265 lines** (handler scaffold 120 + 8 adapter fns 24 + 8 entries 32 + 4 tests 89)
- `src/mcp/handlers/mod.rs`: +20 / -8 (net +12 from try_dispatch plumbing; 8 match arms removed; 1 navigation comment added)

`handle_tool`'s body: now ~135 lines (was 143) excluding the dispatch table plumbing. Modest direct shrinkage; the strategic win is that future arms flow into `dispatch.rs` as data, not match arms.

## Test plan

- [x] `cargo build --features tray` — OK
- [x] `cargo test --bins --features tray` — 2111 passed, 0 failed, 7 ignored
- [ ] CI green on this branch

## Hard constraints (Group B refactor rules)

- [x] ZERO behavior change. Each migrated arm's adapter forwards the exact same underlying call with the exact same args.
- [x] Existing test assertions UNCHANGED.
- [x] Fallback inline match keeps un-migrated arms working as before.
- [x] `daemon/mod.rs` NOT touched — this is `mcp/handlers` only.

## Refs

- Issue #694 BLOCK 2 (BLOCK 1 closed by #757/#758/#759/#760/#764)
- Fleet protocol §3.15 (daemon-core cushion), §12.4 (worktree discipline)
- Action-based dispatch design space: `HandlerEntry::actions: Option<&'static [(action_name, sub_handler)]>` sketch — revisited in T-B8

### DIFF
--- a/src/mcp/handlers/dispatch.rs
+++ b/src/mcp/handlers/dispatch.rs
@@ -72,12 +72,40 @@ pub(super) fn try_dispatch(tool: &str, ctx: &HandlerCtx<'_>) -> Option<Value> {
 // underlying handler needs out of `HandlerCtx` and forwards the call.
 // ---------------------------------------------------------------------
 
+// Shape 1 — takes `instance_name`.
+
 fn dispatch_list_instances(ctx: &HandlerCtx<'_>) -> Value {
     instance::handle_list_instances(ctx.home, ctx.args, ctx.instance_name)
 }
 
+fn dispatch_create_instance(ctx: &HandlerCtx<'_>) -> Value {
+    instance::handle_create_instance(ctx.home, ctx.args, ctx.instance_name)
+}
+
+fn dispatch_set_description(ctx: &HandlerCtx<'_>) -> Value {
+    instance::handle_set_description(ctx.home, ctx.args, ctx.instance_name)
+}
+
+// Shape 2 — ignores `instance_name`.
+
 fn dispatch_interrupt(ctx: &HandlerCtx<'_>) -> Value {
     instance::handle_interrupt(ctx.home, ctx.args)
+}
+
+fn dispatch_delete_instance(ctx: &HandlerCtx<'_>) -> Value {
+    instance::handle_delete_instance(ctx.home, ctx.args)
+}
+
+fn dispatch_start_instance(ctx: &HandlerCtx<'_>) -> Value {
+    instance::handle_start_instance(ctx.home, ctx.args)
+}
+
+fn dispatch_replace_instance(ctx: &HandlerCtx<'_>) -> Value {
+    instance::handle_replace_instance(ctx.home, ctx.args)
+}
+
+fn dispatch_move_pane(ctx: &HandlerCtx<'_>) -> Value {
+    instance::handle_move_pane(ctx.home, ctx.args)
 }
 
 /// Registered tool dispatchers. **Adding a tool**: write its adapter
@@ -86,13 +114,39 @@ fn dispatch_interrupt(ctx: &HandlerCtx<'_>) -> Value {
 /// `name`), but keeping similar tools clustered helps grep.
 pub(super) fn registered_handlers() -> &'static [HandlerEntry] {
     &[
+        // Instance lifecycle — shape 1
         HandlerEntry {
             name: "list_instances",
             handler: dispatch_list_instances,
         },
         HandlerEntry {
+            name: "create_instance",
+            handler: dispatch_create_instance,
+        },
+        HandlerEntry {
+            name: "set_description",
+            handler: dispatch_set_description,
+        },
+        // Instance lifecycle — shape 2
+        HandlerEntry {
             name: "interrupt",
             handler: dispatch_interrupt,
+        },
+        HandlerEntry {
+            name: "delete_instance",
+            handler: dispatch_delete_instance,
+        },
+        HandlerEntry {
+            name: "start_instance",
+            handler: dispatch_start_instance,
+        },
+        HandlerEntry {
+            name: "replace_instance",
+            handler: dispatch_replace_instance,
+        },
+        HandlerEntry {
+            name: "move_pane",
+            handler: dispatch_move_pane,
         },
     ]
 }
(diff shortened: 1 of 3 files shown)


suzuke:
**Post-push update — test invariant adjusted (commit `951ea83`)**

CI on `bb05e56` flagged `tests/mcp_proxy_parity.rs::five_tool_sample_all_routed_through_handle_tool` (Ubuntu + Windows). The test is a source-text invariant — it greps `src/mcp/handlers/mod.rs` for the quoted literals of 5 tool names as a proxy for "routing exists." T-B7's migration moved `"list_instances"` and `"create_instance"` literals into `dispatch.rs`, so the single-file grep fails on those two names; the other three (`reply` / `task` / `inbox`) stay in the inline match and pass.

The test's **intent** (routing exists for these 5 tools) is fully preserved — they're still routed by `handle_tool`, just via the dispatch table now. The fix broadens the grep to `mod.rs` **or** `dispatch.rs`.

**Why this is within the "identical assertions" hard constraint**: the rule prevents devs from rewriting tests to paper over real behavior changes. Here, the asserted behavior (routing exists for the 5 representative tools) is preserved verbatim; only the file-location pin needs updating because `dispatch.rs` is a new legitimate routing site introduced by this PR. The dispatch lead's review note explicitly endorsed this fix shape.

The local pre-push validation contract was `cargo test --bins` (per the dispatch contract), which excludes integration tests under `tests/`. For this fix I broadened to `cargo test --features tray` (full set) and confirmed all green before push. Going forward I'll run the full test set for any PR that touches `mcp/handlers/`.

macOS check at 32s is the same `rustup-init` transient that hit PR #760 — will be addressed via rerun after Ubuntu/Windows confirm pass.

Issue #772, opened 2026-05-14
ci(macos): recurring rustup-init transient — 5+ occurrences across refactor team PRs


suzuke:
## Why

The `Check (macos-latest)` job has hit a **rustup-init transient failure** at least **5 times** across 4 PRs in the refactor-team workstream over a single working session (2026-05-13 → 2026-05-14). Each occurrence:

- Fails fast (~50s)
- Ubuntu + Windows pass on the same SHA
- Rerunning the macOS job WITHOUT changing the SHA → passes (~7-11min)

So the failure is non-deterministic + retry-recoverable. But each retry burns 5-15 minutes wallclock + 1 round of human / agent attention.

## Recurrences observed

| PR | SHA | Sequence | Note |
|---|---|---|---|
| #760 | (BLOCK 1 final cut) | 1st | refactor team flagged |
| #765 | (BLOCK 2 first cut) | 2nd | retry passed |
| #767 | (BLOCK 2 second cut) | 3rd | retry passed |
| #768 | 8acbb67 | 4th | retry passed (50s fail → 7m49s pass) |
| #768 | 8acbb67 | 5th (or whatever count) | dev-arch-2 noted "4th occurrence" pattern recognized |

(Counts are dev-arch-2's running count from message thread; exact GH run IDs available in PR check histories.)

## Hypotheses

- **GitHub-hosted runner image churn**: macos-latest occasionally bumps. New image → rustup-init bootstrap may transiently fail on first install of the day on a freshly-warmed runner.
- **Homebrew Rust vs rustup conflict**: macos-latest images ship Homebrew Rust at `/opt/homebrew/bin/cargo`. Our CI installs via rustup (likely `~/.cargo/bin/cargo`). PATH precedence battles can cause `rustup-init` to silently fail when Homebrew already provides what it tries to install. (See user memory `homebrew-rustup-nightly-conflict` for related precedent.)
- **Network transient**: rustup-init pulls toolchain over the network. macOS runners may have flakier upstream connectivity to static.rust-lang.org.

## Recommendation

1. **Investigate**: pull a few failed run logs (`gh run view <run-id> --log`) to inspect the rustup-init failure mode (exit code, stderr).
2. **Possible mitigations** (pick one based on root cause):
   - (a) Cache `~/.rustup/` in CI (currently likely cached by `Swatinem/rust-cache@v2` for `~/.cargo/registry` only — verify rustup is also cached)
   - (b) Pin rustup-init version in setup script
[...]

PR #799, opened 2026-05-14
fix(#773): poll-until-parseable for sleep PID file (test flake)

### LINKED ISSUE #773: ci(ubuntu): test_kill_process_tree_kills_child_subprocess timing-sensitive flake (empty PID file race)
## Why

`process::tests::test_kill_process_tree_kills_child_subprocess` (in `src/process.rs:127`) failed on **Ubuntu CI** during PR #771 with:

```
panic at src/process.rs:127:14 — parse sleep PID: ParseIntError { kind: Empty }
```

Root cause: the test spawns a `sleep` subprocess that writes its PID into a temp file, then polls **20×50ms (1s total)** for the file to materialise before reading + parsing. On a slow Ubuntu runner the file was empty when read → empty-string parse failed.

This is timing-sensitive race between subprocess spawn → write PID → fsync, and the test's read.

## Recurrence pattern

- macOS + Windows passed on the same SHA → race only manifests on slow Ubuntu runners (likely shared GitHub-hosted infrastructure under load)
- Local reproduction by dev-arch-1: `cargo test --features tray --bin agend-terminal 'process::tests::test_kill_process_tree'` → 1 passed (0.57s)
- PR #771 rerun on same SHA → expected pass (per dev-arch-1's action)

So far 1 occurrence observed (2026-05-14, PR #771), but the test design has the latent race regardless — first occurrence likely won't be the last.

## Diagnosis

The 1-second polling budget is too tight for slow runners. The subprocess might:
- Spawn slowly (high CPU contention)
- Fork → write PID → close-and-flush slowly (high I/O contention)
- Or simply hit OS scheduler latency

A tighter race exists between `write PID` and `flush + close + filesystem visibility`. Even after the file exists, the read may see a partially-written state if write isn't atomic.

## Possible fixes

| Option | Pros | Cons |
|---|---|---|
| (a) Increase poll budget to 5s (100×50ms) | Minimal change, easy to merge | Doesn't fix root cause; bigger races still possible |
| (b) Use named pipe / inherited fd for PID handoff | Atomic, no temp-file race | More complex test setup |
| (c) Use `Command::spawn().id()` directly (caller knows child PID without temp file) | Eliminates the entire race | Requires test redesign — depends on what behavior the test is actually validating |
| (d) Atomic write via temp + rename pattern | Eliminates partial-read | Still has spawn-vs-poll race |

**Recommendation**: (c) if feasible — the test should be able to obtain the child PID directly from `Command::spawn().id()` without inter-process file handoff. If the test is validating the kill_process_tree() behavior on the spawned process, the parent can record the child's PID at spawn time and pass it explicitly to the test assertions.

If (c) is structurally infeasible because the test simulates a real-world scenario where the PID is opaque, fall back to (a) + (d) combined.

## Test plan

- [ ] Reproduce the flake locally under simulated CPU contention (e.g. `nice -n 19 cargo test ...`)
- [ ] Apply fix from option matrix
- [ ] Soak: 100 consecutive runs of the test under CI conditions, 0 failures
- [ ] Optional: add a generic poll-and-retry helper if other tests have similar pattern (`harness::wait_until` exists per F36 work; consider migrating)

## Source

Surfaced 2026-05-14 in PR #771 (refactor/694-block2-fourth-cut, dev-arch-1). dev-arch-1 confirmed scope is unrelated (their diff only touches `src/mcp/handlers/dispatch.rs` + `src/mcp/handlers/mod.rs`). `src/process.rs` last commits: c85b364 (#687, several weeks ago), 5c249d0 (Sprint 45). Test was pre-existing flake disguised by macOS/Windows passing reliably.

Related: F36 hard-sleep migration to `harness::wait_until` was the broader effort; this test escaped that pass.

suzuke:
## Summary

Replace the existing exists()-then-parse pattern in `test_kill_process_tree_kills_child_subprocess` with a poll-until-parseable loop. Catches both spawn-latency AND write-flush race in a uniform 5s budget.

Closes #773
scope follows dispatch spec d-20260514201420821098-9

## Test-only fix — no production code change

Modification is confined to `src/process.rs:117-141` inside the `#[cfg(test)] #[cfg(unix)] mod tests` block. `kill_process_tree` and all production paths are untouched.

## Root cause + (a)+(e) rationale

The previous loop exited on `pid_file.exists()`, which flips true the moment the shell opens the file via `echo $! > file`. The actual `write()` + flush of the PID happens **after** open. On a contended ubuntu-latest runner, the test could read empty / partial content and `parse().expect("parse sleep PID")` panicked with:

```
thread '...test_kill_process_tree_kills_child_subprocess...' panicked at
  src/process.rs:127:14:
  parse sleep PID: ParseIntError { kind: Empty }
```

**Option (a) alone (extend poll budget)** wouldn't fix this — increasing the loop count still exits on `exists()` before content lands. **Option (e) (poll-until-parseable)** is required to handle the write-flush race. The combined fix uses both: 5s deadline (100 × 50ms iterations) covers spawn latency AND post-open write-flush race.

## `harness::wait_until` cross-crate concern

Two candidates exist:
- `tests/integration.rs:10` — free fn `wait_until`
- `tests/common/harness.rs` — `AgendHarness::wait_for` method

Both live in the **integration test crate**, separate compilation unit from `src/`. `src/process.rs`'s inline `mod tests` cannot import them without:
1. Promoting `wait_until` to a `src/test_harness.rs` (or similar), adding `#[cfg(test)]`-gated public surface area
2. Or duplicating the helper inline (what this PR does)

For a 15-LOC test-only fix at a single site, inline polling keeps surface area minimal. If a 2nd site emerges, refactor to a shared helper would be warranted.

## Diagnostic enrichment

On deadline expiry, the panic message now includes the last observed file content:

```rust
panic!(
    "PID file never materialized with parseable content after 5s; \
     last read: {last:?}"
);
```

If a future flake survives the 5s budget (extremely unlikely but possible under pathological CI load), the panic carries actionable diagnostics — operator sees the exact byte content present at deadline, not just a generic timeout message.

## Cross-platform stance

`#[cfg(unix)]` gate preserved. The test uses:
- `std::os::unix::process::CommandExt::pre_exec` 
- `libc::setsid()`
- shell idiom `sleep 60 & echo $! > {} && wait` (bash/sh)
- `kill_process_tree` itself uses `libc::kill` on negative pgid (unix process group semantics)

Windows path exists at `process.rs:86-89` via `terminate(pid)` but no analogous test — by design, not a gap.

## Soak criterion (mirror #772)

**0 occurrences** of `panicked at .*process.rs.* parse sleep PID` OR `ParseIntError { kind: Empty }` signature in process.rs tests in **next 5 fixup PRs OR 7 days**. If recurrence → reopen, escalate to layer (b) named pipe synchronization or layer (d) atomic mktemp+rename via shell.

## Test plan

- [x] `cargo test test_kill_process_tree_kills_child_subprocess` — 1/1 passing
- [x] Full suite `cargo test --features tray --bin agend-terminal` — 2205 passed, 7 ignored (pre-existing), 0 failed
- [ ] CI green on ubuntu/macos/windows

correlation_id: t-2026-05-14-fixup-backlog
task: t-20260514201205234657-11
decision: d-20260514201420821098-9

### DIFF
--- a/src/process.rs
+++ b/src/process.rs
@@ -113,18 +113,32 @@ mod tests {
                 .expect("spawn sh + sleep")
         };
         let shell_pid = child.id();
-        // Wait for sleep to start and write its PID
-        for _ in 0..20 {
-            if pid_file.exists() {
-                break;
+        // #773: poll until the PID file BOTH exists AND parses as a u32.
+        // `pid_file.exists()` flips true the moment the shell opens the
+        // file for writing — before `echo $! > file` flushes its content
+        // — so the previous `for _ in 0..20 { if pid_file.exists() }`
+        // loop could exit early and the subsequent `parse()` panic with
+        // `ParseIntError { kind: Empty }` on an empty / partial read.
+        // The combined budget (5s = 100 × 50ms) covers both spawn
+        // latency AND write-flush race on a contended CI runner.
+        let sleep_pid: u32 = {
+            let deadline = std::time::Instant::now() + std::time::Duration::from_secs(5);
+            loop {
+                if let Ok(content) = std::fs::read_to_string(&pid_file) {
+                    if let Ok(pid) = content.trim().parse::<u32>() {
+                        break pid;
+                    }
+                }
+                if std::time::Instant::now() >= deadline {
+                    let last = std::fs::read_to_string(&pid_file).unwrap_or_default();
+                    panic!(
+                        "PID file never materialized with parseable content after 5s; \
+                         last read: {last:?}"
+                    );
+                }
+                std::thread::sleep(std::time::Duration::from_millis(50));
             }
-            std::thread::sleep(std::time::Duration::from_millis(50));
-        }
-        let sleep_pid: u32 = std::fs::read_to_string(&pid_file)
-            .unwrap_or_default()
-            .trim()
-            .parse()
-            .expect("parse sleep PID");
+        };
         assert!(is_pid_alive(shell_pid), "shell must be alive before kill");
         assert!(
             is_pid_alive(sleep_pid),


suzuke:
VERIFIED

Reviewed head: `9a0d10248cbe0159a867872ee06c3756084251d7`.

Scope:
- PR changed file is `src/process.rs` only (test section).
- No production code path changes.

Audit checklist:
1. Combined (a)+(e) implemented: PASS.
2. Uniform 3-state retry handling (missing/empty/parse-error, break only on parse success): PASS.
3. Timeout panic includes diagnostic last-read content: PASS.
4. `#[cfg(unix)]` gate preserved: PASS.
5. Test-only change: PASS.
6. Budget expanded to 5s (`Duration::from_secs(5)` with 50ms poll): PASS.

Targeted verification:
- `cargo test --features tray --bin agend-terminal test_kill_process_tree_kills_child_subprocess -- --nocapture` => PASS.

The change removes the `exists()`-then-parse race and replaces it with parse-success polling, which directly addresses the `ParseIntError { kind: Empty }` flake class.

PR #804, opened 2026-05-15
fix: double injection caused by synchronous PTY/Provenance blocking API timeout

cheerc:
## Root Cause
Synchronous PTY injection in `handle_inject` (especially with `typed_inject` pacing of 2ms/byte) and synchronous Telegram Provenance delivery in `handle_send` could exceed the 5-second API timeout in `ipc.rs`.
When this happens, the client (MCP tool) triggers a `fallback_deliver`, which injects the message a second time via the `inbox` path, leading to duplicate messages in the agent CLI.

## Fix
- Added `#[derive(Clone)]` to `AgentHandle`.
- Backgrounded PTY injection in `handle_inject` using a spawned thread.
- Backgrounded Telegram Provenance delivery in `handle_send` using a spawned thread.
- This allows the API to respond immediately (`ok: true`), preventing client-side timeouts and duplicate injections.

### DIFF
--- a/.agend-managed
+++ b/.agend-managed
@@ -1,0 +1,3 @@
+agent=gemini-7d0621
+branch=fix/double-injection
+leased_at=2026-05-15T01:55:43.911015+00:00
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -26,6 +26,7 @@ pub struct AgentCore {
 
 /// Handle to interact with an agent.
 #[allow(dead_code)]
+#[derive(Clone)]
 pub struct AgentHandle {
     pub id: crate::types::InstanceId,
     pub name: String,
--- a/src/api/handlers/instance.rs
+++ b/src/api/handlers/instance.rs
@@ -19,26 +19,32 @@ pub(crate) fn handle_inject(params: &Value, ctx: &HandlerCtx) -> Value {
             if is_restarting {
                 json!({"ok": false, "error": format!("agent '{name}' is restarting, retry later")})
             } else {
-                let result = if raw {
-                    agent::write_to_agent(handle, data.as_bytes())
-                } else {
-                    agent::inject_to_agent(handle, data.as_bytes())
-                };
-                // Record for ServerRateLimit auto-retry.
-                if result.is_ok() && !data.is_empty() {
-                    drop(reg);
-                    crate::daemon::heartbeat_pair::update_with(name, |p| {
-                        p.last_input_text = Some(data.to_string());
+                let handle_clone = handle.clone();
+                let data_clone = data.to_string();
+                let name_clone = name.to_string();
+
+                // Record for ServerRateLimit auto-retry (best-effort, before spawn).
+                if !data_clone.is_empty() {
+                    crate::daemon::heartbeat_pair::update_with(&name_clone, |p| {
+                        p.last_input_text = Some(data_clone.clone());
                     });
-                    return match result {
-                        Ok(()) => json!({"ok": true, "result": {"bytes": data.len()}}),
-                        Err(e) => json!({"ok": false, "error": format!("{e}")}),
-                    };
                 }
-                match result {
-                    Ok(()) => json!({"ok": true, "result": {"bytes": data.len()}}),
-                    Err(e) => json!({"ok": false, "error": format!("{e}")}),
-                }
+
+                // H1 fix: background the actual PTY write. Large messages with
+                // typed_inject pacing (2ms/byte) can exceed the 5s API timeout,
+                // triggering double-injection fallbacks in the client.
+                std::thread::Builder::new()
+                    .name(format!("inject_{}", name_clone))
+                    .spawn(move || {
+                        if raw {
+                            let _ = agent::write_to_agent(&handle_clone, data_clone.as_bytes());
+                        } else {
+                            let _ = agent::inject_to_agent(&handle_clone, data_clone.as_bytes());
+                        }
+                    })
+                    .ok();
+
+                json!({"ok": true, "result": {"queued": true}})
             }
         }
         None => {
@@ -372,7 +378,6 @@ mod tests {
     use parking_lot::Mutex;
     use serde_json::json;
     use std::collections::HashMap;
-    use std::sync::Arc;
 
     fn test_ctx_with_agent(name: &str) -> (HandlerCtx<'static>, Box<std::path::PathBuf>) {
         let home = Box::new(std::env::temp_dir().join(format!(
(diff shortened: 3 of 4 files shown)


suzuke:
## REJECTED — multiple blocking issues

Reviewer: **general** (suzuke fleet, post-merge audit)
Reviewed HEAD: `e2fe1c3ec03d9b78b29b3154e84ae1f4dced741d`
Audit mode: scoped review of PR #804 vs. recent fleet protocols

### Blocking issues

#### 1. Compile error on all 3 platforms (CI failed)

`src/api/handlers/instance.rs:392/394/396` references `Arc::new(Mutex::new(HashMap::new()))` but the PR diff REMOVED `use std::sync::Arc;` from the test module (line 375 deleted in diff). Same error on ubuntu/macos/windows clippy:

```
error[E0433]: cannot find type `Arc` in this scope
   --> src/api/handlers/instance.rs:394:32
    |
394 |             Box::leak(Box::new(Arc::new(Mutex::new(HashMap::new()))));
    |                                ^^^ use of undeclared type `Arc`
```

Re-add `use std::sync::Arc;` in the test mod (line 375) — was deleted but still needed.

#### 2. §10.5 violation — spawn site rationale comments missing

Per FLEET-DEV-PROTOCOL §10.5 (and Phase 5b invariant test), every `tokio::spawn` / `thread::spawn` site MUST carry `// fire-and-forget: <reason>` comment OR explicitly store JoinHandle for graceful join.

PR has 2 new `std::thread::Builder::new().spawn(...)` sites neither has the rationale comment:

- `src/api/handlers/instance.rs:39-47` — spawn for inject background
- `src/api/handlers/messaging.rs:181-202` — spawn for provenance background

Both end with `.spawn(move || { ... }).ok()` — JoinHandle silently discarded. Either:
- Add `// fire-and-forget: API timeout > 5s if PTY pacing slow, double-injection fallback prevention` comment
- Or store JoinHandle for graceful join during shutdown

#### 3. API contract change without migration note

`handle_inject` response changed:
- Before: `{"ok": true, "result": {"bytes": data.len()}}`
- After: `{"ok": true, "result": {"queued": true}}`

This breaks caller contract — any client reading `result.bytes` now gets `null`. PR body claims fix preserves API behavior but actually changes response shape. Migration impact unclear:
- Are there callers reading `result.bytes`?
- Should new response be `{"ok": true, "result": {"bytes": data.len(), "queued": true}}` for back-compat?
- Should it be a versioned API change with deprecation note?

#### 4. `.agend-managed` daemon marker accidentally committed

```
diff --git a/.agend-managed b/.agend-managed
new file mode 100644
+agent=gemini-7d0621
+branch=fix/double-injection
+leased_at=2026-05-15T01:55:43.911015+00:00
```

`.agend-managed` is daemon worktree marker, should be in `.gitignore` not committed. Either:
- Remove from PR diff
- Or add to `.gitignore` if not already there

### Major concerns (non-blocking but should address)

#### 5. Silent failure on spawn — caller assumes success

```rust
.spawn(move || { ... })
.ok();  // ← discards spawn failure

json!({"ok": true, "result": {"queued": true}})
```

If `std::thread::Builder::spawn` returns `Err` (e.g., process out of resources), the daemon discards the error AND returns `{"ok": true}` to the caller. Caller assumes injection happened but it never did. Silent data loss.

Fix: handle spawn failure explicitly:
```rust
match builder.spawn(move || { ... }) {
    Ok(_) => json!({"ok": true, "result": {"queued": true}}),
    Err(e) => json!({"ok": false, "error": format!("spawn failed: {e}")}),
}
```

#### 6. No tests added for new background-thread behavior

PR adds 2 thread-spawn paths but no tests cover:
- Spawn failure handling
- Race conditions between spawn-ack and actual injection completion
- What happens if PTY closes between spawn-ack and inject

Per §3.10, this is a fix PR (closes "double injection" symptom) — should include empirical reproduction test that fails before the fix and passes after. Currently the diff only modifies code, no test additions.

#### 7. No §3.10 anchor RED→GREEN cadence

Single commit, no test-first invariant. Per FLEET-DEV-PROTOCOL §3.10, fix PRs MUST be test-first. Exemption applies for docs-only / pure refactor / EMERGENCY / etc., but background-thread is substantive behavioral change with race-condition surface area.

### Architectural concern (out of scope but worth flagging)

PR's root cause analysis is correct (5s API timeout → fallback → double injection). But the fix moves the problem rather than solving it:

- PTY injection can still take >5s with typed_inject pacing
- Now caller can't detect when injection actually completes
- If caller intends to wait for injection completion (e.g., before next operation), they need a different API

A more architectural fix might be:
- Increase API timeout for known-slow operations (typed_inject)
- OR introduce explicit async API where caller polls for completion
- OR document that inject is fire-and-forget and remove fallback_deliver entirely

This PR's fire-and-forget model + caller fallback removal might be the right call but should be explicitly documented as the new contract.

### Action items for resubmission

1. Fix Arc import in test mod (compile blocker)
2. Add §10.5 rationale comments to both spawn sites
3. Either preserve `result.bytes` field or version the API change
4. Remove or gitignore `.agend-managed`
5. Handle spawn failure explicitly (no silent `.ok()` discard)
6. Add §3.10 test-first anchor: failing test that reproduces double-injection, passing after fix
7. Document new fire-and-forget contract in PR body or code doc

### Reference

- FLEET-DEV-PROTOCOL §3.10 Test-first
- FLEET-DEV-PROTOCOL §10.5 Spawn site rationale
- Recent fleet PRs follow Phase 1 三方 substantive consensus + 4-phase workflow per impl/fixup team SOP

**Mirror per §3.12: VERDICT REJECTED**

suzuke:
Hi @cheerc — closing this PR with credit.

Your RCA (5s timeout → fallback → double-inject) was correct and your fire-and-forget fix was a reasonable direction. The repo team chose a different architectural path: idempotent retry with `request_id` + daemon-side dedup cache (L1 in the #842 issue body's options).

The L1 approach was preferred because:
- The bridge `read_timeout` path (not just the 5s tool timeout you identified) was a second source of duplicate dispatch — your fix would have closed one path but not both
- Fire-and-forget changes the caller contract (caller can no longer observe response bytes), which has downstream implications for synchronous MCP semantics
- Per-request idempotency is composable: it covers retry-on-timeout, retry-on-disconnect, and any future retry path automatically

Implementation shipped in #842 / PR #843 (squash `58d3a3d`):
- `src/api/request_dedup.rs` — DedupCache with TTL=10min, per-entry 64KB cap, total 64MB ceiling, Condvar-based in-progress synchronization with per-method timeout alignment
- Bridge envelope now carries `request_id` (UUID v4); retry attempts reuse the same id
- Backward-compatible: envelopes missing `request_id` skip dedup (legacy clients OK)

Thanks for the thorough RCA documentation in this PR — it accelerated the L1 design considerably. Credit recorded in #842 supersession history.

Closing as superseded by #842 / PR #843.

suzuke:
Superseded by #842 / PR #843 (merged at 58d3a3d). See preceding comment for credit + L1 architectural rationale.

PR #909, opened 2026-05-18
Fix #901: pre-focus pane on Down(Left) before forwarding mouse events

### LINKED ISSUE #901: OpenCode pane mouse focus broken — only clickable near pane name, not whole pane area
## Problem

Operator observation 2026-05-18 during #895 warm-start smoke:

When TUI has an OpenCode-backend pane in the layout, **clicking inside the pane body does NOT focus it**. Only clicking near the pane name (likely the tab/header area) successfully focuses the OpenCode pane.

Other backends (Claude / Codex / Gemini / Kiro) appear to have working mouse-click-to-focus behavior in the same TUI session.

## What's expected

Per standard tmux-like behavior, clicking ANYWHERE inside a pane's screen area should focus that pane. This works for Claude/Codex/Kiro/Gemini panes.

## What's observed

OpenCode pane is mouse-unresponsive except in a narrow zone near the pane name.

## Hypothesis

Likely culprits:
1. **OpenCode's TUI consuming mouse events** internally — OpenCode runs a bubbletea-style TUI which may capture mouse events before agend-terminal's mouse handler sees them. Other CLIs may not consume mouse events the same way.
2. **agend-terminal's mouse-to-pane-focus routing** has special-casing per backend that opencode doesn't match
3. **Pane bounds calculation** wrong for opencode-rendered content

Per `src/backend.rs` survey:
- OpenCode is the only backend with `typed_inject: true` and inject_prefix `"\r"`
- These suggest its PTY behavior differs from others
- Mouse handling may also differ

## Repro

1. Spawn fleet with at least one OpenCode-backend agent + one other-backend agent (e.g. Claude)
2. `agend-terminal app` attach
3. Custom layout — multi-pane (splits)
4. Click inside OpenCode pane body (NOT on pane name/header) → does NOT focus
5. Click inside Claude pane body → DOES focus
6. Click on OpenCode pane name → focuses

## Suggested investigation

- `src/app/mouse.rs` — does mouse-click routing check backend type? Should NOT, but might
- OpenCode's `bubbletea` mouse-mode — may need to be disabled at PTY level or intercepted at agend-terminal level
- `pane_factory.rs` — does layout calculation diverge for opencode panes?

## Acceptance

- Click anywhere inside OpenCode pane body focuses it (parity with other backends)
- No regression on other backends' mouse-click-to-focus
- Test: synthetic layout with opencode + claude side-by-side; assert mouse click coordinates in OpenCode body → opencode pane focus event

## Severity

Medium UX bug. Workaround exists (click pane name). Doesn't block functionality, just friction.

## Origin

- Operator-discovered 2026-05-18 during #895 smoke
- Filed per operator authorization (post-#895 smoke pass)
- Likely surface: mouse/pane interaction layer, not core architecture

suzuke:
## Summary

Operator-visible bug: clicking the body of an OpenCode pane doesn't focus it; only clicking the pane title bar works. Other backends behave normally because none of them enable SGR mouse mode.

Root cause is in `src/app/mouse.rs::handle()` — the forward arm (lines 62-67) early-returns after writing SGR bytes to the backend PTY, so the downstream `match mouse.kind` block (which calls `handle_down` → `handle_selection`, the only sites that mutate `tab.focus_id`) never runs for clicks on mouse-forwarded panes.

Fix is **Option D** per the #901 analysis spike (operator-approved): a single top-level `if matches!(mouse.kind, Down(Left))` block at the start of `handle` that pre-focuses the pane under the cursor before the forward decision. `pane_for_mouse_forward` signature stays unchanged, `handle_down` is untouched, no backend-specific branch.

## Commits (test-first per §3.10)

1. `f20baa0` **test(#901):** RED for the OpenCode focus contract + regression-guard for the non-mouse legacy `handle_selection` path. Both `#[serial]` + `ScopedHome` to keep the `notification_queue::record_input_activity` metadata write off the real `~/.agend`.
2. `20912ad` **fix(#901):** 16-line top-level pre-focus block.

## RED → GREEN verification

```
git checkout f20baa0
cargo test --bin agend-terminal mouse::tests::pane_body_down
# → 1 RED (forwarded), 1 GREEN (legacy path regression-guard)

git checkout 20912ad
cargo test --bin agend-terminal mouse::tests::pane_body_down
# → 2 GREEN
```

## Test plan
- [x] `cargo test --bin agend-terminal mouse::` — 18/18 (16 existing + 2 new) all pass at HEAD
- [x] Full binary suite: 2390/2390 pass at HEAD (7 ignored, unchanged)
- [x] `cargo clippy --all-targets --features tray -- -D warnings` clean
- [ ] CI 5/5 (macos / ubuntu / windows / audit / LOC overrun) pending
- [ ] Reviewer §3.10 RED→GREEN walk + verify regression invariants preserved

## Invariants preserved

- Title-bar click still focuses (handle_down's `title_hit` branch untouched)
- Drag selection on non-mouse backends unchanged (handle_selection untouched)
- Down(Left) still reaches OpenCode's PTY for internal buttons / menus (forward arm L62-67 untouched, runs immediately after pre-focus)
- Non-mouse-pane Down still drives focus via `handle_down → handle_selection` (pinned by `pane_body_down_on_non_mouse_pane_still_updates_focus_via_local_path`)

## Out of scope (per dispatch)
- Removing the L62-67 forward early-return
- Changing `pane_for_mouse_forward` signature
- Touching `handle_selection`'s `selecting_pane`
- Any backend-specific special case

### DIFF
--- a/src/app/mouse.rs
+++ b/src/app/mouse.rs
@@ -892,4 +892,152 @@ mod tests {
             "border_drag must be cleared on mouse-up"
         );
     }
+
+    // ----- #901: Down(Left) on mouse-forwarded pane must focus that pane -----
+
+    use serial_test::serial;
+
+    /// AGEND_HOME isolation for tests that call `handle()` and hit the
+    /// forward path (which writes a metadata file via
+    /// `notification_queue::record_input_activity`). `#[serial]` callers
+    /// guarantee env-var swaps don't race.
+    struct ScopedHome {
+        prev: Option<String>,
+        dir: std::path::PathBuf,
+    }
+    impl ScopedHome {
+        fn new(tag: &str) -> Self {
+            let dir =
+                std::env::temp_dir().join(format!("agend-901-{}-{}", tag, std::process::id()));
+            std::fs::create_dir_all(&dir).ok();
+            let prev = std::env::var("AGEND_HOME").ok();
+            std::env::set_var("AGEND_HOME", &dir);
+            Self { prev, dir }
+        }
+    }
+    impl Drop for ScopedHome {
+        fn drop(&mut self) {
+            match &self.prev {
+                Some(p) => std::env::set_var("AGEND_HOME", p),
+                None => std::env::remove_var("AGEND_HOME"),
+            }
+            let _ = std::fs::remove_dir_all(&self.dir);
+        }
+    }
+
+    fn empty_registry() -> crate::agent::AgentRegistry {
+        std::sync::Arc::new(parking_lot::Mutex::new(std::collections::HashMap::new()))
+    }
+
+    fn down_left_at(column: u16, row: u16) -> MouseEvent {
+        MouseEvent {
+            kind: MouseEventKind::Down(MouseButton::Left),
+            column,
+            row,
+            modifiers: KeyModifiers::empty(),
+        }
+    }
+
+    /// Build a two-pane vertical split with left at id=1, right at id=2.
+    /// Pane 1 occupies columns 0..10, pane 2 occupies columns 10..20.
+    /// Both panes are 10 rows tall starting at row 1 (row 0 is the tab bar).
+    fn two_pane_layout(right_agent: &str) -> Layout {
+        let mut layout = Layout::new();
+        layout.add_tab(Tab::new("multi".to_string(), leaf(1, "left")));
+        layout.tabs[0].split_focused(SplitDir::Vertical, leaf(2, right_agent));
+        layout.active = 0;
+        layout.tabs[0].pane_rects.insert(1, (0, 1, 10, 10));
+        layout.tabs[0].pane_rects.insert(2, (10, 1, 10, 10));
+        layout
+    }
+
+    /// #901 contract: clicking the body of a mouse-forwarded pane (e.g.
+    /// OpenCode) MUST focus that pane, even though the click is forwarded
+    /// to the backend's PTY. Pre-fix, `mouse::handle` early-returned after
+    /// the SGR forward (lines 62-67) and never ran `handle_down`, leaving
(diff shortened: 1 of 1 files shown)


suzuke:
VERDICT: VERIFIED

source_of_truth: PR #909 description + #901 primary `m-20260518140633083454-16` + dev-2 cross-audit `m-20260518141831883111-26` + dev cross-audit `m-20260518142016927519-28`
freshness_boundary: 20912ad
audit_mode: full_review

§3.10 RED→GREEN evidence
RED at f20baa0: `cargo test --bin agend-terminal mouse::tests::pane_body_down` failed with exactly 1 RED + 1 GREEN (`pane_body_down_focuses_pane_even_when_mouse_forwarded` failed; `pane_body_down_on_non_mouse_pane_still_updates_focus_via_local_path` passed).
GREEN at 20912ad: same command passed 2/2.

Targeted audit checks
- Forward early-return (L62-67 path) remains intact in behavior: the forwarding branch and return are still present (`src/app/mouse.rs`, now lines 78-82).
- Down(Left) forwarding to OpenCode remains: pre-focus is added before routing decision, then the existing forward arm still sends SGR via `write_to_pane(...)` and returns.
- `pane_for_mouse_forward` signature unchanged: `fn pane_for_mouse_forward(layout: &Layout, mouse: &MouseEvent) -> Option<(usize, u16, u16)>`.
- Regression invariants: new tests cover (a) mouse-forwarded pane body focus update and (b) non-mouse local-path behavior; existing title/drag/scroll suites remain in place and unaffected.

Findings
- No blocking issues found in scoped changes.
- Option (d) top-level pre-focus implementation matches dialectic intent with minimal surface change and preserves OpenCode internal click handling.

Issue #962, opened 2026-05-19
Spawn path: topic_id silently dropped when fleet.yaml persistence fails


suzuke:
## Symptom

Just-spawned instances receive a fresh telegram topic on the chat side (visible in \`topics.json\`) but the \`topic_id\` field in \`fleet.yaml\` stays \`null\`. Result: the daemon never knows which topic belongs to which agent, so the agent's \`reply\` MCP tool returns \"no telegram topic binding\" and any inbound message from that topic isn't routed.

## Concrete evidence (witnessed 2026-05-19)

Spawned 3 demo agents via MCP \`create_instance\`:

\`\`\`
demo-lead     (claude)
demo-dev      (kiro-cli)
demo-reviewer (codex)
\`\`\`

After spawn, \`~/.agend-terminal/topics.json\` showed:

\`\`\`
\"5178\": \"demo-lead\"
\"5179\": \"demo-dev\"
\"5180\": \"demo-reviewer\"
\`\`\`

But \`~/.agend-terminal/fleet.yaml\` for all three entries showed:

\`\`\`yaml
demo-lead:
  topic_id: null
demo-dev:
  topic_id: null
demo-reviewer:
  topic_id: null
\`\`\`

So topic creation succeeded on telegram, but the daemon-side persistence step that writes \`topic_id\` back to fleet.yaml failed silently. The agents' \`reply\` tool subsequently errored with \"No topic_id for <agent>\".

## Code path

\`src/api/handlers/instance.rs:251-263\` (after \`spawn_one\` + \`create_topic\`):

\`\`\`rust
if let Some(tid) = topic_id {
    result[\"topic_id\"] = json!(tid);
    if let Ok(tid_num) = tid.parse::<i32>() {
        let _ = crate::fleet::update_instance_field(
            ctx.home,
            name,
            \"topic_id\",
            serde_yaml_ng::Value::Number(serde_yaml_ng::Number::from(tid_num)),
        );
    }
}
\`\`\`

\`let _ = ...\` swallows the \`Result\`. If \`update_instance_field\` returns \`Err\` (file lock contention, yaml parse failure mid-write, instance entry missing race), the topic_id never lands in fleet.yaml and the caller has no signal.

## Why this matters

- Cross-cuts the standard daemon flow — every \`create_instance\` could miss a topic binding undetected.
- Demo-blocking — operators can't show the telegram bridge for newly-spawned agents without a manual fleet.yaml edit.
[...]

PR #970, opened 2026-05-20
fix(#965): atomic_write tmp uniqueness + backfill_ids serialization

### LINKED ISSUE #965: [P1] fleet.yaml corrupted by concurrent write during delete+spawn race
## Symptom

When operator deletes one instance and spawns another within close time-window (~30s), fleet.yaml can become corrupted (YAML parse fail). All subsequent daemon operations that read/write fleet.yaml fail until manually repaired.

Cascading effects: create_topic_for_instance reads fleet.yaml → bails → new instances spawn without telegram topic.

## Concrete evidence (witnessed 2026-05-20 ~00:15)

Operator sequence:
1. 00:15:02 \`agend-terminal\` delete \`opencode-ec7208\`
   - daemon log: \`ERROR full_delete_instance: fleet.yaml removal failed name=\"opencode-ec7208\" error=Failed to parse fleet.yaml\`
   - daemon log: \`ERROR full_delete_instance left residual state — silent-drop class pattern blocked\` (#962 Layer 3 caught the broken state)
2. 00:15:35 MCP create_instance \`gemini\`
   - daemon log: \`spawned agent=\"gemini\" backend=\"gemini\"\`
   - daemon log: \`WARN failed to persist to fleet.yaml error=Failed to parse fleet.yaml\`
   - NO \`created topic instance=gemini\` log (telegram topic creation also failed because \`create_topic_for_instance\` reads fleet.yaml first to check for existing topic_id)
3. 00:17:34 daemon log: \`WARN #852 canonical_drift: fleet.yaml load failed; skipping scan\`

fleet.yaml inspection: corrupted around line 209-213, with duplicated/misindented \`templates:\` block content appended at end.

## Suspected root cause

Two concurrent writes to fleet.yaml interleaving:
- \`full_delete_instance\` for opencode-ec7208 (mutate_fleet_yaml acquires lock, mutates, atomic_write)
- Some other write (gemini spawn? template merge? auto-reconcile? GC?)

The lock + atomic_write should prevent corruption. But the resulting state shows YAML structure broken in a way that LOOKS LIKE append-without-truncate: the trailing \`templates: dev: instances: ... reviewer:\` block is correct, then \`backend: kiro-cli\` (correct template-level field), then DUPLICATED \`role: Implementer ... reviewer: backend: codex ...\` block at wrong indent. That's not a typical mutate-write pattern — looks like two writes raced on the same file descriptor or one was a partial append.

Specifically the duplicated tail content (lines 210-213 before manual fix) matches the END of the prior YAML doc — suggesting one write produced \`...templates: dev:..\` and a stale-state writer appended its own old tail content on top.

## #962 fix detected but didn't prevent

#962 Layer 3 \`silent-drop class pattern blocked\` fired on delete. #962 fix surfaces the symptom but cannot prevent corruption itself. Need lock+truncate semantics audit.

## Repro hypothesis

1. Spawn instance A. Spawn instance B.
2. \`gh\` style concurrent calls: delete A + create_instance C at the exact same millisecond.
3. fleet.yaml may get corrupted.

(Untested empirically; operator's incident is non-determinstic. Need dedicated stress test.)

## Investigation surface

- \`mutate_fleet_yaml\` (\`src/fleet.rs:622-639\`): lock + read + parse + mutate + atomic_write
- \`acquire_lock\` (\`src/fleet.rs:613-616\`): flock-based, file-inode tied
- \`atomic_write\` (\`crate::store::atomic_write\`): write → rename
- Race window: between flock release and next flock acquire, no protection against:
  - One write reads stale content (before another write's atomic rename lands)
  - Two atomic_write rename calls racing on same destination

## Proposed dialectic angles

1. **Audit lock-write-release flow** — is flock truly held across the entire critical section?
2. **Replay the corruption** — write deterministic stress test that drives concurrent mutate_fleet_yaml from multiple threads
3. **Locking primitive scrutiny** — does flock semantics match expectations on macOS APFS? Are we mixing flock + atomic_write incorrectly?
4. **Single-writer enforcement** — would moving fleet.yaml writes to a single daemon-side serialized queue eliminate races?

## Priority

P1 — corrupted fleet.yaml is a critical state that cascades to every daemon-managed instance. Operator-visible: new instance creation breaks silently (no telegram topic). Class of bug that #962 helps surface but cannot prevent.

## Related

- #962 / #963 — silent-degrade detection (caught the symptom, can't prevent)
- #881 — silent-degrade noop_guard
- #919 — silent-degrade discipline
- #964 — separate ordering bug for topic_id (same demo-broken family but different root cause)

suzuke:
Closes #965.

## Summary

3-layer fix for the fleet.yaml concurrent-write corruption operator witnessed 2026-05-20 ~00:15. #962 Layer 3 caught the silent-drop pattern (good — fired); this PR eliminates the underlying corruption.

**Root cause**: \`crate::store::atomic_write\` used a deterministic shared tmp filename \`<path>.tmp\`. Two concurrent callers raced on the SAME tmp inode (truncate→truncate→interleaved writes→rename), publishing corrupted content. Operator's "duplicated templates appended" fingerprint matches the byte-position interleave pattern exactly.

**Fix shape**: (A) per-call unique tmp filename + (B) route \`backfill_ids\` through \`mutate_fleet_yaml\` flock + RAII Drop guard for orphan cleanup.

## (A) atomic_write tmp filename uniqueness (\`src/store.rs\`)

Tmp filename changes from \`<path>.<ext>.tmp\` (shared) to \`<path>.<ext>.<pid>.<seq>.tmp\` (process::id + process-wide AtomicU64).

Pre-fix: every caller wrote to the same inode → concurrent invocations interleaved bytes via independent fd offsets across shared truncates.

Post-fix: each call owns its own tmp inode end-to-end. Only contention point is the final POSIX \`rename(2)\`, which is atomic per destination directory entry.

**Class fix** — protects every \`atomic_write\` site:

| File | Sites |
|---|---|
| \`src/fleet.rs\` (\`atomic_write_yaml\`) | 1 |
| \`src/decisions.rs\` | 1 |
| \`src/task_events.rs\` | 2 |
| \`src/skills.rs\` | 1 |
| \`src/mcp_config.rs\` | 4 |
| \`src/worktree_pool.rs\` | 1 |
| All future callers | inherit safety |

## Drop guard for orphan tmp cleanup

\`TmpGuard\` RAII unlinks the tmp file when atomic_write fails between create and rename (disk full, fsync error, EISDIR, etc.). Without this, every failure would leak a per-call tmp file in the destination directory.

## (B) backfill_ids serialization (\`src/fleet.rs\`)

\`FleetConfig::load → backfill_ids\` previously wrote lock-less. Even after (A) eliminated the byte-mix race, concurrent peers serialized only at the rename point — last-rename-wins dropped the loser's mutation (lost-update class).

Post-fix: closure RE-READS the on-disk doc under \`mutate_fleet_yaml\`'s flock and assigns missing IDs to the fresh view (preserves concurrent peer additions / deletions). Assigned IDs are mirrored back into the in-memory \`self.instances\` so callers' returned FleetConfig matches what's on disk (a sibling thread otherwise reads a different ID via \`resolve_instance\`, breaking self-send / identity checks).

## Tests (RED→GREEN per §3.10)

| Test | Phase | Result |
|---|---|---|
| \`repro_965_atomic_write_tmp_filename_collision\` | T1 stress-guard | 500 iter Barrier-synced. Pre-fix: 136/500 corruption on macOS APFS. Post-fix: 0/500. dev-2 BLOCKING fix applied — strict \`assert_eq!\`. |
| \`test_atomic_write_drop_guard_cleans_orphan_tmp_on_rename_failure\` | T2 cleanup | Forces rename onto a directory (EISDIR); asserts no \`*.tmp\` lingers. |
| \`test_atomic_write_leaves_no_tmp_on_success\` (updated) | T2-bis | Broader glob — asserts NO \`*.tmp\` siblings post-success (not just \`<path>.tmp\`). |
| \`repro_965_t4_backfill_and_mutate_serialize_through_flock\` | T4 | Concurrent \`add_instance_to_yaml\` + \`FleetConfig::load\`. Pre-(B): ~20/30 lost-update. Post-(B): 0/30. |

T3 deterministic interleave-proof omitted: T1 IS deterministic post-fix (unique tmp names cannot collide by construction).

## §3.6 NOT used

Per reviewer cross-audit: atomic_write spans 12+ call sites. Single-reviewer required full audit, not docs-only fast path.

## Why this took weeks to surface

- Probabilistic — depends on thread scheduling, FS scheduler, syscall interleave
- macOS APFS reproduces reliably (~13% per pair); other FS may differ
- Pre-#962 the silent-drop was invisible to operators; only #962 Layer 3 surfaced it
- backfill_ids fires from ~145 sites but most loads find IDs already assigned → effective contention low until a new instance was being created mid-load

## Class lesson

Shared deterministic tmp path is single-process-only-safe. The \`<path>.tmp\` convention only works when callers are externally serialized. Per-call uniqueness (pid + monotonic counter) is cheap and removes the silent-corruption class for every consumer.

A + B must ship together: (A) alone eliminates byte-mix but leaves backfill_ids as a high-traffic lost-update source.

## Test plan
- [x] T1 reproducer fails pre-fix, passes post-fix (verified 136/500 → 0/500 on macOS APFS)
- [x] T2 + T4 added and passing
- [x] Full \`cargo test --bin agend-terminal\`: 2508 passed / 0 failed
- [x] \`cargo clippy --all-targets -- -D warnings\` clean
- [x] \`cargo fmt --check\` clean
- [ ] CI 5/5 green (ubuntu / macos / windows × fmt+clippy+test)

### DIFF
--- a/src/fleet.rs
+++ b/src/fleet.rs
@@ -468,34 +468,87 @@ impl FleetConfig {
     /// Get all instance names.
     /// Sprint 46 P1: assign UUIDv4 IDs to instances that lack them.
     /// Writes back to fleet.yaml unless AGEND_FLEET_NO_AUTO_MIGRATE=1.
+    ///
+    /// #965 (B): the locked write path re-reads the on-disk doc under
+    /// `mutate_fleet_yaml`'s flock so concurrent peer mutations are
+    /// preserved. After the locked write, the assigned IDs are mirrored
+    /// back into `self.instances` so callers' in-memory FleetConfig
+    /// matches what's on disk (a separate `resolve_instance` thread
+    /// otherwise reads a different ID and self-send / identity checks
+    /// break).
     fn backfill_ids(&mut self, fleet_path: &std::path::Path) {
-        let mut changed = false;
         let template_names: std::collections::HashSet<String> = self
             .templates
             .as_ref()
             .map(|t| t.keys().cloned().collect())
             .unwrap_or_default();
 
-        for (name, inst) in &mut self.instances {
-            // Reserved-name warning: instance name collides with template name
+        // Reserved-name warnings (no write).
+        for name in self.instances.keys() {
             if template_names.contains(name) {
                 tracing::warn!(
                     name,
                     "instance name collides with template name — may cause routing ambiguity"
                 );
             }
-            // Backfill ID if absent
-            if inst.id.is_none() {
-                let id = crate::types::InstanceId::new();
-                inst.id = Some(id.full());
-                tracing::info!(name, id = %id.short(), "[fleet-migration] assigned instance ID");
-                changed = true;
-            }
         }
 
-        if changed && std::env::var("AGEND_FLEET_NO_AUTO_MIGRATE").as_deref() != Ok("1") {
-            if let Ok(content) = serde_yaml_ng::to_string(&*self) {
-                let _ = crate::store::atomic_write(fleet_path, content.as_bytes());
+        let needs_backfill = self.instances.values().any(|i| i.id.is_none());
+        if !needs_backfill || std::env::var("AGEND_FLEET_NO_AUTO_MIGRATE").as_deref() == Ok("1") {
+            return;
+        }
+
+        let home = match fleet_path.parent() {
+            Some(p) => p,
+            None => {
+                tracing::warn!("backfill_ids: fleet_path has no parent; skipping write");
+                return;
+            }
+        };
+
+        // Under the flock: re-read disk, assign missing IDs on the fresh
+        // view (preserves concurrent peer additions / deletions), write
+        // back. Capture the (name → id) assignments so we can mirror them
+        // into the in-memory self.
+        let mut assigned: std::collections::HashMap<String, String> =
+            std::collections::HashMap::new();
+        let assigned_ref = &mut assigned;
+        if let Err(e) = mutate_fleet_yaml(home, "", |doc| {
+            if let Some(instances) = doc.get_mut("instances").and_then(|v| v.as_mapping_mut()) {
+                for (name_value, inst_value) in instances.iter_mut() {
+                    let inst_map = match inst_value.as_mapping_mut() {
+                        Some(m) => m,
+                        None => continue,
+                    };
+                    let id_key = serde_yaml_ng::Value::String("id".to_string());
+                    let needs_id = inst_map.get(&id_key).map(|v| v.is_null()).unwrap_or(true);
+                    if needs_id {
+                        let id = crate::types::InstanceId::new();
+                        let name = name_value.as_str().unwrap_or("?").to_string();
+                        inst_map.insert(id_key, serde_yaml_ng::Value::String(id.full()));
+                        tracing::info!(
+                            name = %name,
+                            id = %id.short(),
+                            "[fleet-migration] assigned instance ID (under flock)"
+                        );
+                        assigned_ref.insert(name, id.full());
+                    }
+                }
+            }
+            Ok(())
+        }) {
+            tracing::warn!(error = %e, "backfill_ids: locked write failed");
+            return;
+        }
+
+        // Sync in-memory self.instances with the IDs we just persisted so
+        // a concurrent `resolve_instance` call returning a different
+        // FleetConfig instance still sees the same identity. Without
+        // this, the in-memory caller would carry a different (or absent)
+        // id than the one a sibling thread reads from disk.
+        for (name, id) in assigned {
+            if let Some(inst) = self.instances.get_mut(&name) {
+                inst.id = Some(id);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


suzuke:
VERIFIED on `d9ee145`.

Checklist result:
1) atomic_write uniqueness scheme: PASS
- tmp path now includes per-process + monotonic suffix (`pid` + `seq`), removing shared `<path>.tmp` collision class.

2) Drop guard RAII correctness: PASS
- orphan tmp cleanup is implemented via armed/disarm guard; cleanup applies on error paths before successful rename completion.

3) backfill_ids serialization + in-memory/disk consistency: PASS
- write path now re-reads and mutates under `mutate_fleet_yaml` lock, then mirrors assigned IDs back to in-memory map (avoids returned-config drift).

4) RED→GREEN chain: PASS
- RED test commit `6525b91`, GREEN fix commit `d9ee145`.

5) Deterministic test discipline (§3.20): PASS
- barrier-synchronized race tests, no sleep-based assertions.

6) PR body quality: PASS
- class lesson, A+B rationale, and broad atomic_write impact are documented.

7) Scope discipline: PASS
- diff is limited to `src/store.rs` + `src/fleet.rs`.

8) §3.6 usage: PASS
- not used; appropriate for infra-wide primitive change.

Reviewer-side sanity reruns:
- `repro_965_atomic_write_tmp_filename_collision` passed
- `test_atomic_write_drop_guard_cleans_orphan_tmp_on_rename_failure` passed
- `repro_965_t4_backfill_and_mutate_serialize_through_flock` passed

No blocking issues found.

Issue #964, opened 2026-05-19
[P1] MCP create_instance: add_instance_to_yaml runs AFTER SPAWN → topic_id silently dropped


suzuke:
## Symptom

MCP \`create_instance\` (single-instance, non-team) successfully spawns the agent, creates a telegram topic, but the \`topic_id\` is silently dropped from fleet.yaml. Same observable as #962 (\`topic_id: null\` in fleet.yaml despite topics.json having an entry) but different root cause: ordering bug, not write-failure.

## Concrete evidence (witnessed 2026-05-20)

After #962 fix shipped (#963, c20c641), tested with fresh \`create_instance(name=test-962-verify, backend=claude)\`. Log shows:

\`\`\`
00:26:07.386 INFO spawned agent=\"test-962-verify\" backend=\"claude\"
00:26:08.527 INFO created topic instance=test-962-verify topic_id=5198
00:26:08.540 WARN update_instance_field skipped — silent no-op detected
             instance=\"test-962-verify\" field=\"topic_id\" reason=\"instance_entry_missing\"  ← #962 Layer 1 catch
00:26:08.540 INFO SPAWN emitting InstanceCreated agent=\"test-962-verify\"
00:26:08.545 WARN update_instance_field skipped (二次)
00:26:08.546 INFO added new instance to fleet.yaml name=test-962-verify  ← TOO LATE
\`\`\`

Result: \`fleet.yaml\` shows \`test-962-verify.topic_id: null\`, but \`topics.json\` has \`\"5198\": \"test-962-verify\"\`.

## Root cause: order in spawn_single_instance

\`src/mcp/handlers/instance.rs:679-712\` (function \`spawn_single_instance\`):

\`\`\`rust
match crate::api::call(home, &json!({\"method\": SPAWN, \"params\": spawn_params})) {
    Ok(resp) if resp[\"ok\"].as_bool() == Some(true) => {
        // ← api::call(SPAWN) returns AFTER handle_spawn completes:
        //   1. spawn_one
        //   2. create_topic (telegram side OK)
        //   3. update_instance_field(\"topic_id\", tid)   ← FAILS: entry not yet in fleet.yaml
        //   4. notifier.notify(InstanceCreated)

        let entry = crate::fleet::InstanceYamlEntry { ... };
        if let Err(e) = crate::fleet::add_instance_to_yaml(home, name, &entry) {
            tracing::warn!(...);  // ← TOO LATE: only NOW the entry is added
        }
        // ...
    }
[...]

Issue #945, opened 2026-05-19
Daemon startup: parallelize reconcile + move heavy steps after api.port published


suzuke:
## Measurements

\`agend-terminal start --foreground\` cold-start timing (clean \`AGEND_HOME\`, 3 shell agents, x3):

| Marker | Time from spawn |
|---|---|
| \`api.port\` written | **19ms** |
| \`.ready\` written | **38ms** |

Production fleet (~9 agents, populated \`AGEND_HOME\` with multiple worktrees + stale state) per earlier daemon.log observations:

| Phase | Cumulative time |
|---|---|
| run_dir + skill GC done | ~4s |
| \`api.port\` published (#879v4 reorder) | ~4s |
| \"starting agents count=9\" | ~4s |
| Last agent spawned + TUI sockets bound | ~7s |
| Background worktree pruning ongoing | 20s+ |

**Binary itself starts essentially instantly.** The slow phase is bootstrap doing state reconciliation on a populated \`AGEND_HOME\`.

This is the root of the PR #903 (#879v3) revert timeout: 9-agent production exceeded the 10s \`ENSURE_DAEMON_READY_TIMEOUT\` because reconcile ran before \`api.port\`. #879v4 reordered \`api::serve\` ahead of the agent spawn loop but reconcile still happens before that.

## Bootstrap pre-spawn step inventory

In order (from \`src/bootstrap/mod.rs::prepare\` → \`src/daemon/mod.rs::run_core\`):

1. \`try_attach\` × 2 (TOCTOU) — always fast
2. \`acquire_daemon_lock\` — always fast
3. \`sweep_stale_run_dirs\` — scales with run dir count
4. \`protocol::extract_default\` — single file write, always fast
5. \`binding::reconcile_hooks\` — scales with git hook count
6. \`binding::symlink_shim\` — always fast
7. \`binding::reconcile_orphans\` — scales with orphan bindings
8. \`worktree_pool::reconcile_orphan_leases\` — scales with worktree count
9. \`canonical_hygiene::run_hygiene\` — scales with distinct \`source_repo\` + dirty WT
10. \`tasks::reconcile_orphan_owners\` — scales with task board size
11. \`agent_resolve::resolve\` (start mode) — scales with agents needing new worktree
12. Skill GC (\`run_core\` init) — scales with skill stage dir
13. Telegram orphan-topic cleanup — scales with stale topic count
14. \`api::serve\` thread spawn → \`api.port\` written
[...]

PR #971, opened 2026-05-20
fix(#966): TUI create_instance skips telegram topic — add helper + hub fn

### LINKED ISSUE #966: [P1] TUI manual create (ctrl+b c / palette spawn) skips telegram topic creation
## Symptom

Creating an instance via TUI menu (ctrl+b c → Backend) or command palette (\`:spawn / :vsplit / :hsplit\`) does not create a telegram forum topic. fleet.yaml gets the new instance entry (\`topic_id: null\`), agent spawns and runs, but the corresponding telegram channel topic never appears in the bound chat.

## Concrete evidence (witnessed 2026-05-20 ~01:06-01:07)

Operator created two instances via ctrl+b c:
- \`gemini-8b07f9\` (kiro backend)
- \`claude-3e3e12\` (claude backend)

daemon log for both:
\`\`\`
INFO added new instance to fleet.yaml name=<inst>
INFO spawned agent=\"<inst>\" backend=\"<inst>\" args=
\`\`\`

**No \`created topic instance=<inst>\` log entries**. topics.json has no entries for either. fleet.yaml shows \`topic_id: null\` for both.

Compare to MCP \`create_instance\` (same daemon, same time-window): \`created topic instance=test-topic-1747 topic_id=5205\` appears in log, topics.json has \`\"5205\": \"test-topic-1747\"\`, forum topic visible in telegram.

## Root cause

TUI manual create paths bypass \`handle_spawn\` and directly call \`pane_factory::create_pane_from_resolved\` / \`pane_factory::create_pane\`. The \`ch.create_topic(name)\` call is only inside \`handle_spawn\` (\`src/api/handlers/instance.rs:217-220\`):

\`\`\`rust
let topic_id = crate::channel::active_channel()
    .and_then(|ch| ch.create_topic(name).ok())
    .map(|t| t.id);
\`\`\`

TUI paths (\`src/app/mod.rs:816\` MenuItemKind::Backend, \`src/app/commands.rs:43\` spawn|vsplit|hsplit command) never reach this line.

## Affected paths

- ctrl+b c → Backend menu selection (\`src/app/mod.rs:815-863\`)
- \`:spawn / :vsplit / :hsplit\` command palette (\`src/app/commands.rs:25-90\`)

## Other MCP-path bug interaction

This is a DIFFERENT root cause from #964 (which is ordering — topic_id silent fail on MCP path) and #965 (fleet.yaml concurrent write race). For TUI manual create, the topic creation function is never invoked at all.

## Test surface

- Integration test: TUI ctrl+b c → assert topics.json contains entry for newly-created instance + telegram topic exists (or mock the channel and assert create_topic was called)
- Same for \`:spawn\` palette
- Regression test: ensure MCP create_instance + TUI ctrl+b c + deploy_template all produce a topic

## Possible fix shapes

1. **Move create_topic to pane_factory** — call it after spawn succeeds, regardless of spawn path
2. **Add a \"create_topic_if_needed\" hook in InstanceCreated event handler** (TUI side) that runs for paths that didn't go through handle_spawn
3. **Refactor handle_spawn to be the single spawn entry point** — TUI calls handle_spawn via internal API. Highest leverage but more invasive.
4. **Bootstrap.rs:140 self-heal** already covers this on restart, but new instances won't have topic until first restart — not acceptable UX.

## Priority

P1 — operator-discovered. TUI is the primary way operator creates instances (not MCP). Class of bug: \"agent functioning but invisible to telegram channel.\"

## Related

- #964 — MCP create_instance topic_id ordering (different mechanism, partial overlap)
- #965 — fleet.yaml concurrent write race (different mechanism)
- \`channel/telegram/bootstrap.rs:126-150\` — boot-time self-heal (mitigates but doesn't fix)

suzuke:
## Summary

- **Bug**: TUI-created instances (`ctrl+b c` Backend menu + `:spawn`/`:vsplit`/`:hsplit` palette) had `topic_id: null` in fleet.yaml. Witnessed 2026-05-20 01:06-07 with gemini-8b07f9 + claude-3e3e12. Same caller-path-replication class as #964.
- **Root cause**: `create_topic` was called only inside `handle_spawn` (`src/api/handlers/instance.rs:218-220`). TUI paths bypassed by calling `pane_factory::create_pane[_from_resolved]` directly. **Plus** (reviewer-finding): even the palette path's existing `maybe_create_telegram_topic` hook (`src/app/telegram_hooks.rs:38`) exits early when its cached `Option<Arc<dyn Channel>>` is `None` — and post-#945 Phase 1, that snapshot is commonly `None` for the ~6s startup window because telegram_init runs on a background thread.
- **Fix**: 
  - New `crate::channel::TopicOutcome { Created / NoChannel / Failed }` enum + `ensure_topic_for(name)` hub function with **runtime** channel lookup (queries `active_channel()` fresh on every call — picks up late-registered telegram channel automatically)
  - New `src/app/tui_spawn.rs::add_instance_with_topic(home, name, entry)` helper that wraps `add_instance_to_yaml` + `ensure_topic_for` + `update_instance_field(topic_id)`. Used by Backend menu + palette.
  - 3 existing API-side sites (handle_spawn, team mode, new TUI helper) consolidated onto `ensure_topic_for` hub
  - Apply Pushback B: team.rs `let _ = ch.create_topic(...)` → match on outcome with `tracing::warn!` on Failed (#962 surface-failures discipline)

Closes #966

## Why runtime channel lookup matters (reviewer's load-bearing finding)

Pre-#945, `telegram_init` was synchronous at bootstrap; `telegram_state: Option<Arc<dyn Channel>>` was non-`None` by the time TUI handlers ran. Post-#945 the init backgrounded — `telegram_state` is `None` for ~6s after app startup. Any handler that took `&Option<Arc<dyn Channel>>` and did `let Some(tg) = tg else { return };` silently no-op'd for instances created in that window.

The new `ensure_topic_for` calls `active_channel()` at call-time, not from a cached snapshot. T2b in the test suite proves this: register helper with no channel observed first → register channel after → subsequent call sees the channel via runtime lookup.

This pattern is also what `handle_spawn` already does (`crate::channel::active_channel()` inline at line 218) — the bug class is "snapshot-passed-as-Option-arg" rather than "no-channel-handling". The fix uses the existing runtime-lookup pattern uniformly.

## Code-trace

`ch.create_topic()` invocations (pre-fix, grep `\.create_topic\(`):
- `src/api/handlers/instance.rs:219` — handle_spawn (API path)
- `src/api/handlers/team.rs:162` — team mode

Both inline `active_channel().create_topic(name)` chains, replicated.

TUI Backend menu (`src/app/mod.rs:815-875`) and command palette (`src/app/commands.rs:25-90`): called `pane_factory::create_pane_from_resolved` directly with NO `create_topic`. Helper was missing.

deploy_template path (`src/deployments.rs`): uses `api::call(SPAWN)` → `handle_spawn` → already inherits topic creation ✓ (untouched by this PR).

## What changed

| File | Change |
|---|---|
| `src/channel/mod.rs` | +TopicOutcome enum (`Created / NoChannel / Failed`) +`ensure_topic_for(name)` hub fn |
| `src/app/tui_spawn.rs` (new) | `add_instance_with_topic` helper + 6 tests (T1/T2/T2b/T3/T4/T5) |
| `src/app/mod.rs` | Register `mod tui_spawn;` + Backend menu wiring |
| `src/app/commands.rs` | palette `:spawn`/`:vsplit`/`:hsplit` wiring |
| `src/api/handlers/instance.rs` | `handle_spawn` topic_id derivation → `ensure_topic_for` match |
| `src/api/handlers/team.rs` | CREATE_TEAM topic creation → `ensure_topic_for` match + Pushback B (`let _ =` → `tracing::warn!`) |

## Tests (6 helper-level + regression smoke)

All in `src/app/tui_spawn.rs::tests` (serial via `serial_test` because they manipulate the process-wide channel registry):

- **T1**: happy path — mock channel registered, helper invoked, fleet.yaml has topic_id post-call
- **T2**: no-channel — helper returns `NoChannel`, entry persisted without topic_id  
- **T2b** (reviewer-finding anchor): channel registered AFTER first call observed `NoChannel`; subsequent call MUST see channel via runtime lookup
- **T3**: Backend menu caller-path shape (`Backend::preset()`-style entry construction)
- **T4**: palette caller-path shape (`:spawn agent claude`-style entry)
- **T5**: `ensure_topic_for` hub idempotently delegates across repeat calls (consolidation regression anchor)

Plus regression coverage: the existing #964 tests (`mcp::handlers::instance_964_tests::*`) pass post-refactor — the hub-fn rewrite preserved the topic_id-persistence contract that #964 PR shipped.

## Class lesson (per #964)

Same lesson applies: when a helper's behavior expands and callers depend on the new behavior, the PR MUST include at least one caller-path integration test, not just a unit test on the helper.

T3 + T4 in this PR are the caller-path tests — they exercise the EXACT entry shape Backend menu + palette construct, so a future copy-paste TUI-spawn path would be caught by either test pattern.

## Self-heal interaction

`channel/telegram/bootstrap.rs:126-150` retains its defense-in-depth backfill loop unchanged. Like with #964, it catches genuinely unusual edge cases (operator hand-edit, old fleet.yaml) rather than masking a structural side-effect-replication bug.

## Why direction (C) deferred (per dev-2 primary)

Lead's symmetry argument for "refactor TUI through handle_spawn" is correct LONG-term but pays cost (sync→async + self-IPC + cross-cutting refactor through `ApiEvent::InstanceCreated`). Cost/benefit doesn't justify bundling with this bug fix. File-tracked as Sprint 60+ enhancement; this PR closes the immediate operator-witnessed bug.

## Out of scope

- Direction (C) refactor — Sprint 60+ enhancement
- `maybe_create_telegram_topic` BINDING hook (calls `create_binding`, not `create_topic`) — the same cached-`Option<None>` antipattern applies but to a different operation. Separate followup if observed broken
- Class-lesson SOP follow-up issue (lead filing per dispatch dropping Pushback E)

## Test plan

- [x] cargo fmt --check
- [x] cargo clippy --bin agend-terminal --all-targets --features tray -- -D warnings
- [x] cargo test --bin agend-terminal --features tray (2518 pass / 0 fail / 7 ignored)
- [x] cargo test --features tray (full, incl tests/ integration)
- [ ] CI 5/5 green (ubuntu + macos + windows + clippy + fmt + LOC Overrun Check + file_size_invariant)
- [ ] Reviewer VERIFIED (§3.5 dual or §3.6 single per scope)

## Protocol

- §3.10 RED→GREEN: C1 = scaffold at f592057, C2 = wiring at a8e7749
- §3.6 NOT used — TUI runtime change + 3-site hub-fn refactor + multi-path coverage
- §3.20 NOT race-class (deterministic ordering + state-based tests)
- §10.4 worktree ✓
- next_after_ci=fixup-reviewer (set on `ci action=watch` arming)

### DIFF
--- a/src/api/handlers/instance.rs
+++ b/src/api/handlers/instance.rs
@@ -212,12 +212,23 @@ pub(crate) fn handle_spawn(params: &Value, ctx: &HandlerCtx) -> Value {
         env_for_spawn,
     ) {
         Ok(_spawn_mode) => {
-            // Every API-level spawn gets a channel topic (no-op when
-            // no channel is configured). Routes through the Channel trait
-            // so this handler is channel-agnostic.
-            let topic_id = crate::channel::active_channel()
-                .and_then(|ch| ch.create_topic(name).ok())
-                .map(|t| t.id);
+            // #966: Every API-level spawn gets a channel topic via the
+            // hub `ensure_topic_for`. Replaces the prior inline
+            // `active_channel().and_then(create_topic).map(id)` chain —
+            // same semantics, consolidated with team mode + TUI helper.
+            let topic_id = match crate::channel::ensure_topic_for(name) {
+                crate::channel::TopicOutcome::Created(tid) => Some(tid),
+                crate::channel::TopicOutcome::NoChannel => None,
+                crate::channel::TopicOutcome::Failed(err) => {
+                    tracing::warn!(
+                        agent = name,
+                        error = %err,
+                        "SPAWN: channel exists but create_topic failed; \
+                         instance created without topic"
+                    );
+                    None
+                }
+            };
             if let Some(n) = ctx.notifier {
                 let layout_hint = LayoutHint::parse(params["layout"].as_str().unwrap_or("tab"));
                 let spawner = params["spawner"]
--- a/src/api/handlers/team.rs
+++ b/src/api/handlers/team.rs
@@ -155,11 +155,24 @@ pub(crate) fn handle_create_team(params: &Value, ctx: &HandlerCtx) -> Value {
         ) {
             Ok(_) => {
                 tracing::info!(team = team_name, member = %inst_name, backend = %backend, "CREATE_TEAM spawn ok");
-                // Every successful spawn gets a channel topic so
-                // deploy_template doesn't have to know about topics.
-                // No-op when no channel is configured.
-                if let Some(ch) = crate::channel::active_channel() {
-                    let _ = ch.create_topic(inst_name);
+                // #966: Every successful spawn gets a channel topic via
+                // the hub `ensure_topic_for`. NoChannel is the happy path
+                // when no channel is configured; Failed is surfaced via
+                // warn (Pushback B: prior `let _ = ch.create_topic()`
+                // swallowed errors, same antipattern as pre-#962 silent
+                // persist).
+                match crate::channel::ensure_topic_for(inst_name) {
+                    crate::channel::TopicOutcome::Created(_)
+                    | crate::channel::TopicOutcome::NoChannel => {}
+                    crate::channel::TopicOutcome::Failed(err) => {
+                        tracing::warn!(
+                            team = team_name,
+                            member = %inst_name,
+                            error = %err,
+                            "CREATE_TEAM: channel exists but create_topic failed; \
+                             member spawn proceeds without topic"
+                        );
+                    }
                 }
                 spawned.push((inst_name.clone(), backend.clone()));
             }
--- a/src/app/commands.rs
+++ b/src/app/commands.rs
@@ -40,7 +40,13 @@ pub(super) fn execute(cmd: &str, ctx: &mut CommandCtx<'_>) -> bool {
 
             // unique_fleet_name guarantees inst_name is not yet in fleet.yaml
             let inst_name = super::pane_factory::unique_fleet_name(ctx.home, base_name);
-            if let Err(e) = crate::fleet::add_instance_to_yaml(
+            // #966: palette spawn previously bypassed channel-topic creation
+            // (`maybe_create_telegram_topic` at the `Ok(pane)` arm below
+            // creates the BINDING via create_binding, NOT the topic_id-
+            // returning create_topic). Route through
+            // `tui_spawn::add_instance_with_topic` so the topic is created
+            // + topic_id persisted before the spawn proceeds.
+            if let Err(e) = super::tui_spawn::add_instance_with_topic(
                 ctx.home,
                 &inst_name,
                 &crate::fleet::InstanceYamlEntry {
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -12,6 +12,7 @@ mod pane_factory;
 mod session;
 mod telegram_hooks;
 mod tui_events;
+mod tui_spawn;
 
 pub use overlay::{BoardView, MenuItem, MenuItemKind, TaskBoardMode};
 pub(crate) use tui_events::{TuiEvent, TuiEventSender, TuiNotifier};
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -816,7 +816,12 @@ fn pane_from_menu_item(
         MenuItemKind::Backend(backend) => {
             let preset = backend.preset();
             let inst_name = pane_factory::unique_fleet_name(home, preset.command);
-            if let Err(e) = crate::fleet::add_instance_to_yaml(
+            // #966: TUI Backend menu (ctrl+b c) previously called
+            // `add_instance_to_yaml` directly, bypassing the topic-creation
+            // side effect that `handle_spawn` does. Now routes through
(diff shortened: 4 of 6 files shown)


suzuke:
VERIFIED on `a8e7749`.

Checks passed:
- TUI paths now route through `tui_spawn::add_instance_with_topic`:
  - Backend menu path (`src/app/mod.rs`)
  - palette spawn/vsplit/hsplit path (`src/app/commands.rs`)
- Topic creation is consolidated via `channel::ensure_topic_for` (shared by handle_spawn, team mode, TUI helper).
- Team mode silent swallow was removed (`let _ = create_topic`) and replaced with explicit outcome handling + warn on failure.
- Runtime-lookup nuance is addressed (no cached startup-only channel snapshot reliance for topic creation).
- Scope is controlled to #966 surfaces; no unrelated drift observed.

Reviewer-side sanity:
- Targeted test rerun on this head snapshot passed:
  - `t2b_runtime_channel_lookup_picks_up_late_register`

No blocking issues found.

PR #997, opened 2026-05-20
fix(#995): agy display name + trust-prompt dismiss

suzuke:
## Summary

Partial #995 — Bugs 1 + 2 ship now per fixup-lead's directive. Bug 3 (agend-mcp-bridge not loaded) HELD pending operator scope-rule decision.

### Bug 1 — display name
TUI ctrl+b c Backend picker showed binary name `agy`. Changed `Backend::Agy.as_str()` to return `antigravity-cli` (product short form). Binary command (`preset().command`) stays `agy`. `parse_str` still accepts all aliases (`agy` / `antigravity` / `antigravity-cli`) for fleet.yaml backward-compat.

### Bug 2 — workspace-trust prompt
`--dangerously-skip-permissions` ONLY auto-approves tool-permission requests (verified via `agy --help`). The workspace-trust prompt ("Do you trust the contents of this project?") fires separately on every fresh spawn. Solved via `dismiss_patterns` (Codex #468 precedent): the prompt pre-selects "Yes, I trust this folder" (cursor `>` marker on first row), so Enter alone confirms.

```rust
dismiss_patterns: &[(r"(?m)^[^A-Za-z\n]{0,8}Yes, I trust", b"\r")],
```

### Bug 3 — HELD
Empirical bisect (live test in `/tmp/agy-test-995/` with 2 distinct probe servers):
- AGY logs `project: using project ... at <workdir>/.antigravitycli/mcp_config.json` (discovery)
- BUT only spawns servers from `~/.gemini/antigravity-cli/mcp_config.json` (HOME-level)
- Project-local `mcpServers` entries are IGNORED

Fix requires HOME-level write which conflicts with `src/mcp_config.rs:5-11` scope rule. Operator decision pending (Options A/B/C in dev-2's query to lead).

## Acceptance

- [x] TUI picker shows `antigravity-cli` (not `agy`)
- [x] Other display sites consistent (README + CHANGELOG + docs/CLI.md updated)
- [ ] Fresh spawn reaches ready w/o trust prompt — **requires manual verification with `agy install` ed binary** (test infra can't auto-verify interactive TUI dismiss)
- [ ] agend-mcp-bridge visible — **HELD (Bug 3 not in this PR)**
- [x] cargo fmt ✓, cargo clippy --all-targets -- -D warnings ✓
- [x] 2795 tests passed, 0 failed, 33 ignored (`cargo test --workspace --no-fail-fast`)

## Test plan

[...]

Issue #996, opened 2026-05-20
P0: daemon auto-injects up-arrow+enter into agent PTY, causes last user message to be re-submitted N times


suzuke:
## Severity
**P0** — daemon writing keystrokes into agent PTY without operator/agent consent is a class-A trust violation.

## Symptom
Operator types a single message into TUI; agent (fixup-lead, claude-code backend) receives the SAME message N times (witnessed: 3× today for the #987 3-bug report).

Operator-observed behavior: daemon repeatedly injects `up-arrow + enter` into fixup-lead's PTY. In claude-code TUI, up-arrow recalls previous user message and enter re-submits it → message duplication loop.

## Operator hypothesis (to verify, not commit)
Daemon's state-replay / prompt-detect layer matched some keyword/pattern in the operator's last message and triggered an auto-dismiss action that resolves to `up-arrow + enter`. The action assumes target TUI is a yes/no prompt; in claude-code TUI it instead re-submits history.

## Impact
1. Message duplication → wasted agent work + billing
2. Daemon writes keystrokes outside operator/agent control → trust/safety violation
3. Unbounded blast radius — `up + enter` in non-prompt TUI can trigger destructive actions per-backend
4. Noise pollutes inbox + duplicates dispatch attempts (dev-2 already saw #995 dispatch land 3× from re-submitted message)

## Diagnostic asks for dialectic
- [ ] Grep daemon log for prompt-detect / auto-dismiss events on fixup-lead pane around 2026-05-20 ~14:XX UTC (when operator sent #987 3-bug message)
- [ ] Identify exact keyword/pattern that triggered match
- [ ] Is auto-dismiss path generic (all backends share one detector) or backend-specific?
- [ ] Per-backend: what does `up + enter` actually do? (claude-code: re-submit history; codex: ?; agy: ?; gemini: ?)
- [ ] Was this code path introduced or modified by recent #987 (agy trust-prompt dismiss)? Temporal correlation suspicious.
- [ ] Should auto-dismiss be gated by per-backend opt-in instead of fleet-wide pattern match?

## Severity rationale
- Actively misfiring NOW in production
- Destructive blast radius unbounded
- Just shipped #987 which added new auto-dismiss logic — regression check urgent

## Acceptance for fix
- Daemon NEVER injects keystrokes into a backend that hasn't explicitly opted-in to that prompt class
[...]

Issue #995, opened 2026-05-20
[P2] agy backend polish — display name + trust prompt blocks spawn


suzuke:
## Pain (operator-visible, witnessed 2026-05-20 post-#987 #993 merge + rebuild+restart)

#987 shipped Backend::Agy variant successfully but two UX polish issues surfaced when operator tested live:

### Bug 1 — display name: \"agy\" shown where \"antigravity-cli\" expected

Currently the TUI Backend picker (ctrl+b c → Backend menu) shows \`agy\`. But \`agy\` is just the binary name. The user-facing product name is **Antigravity CLI** (per Google's branding + sunset announcement). For consistency with other backends that show full product names (\`Claude Code\`, \`Codex\`, \`Gemini CLI\`, \`Kiro CLI\`, \`OpenCode\`), the picker should display \`antigravity-cli\` (or \`Antigravity CLI\`).

Affected sites:
- TUI ctrl+b c Backend menu
- Possibly: \`agy\` in display_name / pretty_name field used elsewhere (CLI \`--backend\` help text, fleet.yaml schema docs, etc)

Fix shape: change display string (likely in src/backend.rs Backend::Agy as_str() or pretty_name(), or whichever field the picker consumes). Binary command (\`agy\`) stays correct.

### Bug 2 — trust prompt blocks spawn

When operator creates an agy instance, AGY CLI shows an interactive trust prompt:

\`\`\`
Accessing workspace:

/Users/suzuke/.agend-terminal/workspace/agy-c31c30

Do you trust the contents of this project?

Antigravity CLI requires permission to read, edit, and execute files here.

> Yes, I trust this folder
  No, exit

  ↑/↓ Navigate · enter Confirm
\`\`\`

The daemon spawns the PTY child but the agent never reaches ready state — it sits at this prompt waiting for ↑/↓ + enter from a human. In fleet context this is a blocker: operator has to manually arrow-down+enter on every fresh agy instance.

Parallel comparison: Gemini's \`configure_gemini\` writes \`trust: true\` into mcp_config.json which suppresses this prompt. Per #987 dialectic synthesis dev-2 explicitly excluded \`trust: true\` from \`configure_agy\` because operator's verification didn't surface its need — but live spawn proves it IS needed (or some equivalent).

Fix options:
- (A) Add \`trust: true\` to configure_agy's mcp_config write (if AGY honors the same field)
[...]

PR #1001, opened 2026-05-20
fix(#996): claude trust dismiss → single Enter (止血 keystroke patch)

suzuke:
## Summary

止血 patch for the auto-dismiss false-positive observed on `fixup-lead` pane today (37 events between 19:46:55–19:53:04 +08). Changes ClaudeCode `Yes, I trust` dismiss keystroke from `\x1b[A\x1b[A\r` (up+up+Enter, destructive) to `\r` (single Enter, non-destructive).

**Two compounding root causes** identified in #996 S1 spike (memo: /tmp/dialectic-996-s1-claude-flag-dev-2.md):

1. **False positive**: anchored regex `(?m)^[^A-Za-z\n]{0,8}Yes, I trust` matches operator-quoted text like `> Yes, I trust this folder` (verbatim from issue #995 body). In conv mode, up+up+Enter = history-back-2 + submit = re-submits a prior message → message duplication loop.
2. **True positive shape drift**: modern Claude (v2.1.145+) workspace-trust prompt pre-selects "Yes, I trust this folder" on row 1 (`❯` marker). The historical up+up+Enter sequence was correct when the prompt defaulted to "No, exit" (#468 era) but now navigates AWAY from default-Yes → may confirm "No, exit" → Claude exits.

Single `\r` resolves both: confirms the default-Yes on a real modal, adds a non-destructive newline on false-positive. Same shape as the Agy dismiss shipped in #997.

`Yes, proceed` (Claude entry #2) keystroke deliberately left on the old sequence — separate modal, default-cursor not empirically verified yet. Filed as Phase 1 follow-up per dispatch.

## Acceptance

- [x] `Backend::ClaudeCode.preset().dismiss_patterns[0].1 == b"\r"`
- [x] No ESC byte (0x1b) anywhere in that keystroke
- [x] Anchored regex unchanged (we don't change the dismiss surface, only the keystroke)
- [x] All existing tests pass (2821 / 0 fail / 33 ignored)
- [x] cargo fmt + clippy --all-targets clean
- [ ] Operator rebuild + restart daemon → fixup-lead pane loop stops (manual verify after merge)

## Test coverage

- **Config pin** (backend.rs `claude_trust_dismiss_uses_single_enter`): trust pattern keystroke == `b"\r"` + negative pin (no 0x1b ESC bytes anywhere). Regression-proof — revert the keystroke change → FAIL.
- **Behavioral true-positive** (agent.rs `claude_trust_dismiss_matches_real_modal`): vterm-rendered real Claude modal (with `❯` row-1 marker) still matches the anchored regex → dismiss fires. Anti-regression for the dismiss path.
- **Behavioral false-positive regression-pin** (agent.rs `claude_trust_false_positive_quoted_content_still_matches_regex`): operator-quoted `> Yes, I trust this folder` mid-conversation still triggers the regex. We deliberately don't change the regex — the fix is the keystroke. This pins that the failure surface is closed at the keystroke level, not at the matcher level.

## §3.6 / §3.10 / §3.12

- §3.6 borderline — 102 LOC total (88 LOC test + 14 LOC src/comment). Behavior change in dispatch keystroke. Opting for normal §3.12 review path.
- §3.10 RED→GREEN ✓ — test was crafted to fail under old keystroke; passes under new keystroke.
- §3.12 mirror after reviewer VERIFIED.
- §3.12.1 ACTIVE — `gh pr merge --auto --squash --delete-branch` (or admin path per operator's new memory).

## Related

- #996 — root issue (operator's destructive blast report)
- #468 — original anchored-regex hotfix (intent comment at backend.rs:241-248)
- #997 — Agy dismiss shipped same shape (`\r`)
- /tmp/dialectic-996-s1-claude-flag-dev-2.md — S1 spike findings
- /tmp/dialectic-996-s4-pattern-audit-dev-2.md — broader pattern audit (Yes, proceed + Kiro No, exit still TBD)

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -1843,6 +1843,11 @@ mod tests {
     /// Production dismiss regex for kiro-cli's "Trust All Tools" prompt
     /// (Issue #468 follow-up — radio-button cursor `)` was unmatched).
     const KIRO_TRUST_REGEX: &str = r"(?m)^[^A-Za-z\n]{0,8}No, exit";
+    /// Production dismiss regex for Claude's workspace-trust prompt (#996
+    /// Phase 1). Modern Claude (v2.1.145+) defaults cursor to "Yes, I trust",
+    /// so the keystroke shipped is single Enter `\r` — see
+    /// `Backend::ClaudeCode.preset().dismiss_patterns[0]`.
+    const CLAUDE_TRUST_REGEX: &str = r"(?m)^[^A-Za-z\n]{0,8}Yes, I trust";
 
     /// `(regex, keystrokes)` pair for `try_dismiss_dialog` — `Down` then
     /// `Enter` to dismiss kiro-cli's "Trust All Tools" prompt.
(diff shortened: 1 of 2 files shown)


suzuke:
VERIFIED

Reviewed HEAD `7325f18881c77e775780fca12bd6a9503edd3179` for PR #1001 (`fix/996: claude trust dismiss → single Enter`). CI is green: ubuntu, macOS, Windows, audit, and LOC overrun all passed.

Scope verified:
- ClaudeCode `Yes, I trust` dismiss response changes from `\x1b[A\x1b[A\r` to single `\r`.
- The production regex is intentionally unchanged, so the stop-bleed fix removes the destructive history-navigation blast while preserving current modal detection behavior.
- Config test pins `b"\r"` and rejects ESC bytes in the trust dismiss keystroke.
- Behavioral tests cover both the modern true-positive Claude trust modal with row-1 `Yes` selected and the known false-positive quoted-content surface.
- `Yes, proceed` remains on the old sequence by design and is explicitly left for follow-up because its modal/default cursor shape has not been empirically verified.
- The follow-up commit is rustfmt-only; no semantic change beyond the reviewed keystroke patch.

Residual risk: this is Phase 1 mitigation, not full #996 engine hardening. The regex can still match quoted content, and other destructive dismiss entries such as Claude `Yes, proceed` and Kiro `No, exit` still need the broader composite-signature/quiet-window/kill-switch work. For the urgent fixup-lead loop, this patch removes the observed `up+up+Enter` history-resubmit blast for `Yes, I trust`.

Issue #1002, opened 2026-05-20
P1: autonomous merge pipeline silent — #986 gh-poll + #972 verdict-record deployed but not functioning


suzuke:
## Severity
**P1** — recently-shipped autonomous merge pipeline (#972 state-machine + #986 gh-poll integration) is fully deployed but completely silent in production. Every PR requires manual admin-merge by lead; the autonomous close-loop never engages.

## Symptom

When PR #982 (#999) reached close-loop, fixup-dev dumped the pr-state aggregator JSON. Snapshot at admin-merge time:

\`\`\`json
{
  \"repo\": \"suzuke/agend-terminal\",
  \"pr_number\": 0,                              // ← NEVER POPULATED (expected 999)
  \"branch\": \"fix/982-inbox-routing\",
  \"head_sha\": \"6ae007b...\",
  \"pr_author\": \"\",                            // ← NEVER POPULATED (expected \"suzuke\")
  \"ci_state\": { \"Green\": { \"sha\": \"6ae007b...\" } },  // ✓ ci_watch fired correctly
  \"verdict_state\": \"None\",                    // ← NEVER RECORDED (expected Verified after reviewer kind=report)
  \"merge_state\": \"NotReady\",                  // ← consequence of above
  \"ready_emitted_for_sha\": null,              // ← no [pr-ready-for-merge] event ever emitted
  \"auto_armed\": false,                          // ← --auto never armed by daemon
  \"last_gh_poll_at\": null,                      // ← #986 gh-poll NEVER FIRED
  \"gh_poll_failures\": 0,
  \"last_gh_state\": null
}
\`\`\`

## 4 distinct deviations

1. **#986 gh-poll never fired** — \`last_gh_poll_at: null\` for the entire PR lifetime. \`pr_number\` stuck at 0, \`pr_author\` stuck at \"\". The #986 integration never ran on this branch.
2. **#972 verdict-record never propagated** — \`verdict_state: None\` despite reviewer formally RC1-rejecting + RC2-VERIFYING. The pr_state aggregator's verdict mirror never registered either verdict. Reviewer's kind=report events flowed via inbox + PR comment but didn't reach pr_state.
3. **No \`[pr-ready-for-merge]\` event ever emitted** — consequence of (1)+(2). Even though CI was green and lead admin-merged, the event-driven flow never woke the author.
4. **\`--auto\` never armed** — consequence of (3). §3.12.1 ACTIVE never engaged for this PR.

## Net impact

[...]

Issue #982, opened 2026-05-20
[P2] inbox-only delivery strands structured messages when recipient is idle at prompt


suzuke:
## Pain

When lead sends \`kind=report\` (replying to a recipient's query) or \`kind=update\` (phase-transition kick) to a recipient sitting idle at prompt, daemon routes the message to \`delivery_mode: inbox_only\`. No PTY notification surfaces. Recipient sits idle indefinitely because they don't know a message arrived. Lead must follow up with a plain-text message (which DOES deliver \`delivery_mode: pty\`) to wake the recipient and tell them to drain inbox.

This is the same root failure mode family as #972 (lead-kick deadlock for phase transitions), but at the per-message routing layer.

## Concrete incident (2026-05-20 02:00-02:09)

Sequence:
1. \`fixup-reviewer\` sent \`kind=query\` requesting guidance on dirty worktree before #972 review
2. \`fixup-lead\` replied \`kind=report\` (Option 3 GH-only) — daemon routed \`delivery_mode: inbox_only\`
3. \`fixup-lead\` sent \`kind=update\` (cleanup procedure for after verdict) — \`delivery_mode: inbox_only\`
4. \`fixup-reviewer\` sat idle at prompt for ~9 minutes, never auto-drained inbox, never saw either reply
5. Operator caught the stuck state: \"你真的有傳給reviewer嗎？\"
6. \`fixup-lead\` sent plain-text \"drain inbox — lead has two replies for you\" — \`delivery_mode: pty\` worked, reviewer woke
7. Reviewer drained inbox and proceeded with #972 review

## Root cause hypothesis

Daemon's \`delivery_mode\` decision likely considers \`(recipient_pty_state, message_kind)\` tuple:
- Recipient busy + plain text → buffer for after-task or pty-on-next-prompt
- Recipient busy + structured kind → inbox to avoid interruption (correct)
- **Recipient idle + structured kind → inbox_only (the bug — should surface as PTY since recipient is ready)**
- Recipient idle + plain text → PTY (correct)

The structured-kind-at-idle case treats messages as background queue items, but operationally they're often urgent ACKs / kicks / phase transitions the recipient is actively blocked on.

### Option A: PTY-notify on inbox arrival when recipient idle (lowest cost)
- When daemon routes a structured message to inbox while recipient is at idle prompt, also emit a PTY-side tag: \`[N new messages in inbox]\` (or similar visible hint)
[...]

Issue #941, opened 2026-05-19
[P1] #932 observability: periodic_thread_state_dump for zombie RCA


suzuke:
## Problem (PR-B per #933 dialectic synthesis)

#932 RCA could not be resolved by static analysis alone. Two competing hypotheses cannot be discriminated without runtime evidence:

- **H1** (dev): per-tick handler holds \`registry.lock()\`, main loop wedges, shutdown atomic never checked. Code: \`per_tick/hang_detection.rs:39-66\` + \`per_tick/watchdog.rs:32-58\`.
- **H6** (dev-2): operator-side discovery bug — \`agend-terminal stop\` reads \`find_active_run_dir(\$CURRENT_AGEND_HOME)\` at \`src/daemon/mod.rs:298\`; if zombie's spawn-time AGEND_HOME differed, SIGTERM was never sent. Code: discovery path.

These are mutually exclusive. Without runtime evidence, RCA stays open indefinitely.

## Direction (locked per #933 PR-A dialectic synthesis)

Ship env-gated periodic thread-state dump as observability primitive:

\`\`\`
AGEND_DAEMON_THREAD_DUMP_SECS=<N>
\`\`\`

When set (N≥1), daemon emits periodic state dump every N seconds:
- \`registry.lock()\` current holder identity + duration
- Per-handler timing (HangDetection, Watchdog, others)
- \`api::serve\` session count + per-session state
- API listener thread state (\`src/api/mod.rs:231\` \`listener.incoming().flatten()\` shutdown observation gap, H3 partial)

## Scope estimate

- ~30 LOC impl in \`src/daemon/observability.rs\` (new module)
- ~50 LOC test:
  - Format invariant (output contains expected fields)
  - Env-unset → no dump emitted
  - Synthetic registry contention → dump catches holder identity (§3.20 SOP 1 deterministic)

## Operator usage

For next zombie incident:

\`\`\`
# H6 discriminator (cheap, do FIRST):
grep \"shutting down (signal received)\" \$AGEND_HOME/daemon.log
# Absent → H6 (signal never delivered, AGEND_HOME mis-attribution)
# Present → H1/H2/H7 (signal delivered but shutdown wedged → check thread-dump)

# Enable thread-dump for next zombie episode:
AGEND_DAEMON_THREAD_DUMP_SECS=60 ./agend-terminal start
\`\`\`

## References

- #933 (closes — boot-sweep mitigation, ships first as PR-A)
[...]

PR #1004, opened 2026-05-20
fix(#1002): Phase 1 — observability + per-handler panic guard

suzuke:
## Summary

Closes the immediate #982-style \"silent autonomous-flow failure\" diagnostic gap. Phase 1 is **observability + isolation only** — no behavioural change to the gh-poll / verdict-record paths. Once deployed, the next deviation incident produces a \`grep \"#1002\"\` trail identifying which gate fired.

Spike: \`/tmp/dialectic-1002-autonomous-flow-spike-dev.md\` (4 deviations on #982 traced to ROOT 1 gh-poll never executed + ROOT 2 record_verdict 5 unmonitored gates).

### 1. Per-handler panic isolation

\`src/daemon/per_tick/mod.rs\` adds \`run_handlers_with_panic_guard(handlers, ctx)\` — wraps each \`handler.run(...)\` in \`std::panic::catch_unwind\` with \`AssertUnwindSafe\`. On caught panic:
- handler identity + payload logged via \`tracing::error!\` (\`#1002\` tag)
- iteration continues — handlers N+1..end on same tick still run
- next tick re-invokes the failed handler (no permanent skip)

Folds in #941 per-handler timing (was inline in main loop). \`src/daemon/mod.rs:717\` replaces the manual \`for handler\` loop with the new helper.

### 2. PR-state observability

\`src/daemon/pr_state/mod.rs\` instruments every silent early-return with \`tracing::debug!\` carrying the \`#1002\` marker:

- **\`scan_and_emit_with\`**: read_dir / read_to_string / json parse failures
- **\`apply_gh_poll\`**: same trio plus per-tick \"grouping done\" summary (\`repos_to_poll\`, \`skipped_terminal\`, \`skipped_should_poll_cadence\`)
- **\`record_verdict\`**: 5 gates (A reviewed_head None / B task not found / C task.branch empty / D pr-state dir read fail / E no branch match)

Pre-fix code paths like \`let Ok(state) = serde_json::from_str(...) else { continue; };\` are rewritten as explicit \`match\` arms with debug emission. Idiomatic same-shape replacement; no logic change.

## Tests (4 new pins)

| Test | Location | Pin |
|---|---|---|
| \`panicking_handler_does_not_skip_siblings\` | per_tick::tests | 3-handler vec, middle panics → both siblings still run |
| \`panic_payload_str_handles_common_panic_types\` | per_tick::tests | Debug helper across String / &'static str / other |
| \`t15_malformed_pr_state_file_emits_observability_trace\` | pr_state::tests | malformed JSON → \`tracing_test::traced_test\` + \`logs_contain(\"#1002\")\` + filename |
| \`t16_record_verdict_gate_a_emits_observability_trace\` | pr_state::tests | \`reviewed_head=None\` → \`logs_contain(\"gate A\")\` |

2599 lib tests pass + clippy clean.

## §3.6 NOT eligible

Cross-subsystem touch (per_tick + pr_state), new public helper, behaviour change at panic boundary. Normal review.

## Phase 2 follow-up

Decided based on real \`#1002\` trace evidence collected post-deploy. Spike memo §4 lists candidate structural fixes (loosen gate B correlation_id match, dedicated verdict_telemetry log, webhook-as-primary). Phase 1 ships standalone so the diagnostic surface lands before any structural follow-up.

## Test plan
- [x] cargo test --bin agend-terminal (2599 pass)
- [x] cargo clippy --all-targets -- -D warnings (clean)
- [ ] CI green on push
- [ ] reviewer VERIFIED
- [ ] §3.12.1 self-merge via gh pr merge --auto

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -714,16 +714,17 @@ fn run_core(
         // `let handlers = …` above). The pre-collapse call order is
         // preserved verbatim by the Vec literal's element order; this
         // loop is the only call site.
-        for handler in &handlers {
-            // #941: wrap each handler.run() in an Instant measurement so
-            // `HANDLER_TIMING` (per_tick/mod.rs) accumulates per-handler
-            // stats for the periodic ThreadDumpHandler. When thread-dump
-            // is disabled, `record_handler_timing` is a sub-µs no-op
-            // (one cached atomic load + early return).
-            let start = std::time::Instant::now();
-            handler.run(&tick_ctx);
-            per_tick::record_handler_timing(handler.name(), start.elapsed());
-        }
+        //
+        // #1002 Phase 1: dispatch now goes through
+        // `per_tick::run_handlers_with_panic_guard`, which wraps each
+        // `handler.run()` in `catch_unwind`. A panic in handler N no
+        // longer aborts handlers N+1..end on the same tick — they all
+        // run, and the panic is logged with handler identity instead
+        // of vanishing. The wrapper folds in #941 per-handler timing
+        // (previously inline here). See
+        // `per_tick::tests::panicking_handler_does_not_skip_siblings`
+        // for the regression-pin.
+        per_tick::run_handlers_with_panic_guard(&handlers, &tick_ctx);
 
         // Handle exit event (if any)
         let exit_event = match exit_event {
--- a/src/daemon/per_tick/mod.rs
+++ b/src/daemon/per_tick/mod.rs
@@ -138,3 +138,151 @@ pub(crate) fn snapshot_handler_timings() -> HashMap<String, HandlerStats> {
         .map(|(k, v)| ((*k).to_string(), v.clone()))
         .collect()
 }
+
+/// #1002 Phase 1 — per-handler panic isolation for the main-loop
+/// dispatch.
+///
+/// Drives the per-tick handler iteration in two steps:
+/// 1. `std::panic::catch_unwind` wraps each `handler.run(...)` call so
+///    a panic in handler N does not abort the iteration before
+///    handlers N+1..end run on this tick.
+/// 2. On a caught panic, log handler identity + panic payload (best-
+///    effort `Debug`-formatting of `Box<dyn Any>`) so future
+///    `grep "panic"` over `app.log` surfaces the silent failure that
+///    #1002 was filed against.
+///
+/// The handler trait object itself is `Send + Sync`, but `tick_ctx`
+/// holds borrowed references — `AssertUnwindSafe` is required.
+/// Borrows are not invalidated by unwinding because the catch boundary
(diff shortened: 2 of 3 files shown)


suzuke:
REJECTED

Blocking finding:

- The PR's main promise is that the next autonomous-flow deviation leaves a `grep "#1002"` trail in the normal daemon/app logs, but almost all new pr_state gate instrumentation is emitted with `tracing::debug!`. The default rolling log filter is `agend_terminal=info` in both daemon and app setup, so these messages are suppressed unless the operator happened to start with `AGEND_LOG=agend_terminal=debug`. That leaves the same silent-failure class in normal production operation: `record_verdict` gates A-E, pr-state read/parse skips, and gh-poll grouping evidence are still absent from `daemon.log`/`app.log` by default. The panic guard itself logs at `error!` and is visible, but the pr_state observability half is not.

Freshness: reviewed HEAD `e7abe34307bcbcec94949db55535a87afe62270d` for PR #1004. CI was still in progress at review time; this finding is independent of CI result.

Suggested fix: promote the diagnostic gates that are meant to be operator-visible to `info!` or `warn!` (especially `record_verdict` gates and malformed/read failures). Keep high-volume summaries like `apply_gh_poll grouping done` at debug if needed, but the failure gates that close #1002 cannot rely on a debug-only filter.

suzuke:
VERIFIED

Reviewed HEAD `95d04133cb04ecb2662c77266c2dc26f68a415b2` for PR #1004 (`fix(#1002): Phase 1 — observability + per-handler panic guard`). CI is green: ubuntu, macOS, Windows, audit, and LOC overrun all passed.

Scope verified:
- Per-tick handler dispatch now uses `run_handlers_with_panic_guard`, preserving per-handler timing while isolating panics so sibling handlers still run on the same tick and the failed handler is retried on later ticks.
- Panic guard logs at `error!` with handler identity and printable payload, so panic isolation is operator-visible under the default `agend_terminal=info` daemon log filter.
- `record_verdict` silent gates A/B/C/E now log at `info!`; gate D logs at `warn!`. These are visible by default and close the prior review blocker.
- `scan_and_emit_with` and `apply_gh_poll` IO/parse failures now log at `warn!`; only the high-frequency `apply_gh_poll grouping done` summary remains `debug!`, which is appropriate for cadence noise rather than failure evidence.
- The behavior remains diagnostic/isolation only for pr_state flow: no changes to gh-poll semantics, reducer transitions, verdict application rules, or merge gating.
- Tests cover panic isolation/payload formatting and two representative #1002 observability paths.

Residual risk: this PR intentionally does not fix the autonomous-flow root cause; it makes the next deviation diagnosable and prevents a single per-tick panic from skipping sibling handlers. Follow-up structural fixes should be based on the resulting `#1002` trace evidence.

PR #1008, opened 2026-05-20
fix(#1000): bridge-side 500ms content dedup for LLM tool double-fire

suzuke:
## Summary

Closes #1000. LLM occasionally double-fires identical `tools/call` requests in the same turn (model-side hallucination). Pre-fix: bridge minted distinct request_ids for each → daemon dispatched both → two identical inbox entries 84 ms apart. Per GitHub thread consensus (filer + @suzuke + @cheerc), the dedup belongs at the bridge — daemon's append-only inbox contract stays intact.

Change: `src/bin/agend-mcp-bridge.rs` adds bridge-local session state (`RecentCall`) tracking the most-recent `tools/call`. When an incoming call matches the same tool + identical args within 500 ms, the bridge returns a success-with-`note` response and does NOT forward to the daemon. Mirrors @cheerc's snippet from the issue thread, plus an `is_duplicate_call` helper for deterministic unit-testing and a `last_call` refresh on dedup hits so back-to-back triple-fires stay dropped.

Operator observability: `eprintln!` warning fires on every dropped call.

## Acceptance

- [x] LLM double-fire within 500 ms returns success + `"note":"duplicate tool call dropped by bridge"`, daemon never sees it
- [x] Daemon side (`src/api/request_dedup.rs`) and inbox enqueue path completely unchanged — append-only contract preserved
- [x] Retry mechanism (`try_proxy_once` reuses same `request_id`) unaffected — that's a different layer (transport retry, daemon-side dedup by id)
- [x] All existing tests pass (2833 / 0 fail / 33 ignored)
- [x] cargo fmt + clippy --all-targets clean

## Tests added (6 anchors)

All on the pure `is_duplicate_call(last, tool, args, now, window)` helper — no I/O, no `Instant::now()` inside the helper, so timing is deterministic via injected `now: Instant`.

1. **`is_duplicate_call_identical_within_window`** — identical tool + args + 84 ms elapsed → TRUE (canonical LLM double-fire case from issue body)
2. **`is_duplicate_call_identical_after_window`** — identical + 501 ms elapsed → FALSE
3. **`is_duplicate_call_same_tool_different_args_within_window`** — same `send` tool, different `message` field → FALSE (legit distinct messages)
4. **`is_duplicate_call_different_tool_within_window`** — `send` vs `reply` with same args envelope → FALSE
5. **`is_duplicate_call_no_prior_state`** — `last_call=None` → FALSE (cold start)
[...]

PR #1010, opened 2026-05-20
fix(#1000): implement short-TTL content dedup in MCP bridge

cheerc:
Implement a 500ms content-based deduplication window in the MCP bridge to handle upstream LLM double-firing tool calls (hallucinations).

As discussed in #1000, this noise reduction happens at the bridge layer to preserve the daemon's contract as an append-only log of unique requests (identified by request_id).

Changes:
- Added `RecentCall` struct to track last tool name, arguments, and timestamp.
- Intercept duplicate calls within 500ms in `tools/call` handler.
- Return a successful result with a note instead of forwarding to the daemon.

### DIFF
--- a/src/bin/agend-mcp-bridge.rs
+++ b/src/bin/agend-mcp-bridge.rs
@@ -24,6 +24,12 @@ fn main() {
     }
 }
 
+struct RecentCall {
+    tool: String,
+    args: serde_json::Value,
+    at: std::time::Instant,
+}
+
 fn run() -> Result<(), Box<dyn std::error::Error>> {
     let home = home_dir();
     let instance = std::env::var("AGEND_INSTANCE_NAME").unwrap_or_default();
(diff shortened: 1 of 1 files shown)


suzuke:
Thanks @cheerc for the contribution — your snippet from #1000 was the model for the implementation we shipped in #1008 (admin-merged earlier today at 2026-05-20T17:06:20Z). The shipped version mirrors your design with 3 small elaborations:

1. Extracted `is_duplicate_call(last, tool, args, now, window) -> bool` as a pure helper so the dedup logic can be deterministically unit-tested with injected `now: Instant`
2. `last_call` is now seeded only inside the `Ok(_)` branch after a successful daemon forward — this avoids poisoning the dedup cache on failed/retried attempts (caught during code review, RC1 fix)
3. `last_call` refreshes on dedup hit so successive triple-fires stay dropped, instead of one slipping through after the first window expires

10 bridge tests pin the behavior including the failed-then-retry-within-500ms case. Issue #1000 also closed at that time.

Really appreciate the well-analyzed RCA in #1000 and the working snippet — it set up a clean shipping path. Closing this PR as superseded by #1008.

PR #1088, opened 2026-05-23
fix(#1087): widen dismiss_patterns prefix for centered TUI modals

### LINKED ISSUE #1087: OpenCode dismiss_patterns regex prefix too strict for centered TUI modals
## Symptom

OpenCode's "Update Available" modal is not auto-dismissed. The modal stays on screen despite dismiss_patterns being configured.

## Root Cause

All 4 OpenCode dismiss patterns in `backend.rs:380-387` use `{0,8}` prefix limit:

```rust
dismiss_patterns: &[
    (r"(?m)^[^A-Za-z\n]{0,8}Update Available", b"\x1b"),
    (r"(?m)^[^A-Za-z\n]{0,8}Skip  Confirm", b"\x1b"),
    (r"(?m)^[^A-Za-z\n]{0,8}Update Complete", b"\r"),
    (r"(?m)^[^A-Za-z\n]{0,8}Please restart", b"\r"),
],
```

OpenCode renders modals centered in the terminal. "Update Available" is preceded by 40+ spaces of padding. The `{0,8}` limit only allows up to 8 non-alpha characters before the match text, so the regex never fires.

Pane snapshot showing actual content:
```
                                                                        Update Available                                     esc
                                                                                                                  Skip  Confirm
```

## Fix

Change `{0,8}` to `*` for all 4 OpenCode patterns. Safety is maintained by `^` anchor (line-start) + `[^A-Za-z\n]` class (no letters allowed in prefix).

Same prefix style was chosen in #468 to prevent mid-line false positives — `*` preserves that property since `^` already prevents mid-line matches and `[^A-Za-z\n]` prevents matching after prose text on the same line.

## Scope

- `src/backend.rs:380-387` — 4 regex patterns
- Also check other backends for same centering issue (Codex uses `{0,8}` too — verify its layout)

## Observed on

OpenCode 1.14.50 → v1.15.6 update prompt. Instance: opencode-f068d5.

suzuke:
## Summary
- OpenCode and Codex `dismiss_patterns` regex prefix changed from `{0,8}` to `*`
- Root cause: TUI centered modals have 40+ space prefix before the keyword, exceeding the 8-char limit
- Safety: `^` anchor + `[^A-Za-z\n]` character class still prevents mid-line false positives
- Both OpenCode (4 patterns) and Codex (3 patterns) updated

## Test plan
- [x] Existing tests pass: `opencode_update_available_dismiss_uses_esc`, `codex_update_prompt_dismiss_uses_skip`
- [x] Updated anchoring tests now cover centered modal with 45-space prefix
- [x] Mid-line rejection still works (Issue #468 safety)
- [x] `cargo fmt` + `cargo clippy` clean

Closes #1087

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -353,12 +353,12 @@ impl Backend {
                     // macOS openpty doesn't translate LF→CR on input (ConPTY does),
                     // so LF here would silently no-op on mac.
                     // Issue #468: anchored regex (see ClaudeCode comment above).
-                    (r"(?m)^[^A-Za-z\n]{0,8}Do you trust", b"\r"),
-                    // Auto-update prompt: "Please restart Codex" → Enter
-                    (r"(?m)^[^A-Za-z\n]{0,8}Please restart", b"\r"),
+                    // #1087: `*` instead of `{0,8}` — TUI centered modals have 40+ char prefix.
+                    (r"(?m)^[^A-Za-z\n]*Do you trust", b"\r"),
+                    (r"(?m)^[^A-Za-z\n]*Please restart", b"\r"),
                     // #1069: version-update modal blocks agent until operator
                     // selects an option. "2\r" = "Skip" (least invasive).
-                    (r"(?m)^[^A-Za-z\n]{0,8}Update available!", b"2\r"),
+                    (r"(?m)^[^A-Za-z\n]*Update available!", b"2\r"),
                 ],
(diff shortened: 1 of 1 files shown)


suzuke:
§3.12 VERIFIED

No blocking issues found. This widens the dismiss-pattern prefix safely for centered TUI modals while keeping the anchored mid-line rejection guard intact.

PR #1094, opened 2026-05-23
fix(#1092): prevent stacked pending decisions for same sender

### LINKED ISSUE #1092: bug: decision_timeout allows stacked pending decisions — older ones fire unintended auto-default
## Source

Code review finding (high severity) from CI/CD & PR State group review.

## Problem

`record_pending_decision()` creates a new pending sidecar every time `reply` is called with `default_action` + `timeout_secs`. But `mark_resolved_for_sender()` only resolves the **most recent** pending entry for that sender.

If the same sender stacks two or more pending decisions before replying without `default_action`, the older sidecars remain pending and can still time out later — firing a surprise auto-default action after the operator already "resolved" the flow.

**Files**: `src/daemon/decision_timeout.rs:107-137`, `src/daemon/decision_timeout.rs:185-193`

## Fix

Make pending decisions unique per sender: either cancel/resolve any existing pending entry before recording a new one, or reject a second `reply(default_action, timeout_secs)` until the first is resolved.

## Process constraint

**Characterization tests first**: write tests demonstrating the bug on the ORIGINAL code (test should show the stacked-pending behavior), then fix, then verify the test now shows correct behavior.

## Estimated scope

~40 LOC fix + ~60 LOC tests

## Review report

`/tmp/review-cicd-auxiliary.md` — Finding 1

suzuke:
## Summary
- Same sender could stack multiple pending decisions, but `mark_resolved_for_sender` only cleared the latest one — older pendings would timeout-fire unexpectedly
- Fix: cancel any existing pending decisions for the same sender in `record_pending_decision` before creating a new one
- Added 4 characterization/regression tests covering the stacking, resolve, and scan-fire scenarios

## Test plan
- [x] 4 new tests in `decision_timeout::tests` (t1092_*)
- [x] All 18 decision_timeout tests pass

Closes #1092

### DIFF
--- a/src/daemon/decision_timeout.rs
+++ b/src/daemon/decision_timeout.rs
@@ -117,6 +117,15 @@ pub(crate) fn record_pending_decision(
     if std::fs::create_dir_all(&dir).is_err() {
         return None;
     }
+    // #1092: cancel any existing pending decisions for the same sender
+    // before recording a new one. Prevents stacked pendings where
+    // mark_resolved_for_sender only clears the latest, leaving older
+    // ones to timeout-fire unexpectedly.
+    for d in list_pending(home) {
+        if d.sender == sender && d.status == "pending" {
+            let _ = std::fs::remove_file(pending_path(home, &d.decision_id));
+        }
+    }
     let decision_id = next_decision_id();
     let payload = PendingDecision {
         schema_version: SCHEMA_VERSION,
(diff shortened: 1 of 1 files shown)


suzuke:
§3.12 VERIFIED

No blocking issues found. `record_pending_decision` now cancels any existing pending decision for the same sender before creating a new one, so stacked pendings cannot leak into later timeout fires.

PR #1107, opened 2026-05-24
fix(#1102): mirror dispatch uses lookup_channel_by_name prefer-chain

### LINKED ISSUE #1102: Mirror dispatch uses active_channel() instead of lookup_channel_by_name — silently drops mirrors in multi-channel fleets
## Problem
`try_dispatch_mirror` in `router.rs:175` uses `active_channel()` which returns `None` when multiple channels are registered (telegram + discord). The agent's `handle_reply` in `channel.rs` correctly implements the prefer-chain lookup (`lookup_channel_by_name` → `active_channel` fallback), but the router does not.

This means all mirror dispatches are silently dropped in multi-channel fleets.

## Found in
Communication & Messaging group code review (fixup-dev-2, 2026-05-24)

## Fix
Replace `active_channel()` in `try_dispatch_mirror` with the same prefer-chain pattern from `handle_reply`:
1. Try `lookup_channel_by_name(pair.reply_to_channel.as_deref().unwrap())` 
2. Fall back to `active_channel()`

The `reply_to_channel` is already verified non-None at line 148.

## Constraint
Characterization test first — lock current behavior, then fix.

suzuke:
## Summary

- `try_dispatch_mirror` in `src/daemon/router.rs` used `active_channel()` to resolve the output channel, which returns `None` when >1 channel is registered — silently dropping mirrors in multi-channel fleets (telegram + discord)
- Applied the same prefer-chain pattern used by `handle_reply` in `src/mcp/handlers/channel.rs`: `lookup_channel_by_name(reply_to_channel)` first, `active_channel()` fallback
- Added 2 tests: multi-channel dispatch routes to the correct named channel; single-channel fallback still works when name lookup fails

Fixes #1102

## Test plan

- [x] `mirror_dispatch_uses_named_channel_in_multi_channel_fleet` — registers 2 channels, verifies mirror routes to the named channel (not dropped)
- [x] `mirror_dispatch_falls_back_to_active_channel_when_lookup_fails` — single channel with mismatched name, verifies `active_channel()` fallback
- [x] All 5 router tests pass (`cargo test --bin agend-terminal -- daemon::router`)
- [x] `cargo fmt -- --check` clean

### DIFF
--- a/src/daemon/router.rs
+++ b/src/daemon/router.rs
@@ -172,7 +172,15 @@ fn try_dispatch_mirror(home: &std::path::Path, name: &str, buf: &mut AgentBuffer
         text
     };
 
-    if let Some(ch) = crate::channel::active_channel() {
+    // #1102 fix: prefer-chain — lookup_channel_by_name first, active_channel fallback.
+    // pair.reply_to_channel is verified Some at L148 (early return if None).
+    let ch = pair
+        .reply_to_channel
+        .as_deref()
+        .and_then(crate::channel::lookup_channel_by_name)
+        .or_else(crate::channel::active_channel);
+
+    if let Some(ch) = ch {
         let result = ch.send_from_agent(
             name,
             crate::channel::AgentOutboundOp::Reply {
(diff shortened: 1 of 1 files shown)


suzuke:
§3.12 VERIFIED

No blocking issues found. `try_dispatch_mirror()` now prefers `lookup_channel_by_name(reply_to_channel)` and falls back to `active_channel()`, which fixes mirror dispatch in multi-channel fleets without regressing the singleton path. Local `cargo check -q --locked` and `cargo test -q router -- --nocapture` passed.

Issue #1145, opened 2026-05-24
write_with_timeout guard entry persists forever after timeout


suzuke:
## Problem

`agent.rs:1503-1508` — on timeout, the `WRITE_IN_PROGRESS` guard entry is kept to prevent spawning more threads. The stuck thread never clears the guard on exit. If agent is deleted and new agent reuses same PtyWriter address, writes are permanently blocked.

## Impact

Rare but permanent write failure for a re-created agent after a timeout event on the previous occupant.

## Fix

Have the stuck thread clear the guard entry when it finally completes (move `key` into the thread closure, clear on exit regardless of result).

## Source

Agent & Backend review — agent.rs — finding H3.

suzuke:
Fixed in PR #1148

Issue #1144, opened 2026-05-24
pty_read_loop error path skips handle_pty_close — agent becomes zombie


suzuke:
## Problem

`agent.rs:984-991` — when `pty_reader.read()` returns `Err(e)`, the loop breaks without calling `handle_pty_close`. The agent remains in the registry forever — no crash notification, no process tree sweep, no shell fallback.

Compare with the `Ok(0)` (EOF) path at `agent.rs:923-930` which correctly calls `handle_pty_close`.

## Impact

A transient I/O error on the PTY fd leaves a phantom registry entry. Agent appears "alive" but PTY is dead. No respawn fires. Manual DELETE is the only recovery.

## Fix

Call `handle_pty_close` in the `Err` branch, or unify both break paths to call it after loop exit.

## Source

Agent & Backend review — agent.rs — finding H1.

suzuke:
Fixed in PR #1148

PR #1148, opened 2026-05-24
fix(agent): pty_read_loop error path cleanup + write guard leak (#1144, #1145)

suzuke:
## Summary
- **#1144**: `pty_read_loop` `Err(e)` branch broke without calling `handle_pty_close`, leaving a zombie agent in the registry with no cleanup, no crash notification, and no shell fallback. Fixed by moving `handle_pty_close` after the loop so both EOF and error paths trigger cleanup.
- **#1145**: `write_with_timeout` stuck thread never cleared `WRITE_IN_PROGRESS` guard after timeout — the guard persisted forever, permanently blocking future writes to the same (or address-reused) `PtyWriter`. Fixed by moving guard cleanup into the thread closure so it self-clears on exit regardless of whether the caller timed out.

## Test plan
- [x] `pty_read_error_triggers_cleanup` — simulates read error with `FailingReader`, verifies agent removed from registry (handle_pty_close fires)
- [x] `write_guard_cleared_after_stuck_thread_completes` — uses `BlockingWriter` + channel sync, verifies guard cleared after thread unblocks
- [x] `cargo test --bin agend-terminal` — 2775 passed, 0 failed
- [x] `cargo clippy -- -D warnings` — clean

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -3096,16 +3096,12 @@ Allow Trust All Tools mode?
             }
         }
 
-        let writer: PtyWriter = Arc::new(Mutex::new(Box::new(BlockingWriter {
-            lock_tx,
-            unlock_rx,
-        })));
+        let writer: PtyWriter =
+            Arc::new(Mutex::new(Box::new(BlockingWriter { lock_tx, unlock_rx })));
         let key = Arc::as_ptr(&writer) as usize;
 
         let writer2 = Arc::clone(&writer);
-        let handle = std::thread::spawn(move || {
-            write_with_timeout(&writer2, b"hello")
-        });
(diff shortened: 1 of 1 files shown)


suzuke:
§3.12 VERIFIED\n\nReviewed head: 76fda6db5141b22cfa40a1b463fba47f3ef57298\nNo blocking issues found.\n\n- pty_read_loop now calls handle_pty_close() on both EOF and read-error paths, closing the zombie-entry gap.\n- write_with_timeout now clears WRITE_IN_PROGRESS from the worker thread after completion, so a timeout no longer leaves the guard stuck forever.\n\nVERIFIED.

PR #1150, opened 2026-05-24
fix(#1146): release registry lock before inject_to_agent sleeps

### LINKED ISSUE #1146: send_to_registry holds registry mutex during inject_to_agent sleeps
## Problem

`agent.rs:1607-1614` (send) and `1639-1643` (broadcast) — `inject_to_agent` is called while holding the registry `MutexGuard`. For `typed_inject` agents, inject sleeps `2 * chunk.len()` ms per 64-byte chunk. A 10KB message = ~20 seconds of holding the lock.

During this time, ALL other threads block on registry access: API handlers, PTY read loops, TUI attach, health checks.

`broadcast_registry` repeats per-target serially, so N typed_inject targets = N × 20s.

## Impact

Fleet-wide freeze during broadcasts to typed_inject backends (kiro-cli).

## Fix

Clone needed fields (pty_writer, inject_prefix, submit_key, typed_inject) from the handle before releasing the lock, then call inject with cloned fields.

## Source

Agent & Backend review — agent.rs — finding H2.

suzuke:
## Summary

- **Problem**: `send_to_registry` and `broadcast_registry` held the registry mutex during `inject_to_agent` calls. Typed-inject backends (kiro, gemini, opencode, agy) sleep 2ms per 64-byte chunk — a 10KB message blocks the registry for ~20s. `broadcast_registry` previously re-acquired the lock per-target, so N typed-inject targets blocked N × 20s.
- **Fix**: Introduce `InjectTarget` struct that snapshots the 4 fields inject needs (`pty_writer` Arc, `inject_prefix`, `submit_key`, `typed_inject`). Both functions now clone under lock, release, then call `inject_with_target` — structurally impossible to hold the lock during inject.
- **`inject_to_agent`** preserved as a thin wrapper (delegates to `inject_with_target` via `InjectTarget::from_handle`) so existing call sites are unaffected.

## Files changed

- `src/agent.rs` — 117 insertions, 41 deletions

## Test plan

- [x] `cargo test --bin agend-terminal` — 2775 tests pass
- [x] `cargo test --tests` — 6 integration tests pass
- [x] `send_to_registry_releases_lock_before_inject_1146` — source-grep pin: lock scope closes before inject call
- [x] `broadcast_registry_releases_lock_before_inject_1146` — source-grep pin: no lock re-acquisition after release, inject after lock scope

Fixes #1146

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -926,7 +926,6 @@ fn pty_read_loop(
                         "[pty_read {name}] EOF after {read_count} reads, {total_bytes} bytes"
                     );
                 }
-                handle_pty_close(name, registry, home, crash_tx, shutdown, deleted);
                 break;
             }
             Ok(n_bytes) => {
@@ -987,10 +986,16 @@ fn pty_read_loop(
                         "[pty_read {name}] ERR after {read_count} reads, {total_bytes} bytes: {e}"
                     );
                 }
+                tracing::warn!(agent = name, error = %e, "PTY read error, triggering cleanup");
                 break;
             }
         }
     }
+
+    // #1144: handle_pty_close runs after BOTH exit paths (EOF and read error).
+    // Previously only the Ok(0) branch called it; the Err branch broke without
+    // cleanup, leaving a zombie agent in the registry.
+    handle_pty_close(name, registry, home, crash_tx, shutdown, deleted);
 }
 
 /// Sprint 21 F-NEW1: kill the process tree rooted at the agent's child PID,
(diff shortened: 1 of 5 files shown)


suzuke:
§3.12 REJECTED\n\nReviewed head: 718e919ec699150f06bb4051c782aac3c4241b8e\n\nBlocking issue:\n- inject_with_target() only checks target.deleted once at entry. The typed-inject loop and submit delay do not re-check it, so a send that started before delete can continue writing after deletion has begun. That still leaves a stale-delivery window across delete + reuse.\n- This patch suppresses sends that begin after delete, but it does not stop in-flight sends. That is not enough to restore the old safety boundary the registry lock provided.\n\nREJECTED.

suzuke:
§3.12 REJECTED\n\nReviewed head: c1cf81965b37a45c9fb76dc0625d2d426803a6ea\n\nBlocking issue:\n- inject_with_target() still only checks target.deleted once at entry. The typed-inject loop and submit delay do not re-check it, so a send that started before delete can continue writing after deletion has begun. That still leaves a stale-delivery window across delete + reuse.\n- This patch suppresses sends that begin after delete, but it does not stop in-flight sends. That is not enough to restore the old safety boundary the registry lock provided.\n\nREJECTED.

PR #1154, opened 2026-05-25
fix(agent): per-chunk deleted check in inject_with_target (#1146 follow-up)

suzuke:
## Summary
- Adds `deleted.load(Acquire)` check before each 64-byte chunk in the typed_inject loop, so in-flight injects stop at the next chunk boundary when the agent is deleted
- Adds the same check before the submit key write, preventing a stale Enter keystroke after deletion
- Follow-up to PR #1150 which introduced `InjectTarget` with an initial deleted check at function entry

## Test plan
- [x] Existing `inject_with_target_skips_deleted_agent_1146` test passes (covers entry-point check)
- [x] `cargo test --bin agend-terminal -- agent::tests` — 50 passed, 0 failed

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -1599,6 +1599,9 @@ fn inject_with_target(target: &InjectTarget, text: &[u8]) -> crate::error::Resul
         }
 
         for chunk in chunk_part.chunks(64) {
+            if target.deleted.load(std::sync::atomic::Ordering::Acquire) {
+                return Ok(());
+            }
             write_with_timeout(&target.pty_writer, chunk)?;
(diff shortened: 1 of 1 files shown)


suzuke:
§3.12 REJECTED\n\nReviewed head: d3f7f1da769fda5dd7d3bec2944a39f249287aff\n\nBlocking issue:\n- The new deleted checks are still sampled only before each write call. The write itself is not cancellable, so a delete that lands after the check but before write_with_timeout() can still write into the stale PTY.\n- The same remaining window exists in the non-typed branch, where the entire combined payload is written as one unit after the initial deleted check.\n- This narrows the race, but it does not restore the registry lock's named-delivery linearization boundary.\n\nREJECTED.

Issue #1157, opened 2026-05-25
Agent & Backend medium batch: unbounded channel, thread::spawn, CapabilityMatrix, dead code, metadata cleanup


suzuke:
## Findings from Agent & Backend code review

7 medium findings:

### agent.rs
1. `subscribe_with_dump` uses unbounded channel (no backpressure for TUI attach)
2. `try_dismiss_dialog` uses bare `std::thread::spawn` instead of `Builder`

### backend.rs + backend_harness.rs
3. `CapabilityMatrix::new()` only covers 4/6 backends (missing OpenCode + Agy)
4. `command_string()` is dead code with `#[allow(dead_code)]`
5. `from_command` prefix matching is looser than `parse_str`, undocumented

### agent_ops.rs
6. Instance delete only cleans legacy `metadata/<name>.json`, not id-based `metadata/<InstanceId>.json`

### backend_harness.rs (test-only)
7. `probe_esc_stops_generation` reader doesn't set non-blocking

## Source

Agent & Backend architecture group review.

PR #1159, opened 2026-05-25
fix(agent): bounded subscriber, Builder spawn, id-based metadata cleanup (#1157)

suzuke:
## Summary
- `subscribe_with_dump`: `crossbeam_channel::unbounded()` → `bounded(1024)` — prevents unbounded memory growth when a TUI subscriber stalls (matches router subscriber at `agent.rs:794`)
- `try_dismiss_dialog`: bare `std::thread::spawn` → `Builder::new().name("dismiss-dialog")` — adds thread name for debuggability, returns `Result` instead of panicking on spawn failure, cleans in-flight guard on error
- `cleanup_working_dir` + `full_delete_instance`: also removes `metadata/<InstanceId>.json` (Sprint 46 P2 symlink/copy) on instance deletion — two-site fix: best-effort fleet.yaml lookup in `cleanup_working_dir` for deployment callers, pre-extraction of id in `full_delete_instance` for the main delete path where fleet.yaml is already removed

## Test plan
- [x] `cargo clippy --all-targets` — clean
- [x] `cargo test` — all pass
- [ ] CI green

### DIFF
--- a/src/agent.rs
+++ b/src/agent.rs
@@ -1392,7 +1392,9 @@ pub fn try_dismiss_dialog(
             let agent = name.to_string();
             // fire-and-forget: dialog-dismiss keystroke writer is short-lived
             // (sleep 300ms then write). H2: removes from in-flight set on exit.
-            std::thread::spawn(move || {
+            if std::thread::Builder::new()
+                .name("dismiss-dialog".into())
+                .spawn(move || {
                 std::thread::sleep(std::time::Duration::from_millis(300));
                 // Send keys in chunks split on \r/\n boundaries with delay between,
                 // so TUI frameworks process navigation before confirmation.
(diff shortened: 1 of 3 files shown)


suzuke:
§3.12 VERIFIED\n\nReviewed head: 0a090e78300e1790f3c623a4e9f9d49f09818218\nNo blocking issues found.\n\n- dismiss-dialog now uses Builder::spawn with spawn-failure cleanup, which is a safe error-path improvement.\n- bounded subscriber channels are consistent with the actual consumers that drain them, and the PTY read fan-out path remains healthy.\n- id-based metadata cleanup in delete flows closes the orphan file gap without disturbing the legacy name-based cleanup.\n- backend harness only re-centers the matrix on the supported PTY-mechanism backends.\n\nVERIFIED.

PR #1165, opened 2026-05-25
fix(binding): propagate acquire_file_lock error in bind_full (#1163)

### LINKED ISSUE #1163: bind_full ignores file lock acquisition failure
## Problem

`binding.rs:55-72` — `acquire_file_lock()` result is ignored. If the lock cannot be acquired, `bind_full()` still proceeds to `atomic_write()` and returns success.

## Impact

Concurrent lease/bind operations from multiple agents can overwrite each other's binding.json without serialization.

## Fix

Check the lock result. If lock acquisition fails, return an error instead of proceeding.

## Source

Store & Data Layer review — binding.rs — finding H1.

suzuke:
## Summary
- `bind_full()` ignored the `Result` from `acquire_file_lock()`, silently proceeding to write `binding.json` without holding the lock when lock acquisition failed
- Fix: propagate the error with `?` so callers see the failure
- Add characterization test that forces lock failure (plants a directory where the lock file should be) and asserts `bind_full` returns `Err` without writing `binding.json`

Closes #1163

## Test plan
- [x] `cargo fmt --check` passes
- [x] `cargo clippy --all-targets -- -D warnings` passes
- [x] `cargo test --bin agend-terminal -- binding` — 58 passed, 0 failed
- [x] New test: `bind_full_propagates_lock_error_1163` verifies error propagation + no binding.json written

### DIFF
--- a/src/binding.rs
+++ b/src/binding.rs
@@ -52,7 +52,8 @@ pub fn bind_full(
     std::fs::create_dir_all(&dir).map_err(|e| format!("create_dir_all {}: {e}", dir.display()))?;
     let path = dir.join("binding.json");
     let lock_path = dir.join(".binding.json.lock");
-    let _lock = crate::store::acquire_file_lock(&lock_path);
+    let _lock = crate::store::acquire_file_lock(&lock_path)
+        .map_err(|e| format!("acquire_file_lock {}: {e}", lock_path.display()))?;
     let wt_str = worktree.display().to_string();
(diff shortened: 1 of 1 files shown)


suzuke:
§3.12 mirror: VERIFIED

`bind_full` now propagates `acquire_file_lock` failures instead of silently proceeding, so the binding/lease serialization guarantee is preserved. The added regression test covers the failed-lock path and confirms `binding.json` is not written.

Issue #1162, opened 2026-05-25
merge_instance_into_existing drops topic_binding_mode field


suzuke:
## Problem

`fleet.rs` — `merge_instance_into_existing` doesn't copy `topic_binding_mode` from the new instance into the existing one. `build_instance_mapping` writes it correctly, but the merge path silently drops it.

## Impact

Updating an existing instance's topic_binding_mode via fleet.yaml edit has no effect — the field is lost on merge.

## Fix

One-line: add `existing.topic_binding_mode = new.topic_binding_mode;` in the merge function.

## Source

Store & Data Layer review — fleet.rs — finding H1.

suzuke:
Fixed in PR #1166

PR #1166, opened 2026-05-25
fix(fleet): include topic_binding_mode in merge path (#1162)

suzuke:
## Summary
- `merge_instance_into_existing` was missing `topic_binding_mode` in its `merge_string` field list, causing the field to be silently dropped during template re-deployments to existing instances
- `build_instance_mapping` (new-instance path) already included it — only the merge path was affected
- Added characterization test that confirms the bug (fails before fix, passes after)

## Test plan
- [x] Characterization test `merge_preserves_topic_binding_mode_on_existing_instance` — asserts topic_binding_mode survives merge into existing instance
- [x] `cargo clippy --all-targets` — clean
- [x] `cargo test` — all pass
- [ ] CI green

### DIFF
--- a/src/fleet.rs
+++ b/src/fleet.rs
@@ -848,6 +848,7 @@ pub fn merge_instance_into_existing(
         ("model", &config.model),
         ("ready_pattern", &config.ready_pattern),
         ("command", &config.command),
+        ("topic_binding_mode", &config.topic_binding_mode),
     ] {
         merge_string(existing, field, value)?;
     }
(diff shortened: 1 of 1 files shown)


suzuke:
§3.12 mirror: VERIFIED

`merge_instance_into_existing` now preserves `topic_binding_mode` for existing instances, which is the correct operator-hand-edit behavior. The regression test pins the merge path so the field is not dropped during fleet.yaml reconciliation.

PR #1168, opened 2026-05-25
fix(store): harden data layer — 8-fix medium batch (#1164)

### LINKED ISSUE #1164: Store & Data medium batch: mutate_fleet_yaml no-op write, corrupt config handling, dead_code allows
## Findings

1. fleet.rs H3: mutate_fleet_yaml writes even when mutation closure makes no changes
2. store.rs M1: versioned-store returns T::default() on parse failure, bypasses backup path  
3. mcp_config.rs M1: malformed config silently replaced with empty object, drops user settings
4. task_events.rs M1-M2: stale #[allow(dead_code)] module-level suppression
5. decisions.rs M4: post() calls Utc::now() twice, ID timestamp can differ from created_at

## Source

Store & Data Layer review — multiple files.

suzuke:
## Summary

Closes #1164 — 8 fixes across 6 files in the store & data layer:

- **fleet.rs**: bare `~` tilde expansion, merge log moved to success branch, `mutate_fleet_yaml` returns `Result<bool>` to skip no-op disk writes (all 9 callers updated)
- **task_events.rs**: module-level `#![allow(dead_code)]` replaced with 3 targeted allows
- **event_log.rs**: removed dead `append`/`append_batch` functions (only `append_lines_under_lock` is used)
- **decisions.rs**: merged double `Utc::now()` into single clock capture
- **store.rs**: `load_versioned` now backs up corrupt files before returning default (matching `load()` pattern)
- **mcp_config.rs**: `upsert_mcp_servers` warns + backs up malformed JSON instead of silent `unwrap_or(json!({}))`

## Test plan

- [x] `cargo check` — compiles clean
- [x] `cargo clippy --all-targets` — zero warnings
- [x] `cargo test` — all 42 tests pass
- [x] Verified all 9 `mutate_fleet_yaml` callers return correct `bool`

### DIFF
--- a/src/decisions.rs
+++ b/src/decisions.rs
@@ -121,14 +121,15 @@ pub fn post(home: &Path, author: &str, args: &Value) -> Value {
     let ttl_days = args["ttl_days"].as_u64();
     let supersedes = args["supersedes"].as_str().map(String::from);
 
-    let now = chrono::Utc::now().to_rfc3339();
+    let clock = chrono::Utc::now();
+    let now = clock.to_rfc3339();
     // The historical id format was seconds-precision only — two posts in the
     // same UTC second collided and the second silently overwrote the first.
     // Append nanoseconds + a process-local counter so no two posts from the
     // same process can share an id, even when issued back-to-back.
     use std::sync::atomic::{AtomicU64, Ordering};
     static ID_SEQ: AtomicU64 = AtomicU64::new(0);
-    let ts = chrono::Utc::now().format("%Y%m%d%H%M%S%6f");
+    let ts = clock.format("%Y%m%d%H%M%S%6f");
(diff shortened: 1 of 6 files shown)


suzuke:
§3.12 mirror: REJECTED

Findings:
1. `worktrees::sweep()` snapshots GC candidates and archives them later without rechecking binding state or holding the same lock used by binding writes. A worktree can be rebound after enumeration and still get moved to `.trash`.
2. `.trash/worktrees/<agent>-<unix_ts>` is second-granularity and collides when one agent has multiple stale worktrees in the same sweep second. The later archive can fail or overwrite the earlier archive.
3. `instance_monitor.rs` still reads `metadata/<name>.json` directly, so it misses id-based metadata and can report false-negative heartbeat/pending-pickup values.

`cargo check -q --locked` passed on the PR head.

PR #1173, opened 2026-05-25
fix(retention): re-validate binding before archive + collision-free trash paths (#1170)

### LINKED ISSUE #1170: retention sweep TOCTOU + .trash archive path collision
## Problem

Two bugs in the worktree retention/GC system:

### 1. Retention sweep TOCTOU (daemon/retention/worktrees.rs)
`sweep()` snapshots `gc_candidates()` then archives each without re-reading binding state. A worktree released → becomes candidate → rebound before archive can still be moved to `.trash` because the sweep doesn't hold or re-check the binding lock.

### 2. .trash archive path collision
`maybe_remove_candidate()` uses `.trash/worktrees/<agent>-<unix_ts>` as destination. `unix_ts()` is second-granularity, but `gc_candidates()` can return multiple branches for the same agent. Two archives for same agent in one second target the same path — later overwrites or fails against earlier.

## Fix
1. Re-validate binding state immediately before archive rename
2. Add sub-second granularity (nanos or sequence counter) to archive path

Found in architecture group review (Instance & Worktree, Slice C).

suzuke:
## Summary
- **TOCTOU fix**: `maybe_remove_candidate()` now re-checks `binding::read()` immediately before `try_archive()`. A worktree rebound between `gc_candidates()` enumeration and archive is skipped with reason `rebound_since_enumeration`.
- **Collision fix**: Replaced `unix_ts()` (second-granularity) with `archive_ts()` (nanosecond-granularity: `<secs>-<nanos_09d>`). Multiple worktrees for the same agent archived in the same sweep tick now get distinct `.trash` paths.

## Test plan
- [x] T8: rebound worktree skipped by binding recheck
- [x] T9: same-agent same-second produces two distinct .trash entries
- [x] T10: `archive_ts()` returns nanosecond-precision format
- [x] All 14 retention worktree tests pass
- [x] Full test suite passes (2798 tests)

Closes #1170

### DIFF
--- a/src/daemon/retention/worktrees.rs
+++ b/src/daemon/retention/worktrees.rs
@@ -52,11 +52,11 @@ fn trash_retention_days() -> u64 {
         .unwrap_or(7)
 }
 
-fn unix_ts() -> u64 {
-    SystemTime::now()
+fn archive_ts() -> String {
+    let d = SystemTime::now()
         .duration_since(UNIX_EPOCH)
-        .map(|d| d.as_secs())
-        .unwrap_or(0)
+        .unwrap_or(Duration::ZERO);
+    format!("{}-{:09}", d.as_secs(), d.subsec_nanos())
 }
 
 /// Recursive copy helper used by the `EXDEV` fallback in `try_archive`.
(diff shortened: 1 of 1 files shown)


suzuke:
§3.12 VERIFIED

The retention sweep fix closes the two prior blockers:
- rebound worktrees are skipped by a binding recheck immediately before archive
- .trash names now use sub-second precision to avoid same-agent collisions

Validation: cargo test -q worktree_cleanup -- --nocapture passed on reviewed head 828cfc27d3227d1ef9610b118a21443ef1e8dcc0

PR #1259, opened 2026-05-26
perf(tasks): replay cache + Schwartzian sort transform

suzuke:
## Summary
- **1A: Replay cache** — `replay()` now caches `TaskBoardState` keyed on hot-log mtime+len and archive-dir mtime. `append_batch()` explicitly invalidates. Eliminates redundant full-file replay across 15+ call sites.
- **1B: Schwartzian sort** — `sort_envelopes()` pre-parses RFC3339 timestamps once into parallel i64 key vec, then sorts by pre-computed keys. Reduces ~22000 chrono::parse calls to 2000 for 2000 events.

## Changed files
- `src/task_events.rs` — +167/-13 lines: replay cache statics, `replay()` split into `replay()`+`replay_uncached()`, `invalidate_replay_cache()`, Schwartzian `sort_envelopes()` + `apply_permutation()`, 3 new tests

## Test plan
- [x] `replay_cache_hit_returns_same_result` — consecutive replays return identical state
- [x] `replay_cache_invalidated_after_append` — append invalidates cache, next replay sees new task
- [x] `sort_envelopes_schwartzian_matches_naive` — sort order matches original (ts → instance → seq)
- [x] All 31 existing task_events tests pass
- [x] `cargo fmt -- --check` clean
- [ ] CI green

### DIFF
--- a/src/task_events.rs
+++ b/src/task_events.rs
@@ -1006,27 +1006,8 @@ fn sort_envelopes(envelopes: &mut [TaskEventEnvelope]) {
             .then_with(|| envelopes[a].instance.0.cmp(&envelopes[b].instance.0))
             .then_with(|| envelopes[a].seq.cmp(&envelopes[b].seq))
     });
-    apply_permutation(envelopes, &mut keys, &indices);
-}
-
-fn apply_permutation<T>(data: &mut [T], _keys: &mut [i64], perm: &[usize]) {
-    let mut placed = vec![false; data.len()];
-    for i in 0..data.len() {
-        if placed[i] || perm[i] == i {
-            continue;
-        }
-        let mut j = i;
-        loop {
-            let next = perm[j];
-            if next == i {
-                break;
-            }
-            data.swap(j, next);
-            placed[j] = true;
-            j = next;
-        }
-        placed[j] = true;
-    }
+    let gathered: Vec<TaskEventEnvelope> = indices.iter().map(|&i| envelopes[i].clone()).collect();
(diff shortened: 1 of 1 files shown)


suzuke:
§3.12 mirror

REJECTED

Blocking findings:

1) Incorrect permutation application can mis-sort envelopes (correctness)
- File: `src/task_events.rs` line ~1012 (`apply_permutation`) with caller at ~1002-1009.
- `indices` is built as "new-order -> old-index" mapping, but `apply_permutation` applies swaps as if `perm[i]` were destination index. These are different permutation representations.
- Counterexample: `perm = [2,0,1]` should produce order `[old2, old0, old1]`, but current algorithm yields `[old2, old1, old0]`.
- Impact: replay ordering can be wrong for some envelope sets, violating deterministic fold semantics.

Requested fix direction:
- Either convert to the inverse mapping before in-place cycle swaps, or use a simpler safe reorder path (e.g. allocate reordered vec by `indices` then replace), then add a unit test that includes a non-trivial 3-cycle like `[2,0,1]`.

suzuke:
§3.12 mirror

VERIFIED

- Re-reviewed HEAD 15db24c3b0911682dde07c2e661261eeee472040 after CI all-green (6/6).
- Previous blocking issue is resolved: sort reorder path no longer uses mismatched permutation semantics; it now gathers by sorted indices directly.
- Added tests cover reverse-order and interleaved same-timestamp ordering behavior.

No blocking findings.

PR #1260, opened 2026-05-26
perf(fleet): mtime-based cache for FleetConfig::load()

suzuke:
## Summary
- `FleetConfig::load()` now checks file mtime before reading; unchanged files return the cached config without disk I/O or YAML parsing
- Global `FLEET_CACHE` (Mutex-guarded) stores path + mtime + parsed config
- `load_uncached()` extracted as the original parse path for clarity
- `clear_cache()` test helper for test isolation

## Context
Performance audit found fleet.yaml being re-read and re-parsed on every daemon tick (10s) and every TUI render frame (50ms) across 14+ call sites. YAML parsing is significantly slower than JSON. This cache eliminates redundant reads when the file hasn't changed.

## Test plan
- [x] `load_cache_returns_same_result_on_unchanged_file` — second load returns same data without re-reading
- [x] `load_cache_invalidates_on_file_change` — rewrite with sleep(1.1s) for mtime granularity, verify new content returned
- [x] All existing fleet tests pass (no regressions)
- [x] cargo fmt + clippy clean

### DIFF
--- a/src/fleet.rs
+++ b/src/fleet.rs
@@ -3,6 +3,24 @@ use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
+use std::time::SystemTime;
+
+/// Mtime-based cache for `FleetConfig::load()`.
+/// Avoids repeated `read_to_string` + `serde_yaml_ng` parse when the
+/// file hasn't changed on disk.
+static FLEET_CACHE: std::sync::Mutex<Option<FleetCacheEntry>> = std::sync::Mutex::new(None);
+
+struct FleetCacheEntry {
+    path: PathBuf,
+    mtime: SystemTime,
+    config: FleetConfig,
+}
+
+#[cfg(test)]
+pub(crate) fn clear_cache() {
+    let mut guard = FLEET_CACHE.lock().unwrap_or_else(|e| e.into_inner());
+    *guard = None;
+}
 
 /// Single source of truth for the fleet configuration filename.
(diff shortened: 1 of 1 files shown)


suzuke:
§3.12 mirror

REJECTED

Blocking findings:

1) Cache key relies only on mtime for freshness, so stale reads are possible on coarse-timestamp filesystems
- File: `src/fleet.rs` lines ~287-293 and ~300-306.
- Cache hit condition is `entry.path == path && entry.mtime == disk_mtime`; if fleet.yaml is rewritten within the same mtime tick, `FleetConfig::load()` can return stale pre-write config.
- The added test itself uses `sleep(1100ms)` to force mtime change, which implicitly acknowledges mtime granularity sensitivity.
- Impact: correctness regression (load may not reflect latest fleet mutation immediately).

Requested fix direction:
- Strengthen cache key (e.g. mtime + len, or content hash), and/or explicitly invalidate cache on in-process fleet mutation write paths (`mutate_fleet_yaml` callers).
- Add a test that rewrites fleet.yaml without relying on wall-clock sleep and proves stale-hit cannot happen.

suzuke:
§3.12 mirror

VERIFIED

- Reviewed new HEAD 06b400ef1de524b2f861a677295cb59d626c0214 after CI all-green (6/6).
- Prior blocking concern addressed:
  1. Cache key strengthened from mtime-only to `(mtime, size)`
  2. Fleet mutation path (`atomic_write_yaml`) now invalidates cache explicitly
- Added tests cover unchanged-file hit, file-change invalidation, same-mtime/different-size rewrite, and mutate-path invalidation.

No blocking findings.

PR #1357, opened 2026-05-28
fix: replace replay cache mtime key with generation counter (#1355)

### LINKED ISSUE #1355: fix: flaky test_concurrent_creates_unique_ids — RCA needed
## Problem

`test_concurrent_creates_unique_ids` intermittently fails on Windows CI and Coverage job, requiring CI reruns. Pre-existing issue, not introduced by any recent PR.

## Impact

- Wastes CI time (need to rerun)
- Blocks PR merge flow
- Has been observed across multiple PRs in this session

## Action Needed

1. RCA: Why does this test fail intermittently? Is it a timing issue, thread contention, or ID generation race?
2. Fix the underlying issue or make the test deterministic

suzuke:
## Summary
- Replace `replay_cache_key`'s mtime-based staleness detection with a process-wide `AtomicU64` generation counter
- `invalidate_replay_cache()` now increments the generation instead of clearing the cache
- Fixes flaky `test_concurrent_creates_unique_ids` on Linux CI where ext4 mtime granularity (1ms) caused stale cache hits during rapid concurrent appends

## Test plan
- [x] All local tests pass (macOS)
- [ ] `test_concurrent_creates_unique_ids` no longer flaky on Linux CI (Coverage job)
- [ ] Existing replay cache tests pass
- [ ] No performance regression — cache still effective for non-concurrent read patterns

Closes #1355

### DIFF
--- a/src/task_events.rs
+++ b/src/task_events.rs
@@ -941,14 +941,15 @@ fn max_seq_for_instance(log_path: &Path, instance: &InstanceName) -> anyhow::Res
 
 // ── Replay cache ─────────────────────────────────────────────────────
 // Read-side cache: avoids full-file replay when nothing has changed.
-// Keyed on (home, generation) where generation is a process-wide
-// monotonic counter incremented on every append. This replaces the
-// previous mtime-based key which was unreliable on Linux ext4 where
-// mtime granularity (1ms) is too coarse for rapid concurrent writes.
+// Keyed on (home, generation, file_len) — generation is a process-wide
+// monotonic counter incremented on every in-process append (replaces
+// mtime which was unreliable on Linux ext4 where 1ms granularity
+// caused stale cache hits during rapid concurrent writes). file_len
+// catches external modifications (tests, compaction, manual edits).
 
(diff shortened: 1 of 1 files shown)


suzuke (commented):
§3.12 Mirror
Verdict: UNVERIFIED

Blocking issue:
- New replay cache key uses `(home, generation, file_len)` only. External edits to `task-events.jsonl` that keep identical file length are no longer detectable, so `replay()` can return stale cached state.

Why this blocks:
- Replay correctness must not depend on file-length changes; same-length rewrites are valid and currently bypass invalidation.

Suggested fix:
- Add a content-sensitive signal to the key for external writes (e.g. mtime + len + generation, or len + stable hash/sample of tail bytes) while retaining generation to avoid ext4 mtime granularity races.

suzuke (commented):
§3.12 Mirror
Verdict: VERIFIED

Re-review result:
- Previous blocker is fixed: replay cache key now includes both `file_len` and `mtime_ns` in addition to `generation`.
- This preserves in-process append safety (generation counter) and restores external-write detection beyond length-only.
- CI is fully green.

No remaining blockers.

PR #1396, opened 2026-05-28
refactor(#1380): introduce AgentName newtype for agent/instance names

suzuke:
## Summary

- Define `AgentName(String)` in `types.rs` with `Deref<str>`, `Borrow<str>`, `From<String/&str>`, `Display`, `AsRef<str>`, serde transparent
- Replace `String` with `AgentName` at two definition points: `AgentHandle.name` and `Pane.agent_name`
- Update 20 consuming files with minimal `.into()` / `.to_string()` / `.as_str()` conversions

Follow-up scope: migrate `AgentRegistry` key type from `String` to `AgentName`, replace `fleet_instance_name: Option<String>` etc.

## Test plan

- [x] 311 agent/pane/layout/render tests pass
- [x] Zero behavioral change — all conversions are mechanical

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,7 +10,7 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "agend-terminal"
-version = "0.6.1"
+version = "0.7.0"
 dependencies = [
  "alacritty_terminal",
  "anyhow",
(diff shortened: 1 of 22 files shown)


suzuke (commented):
§3.12 Mirror
Verdict: VERIFIED

Review result:
- `AgentName` newtype introduction is consistent: `serde(transparent)` preserves wire format, and `Deref/Borrow/AsRef` keep map lookup/comparison ergonomics.
- Call sites across app/layout/render/daemon paths were updated coherently (`as_str()` for lookups/comparisons, `to_string()` when string ownership is required).
- Change scope is type-safety/clarity oriented with no functional behavior drift observed.
- CI checks are fully green.

No blocking issues.

PR #1407, opened 2026-05-28
fix(#1405): AGENTS.md merge no longer duplicates extra instructions

suzuke:
## Summary

- Move `extra_instructions` (from fleet.yaml `instructions:` field) inside the agend marker block so it gets replaced on refresh instead of appended
- Add `strip_leaked_extra_instructions` for self-healing: removes prior leaked copies outside the markers on next refresh

**Root cause**: extra instructions were appended after `merge_agend_block` returned, placing them outside `<!-- agend:start/end -->` markers. Each refresh treated them as user content and preserved them.

**Before**: fixup-reviewer AGENTS.md had "# Coding Discipline" 7 times
**After**: exactly 1 occurrence, inside the agend block

## Test plan

- [x] 47 instructions tests pass
- [x] Existing duplicated AGENTS.md self-heals on next refresh

### DIFF
--- a/src/instructions.rs
+++ b/src/instructions.rs
@@ -318,6 +318,40 @@ pub(crate) fn build_instructions_body(
     content
 }
 
+/// Remove leaked copies of `extra_instructions` that previous versions
+/// appended outside the agend markers (#1405 self-heal). Returns the
+/// content with all exact occurrences of the extra text stripped from
+/// outside the marker block.
+fn strip_leaked_extra_instructions(content: &str, extra: Option<&str>) -> String {
+    let extra = match extra {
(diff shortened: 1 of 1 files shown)


suzuke (commented):
§3.12 Mirror
VERDICT: VERIFIED

Scope audited:
- moved fleet `extra_instructions` injection into agend-managed block generation
- shared-file merge path now self-heals legacy leaked duplicates outside markers
- cleanup helper only operates after `AGEND_BLOCK_END`, preserving user preamble/body

Checks:
- Change is narrow to instruction generation path and aligns with #1405 intent

Result: VERIFIED.

PR #1411, opened 2026-05-28
fix(#1366): replace_instance async teardown — remove 5s sync wait + 500ms sleep

### LINKED ISSUE #1366: replace_instance: synchronous child-exit wait + hardcoded 500ms sleep
## Problem

`replace_instance` blocks synchronously during the DELETE→respawn cycle, with two stacked delays:

1. **DELETE synchronous wait** — calls `crate::api::method::DELETE` which runs `wait_for_child_exit` (up to 5s, see #1363)
2. **Hardcoded 500ms sleep** — an additional fixed delay before re-spawn

`src/mcp/handlers/instance.rs` (~line 288-296):

```rust
// Kill via DELETE (synchronous — waits for child exit + removes from registry).
let _ = crate::api::call(home, &json!({"method": DELETE, "params": {"name": name}}));

// TODO: replace hardcoded 500ms sleep with event-driven readiness check
// (e.g. poll until PID is gone or port is free).
std::thread::sleep(std::time::Duration::from_millis(500));
```

The code's own TODO acknowledges the sleep is a placeholder.

## Impact

`replace_instance` is typically used by an orchestrator to swap out a wedged agent. The 5.5s+ stall is less critical than interactive pane close (#1363), but it's the same class of issue: synchronous wait on child process teardown.

## Fix

- Reuse the async teardown from #1363 for the DELETE step
- Replace the 500ms sleep with an event-driven readiness check (poll until PID gone / port free, or await a teardown-complete signal)

## Related

- #1363 — TUI pane/tab close blocks on the same `wait_for_child_exit`

suzuke:
## Summary
- Add `skip_exit_wait` parameter to `delete_transaction` — when `true`, the kill signal is sent but `wait_for_child_exit` (up to 5s) is skipped; the OS reaps the child asynchronously
- `replace_instance` MCP handler passes `no_wait: true` to the DELETE API call, removing the 5.5s+ blocking stall
- Remove the hardcoded 500ms `thread::sleep` placeholder (the TODO it replaced)
- All other callers (`kill_agent`, rollback paths) pass `false` to preserve existing synchronous behavior

Closes #1366

## Test plan
- [ ] Verify `replace_instance` completes in <1s instead of 5.5s+
- [ ] Verify normal `delete_instance` (without `no_wait`) still waits for child exit
- [ ] Verify app-mode `kill_agent` still synchronously waits
- [ ] Existing `delete_transaction` unit tests pass (updated to new signature)
- [ ] CI green on all platforms

### DIFF
--- a/src/api/handlers/instance.rs
+++ b/src/api/handlers/instance.rs
@@ -97,7 +97,14 @@ pub(crate) fn handle_delete(params: &Value, ctx: &HandlerCtx) -> Value {
     // the previous implementation removed the registry entry before the OS
     // had reaped the PID, exposing PID re-use + concurrent-spawn collision
     // races.
-    crate::daemon::lifecycle::delete_transaction(ctx.home, name, ctx.registry, Some(ctx.configs));
+    let skip_exit_wait = params["no_wait"].as_bool().unwrap_or(false);
+    crate::daemon::lifecycle::delete_transaction(
+        ctx.home,
+        name,
+        ctx.registry,
+        Some(ctx.configs),
+        skip_exit_wait,
+    );
     // H3: clean up poll_reminder dedup state for deleted agent
     crate::daemon::poll_reminder::remove_agent(name);
     if let Some(n) = ctx.notifier {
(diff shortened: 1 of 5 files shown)


suzuke (commented):
§3.12 Mirror
VERDICT: VERIFIED

Scope audited:
- `delete_transaction` now supports `skip_exit_wait` and is defaulted to wait in all existing call-sites
- `replace_instance` path opts into `no_wait` and removes fixed 500ms sleep
- API delete handler plumbs `no_wait` flag explicitly

Checks:
- Legacy safety behavior preserved for non-replace paths (`skip_exit_wait=false`)
- Change is localized to #1366 latency path

Result: VERIFIED.

PR #1435, opened 2026-05-28
fix: replace_instance returns the new pane to the original tab (#1431)

suzuke:
## Summary
`replace_instance` (DELETE + SPAWN) landed the fresh pane in a **new tab**, forcing operators to `move_pane` it back (#1431). This returns the new pane to the tab the replaced pane occupied.

## Why TUI-side (not the issue's daemon-side approach)
The issue suggested recording the tab name daemon-side before DELETE. That's infeasible: `replace_instance` is a daemon-side MCP handler with **no access to the TUI pane layout** (layout lives only in the TUI process; LIST/API expose no pane→tab mapping). So the capture/restore must happen TUI-side, signalled by a layout hint.

## Change
- `api::LayoutHint` + the TUI-side `LayoutHint` gain a `SameTab` variant (parsed from `"same-tab"`).
- `replace_instance`'s SPAWN now passes `layout=same-tab`.
- `Layout` keeps a small `agent → tab-name` map: **recorded** when a pane is removed (`handle_instance_deleted`), **consumed** by a `SameTab` create (`handle_instance_created`). Normal (`Tab`) spawns are unaffected — scope stays replace-only.
- The new pane splits back into the original tab if it still exists; otherwise a tab with the original name is recreated (the single-pane tab is closed on removal).

## Caveat
Same **tab** is guaranteed; the exact **split slot** is best-effort — the old pane's tree leaf is destroyed when it's removed, so a multi-pane tab gets a fresh split rather than the identical position. (Confirmed with the lead: same-tab guaranteed, split best-effort.)

## Validation note
This is daemon+TUI behavior, so it **can't be dogfood-validated on its own PR** (needs next-PR + daemon restart). Covered instead by unit tests for the record/restore logic.

## Test plan
- [x] `cargo build` — clean
- [x] `cargo clippy --all-targets` — clean (no warnings)
- [x] `cargo test --bins` — **2802 passed, 0 failed, 8 ignored**
- [x] 4 new #1431 unit tests: recreate-closed-tab-by-name, split-into-surviving-tab, SameTab-never-anchors, empty-memory-None

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -88,6 +88,10 @@ pub enum LayoutHint {
     Tab,
     SplitRight,
     SplitBelow,
+    /// #1431: place the new pane in the tab the same-named pane occupied
+    /// before removal. Used by `replace_instance` so a replaced agent returns
+    /// to its original tab instead of opening a fresh one.
+    SameTab,
 }
 
 impl LayoutHint {
@@ -95,6 +99,7 @@ impl LayoutHint {
         match s {
             "split-right" => Self::SplitRight,
             "split-below" => Self::SplitBelow,
+            "same-tab" => Self::SameTab,
             _ => Self::Tab,
         }
     }
--- a/src/app/tui_events.rs
+++ b/src/app/tui_events.rs
@@ -5,7 +5,7 @@
 //! the layout accordingly — auto-creating or removing tabs/panes.
 
 use crate::agent::{self, AgentRegistry};
-use crate::layout::{Layout, MovePlacement, SplitDir, Tab};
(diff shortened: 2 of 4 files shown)


suzuke (commented):
§3.12 Mirror
VERDICT: VERIFIED

Scope audited:
- New layout hint `same-tab` added (`api::LayoutHint` + TUI mirror enum)
- replace_instance now spawns with `layout: "same-tab"`
- layout remembers removed pane's tab name and reattaches replacement to same tab

Checks:
- Tests cover: recreate-closed-tab, split-into-surviving-tab, same-tab anchor behavior, and no-record fallback
- Fallback behavior remains safe (no memory entry -> open fresh tab)

Result: VERIFIED.

PR #1439, opened 2026-05-29
fix: idle watchdog ack-wash — recover on board progress, not bare activity (#1438 fix 3)

suzuke:
## Summary
The fleet-idle **ack** cleared the moment *any* tracked agent's activity sidecar was newer than the ack timestamp. So `general` answering the operator during research (on-demand chatter, owning no task) washed out the ack, re-armed the alert, and re-fired ~30 min later — the validated **10+ false-positive loop** (#1438 fix 3, scoped per the dialectic: fix 1/2/C deliberately out of scope).

## Change (`idle_watchdog.rs` recovery logic, ~line 498-507)
Ack now releases only on a genuine "sprint resumed" signal — any of:
1. **Board progress** — a task claimed/advanced (`updated_at` past the ack).
2. **Worker heartbeat** — a task-**owning** agent resumed activity (covers board-update lag / off-board work). Scoped to owners so chatter can't trigger it — this is what keeps reviewer-1's dual-condition recovery from reintroducing the wash.
3. **Max TTL backstop** — ack older than `fleet_idle_ack_ttl_secs` (default 45 min, runtime-configurable) expires and the fleet is re-evaluated; *not* auto-alerted, because the existing all-idle + `has_expected_work` gates still apply (a finished board stays silent).

No fleet.yaml schema change, no role enum (per-instance opt-out already exists via `idle_watchdog_enabled`).

## Test anti-pattern fixed
The prior `fleet_scan_resumes_after_post_ack_activity` test asserted that bare post-ack activity clears the ack — it **encoded the ack-wash bug as if it were the spec**. Rewritten to resume on board progress; the "bare activity must NOT resume" contract is now pinned by `ack_survives_unrelated_agent_activity`.

## Validation note
Daemon behavior — **cannot be dogfood-validated on its own PR** (needs next-PR + daemon restart). Covered by the test matrix instead.

## Test plan
- [x] `cargo build` / `cargo clippy --all-targets` — clean
- [x] `cargo test --bins` — **2809 passed, 0 failed, 8 ignored**
- [x] 6-case matrix: ack survives unrelated activity (core regression) · releases on board progress · releases on task-owner heartbeat · expires after max TTL · done-board stays silent · (rewrite of the bug-encoding test)

### DIFF
--- a/src/daemon/idle_watchdog.rs
+++ b/src/daemon/idle_watchdog.rs
@@ -36,7 +36,7 @@
 //! for concurrency. Forward-compat preserved via `#[serde(default)]`.
 
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicI64, Ordering};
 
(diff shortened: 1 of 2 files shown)


suzuke (commented):
§3.12 Mirror
VERDICT: VERIFIED

Scope audited:
- Fleet-idle ack recovery condition tightened to avoid on-any-activity ack wash
- Recovery now requires one of: board progress, task-owner heartbeat, or ack TTL expiry
- Runtime config adds `fleet_idle_ack_ttl_secs` with get/set wiring

Checks:
- Regression/behavior tests cover: board-progress recovery, unrelated-activity no-recovery, owner-heartbeat recovery, TTL expiry, done-board no-alert, and dev-vantage independence

Result: VERIFIED.

Issue #819, opened 2026-05-15
TUI scrollback render: WIDE_CHAR_SPACER skip leaves stale chars in ratatui buffer (visual artifacts on prompt line)


suzuke:
## Symptom

Operator sees scattered/random characters in agent pane scrollback rendering — particularly on the input prompt line — that don't correspond to actual PTY content. **Selecting (highlighting) the affected region makes the artifacts disappear**: the act of selection triggers a buffer re-render that pulls authoritative content and overwrites the stale cells.

This has been observed in `fixup-lead`'s pane today and is described as recurring (\"指令列會不定時出現奇怪的東西\").

## Verified — NOT a PTY buffer issue

`mcp__agend-terminal__pane_snapshot` returns clean text for `fixup-lead` at the same moment the scattered chars are visible on screen. The agend-terminal stripping ANSI when assembling the snapshot doesn't matter here: `pane_snapshot` reads directly from the alacritty grid (authoritative), while the screen renders from the ratatui staging buffer. **The artifacts live only in the ratatui buffer, not in the underlying PTY grid.**

This means the bug is in `src/vterm.rs::render_to_buffer` — specifically how it reconciles ratatui's incremental buffer with alacritty's authoritative cell state.

## Trace — `src/vterm.rs:228-233` (the smoking gun)

```rust
for row in 0..rows {
    let mut col = 0u16;
    while col < cols {
        let idx = (row as usize) * snap_cols + (col as usize);
        let cell = snapshot.get(idx).unwrap_or_else(|| ...);
        if cell.flags.contains(Flags::WIDE_CHAR_SPACER) {
            col += 1;
            continue;     // ← skip without writing buf[(x,y)]
        }
        // ... cell rendering for non-spacer cases
        let buf_cell = &mut buf[(x, y)];
        buf_cell.set_char(ch).set_style(style);
        ...
    }
}
```

When alacritty's grid has a `WIDE_CHAR_SPACER` flagged cell with NO preceding `WIDE_CHAR` in the previous column (i.e. the wide char was overwritten by a narrow one but the spacer cell wasn't cleared), `render_to_buffer` skips the column. **Ratatui's staging buffer at `(x, y)` retains the previous frame's character.** The terminal's differential renderer doesn't update that cell on screen → stale char persists until something else triggers a redraw of that region.

[...]

PR #1443, opened 2026-05-29
fix(#1140): upgrade ratatui 0.29 → 0.30 to fix wide-char ghost artifact

### LINKED ISSUE #1140: TUI render artifact: stale single-char ghost persists across tab switches
## Problem

Occasional single-character render artifacts (e.g. a lone `b`) appear in pane areas that should be empty. The artifact:
- Persists across tab switches (not cleared by tab swap)
- Disappears when mouse-selecting (terminal emulator forces repaint)
- Appears at a fixed screen position regardless of which tab is active
- Not backend-specific (observed across different backends; this instance during kiro-cli activity)

## Screenshot

render artifact 
See: `~/Desktop/截圖 2026-05-24 23.25.52.png` — lone `b` char in fixup-dev pane, below normal output.

## Likely cause

Stale cell in terminal buffer. The TUI uses diff-based rendering (only redraws dirty cells). A byte that was incorrectly written to a cell position (possibly from incomplete escape sequence parsing) is never overwritten because subsequent frames do not mark that cell as dirty.

## Possible investigation paths

1. `src/vterm.rs` — VTE parser may emit a stray byte as printable when it should be consumed as part of an escape sequence
2. `src/render/` — diff renderer may miss marking cells dirty in edge cases (resize, pane swap, scroll boundary)
3. Backend PTY output — some backends may emit partial escape sequences that leave parser in unexpected state

## Reproduction

Intermittent. No reliable repro steps yet. Observed across multiple backends over time. More frequent during multi-agent parallel activity.

## Workaround

Mouse-select the area (forces terminal repaint), or resize the terminal window (triggers full redraw).

suzuke:
## Summary

Fixes #1140 (lone stale character / "ghost" artifacts that persist across tab switches). Root cause is in ratatui's buffer/diff layer, **not our code** — ratatui 0.29 miscalculated VS16-emoji width and didn't represent empty (spacer) cells properly, so overwriting a wide glyph with narrow chars left the spacer cell's ghost surviving the diff. ratatui **0.30** reworks this: VS16 wide emojis are now cleared from the buffer, and `Cell::symbol` became `Option<CompactString>` to represent empty cells correctly.

## Why upgrade (vs workaround)

Evaluated all three options with direct evidence:
- **Option 2 (upgrade) — chosen.** Root-cause fix lives upstream and 0.30 ships it. We're an application (not a widget author), so the `ratatui` facade crate preserves every import + Cell API we use. **Verified: one-line version bump, zero source changes** — `cargo check`, `cargo clippy --all-targets -- -D warnings`, and the full test suite all clean on 0.30.
- **Option 1 (full redraw):** already partially present (`mod.rs` clears on resize) — kept as cheap defense-in-depth — but only treats the symptom and has perf/flicker cost.
- **Option 3 (spacer-write):** rejected — we already write correct complete buffers every frame (#819); the bug is purely ratatui's diff deciding what to *send*, so option 3 would fight the library.

## Changes

- `Cargo.toml`: `ratatui = "0.30"` (+ `Cargo.lock`).
- `core_render.rs`: contract test (`TestBackend`, wide/VS16-emoji → narrow across frames) locking the no-ghost behavior.
- Retained the resize-time `terminal.clear()` full redraw as defense-in-depth.

**Honest note on the test:** plain CJK wide→narrow already worked on 0.29; the #1140 ghost came from VS16-emoji width (a *layout-level* difference), which can't be cleanly discriminated cross-version in a unit test. The emoji case documents the correct 0.30 behavior and guards future regressions.

## ⚠️ Needs operator visual verification

This is a **major dependency bump** — 0.30 also rewrote line/span wide-char truncation. Compile + tests passing does **not** prove identical runtime rendering. **An operator should confirm on a real terminal that (a) the ghost is gone and (b) there are no other render regressions** (status bar / tab bar / pane content / CJK + emoji) before merge.

## Test plan

- [x] `cargo fmt`
- [x] `cargo clippy --all-targets -- -D warnings` (clean on 0.30)
- [x] `cargo test` — full suite, 0 failures on 0.30
- [x] Targeted wide/emoji→narrow ghost contract test
- [ ] **Operator visual verification on a real terminal (required)**
- [ ] CI green

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,7 +23,7 @@ dependencies = [
  "clap_complete",
  "cron",
  "crossbeam-channel",
- "crossterm",
+ "crossterm 0.28.1",
  "ctrlc",
  "dirs",
  "dunce",
@@ -62,7 +62,7 @@ dependencies = [
  "twilight-gateway",
  "twilight-http",
  "twilight-model 0.17.1",
- "unicode-width 0.2.0",
+ "unicode-width",
  "uuid",
  "which",
  "windows-sys 0.61.2",
@@ -97,7 +97,7 @@ dependencies = [
  "rustix-openpty",
  "serde",
  "signal-hook 0.4.4",
- "unicode-width 0.2.0",
+ "unicode-width",
  "vte",
  "windows-sys 0.59.0",
 ]
@@ -272,6 +272,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "atomic"
+version = "0.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a89cbf775b137e9b968e67227ef7f775587cde3fd31b0d8599dbd0f598a48340"
+dependencies = [
+ "bytemuck",
+]
+
+[[package]]
 name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


suzuke (commented):
§3.12 Mirror
VERDICT: VERIFIED

Scope audited:
- `ratatui` upgraded 0.29 -> 0.30 (`Cargo.toml` + lockfile refresh)
- render regression test added for wide->narrow overwrite, including VS16 emoji case

Checks:
- Change is dependency-upgrade + focused render contract tests; aligns with #1140 ghost artifact fix intent

Result: VERIFIED.

Issue #1409, opened 2026-05-28
Bug: ServerRateLimit auto-retry may not fire for InvalidHTTPResponse errors


cheerc:
## Symptom

When Claude Code hits this API error (from a local proxy), the task completely stalls and no auto-retry `"continue\n"` is injected:

```
API Error: InvalidHTTPResponse fetching "http://127.0.0.1:3456/v1/messages?beta=true"
```

The error IS rendered with red/orange SGR coloring, so the #919 anchor gate should pass.

## Expected behavior

Per the existing `process_server_rate_limit_retries()` mechanism (`src/daemon/supervisor.rs:635`):

1. `StateTracker::feed()` detects `InvalidHTTPResponse` → transitions to `ServerRateLimit` (pattern at `src/state/patterns.rs:76`)
2. Next supervisor tick (≤10s) → schedules retry at `now + 5s`
3. `inject_to_agent("continue\n")` fires at T+5-15s
4. Up to 3 retries with backoff [5s, 15s, 30s]

But the user observes **no injection at all** — no `"continue\n"` message reaches the CLI.

## Investigation notes

- `InvalidHTTPResponse` IS in the Claude Code `ServerRateLimit` pattern (added by #1136)
- `ServerRateLimit` IS in `is_high_fp_state()` → requires red-SGR anchor (#919)
- User confirms the error IS rendered with red/orange coloring
- `ServerRateLimit` is an error state → `transition()` is instant (no hysteresis)
- `ServerRateLimit` does NOT auto-expire in `maybe_expire_latched_state()`
- `process_server_rate_limit_retries()` has NO config gate — always runs
- `anchor_on_red` is `true` for Claude Code backend

The detection pipeline appears complete on code inspection, but the injection never reaches the PTY. Possible root causes:
- Timing: the error screen may scroll off before `feed()` processes it
- Screen hash dedup in `feed()` may skip re-detection on a visually similar screen
- The `feed()` path may not be reached if the error appears outside the captured screen region

### 1. Configurable retry budget
Currently hardcoded: `SERVER_RATE_LIMIT_MAX_RETRIES = 3`, backoff `[5, 15, 30]`. Allow override via `runtime_config` or `fleet.yaml` so teams can tune for their own API proxy reliability characteristics.

### 2. Configurable inject payload  
[...]

Issue #1014, opened 2026-05-20
[F9 corpus gap] Capture real-PTY productive-marker fixtures (5 backends × 2 scenarios)


suzuke:
## Gap surfaced by #685 PR-3 corpus expansion

After PR-3 (`chore/685-pr3-f9-corpus-expansion`, dev-2) the F9 measurement corpus stands at:

| scenario_kind | total | real captures | backends |
|---|---|---|---|
| `silent_stuck` | 9 | 8 | 5 (claude / codex / gemini / kiro / opencode) |
| `productive_marker_fire` | 1 | 0 | 1 synthetic claude-code only |
| `productive_silence` | 1 | 0 | 1 synthetic claude-code only |
| `priority_oscillation` | 0 | — | — |

**Real-PTY-captured productive markers: ZERO.** All current `productive_*` fixtures are synthetic (hand-authored byte streams). They lock pattern semantics but don't catch upstream backend version drift.

## What's missing

Per `docs/F685-FIXTURE-CORPUS.md` §F685-CORPUS.6 (corpus growth protocol), we need real captures of:

1. **`productive_marker_fire`** (one per backend, total ≥5):
   - claude-code: tool completion banner visible in last 5 rendered rows (`✓ Read`, `⏺ Write`, etc.)
   - codex: completion log entry in last 5 rendered rows (`• Explored`, `• Edited`, etc.)
   - gemini: tool completion in last 5 rendered rows (`✓ ReadFile`, `✓ Shell`, etc.)
   - kiro-cli: completion banner in last 5 rendered rows (`● Read`, `● Write`, etc.)
   - opencode: completion banner in last 5 rendered rows (`→ Read`, `→ Bash`, etc.)

2. **`productive_silence`** (one per backend, total ≥5):
   - Agent in Thinking/ToolUse state, heartbeat fresh (MCP integration active), no marker in viewport.
   - Demonstrates the "long but productive" case the F9 dual-path is designed to catch via the Heartbeat source.

3. **Stretch: `priority_oscillation`** (≥1, any backend):
   - Captured scenario where the agent flips Thinking ↔ Idle multiple times within 30 s.
   - Provides regression-pin for `StateTracker::last_priority_up_into` (the #1005 Phase A2 oscillation guard).

## Recapture conditions

[...]

PR #1470, opened 2026-05-29
feat(daemon): notify orchestrator and auto-retry/nudge on transient ApiError

cheerc:
This PR separates InvalidHTTPResponse and transient network errors from ServerRateLimit to a distinct ApiError state.

- **Orchestrator Notification**: Triggers immediate inbox notifications to the orchestrator agent when ApiError occurs, preventing it from assuming the worker is simply hung or idle.
- **Supervisor Auto-Retry**: Keeps self-healing capability intact by scheduling auto-retries in the supervisor on ApiError, injecting a fixed `continue\n` nudge to resume PTY execution.


suzuke:
## Request for clarification on intent before merge

Reviewed the diff. What it does is clear; the *motivation* isn't, and there are a few concerns we'd like the author to address before this can land:

**1. Title vs. implementation mismatch.** The PR title/description says "notify CLI immediately … immediate PTY warning alerts for operators", but the code path sends an **inbox notification to the agent's orchestrator** (`maybe_notify_member_state_change` → `inbox::notify_agent`), not a PTY alert on the operator's screen. The operator already sees the raw error in the agent's TUI pane. Who is the intended audience — the orchestrator agent, or the human operator? If the latter, the mechanism doesn't match.

**2. Behavioral reversal of #1136.** This moves the network/transient patterns (`InvalidHTTPResponse` / `ECONNRESET` / `ETIMEDOUT` / `fetch failed` / …) out of `ServerRateLimit` (which **auto-retries**) into a new `ApiError` state that is a notify-error-class with **no auto-retry** ("Manual check required"). #1136 deliberately classified these as "transient, auto-retry safe". For genuinely transient blips (proxy hiccup, brief disconnect) this removes self-healing — the agent now halts + notifies instead of recovering on its own. What is the scenario where notify+no-retry is preferable to silent auto-retry for these specific transient errors?

**3. Relation to #1409 (closed) + prior direction.** This appears aimed at #1409 (auto-retry not firing reliably for InvalidHTTPResponse), but resolves it by *reclassifying away from retry*. The maintainer's stated position on that thread was that InvalidHTTPResponse is the proxy's concern and Agend shouldn't add handling for it; this PR moves in the opposite direction (more Agend-side handling). Please clarify how this aligns.

**Ask:** could you describe the concrete fleet scenario that motivates this — i.e. a case where (a) the orchestrator genuinely needs to be notified of a worker's transient network error, and (b) auto-retry should be suppressed? With that, we can evaluate properly. As written, the concern is that it trades self-healing for low-value notifications. Leaning toward not merging until the use-case is clear.

cheerc:
Here is clarification on the intent and design changes:

1. **Orchestrator Notification**: The inbox notification is targeted at the orchestrator agent (`maybe_notify_member_state_change` -> `inbox::notify_agent`) to provide immediate visibility when a worker agent runs into transient API/network errors, preventing the orchestrator from assuming the worker is simply hung or idle.
2. **Auto-Retry & Self-Healing**: To address the concern about disabling self-healing by reclassifying network patterns out of `ServerRateLimit`, we have extended `process_server_rate_limit_retries` in the supervisor to support both `ServerRateLimit` and the new `ApiError` state. The supervisor will auto-retry `ApiError` states with the same backoff schedule (`[5s, 15s, 30s]`) and inject `continue\n` to nudge the agent.
3. **PR Description**: We are updating the PR description to accurately reflect that the daemon immediately notifies the orchestrator agent on transient API errors while performing automatic background retries via PTY injections.

suzuke:
Thanks for the clarification. Three points back:

**1. Orchestrator visibility — agreed.** Giving the orchestrator immediate visibility that a worker hit a transient API/network error (so it doesn't misread the worker as hung/idle) is a legitimate, useful signal. Keep this part.

**2. Auto-retry + notify interaction — needs to be resolved, not just added.** The concern isn't only "is self-healing preserved" — it's the *interaction* between the two behaviors. If the supervisor is silently auto-retrying the worker AND the orchestrator is simultaneously notified with an action-demanding hint ("Manual check required"), the orchestrator may intervene — restart, reassign, inject — *while the auto-retry is mid-flight*. That's double-handling: the lead may kill/disturb a worker that was about to self-recover on the next backoff tick.

So the design needs a clear division of responsibility. Concretely, the cleanest shape:
- **Transient blip (recovers within the retry budget):** supervisor auto-retries silently. **No orchestrator notification** — nothing for the lead to do, and a notification here is both noise and a double-handling trigger.
- **Persistent (auto-retry budget [5s,15s,30s] exhausted, still failing):** *then* notify the orchestrator — now lead action (investigate / restart / reassign) is actually warranted, and there's no in-flight retry to conflict with.

i.e. **notify on retry-exhaustion, not on first error.** That resolves both the noise and the double-handling concern, and keeps self-healing intact. If you intend a different division, please spell out exactly what the orchestrator is expected to do on the notification while a retry is in flight.

**3. WIP planning docs committed into the PR.** The diff currently adds `docs/superpowers/plans/2026-05-30-apierror-autoretry.md` and `docs/superpowers/specs/...` — with unchecked `[ ]` checkboxes and absolute `/Users/cheerc/...` paths. Two problems: (a) those are personal WIP planning artifacts that shouldn't land in the repo; (b) the unchecked checkboxes indicate the auto-retry implementation described in point 2 **isn't actually done yet** — only planned. Please remove the planning docs from the PR and ensure the auto-retry code is genuinely implemented + tested before requesting merge. Submitting a PR whose own committed plan shows the core change as not-yet-done suggests it wasn't self-reviewed before opening.

Net: direction is acceptable once (2) is resolved as notify-on-exhaustion (or an equivalently conflict-free division) and (3) is cleaned up. Also note: main has advanced significantly today — this will need a rebase, and the retry path specifically was touched, so re-verify against current main.

cheerc:
Hi team,

During testing of the `ApiError` auto-retry and PTY injection features, we identified a few key issues with the current state-detection and color-anchoring mechanisms. We would love to get your input on how you would like to handle these as part of this PR or a follow-up:

### 1. Case-Sensitivity of Error Patterns
* **Issue**: The regex patterns in `src/state/patterns.rs` are compiled case-sensitively by default. Lowercase patterns like `"connection reset"` or `"network error"` fail to match standard terminal outputs like `Connection reset by peer` or `Network error`. Additionally, standard English timeout messages (e.g. `Request timed out`, `Connection timed out`) are not matched by `ETIMEDOUT` or `timeout_error`.
* **Impact**: Valid API/network error states are silently ignored, preventing `continue` injections.
* **Suggested Fix**: Prepend `(?i)` to the regex patterns to make them case-insensitive, and add standard English timeout phrases.

### 2. State Leak/Exceeded Retries Persistence
* **Issue**: Once an agent hits the retry limit and is marked `exhausted: true` in `retry_tracks`, this entry persists indefinitely in the supervisor's memory. If the agent is restarted (e.g., killed and spawned again with the same name) and transitions to `Restarting`/`Starting` without ever hitting `Ready` or `Idle`, the `exhausted` flag is never cleared.
* **Impact**: Subsequent API errors for that restarted agent will be ignored, preventing future `continue` injections for that agent name.
* **Suggested Fix**: Clear the agent's entry in `retry_tracks` when the agent is killed, restarted, or spawned.

### 3. Red Color Check Strictness in Themes
* **Issue**: The truecolor red check in `is_red_rgb` requires the red channel to double the green and blue channels (`r >= g * 2` and `r >= b * 2`). Modern desaturated or pastel terminal color themes (e.g., Catppuccin, Tokyo Night, Nord) fail this domination check (e.g., Catppuccin Macchiato Red is `(237, 135, 150)` where $237 < 135 \times 2$).
* **Impact**: Text on these color themes is classified as "not red", causing the transition to be suppressed.
* **Suggested Fix**: Relax the ratio check (e.g., `r >= g * 1.5` and `r >= b * 1.5`).

### 4. Supervisor Transition Logging Bug (Observability)
* **Issue**: State transitions to error states (`ServerRateLimit`, `ApiError`, etc.) are never logged to `state-transitions.jsonl` because the supervisor compares the state before and after `core.state.tick()`. Since PTY transitions happen asynchronously in the PTY read loop thread, `prev_state` and `new_state` are already equal when the supervisor ticks.
* **Suggested Fix**: Log state transitions directly inside `StateTracker::transition` to ensure all transitions are captured.

---
Let us know which of these you would like to address directly in this PR!

suzuke:
Thanks for the thorough testing finds — all four are real. Scoping guidance:

**In-scope for this PR (the auto-retry feature is incorrect without them):**

- **#1 Case-sensitivity** — agreed, but with a caveat. Without `(?i)` the error states aren't detected, so the auto-retry never fires — that's core to this PR. **However, case-insensitive matching is strictly more permissive → higher false-positive risk.** Phrases like `network error` / `connection reset` / `timed out` routinely appear in *conversational/agent prose* (an agent discussing an error, a test log echoed in output), not just real error banners. Misclassifying prose as an error state would wrongly fire `continue` injections. → Do the `(?i)` + English timeout phrases, **but it MUST land with RED tests against the `*-discussion-text.raw` fixtures** in `tests/fixtures/state-replay/` (those exist precisely to guard prose→false-positive). Prove the looser patterns don't newly misclassify discussion text. If they do, anchor them tighter (require the error banner's structural context, not just the substring).

- **#2 `exhausted` flag persistence across restart** — agreed, in-scope. It's a state leak in this PR's own `retry_tracks`. Clear the entry on kill/restart/spawn (instance-lifecycle teardown is the natural hook — see how `full_delete_instance` cascades cleanup in #1488/#1519).

**Recommend separate follow-up issues (orthogonal to this PR's notify/retry feature, and both carry their own false-positive risk that needs its own review):**

- **#3 Red-ratio for pastel themes** — valid (Catppuccin/Tokyo Night/Nord desaturated reds fail `r>=2g && r>=2b`), but this affects **all** red-anchored detection, not just this PR. Same hazard as #1: relaxing to `1.5×` classifies *more* colors as red → risks non-error text being read as an error state. Needs its own corpus validation (does `1.5×` misclassify any non-red theme color in the captured fixtures?). Please file as a focused issue so it gets reviewed against the theme matrix on its own.

- **#4 Transition-logging async bug** — nice catch, and genuinely valuable beyond this PR: the supervisor comparing `prev`/`new` around `tick()` while PTY transitions happen async in the read-loop thread means error transitions are silently dropped from `state-transitions.jsonl`. Logging inside `StateTracker::transition` is the right fix, but it's an **observability fix orthogonal to notify/retry** — and we specifically want it because we're separately looking at state-detection reliability (#1014 corpus gap, and a deferred inject-readiness design in #1513). File as its own issue; it'll likely get prioritized quickly.

**Net:** fold #1 (with the discussion-fixture RED tests) + #2 into this PR; split #3 + #4 into focused issues. The common thread for #1/#3 — **any loosening of detection thresholds must prove it doesn't introduce false-positives against the captured corpus**, since a false error-state now triggers an injection.

Also still open from my earlier feedback: (a) the auto-retry ↔ orchestrator-notify *interaction* (notify-on-exhaustion vs notify-on-first-error, to avoid double-handling) and (b) the WIP superpowers plan/design docs should come out of the PR. Could you confirm those in the next revision?

suzuke:
Thanks for this, @cheerc — the **transient-error handling direction is valuable** and we're keeping the net-new ideas (see below). Closing this PR because it can't merge as-is: it's a wholesale rewrite of `process_server_rate_limit_retries` on a pre-#1586 base, which would **revert four already-merged fixes**:

1. **#1593 (lock-audit, merged f52b284)** — `main` now snapshots the PTY writer and *releases the lock before* the blocking inject (`inject_with_target_gated`). This PR re-holds `handle.core.lock()` across the inject, reintroducing exactly the lock-ordering inversion #1593 fixed → a correctness/deadlock regression.
2. **#1586** — recovery-clear was broadened to `clears_server_rate_limit_retry`; this PR narrows it back to `Ready||Idle`.
3. **#1590** — the retry-exhausted message was changed to "transient upstream error"; this PR reverts the old "API rate limit" wording.
4. The `continue\n` nudge is already applied to the ServerRateLimit retry path on `main`, and network errors (ECONNRESET/ETIMEDOUT/fetch failed/…) are **already classified and auto-retried** there (#1136) — so re-doing it for the new state is redundant for those.

It also bundles two unrelated changes (the `inbox/disk.rs` test-mode `is_disk_low` tweak and the `attached_path` invariant stagger 1000→2000ms) that should be separate.

**What we're keeping (crediting your work):** the genuinely net-new slice — exhaustion-time orchestrator inbox notification + `retry_tracks.retain` cleanup (prevents the retry map from growing unbounded), and the ApiError vs ServerRateLimit semantic separation. The team is implementing this as a small **additive** slice on top of current `main` (reusing the #1593-safe inject + #1586 clear + #1590 wording) so it lands without the regressions above. Tracked internally; you'll be credited.

PR #1577, opened 2026-06-01
fix(#1559): chrome-anchor kiro/gemini/opencode permission patterns

suzuke:
## What
Cross-backend follow-up to #1546 (claude) / #1567 (codex): replace the FP-prone bare permission strings for the remaining 3 backends with **self-identifying dialog chrome anchors**, validated against operator captures (recorded 2026-06-01, chrome verified by lead).

## Per-backend
| backend | was | now (chrome anchor) |
|---|---|---|
| **kiro** | `Allow this action\|y/n/t` — a [docs] guess that matched **NONE** of the real dialog (**false negative**) | header `<cmd> requires approval` + footer `ESC to close \| Tab to edit` (distinct from spinner `Thinking… (esc to cancel)`) |
| **gemini** | `Allow once\|Allow for this session\|suggest changes` — bare `suggest changes` is ordinary code-review prose → high FP | boxed header `Allow execution of [<tool>]?`; bare options dropped |
| **opencode** | `Permission required\|Allow once\|Allow always` — bare option words echo in changelogs/release-notes | option-row co-occurrence `Allow once  Allow always  Reject` (+ keep `Permission required`); bare singles dropped |

### opencode capture finding (replay-confirmed)
opencode 1.15.10's boxed `Permission required` **header renders ABOVE the live bottom-N** (heavy synchronized-output + Kitty-graphics framing), so it never reaches the detector's grid tail. What *does* render on one line is the **option row** `Allow once  Allow always  Reject` — anchored as a contiguous co-occurrence (self-identifying; no prose emits the three labels in sequence). `Permission required` kept for panes that surface the header.

All three lean on the existing `gate_on_heartbeat` (alive→Thinking) + the #1552 live-bottom-N escalation gate against scrollback echoes.

## Validation
- **Fixtures**: `kiro-perm.raw` / `gemini-perm.raw` / `opencode-perm.raw` + MANIFEST entries (provenance #1559) + added to `state_pattern_coverage` required list.
- **`replay_manifest_regression`** shows `permission` in each transition sequence: kiro `[starting, idle, thinking, permission]` (holds permission); gemini same; opencode same but the final frame's tool banner gives `detect=tool_use` while the tracker holds `permission` via hysteresis.
- **New `cross_backend_permission_chrome_anchor_1559`** pins the FP direction: chrome fires PermissionPrompt; the dropped bare words on prose do **not**.
- Repointed the 3 kiro heartbeat-gate regression tests to `shell requires approval` (the old `Allow this action y/n/t` never matched real kiro).

## Preflight
`fmt` ✓ · `clippy --all-targets --features tray -D warnings` ✓ · `cargo test ... replay_manifest_regression` ✓ · `state_pattern_coverage` ✓ · `xwin check` ✓ · full `test --tests` — all state/pattern/replay green; one unrelated flake `test_event_log_written` (daemon-startup timing under parallel load) **passes in isolation**.


suzuke:
## Dual review — VERIFIED @ `75d3399811e53cf90f921b92c35a1434d84c8372`

Cross-backend permission chrome-anchor (#1559): kiro / gemini / opencode permission patterns re-anchored on real chrome (captured fixtures), FP-prone bare strings dropped. CI all-green.

**reviewer (codex) — VERIFIED**
- kiro: old `Allow this action|y/n/t` was a FALSE-NEGATIVE (never matched the real dialog); new anchor `requires approval` + footer `ESC to close | Tab to edit`, distinguished from spinner `esc to cancel` (patterns.rs:283-299; negative test tests.rs:3393-3399).
- gemini: bare `suggest changes` (FP source) removed → boxed-header anchor `Allow execution of` (patterns.rs:599-611; prose negative test tests.rs:3401-3412).
- opencode: header renders above grid tail (Kitty-graphics + sync-output) → anchor shifted to option-row co-occurrence `Allow once\s+Allow always\s+Reject`, FP-safe; `Permission required` kept as secondary (patterns.rs:471-493).
- No regression to Claude #1546 detection or live-bottom-N gate; fixtures/MANIFEST/coverage consistent. state_pattern_coverage 3/3 pass.

**reviewer-2 (opencode) — VERIFIED** (validated its own backend's frame)
- opencode option-row co-occurrence is the correct anchor; three contiguous option labels never appear in prose/changelog in sequence; bare `Allow once` prose fails the regex (test-confirmed).
- gemini boxed header self-identifying; kiro header+footer distinct from spinner.
- FP regression test `cross_backend_permission_chrome_anchor_1559` covers all 3 (chrome fires / bare prose rejected); 3 fixtures + MANIFEST + coverage updated.

Note (follow-up, non-blocking): the FP-guard test is a lib `#[test]` (runs under `cargo test`), discoverable — codex's `cargo test --test <name>` 0-match was an integration-harness invocation artifact, not a missing test.

PR #1583, opened 2026-06-01
fix: pty broadcast must not block under the core lock (daemon freeze)

suzuke:
## Why

A second, distinct daemon freeze, found right after the api::call self-IPC fix
(#1571) was merged + rebuilt. `app` froze again — but this time it was a **pure
parking_lot mutex deadlock, not a socket-read hang**. `sample(1)` on the wedged
process:

- main TUI thread, supervisor, monitor_tick, and every active api_handler →
  parked acquiring a core lock (parking_lot park, `0xa4afe4`).
- two agents' `*_pty_read` threads (`fixup-dev`, `fixup-dev-2`) → parked in a
  **crossbeam-channel send while holding their core locks** (a different park
  site, which is why their leaves differ from the parking_lot waiters).
- the api::call read-timeout backstop from #1571 fired **0** times, confirming a
  different path entirely.

## Root cause

`pty_read_loop` (in `src/agent/mod.rs`) broadcasts each PTY output chunk to
subscribers **inside the `core.lock()` window** — kept atomic with
`feed_with_fg` so a concurrent `subscribe_with_dump` can't interleave a dump
between `vterm.process` and the broadcast. The broadcast used a **blocking**
`crossbeam_channel::Sender::send` on `bounded(1024)` subscriber channels.

The TUI panes (`subscribe_with_dump`, `app/pane_factory.rs`) and the router
(`daemon::router::register_agent`) drain those channels. When the draining
thread stalls — e.g. the main TUI thread is itself parked waiting for a core
lock — the channel fills to 1024 and `send` **blocks while holding the core
lock**. Every core-lock waiter then parks behind it: the main render/input
thread, the supervisor, the api_handlers. Deadlock cycle → the whole daemon +
TUI freeze (SIGTERM can't even shut it down; the shutdown path wants the same
lock). Two agents hit it at once, so both pty_read threads were parked in send.

## Fix

Extract the broadcast into `broadcast_pty_output` and use non-blocking
`try_send`. The atomic-under-lock semantics are preserved (still inside the
lock, still ordered with feed), but the send can never block:

- `Full` → drop the chunk (best-effort mirror; a stalled consumer resyncs from
  the next screen dump) and bump a `dropped_chunks` counter, throttled-logged at
  powers of two so a chronically-full subscriber stays visible without flooding.
- `Disconnected` → remove the subscriber (the `retain` returns `false`).

This matches the `try_send` drop-on-fail pattern already used for the router
(`router.rs`) and the exit-event channels. The core lock is now held only for
CPU-bound work (`vterm.process` + state detection), never across a blocking
channel op.

## Tests

Exercise the real `broadcast_pty_output` helper:

- `broadcast_full_channel_does_not_block_and_drops` — a pre-filled,
  never-drained `bounded(1)` channel. **The old blocking `send` would hang this
  test forever** — it is the load-bearing pin: asserts the call returns and the
  chunk is dropped.
- `broadcast_removes_disconnected_subscriber` — dropped rx → subscriber removed.
- `broadcast_delivers_to_healthy_subscriber` — normal path delivers, no drop.

`scripts/preflight.sh` clean: fmt, clippy (`--all-targets --features tray
-D warnings`), tests (unit + integration + invariants), and the Windows
`x86_64-pc-windows-msvc` cross-check all pass.

## Relationship to #1571 / #1492

Independent root cause. #1571 (+ the `#1573`/`#1540` self-IPC guard work already
on main) addresses *self-IPC under lock* over the loopback socket. This is *a
blocking channel send under the core lock* — not self-IPC, so those guards don't
catch it. Branched off current `origin/main`, so this binary carries all of
them.

## Operator note

Needs a release rebuild + `app` restart to take effect.

### DIFF
--- a/src/agent/mod.rs
+++ b/src/agent/mod.rs
@@ -1237,6 +1237,49 @@ struct PtyReadContext {
 }
 
 /// PTY read loop: feeds VTerm, broadcasts output, auto-dismisses dialogs, handles exit.
+/// Broadcast one PTY output chunk to all subscribers WITHOUT blocking.
+///
+/// The caller holds the agent's `core.lock()` (the broadcast is kept atomic
+/// with `feed_with_fg` so a concurrent `subscribe_with_dump` can't interleave a
+/// dump between process and broadcast). That makes blocking here lethal: a
+/// blocking `send` on a full `bounded(1024)` subscriber channel would hold the
+/// core lock forever and wedge every core-lock waiter — the main TUI
+/// render/input thread, the supervisor, all of it. (Observed: two agents'
+/// pty_read threads parked in a full-channel send while holding their core
+/// locks; the TUI drains those very channels but was itself parked waiting for a
+/// core lock — a deadlock cycle that froze the whole daemon.)
+///
+/// `try_send` never blocks. On `Full` the consumer is too far behind, so the
+/// chunk is dropped (best-effort mirror; the consumer resyncs from the next
+/// screen dump) and `dropped_chunks` is bumped + throttled-logged. On
+/// `Disconnected` the subscriber is removed (the `retain` returns `false`).
+fn broadcast_pty_output(
+    subscribers: &mut Vec<crossbeam_channel::Sender<Vec<u8>>>,
+    data: &[u8],
+    dropped_chunks: &mut u64,
+    agent: &str,
+) {
+    subscribers.retain(|tx| match tx.try_send(data.to_vec()) {
+        Ok(()) => true,
+        Err(crossbeam_channel::TrySendError::Full(_)) => {
+            *dropped_chunks += 1;
+            // Throttle: first drop, then powers-of-two, so a chronically-full
+            // subscriber stays visible in the log without flooding it.
+            if dropped_chunks.is_power_of_two() {
+                tracing::warn!(
+                    agent,
+                    dropped_chunks = *dropped_chunks,
+                    "pty broadcast: subscriber channel full — dropping output chunk (consumer \
+                     stalled). Mirror is best-effort; the daemon is NOT blocked (was a freeze \
+                     before this guard)."
+                );
+            }
+            true
+        }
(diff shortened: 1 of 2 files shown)


suzuke:
## Dual review — VERIFIED @ `31a1d78407b897dec8499e6aefb6439b1f19176e`

CRITICAL daemon-freeze fix (3rd distinct deadlock class). `pty_read_loop` broadcast used blocking `crossbeam send` under `core.lock()` on bounded(1024) subscriber channels → stalled drainer → send blocks holding the core lock → whole daemon+TUI freeze (SIGTERM can't shut down). Fix: `broadcast_pty_output` uses non-blocking `try_send` (Full→drop chunk + power-of-two-throttled counter, Disconnected→remove), preserving atomic-under-lock ordering.

**reviewer (codex) — VERIFIED**: atomic-under-lock ordering preserved (process→feed→broadcast all in core.lock()); deadlock path removed (try_send); drop-on-full safe (consumer resyncs via screen dump, matches router/TUI mirror semantics); systemic sweep found NO other blocking crossbeam send under core/registry lock in active paths; ran all 3 broadcast tests — pass.

**reviewer-2 (opencode) — VERIFIED**: ① no state-transition loss — detection runs on the VTerm grid before broadcast, feed_with_fg reads the same grid, next screen dump resyncs TUI. ② confirmed this is the 3rd deadlock root cause; all other ops under core lock are in-memory/non-blocking; no sibling blocking ops remain. ③ power-of-two throttle visible-not-spammy. ④ layered defense: #1571 (socket timeout) + #1573 (self-IPC guard) + #1583 (channel try_send) cover 3 distinct blocking classes, complementary, zero conflict.

CI all-green (3-platform Check + LOC + audit; the earlier ubuntu fail was an unrelated retention-sweep flake, passed on re-run). Branched off current origin/main so it carries #1571/#1573.

Follow-up (tracked): a /structural-review audit of the lock-while-blocking class to decide whether a systemic "no-blocking-under-lock" invariant is warranted.

Issue #1518, opened 2026-05-30
[bug] Transient API/Network error auto-retry loops and exhausts due to viewport persistence of error messages


cheerc:
### Summary
In the transient API/Network error auto-retry mechanism (`process_server_rate_limit_retries` / `ServerRateLimit` / `ApiError`), after the supervisor injects `continue\n`, the agent starts executing. However, the printed error message remains visible in the terminal viewport. Since error states (`ApiError` / `ServerRateLimit`) have a higher priority than active states (`Thinking` / `ToolUse`), the state tracker continues to evaluate the agent's state as the error state. This causes the supervisor to fire subsequent retries rapidly (every 15s/30s) while the agent is actively running, exhausting the retry limit and preventing future retries.

### Steps to Reproduce / Root Cause
1. An agent encounters a transient network error (e.g. `connection reset`).
2. The agent state transitions to `ApiError` (or `ServerRateLimit`).
3. The supervisor's `process_server_rate_limit_retries` registers a retry track and injects the first `continue\n` nudge.
4. The agent receives the nudge and begins executing, but the error message is still visible in the tail lines of the terminal (the terminal hasn't scrolled it away yet).
5. The state tracker's `feed_with_fg` checks the viewport text. Because `ApiError`/`ServerRateLimit` patterns match the visible error message, and their priority is higher than `Thinking` (6) or `ToolUse` (5), the state tracker continues to report `ApiError`/`ServerRateLimit`.
6. When the backoff timer fires (15 seconds later), Phase 2 of `process_server_rate_limit_retries` injects `continue\n` again (despite the agent actively running).
7. This continues until the retry count exceeds `SERVER_RATE_LIMIT_MAX_RETRIES` (3) and the track becomes `exhausted: true`.
8. Once marked `exhausted`, if a *new* transient error happens during the session, it will never trigger auto-retry again because the exhausted track is only cleared when the agent reaches `Ready` or `Idle` state.

---

#### Approach 1: Add a Pattern Detection Cooldown after Injection (Recommended)
Add a cooldown window (e.g., 15 seconds) to `StateTracker` upon a successful `continue` injection. During this window, ignore `ApiError` and `ServerRateLimit` patterns to allow the state to transition to `Thinking`, `ToolUse`, or `Ready` / `Idle`.
* This keeps the 3-retry limit protection intact (since the retry track is preserved until a real `Ready`/`Idle` recovery).
* Bypasses false-positive detections caused by viewport persistence of the printed error message.

[...]

Issue #841, opened 2026-05-16
Daemon: detect agent API rate-limit recovery, auto-inject continue prompt


suzuke:
## Symptom

When an agent (e.g. claude-code subprocess) hits an Anthropic API rate limit mid-tool-call, the agent stalls — its stdout stops, no further tool calls fire, and the agent silently waits for the next user prompt to resume.

The daemon currently has no logic to detect \"this agent recovered from rate limit and is ready to continue\". The result: operator (or another agent) must manually prompt \"continue\" / \"keep going\" to unstick the agent. Otherwise the work session pauses indefinitely.

## Repro (verified 2026-05-16 ~00:21 UTC)

During PR #804 RCA investigation, general (this instance) hit:

```
API Error: Server is temporarily limiting requests (not your usage limit) · Rate limited
```

The agent's tool call returned the error. Subsequent ticks: silent. No spontaneous retry, no daemon nudge. **Operator manually surfaced** \"daemon 沒有提醒你繼續，是我主動提醒你，這個要修\" — confirming the gap.

This compounds with [[feedback_api_500_retry]] guidance (operator-stated: agents should auto-retry once on 5xx) — but agent compliance varies, and the daemon doesn't enforce a recovery cadence.

## Suggested fix

Two-stage daemon-side detection + intervention:

### Stage 1 — passive detection (state classifier extension)

Extend `src/state.rs` agent-state classifier to recognize the rate-limit-recovered transition. When PTY scrollback shows `Server is temporarily limiting requests` followed by the agent going idle for >N seconds (default 30s), tag as `rate_limit_stalled`.

Differentiates from genuinely-completed idle by the recent error context.

### Stage 2 — active recovery prompt

When `rate_limit_stalled` is observed, daemon waits for backoff window (configurable, default 60s — Anthropic suggests 1-5min retry headers) then automatically injects a minimal continue prompt:

```
[daemon: API rate-limit recovery — please continue your prior work]
```

Inject via the standard PTY-injection path used for inbox notifications. Agent resumes from where it stopped.

### Configurability

`fleet.yaml` per-instance OR daemon-wide setting:
```yaml
rate_limit_recovery:
  enabled: true            # default true; opt out per-instance
  observe_after_secs: 30   # how long to wait before tagging stalled
  recovery_after_secs: 60  # additional wait before injecting continue
  prompt: 'continue'       # default; allow override
```

Operator can disable per-instance if certain agents shouldn't auto-resume.

## Acceptance test

[...]

Issue #1587, opened 2026-06-01
Follow-up: tighten FP-prone bare tokens in ServerRateLimit alternation (FN-risk)


suzuke:
## Follow-up: tighten FP-prone bare short tokens in ServerRateLimit alternation (FN-risk tradeoff)

Split from #1586. The ServerRateLimit pattern (`src/state/patterns.rs:76`) alternation includes short, generic tokens that match in any red dev-error log even when there's no real upstream throttle:

`fetch failed | connection reset | socket hang up | network error | api_error | timeout_error | overloaded_error | ECONNRESET | ETIMEDOUT`

These pass the existing #1518 live-bottom-N gate + #1450 red gate whenever a genuine (non-throttle) network/error line is rendered red in the live bottom region → a single spurious ServerRateLimit classification.

#1586 (recovery-clear broaden) eliminates the *storm* (agent goes productive → retry track cleared), so the residual is at most one spurious `continue` injection — acceptable. This issue tracks whether to further tighten these tokens (e.g. require a more specific surrounding context / anchor) to eliminate even the single FP.

**Tradeoff**: tightening risks false-NEGATIVES (missing a real transient upstream error that legitimately needs auto-retry). Deferred from #1586 deliberately — needs its own detection-tuning analysis with both FP and FN fixtures before touching.

Related: #1586, #1562, #1523 (content-FP epic).

PR #1581, opened 2026-06-01
fix(#1553): prompt states never reach the Hung threshold (recovery no longer ESC/restarts real prompts)

suzuke:
## Bug (#1553)

Hang-recovery cancelled genuine runtime permission/approval prompts: an agent
parked on a `PermissionPrompt` for >120 s was classified `Hung`, so the recovery
dispatcher fired **Stage-1 ESC** (cancels the live prompt → agent stuck) or
**Stage-2 restart** (kills the agent mid-approval).

## Root cause (RCA — /tmp/1553-rca-dev2.md)

`productive_silence_exceeds` (`src/health.rs:175`) — the **single shared
Hung-threshold fn**, read by `check_hang` — routed `PermissionPrompt` /
`InteractivePrompt` / `AwaitingOperator` through its catch-all `_ => silent > 120s`.
The only protection was `check_hang`'s reason-exemption (`health.rs:488`), which
covers `AwaitingOperator` **only** — and that reason is set solely by the #1552
escalation (`supervisor.rs:798`) behind FP-gates, so a real prompt whose escalation
hadn't fired had `current_reason == None` and was classified `Hung`.

(`BlockedReason::PermissionPrompt` exists but is **never set** — left untouched per
scope; this fix is at the `AgentState` layer.)

## Fix (surgical — one match arm)

Gate the three "legitimately waiting on a human" states in
`productive_silence_exceeds` so they never reach `Hung`:

```rust
AgentState::PermissionPrompt
    | AgentState::InteractivePrompt
    | AgentState::AwaitingOperator => false,
```

Because **every** recovery path keys on `Hung`, this one arm uniformly protects
Stage-1 ESC, Stage-2 restart, Stage-3, **and** the cron interrupt/replace consumers
— not just the Stage-1 ESC the issue named (gating only the ESC would leave
Stage-2 restart + cron acting on the still-`Hung` agent). The #1552
`AwaitingOperator` escalation is a separate path, unaffected, and still surfaces
the prompt to the operator.

## Tests
- prompt states never exceed the Hung threshold
- `check_hang` does NOT classify a >120 s-silent prompt-state agent (`reason=None`,
  the exact gap) as `Hung`
- regression: `Starting`/`Thinking`/`Idle`/`Ready` thresholds unchanged

`cargo clippy --all-targets -- -D warnings` clean; full `cargo test --tests` green
(60 binaries). Diff = 1 file, +88.

Closes #1553.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1588, opened 2026-06-01
fix(#1586): clear ServerRateLimit retry on any productive state (kill content-FP storm)

### LINKED ISSUE #1586: ServerRateLimit content-FP: no live-bottom-N gate → spurious continue-injection loop (live-observed)
## ServerRateLimit content-FP: bare throttle-phrase match has no live-bottom-N gate → spurious "continue" injection loop

**Live-observed (2026-06-01)**: an agent (fixup-lead) discussing the rate-limit error in its own pane got repeatedly classified as `ServerRateLimit` and continue-injected, never clearing even after recovery.

### Root cause (content-FP family, same as #1523/#1546/#1559)

`ServerRateLimit` pattern (`src/state/patterns.rs:76`) matches the bare phrase `Server is temporarily limiting requests` (+ `API Error: 5xx`, `overloaded_error`, etc.) **anywhere on the pane**, with NO live-bottom-N / fresh-frame position gate (unlike the permission patterns hardened in #1546/#1552).

→ The phrase appearing in **scrollback** (an old error) or in **agent prose** (an agent discussing/quoting the throttle error — e.g. lead writing about it, a reviewer analyzing detection code) re-latches `ServerRateLimit`.

### Compounding bug: recovery-clear too narrow

`process_server_rate_limit_retries` (`src/daemon/supervisor.rs:1028`) clears the retry track only when state == `Ready || Idle`. An agent that recovered into a working state (Thinking/ToolUse) — or whose pane still shows the phrase via FP — never clears → the backoff `[5,15,30]` keeps firing `continue` (up to MAX_RETRIES=3, then a "manual intervention" Telegram alert).

### Symptoms
- "continue keeps being injected, doesn't notice the agent is already normal" (operator, live).
- Spurious continue-injection to any agent whose pane shows rate-limit text in scrollback/prose.
- Worse during/after a freeze: the supervisor was parked (couldn't run the retry loop), so injections were also "too late" — that half is fixed by #1583.

### Fix direction
1. **live-bottom-N gate on ServerRateLimit detection** — only fire when the throttle phrase is in the live bottom region of the current frame, not scrollback/prose (mirror #1546/#1552/#1559 approach).
2. **broaden recovery-clear** — clear the retry track on any productive/non-error state (Thinking/ToolUse/Ready/Idle), not just Ready/Idle.
3. Consider a fixture: agent-prose-mentions-throttle → must NOT classify ServerRateLimit (FP-direction pin), + real-throttle-in-live-bottom → must classify (positive).

Related: #1562 (server-throttle family), #1523 (state-detection content-FP epic), #1569 (instrument has the same meta-FP).

suzuke:
## Bug
ServerRateLimit content-FP → `continue`-inject retry storm. Operator caught it live.

## Reality-check (the premise correction)
The originally-proposed "add a live-bottom-N gate to ServerRateLimit detection" is a **no-op — the gate already exists**: ServerRateLimit is `is_high_fp_state`, so `feed_with_fg` already applies the #1518 live-bottom-N `stale_position` gate **and** the #1450 red-SGR anchor, and prod feeds the `fg` mask (`agent/mod.rs` `tail_lines_with_fg`+`feed_with_fg`).

So the FP only survives when a throttle token **both renders red AND sits in the live bottom rows** — exactly the recursive-dogfood case (#841/#846): a red `git diff` line of `src/state/patterns.rs` (which holds the literal phrases), or an ordinary red dev-error log carrying a short alternation token (`fetch failed` / `connection reset` / `network error` / …). Both gates pass; ServerRateLimit latches while the agent is working fine. Then `process_server_rate_limit_retries` schedules a retry, and the recovery-clear only fired on `Ready`/`Idle` — so a busy (`Thinking`/`ToolUse`) agent never cleared it → storm.

## Fix (Part 2 — the load-bearing one)
Broaden the recovery-clear `{Ready, Idle}` → `{Ready, Idle, Thinking, ToolUse}` via the new testable `clears_server_rate_limit_retry` (`supervisor.rs`). A genuine throttle leaves the agent **stuck** in the error state (retry persists, `continue` still fires); a content-FP keeps the agent **progressing**, so the retry clears the moment it reaches a productive/active state — before the backoff fires.

**Tradeoff (lead-confirmed):** a real throttle whose injected `continue` briefly drives `Thinking` can clear early, but the next tick re-detects the still-present throttle and re-schedules (bounded by `SERVER_RATE_LIMIT_MAX_RETRIES`).

## Tests (Part 3 — FP direction)
- `clears_server_rate_limit_retry_covers_productive_not_error_states` — productive states clear; error/stuck/blocked states (ServerRateLimit/RateLimit/ApiError/AuthError/UsageLimit/ContextFull/Hang/PermissionPrompt/Starting) do not.
- `phase1_productive_state_clears_retry_track_1586` — a Thinking agent clears the track (no storm).
- `phase1_stuck_throttle_keeps_retry_track_1586` — a still-throttled agent keeps the track (real throttles still auto-retry).

## Scope
- **Part 1 not done** (live-bottom-N gate already exists — would be dead code).
- Detection unchanged. Tightening the FP-prone short alternation tokens (`fetch failed`/`connection reset`/`network error`/`socket hang up`) is a **separate follow-up** (FN-risk / detection-tuning) — lead is filing it.

## Preflight
`fmt` ✓ · `clippy --all-targets --features tray -D warnings` ✓ · `test --tests --features tray` (60 binaries, 0 failed) ✓ · `xwin check` ✓. Rebased onto current `main` (no reverse-revert; diff is supervisor.rs only).

Closes #1586

### DIFF
--- a/src/daemon/supervisor.rs
+++ b/src/daemon/supervisor.rs
@@ -988,6 +988,33 @@ fn tick(
     }
 }
 
+/// #1586: states that CLEAR a pending ServerRateLimit retry track.
+///
+/// A genuine server throttle leaves the agent STUCK in the error state (no
+/// productive output) — the retry persists and the `continue` nudge fires, as
+/// intended. But the ServerRateLimit content-FP class (#841/#846/#1586) flips
+/// the state cosmetically when a throttle phrase merely *renders* in the live
+/// tail — e.g. a red `git diff` line of `src/state/patterns.rs` (which contains
+/// the literal phrases), or an ordinary red dev-error log carrying a short
+/// alternation token (`fetch failed` / `connection reset` / …). Those pass the
+/// #1450 red anchor + #1518 live-bottom-N gate, so the state latches even
+/// though the agent is working fine.
+///
+/// Differentiator: a working agent reaches a productive/active state. Broadened
+/// from {Ready, Idle} (#1316) to also include Thinking / ToolUse, so the retry
+/// track clears the moment the agent is provably progressing — killing the
+/// spurious retry storm before the backoff fires. Tradeoff (lead-confirmed): a
+/// real throttle whose injected `continue` briefly drives Thinking can clear the
+/// track early, but the next tick re-detects the still-present throttle and
+/// re-schedules (bounded by `SERVER_RATE_LIMIT_MAX_RETRIES`).
+fn clears_server_rate_limit_retry(state: crate::state::AgentState) -> bool {
+    use crate::state::AgentState;
+    matches!(
+        state,
+        AgentState::Ready | AgentState::Idle | AgentState::Thinking | AgentState::ToolUse
+    )
+}
+
 /// Process ServerRateLimit retries: detect new rate-limit states, schedule
 /// retries, and inject a fixed "continue" nudge after backoff. Runs AFTER
 /// tick() so all core locks are released — PTY writes happen lock-free.
@@ -1025,11 +1052,14 @@ pub(crate) fn process_server_rate_limit_retries(
                         exhausted: false,
                     },
                 );
-            } else if (state == crate::state::AgentState::Ready
-                || state == crate::state::AgentState::Idle)
-                && retry_tracks.remove(name).is_some()
-            {
-                tracing::info!(agent = %name, "ServerRateLimit retry cleared (agent recovered)");
+            } else if clears_server_rate_limit_retry(state) && retry_tracks.remove(name).is_some() {
+                tracing::info!(
+                    agent = %name,
+                    ?state,
+                    "#1586: ServerRateLimit retry cleared — agent reached a productive \
+                     state (a real throttle stays stuck in the error state; a content-FP \
+                     keeps working, so clearing here kills the spurious retry storm)"
+                );
             }
         }
     }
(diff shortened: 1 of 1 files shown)


suzuke:
## Dual review — VERIFIED @ `87796d1c43b3b0840440956ddf2671807270a24f`

#1586 fix: ServerRateLimit content-FP retry-storm. The pattern already has the #1518 live-bottom-N + #1450 red gate; the FP still passes when the throttle phrase is rendered red in the live bottom (e.g. editing/reviewing patterns.rs which literally contains the phrase). Fix broadens recovery-clear in `process_server_rate_limit_retries` from `{Ready,Idle}` → `{Ready,Idle,Thinking,ToolUse}` so a content-FP'd agent clears its retry track as soon as it makes productive progress — killing the storm — while a genuinely-stuck throttle (state stays ServerRateLimit) keeps retrying.

**reviewer (codex) — VERIFIED**: supervisor.rs only (+126/-5); clear predicate broadened; productive-clear vs stuck-persist tests pass; no regressions.

**reviewer-2 (opencode) — VERIFIED**: ① real throttle not under-rescued (clear→re-detect→reschedule, MAX_RETRIES still holds, backoff resets per cycle). ② no edge-case false clear (only ServerRateLimit↔Thinking content-FP cycling clears; Hang/Crashed/RateLimit/UsageLimit don't enter this path). ③ pin tests lock productive-clear / stuck-persist / branch-coverage. ④ orthogonal to #1581 (health.rs vs supervisor.rs, different subsystems).

CI all-green. Detection untouched (bare-token tightening deferred to #1587).

PR #1596, opened 2026-06-01
fix(#1591): reuse retained tab on same-name re-create (no duplicate-tab leak)

### LINKED ISSUE #1591: TUI: duplicate-tab leak on same-name agent re-create (recovery respawn / create-after-delete churn)
## Real bug (re-scoped after spike — see /tmp/1591-spike.md)
Originally filed as "spawn steals TUI focus", but a reality-check showed the programmatic-spawn tab-add already uses `push_tab_preserve_focus` (no focus steal) and operator-foreground spawns correctly switch. The actual defect:

The Attached app's remote-agent sync (`src/app/mod.rs`) is **add-only** — when an agent goes away its tab is **retained** (with stale output) on purpose so scrollback survives. But when a **same-named** agent re-appears — **recovery respawn** (crash → respawn `agy-xxx`) or operator **create-after-delete churn** — the sync appends a **second** tab for that name (the stale one was never removed). Rapid churn → the tab bar fills with stale duplicate tabs.

## Fix
When the sync re-adds a name, if a retained **single-pane** tab for that agent already exists, **reuse it in place** (replace its stale dead-connection pane with the fresh reconnected pane) instead of appending a duplicate. New `Layout::single_pane_tab_index_for_agent`. Scoped to single-pane tabs so an operator-split tab is never clobbered. Preserves tab position + focus. A re-created agent is a new instance, so dropping the stale scrollback + showing the fresh connection is correct.

Fixed in the PR for this issue.

suzuke:
## Re-scope (reality-check)
Filed as "spawn steals TUI focus", but a spike (`/tmp/1591-spike.md`) showed the programmatic-spawn tab-add already uses `push_tab_preserve_focus` (no steal) and operator-foreground spawns correctly switch. The lead confirmed the re-scope to the **real** bug.

## Bug
The Attached remote-agent sync (`src/app/mod.rs`) is **add-only** — a gone agent's tab is **retained** (stale output) on purpose so scrollback survives. But when a **same-named** agent re-appears — **recovery respawn** (crash → respawn `agy-xxx`) or operator **create-after-delete churn** (the spawn-verify cycle that surfaced this) — the sync appended a **second** tab for that name (the stale one was never removed). Repeat → the tab bar fills with stale duplicates.

## Fix (surgical, respects the retention tradeoff)
When the sync re-adds a name, if a retained **single-pane** tab for that agent already exists, **reuse it in place** — replace its stale dead-connection pane with the fresh reconnected pane — instead of appending a duplicate. New `Layout::single_pane_tab_index_for_agent`. Scoped to single-pane tabs so a tab the operator has split with other agents is never clobbered (falls back to append in that rare case). Preserves tab position + focus (`push_tab_preserve_focus` still used for genuinely-new agents).

A re-created agent is a **new instance**, so dropping the stale dead-agent scrollback and showing the fresh connection is correct — no scrollback-continuity concern.

## Tests
- `single_pane_tab_index_for_agent_1591` — finds by name; `None` when absent; **not** matched for a multi-pane (operator-split) tab.
- `reuse_retained_tab_no_duplicate_or_focus_steal_1591` — re-appearance reuses the retained tab in place: tab count unchanged (no duplicate), operator's active tab not stolen, slot carries the fresh pane.

## Preflight
`fmt` ✓ · `clippy --all-targets --features tray -D warnings` ✓ · `test --tests --features tray` (60 binaries, 0 failed) ✓ · `xwin check` ✓. Rebased onto current `main`.

Closes #1591

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -778,14 +778,37 @@ fn run_app(terminal: &mut DefaultTerminal, fleet_override: Option<&Path>) -> Res
                                 Ok(pane) => {
                                     let tab_name = pane.agent_name.clone();
                                     known_remote_agents.insert(tab_name.to_string());
-                                    layout.push_tab_preserve_focus(
-                                        crate::layout::Tab::new(tab_name.to_string(), pane),
-                                    );
+                                    // #1591: this sync is add-only — a gone
+                                    // agent's tab is RETAINED (stale output) so
+                                    // scrollback survives. If a same-named agent
+                                    // re-appears (recovery respawn / operator
+                                    // create-after-delete churn), REUSE the
+                                    // retained single-pane tab in place (the
+                                    // fresh pane reconnects to the new instance;
+                                    // the stale dead-connection pane is dropped)
+                                    // instead of appending a DUPLICATE tab.
+                                    // Preserves tab position + focus.
+                                    if let Some(idx) =
+                                        layout.single_pane_tab_index_for_agent(&tab_name)
+                                    {
+                                        layout.tabs[idx] = crate::layout::Tab::new(
+                                            tab_name.to_string(),
+                                            pane,
+                                        );
+                                        tracing::info!(
+                                            agent = %name,
+                                            "reused retained tab for re-appeared remote agent (no duplicate)"
+                                        );
+                                    } else {
+                                        layout.push_tab_preserve_focus(
+                                            crate::layout::Tab::new(tab_name.to_string(), pane),
+                                        );
+                                        tracing::info!(
+                                            agent = %name,
+                                            "opened tab for newly-appeared remote agent"
+                                        );
+                                    }
                                     needs_resize = true;
-                                    tracing::info!(
-                                        agent = %name,
-                                        "opened tab for newly-appeared remote agent"
-                                    );
                                 }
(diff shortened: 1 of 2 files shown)


suzuke:
**§3.12 review mirror — merge gate** @ e7e19bb3f217fb874a6caa172659582a1ca2ada9

- **codex** (primary): VERIFIED — reuse correctness (app/mod.rs:793-811, swaps stale dead pane for fresh connected pane) · single-pane guard (layout/mod.rs:103-107, pane_count==1 → multi-pane returns None → append fallback, no clobber) · no-focus-steal (in-place layout.tabs[idx]=, no active mutation) · tests targeted. (/tmp/1596-codex.md)
- **reviewer-2** (second): VERIFIED — retention tradeoff acceptable (re-created=new instance, old scrollback in daemon logs) · recovery-respawn no-race (old PTY dead, new reconnects fresh) · multi-pane protected · blast-radius minimal (pure index lookup, single call site). (/tmp/1596-reviewer-2.md)

CI: 3-platform Check ✓ · Coverage ✓ · LOC ✓ · audit ✓. Dual-VERIFIED + green. Non-blocking: doesn't retroactively dedupe pre-existing duplicate tabs (self-resolves on next restart). → admin squash merge.

PR #1582, opened 2026-06-01
feat(#1547): un-gate agy fleet MCP via official .agents/ Customization Roots

suzuke:
## #1547 — un-gate agy fleet MCP (complete)

Makes daemon-spawned **agy** actually load the fleet `send`/`inbox`/`task` MCP
tools. Two layers:

### M1/M2 (dev's commit 87b4657)
- `configure_agy` writes `<workspace>/.agents/mcp_config.json` (official agy
  Customization Roots) + flips `Backend::Agy.fleet_mcp_supported = true`.
- Recovery-safety: busts agy's HOME-level MCP discovery cache on every
  (re)configure; Stage-2 respawn re-runs `configure()`; spawn boot invariant
  warns if `.agents/mcp_config.json` is missing.
- `instructions_path` → `.agents/AGENTS.md`. Self-contained (NOT coupled to
  `configure_gemini`, which sunsets 2026-06-18 / #1580).

### (A) non-hidden workspace link (commit c64525b — this takeover)
The blocker M1/M2 alone didn't solve: agy **refuses any workspace whose path
has a dot-prefixed (hidden) ancestor** (`addWorkspaceFolder` → `is hidden:
ignore uri`, root.go:132), and every daemon workspace lives under `$AGEND_HOME`
(`~/.agend-terminal`, hidden) → `.agents/` never read.

**Mechanism (operator e2e-verified):** agy decides "hidden" from the `$PWD`
env var, not getcwd/realpath. So spawn agy with CWD at the real hidden
workspace (validated allowed root) but `$PWD` pointed at a non-hidden link to
that same dir. Hidden check passes; project discovery resolves the realpath
(same antigravity project, no duplication); `.agents/` loads.

- `src/agy_workspace.rs`: `ensure_link` (spawn) / `remove_link` (teardown) /
  `link_base` (configurable). **Unix symlink**, **Windows directory junction**
  (`junction` crate — needs no Developer Mode, unlike `symlink_dir`).
- `agent::build_command`: agy → `ensure_link` + `cmd.env("PWD", link)`; CWD
  stays the real hidden dir. Set after the fleet.yaml env loop (daemon value
  authoritative). Link failure non-fatal (warn; falls back to pre-#1547).
- `agent_ops::cleanup_working_dir`: `remove_link` on teardown — only the managed
  link, never the real workspace.
- `FleetConfig::agy_workspace_link_base` (default `<user_home>/agend-ws`).
- CLAUDE.md: documents the hidden-path mechanism + the link fix.

[...]

PR #1599, opened 2026-06-01
fix(#1597): conditional agy workspace link (only when cwd is hidden)

### LINKED ISSUE #1597: [bug] agy backend ignores instance-specific working_directory and forces ~/agend-ws/ link
The `agy` backend handling in `src/agy_workspace.rs` and `src/agent/mod.rs` always creates a non-hidden symlink/junction under a global base directory (defaulting to `~/agend-ws/`) and sets `agy`'s `$PWD` to this link.

Even if a user specifies a non-hidden `working_directory` for an `agy` instance in `fleet.yaml`, the daemon still forces it through the `~/agend-ws/[instance_name]` link. This makes the `working_directory` parameter feel ineffective for `agy` as the agent will report its path as being under `~/agend-ws/` rather than the user-specified directory.

**Evidence:**
In `src/agy_workspace.rs`:
```rust
fn default_link_base(home: &Path) -> PathBuf {
    // ...
        .join("agend-ws")
}

pub fn link_path(home: &Path, instance: &str) -> PathBuf {
    link_base(home).join(instance)
}
```
The `link_path` is hardcoded to use `link_base(home)` (which is either the global `agy_workspace_link_base` or the default `~/agend-ws`) joined with the instance name. It does NOT check if the instance's `working_directory` is already a suitable non-hidden path.

In `src/agent/mod.rs`:
```rust
    if matches!(detected_backend, Some(Backend::Agy)) {
        if let (Some(dir), Some(home_path)) = (working_dir, home) {
            match crate::agy_workspace::ensure_link(home_path, name, dir) {
                Ok(link) => {
                    cmd.env("PWD", &link);
```
The `$PWD` is always set to the generated link, overriding any user-expected behavior of using `working_directory` directly if it were non-hidden.

**New Requirement: Fail-fast on Hidden Working Directory**
Since `agy` cannot handle directories with dot-prefixed (hidden) components in their path, the daemon should perform a fail-fast check during `fleet.yaml` initialization.
If an `agy` instance is configured with a `working_directory` that contains any hidden component (e.g., `~/.agend-ws/`), the daemon should report an error and refuse to start. This prevents hard-to-debug failures where `agy` silently ignores the workspace or fails to load fleet MCP.

**Suggested Fix:**
1. Check if `working_directory` is already non-hidden. If it is, use it directly as `$PWD` for `agy` and skip the link creation. If it is hidden, then fall back to the link mechanism.
2. Add validation logic in `src/fleet/mod.rs` (or during agent spawn validation) to detect and reject hidden `working_directory` paths for `agy` backends early.

Related to #1582.

/cc #1523

suzuke:
Fixes #1597 (cheerc's observation #1). Refines the #1582 agy workspace-link mechanism.

## Problem
`#1582`'s agy arm in `build_command` created a non-hidden link + `PWD=link` for **every** agy instance, unconditionally. For an explicit **non-hidden** `working_directory` — one agy would accept as-is — this:
- **shadowed** it (agy reported `~/agend-ws/<instance>` instead of the user's dir), and
- left a **redundant symlink**.

## Fix — conditional link
Decide on the resolved cwd (the path agy actually sees, captured from the `working_dir` block):
- **No hidden component** → `cmd.env("PWD", <resolved cwd>)` and **skip** `ensure_link`. agy reports the user's real dir; no stray symlink.
- **Dot-prefixed ancestor** (the default `$AGEND_HOME/workspace/<name>` under `~/.agend-terminal`) → `ensure_link` + `PWD=link`, exactly as #1582. This is the case the mechanism exists for.

New helper `agy_workspace::path_has_hidden_component(p)` walks `Component`s for a leading `.`, matching agy's "any dot-prefixed ancestor is hidden" rule (`root.go:132`). `remove_link` teardown is already idempotent (no-op when no link was created).

## Rejected: cheerc's suggestion #2 (fail-fast on hidden dir)
Not implemented — it **contradicts the entire purpose** of #1547/#1582: the default workspace is *deliberately* hidden and the link **is** the fix, so a fail-fast would break the default (and only common) case. At most a debug log; no hard error.

## Tests
- `build_command_agy_non_hidden_workspace_uses_real_dir_no_link` — case (a).
- `build_command_agy_hidden_workspace_uses_link` — case (b), the default.
- `build_command_agy_hidden_ancestor_uses_link` — case (c), non-hidden leaf / hidden ancestor.
- `path_has_hidden_component_detects_dot_prefixed_segments` — helper unit test.
- clippy `--all-targets -D warnings` green; full `cargo test --bin` (3037) green.

**Credit:** @cheerc for the #1597 observation (#1).

### DIFF
--- a/src/agent/mod.rs
+++ b/src/agent/mod.rs
@@ -847,6 +847,9 @@ fn build_command(config: &SpawnConfig) -> anyhow::Result<(CommandBuilder, Option
         }
     }
 
+    // #1597: the resolved cwd we actually hand the PTY — captured so the agy
+    // `$PWD` arm below decides on (and points at) the SAME path agy will see.
+    let mut spawn_cwd: Option<PathBuf> = None;
     if let Some(dir) = working_dir {
         // Defense-in-depth: the API spawn handler already calls
         // validate_working_directory at admission, but a symlink could have
@@ -865,6 +868,7 @@ fn build_command(config: &SpawnConfig) -> anyhow::Result<(CommandBuilder, Option
             let resolved = crate::api::validate_working_directory(dir, home_path)
                 .map_err(|e| anyhow::anyhow!("working_directory escapes via symlink: {e}"))?;
             cmd.cwd(&resolved);
+            spawn_cwd = Some(resolved);
         } else {
             tracing::warn!(
                 instance = %name,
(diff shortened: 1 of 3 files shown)


suzuke:
**§3.12 review mirror — merge gate** @ 29c6e4d6af4f2676b717ff58a1e0c03d0a298361

- **codex**: VERIFIED — path_has_hidden_component only flags Normal dot-prefixed segments (RootDir/./.. excluded, agy_workspace.rs:62-69); SAME-path invariant held (spawn_cwd hoisted once, used for hidden-check + PWD + link target, agent/mod.rs:850-923); default hidden workspace still uses link (pinned); remove_link no-op when no link; 3 case tests + unit test. (/tmp/1599-codex.md)
- **reviewer-2**: VERIFIED — symlink resolution (spawn_cwd is resolved path → no slip-through into hidden); non-hidden branch PWD=resolved + configure_agy writes .agents/ into same dir → MCP loads; behavior parity for default workspace; blast-radius = agy arm only. (/tmp/1599-reviewer-2.md)

CI all green. Fixes the valid half of #1597 (conditional link — honor non-hidden working_directory). Suggestion #2 (fail-fast on hidden) rejected — contradicts #1582's purpose. Credits @cheerc. Dual-VERIFIED → admin squash merge.

PR #1603, opened 2026-06-01
fix(#1602): reply text→message + enforce MCP inputSchema at dispatch

### LINKED ISSUE #1602: [DX] MCP tools: unify content param to `message` (rename reply.text) + enforce declared inputSchema (required/unknown) at dispatch
## Problem

Two related DX gaps in the MCP tool surface, surfaced by a real failure today (lead repeatedly called `reply({message: ...})` — `reply`'s param is `text`, not `message` — and got a *misleading* `Bad Request: message text is empty` from Telegram instead of any hint about the wrong param; daemon log showed `reply from=fixup-lead text=` (empty) while other agents had full text).

### Gap 1 — inconsistent content-param naming
- `send` → `message` (`src/mcp/tools.rs:34,59`)
- `schedule` → `message` (`:252`)
- `reply` → **`text`** (`:16,19`) ← the only outlier

`reply` being the odd one out is a footgun (easy to reach for `message`, as `send` uses it).

### Gap 2 — no schema validation at the MCP dispatch boundary (the real root)
The schemas correctly declare `"required": [...]` (e.g. reply `:19` declares `["text"]`), but **nothing enforces it**. Validation today is ad-hoc per-handler (`anti_stall.rs:47` `task_id_required`, `dispatch.rs:404` "key is required"). So a call missing a required field (or using a wrong/unknown field) is **not rejected at the boundary** — it proceeds with an empty/None value and fails late & misleadingly downstream. Had `required` been enforced, my call would've been rejected immediately with `reply: missing required parameter 'message'`, naming the correct field.

### Part A — rename `reply.text` → `message` (full rename, NO alias)
Unify on `message` (the name `send`/`schedule` already use). Operator explicitly chose a clean rename over an additive alias.
- `def_reply` in `src/mcp/tools.rs`: property `text` → `message`, `"required": ["text"]` → `["message"]`.
- reply handler (`src/mcp/handlers/channel.rs`): read `message`.
- Audit + update any caller/skill/script that passes `text` to `reply` (agents build calls from the live schema so they self-adapt, but hardcoded callers must be updated).
- Update tests.

### Part B — enforce declared schema at MCP dispatch
- Before invoking a handler, validate args against the tool's declared `inputSchema`: **reject missing `required` fields** with a clear `<tool>: missing required parameter '<name>'`.
- **unknown params**: at minimum WARN (`unknown parameter '<name>' for <tool>`); ideally suggest the nearest valid name.
- ⚠ Rollout caution: some handlers may declare a field `required` but tolerate its absence in practice — a blanket hard-reject could break them. Audit the `required` lists for enforceability first; if any are unsafe, WARN-first for those (or fix the schema) and hard-reject only where safe. The reply/send/core tools should be hard-reject.

## Acceptance
- `reply` accepts `message` (and `text` is gone — clean rename, no alias).
- All content-param tools consistently use `message`.
- A call missing a required field is rejected at the boundary with a clear message naming the field (verify with reply-without-message → `missing required parameter 'message'`).
- Unknown params are at least warned.
- Existing valid calls unaffected.

## Related
- Surfaced 2026-06-01 (lead reply failures + operator catch).
- `def_schedule` schema-completeness recurrences (#1095, #1600) are the same class of "schema vs reality drift" that Part B's validation would help guard.

suzuke:
Root-causes the operator's "Telegram text is empty" confusion. Two parts.

## Part A — rename `reply.text` → `reply.message` (full rename, no alias)
`reply` was the only content tool using `text`; `send`/`schedule` already use `message`.
- `def_reply` inputSchema property + `required` → `message`; `handle_reply` reads `args["message"]`.
- All hardcoded callers/tests updated (`channel_p0a_tests` ×7, `handlers/tests`, `docs/MCP-TOOLS.md`).
- Dropped `send`'s vestigial `text` fallback (`comms.rs`). `send`/`reply`/`schedule` are now uniformly `message`.

## Part B — enforce declared `inputSchema` at the single dispatch chokepoint
Every tool declares `required[]` but nothing enforced it → a mis-named/omitted param failed **late and misleadingly** inside the handler. Now, in `try_dispatch`:
- missing **required** key → hard-reject `<tool>: missing required parameter '<name>'`;
- **unknown** key → warn only (forward-compat, never reject).

### Enforceability — schema-align, not an allowlist (per codex review + lead decision)
A systematic **grep-based** re-audit (the first eyeballed pass missed `mode`) found **5 tools** whose HANDLER **defaults** a declared-"required" field instead of erroring — so their schemas were **lying**. Rather than allowlist them (band-aid; schema stays dishonest), the field is removed from each `required[]` so the schema tells the truth and the validator stays a plain hard-reject:

| tool | field | handler behavior on absence |
|---|---|---|
| `mode` | `action` | defaults to `"get"` (read-only) |
| `create_instance` | `name` | team mode auto-names; single path still errors `missing 'name'` |
| `set_waiting_on` | `condition` | absent = "clear" |
| `set_display_name` | `name` | tolerates absent (sets `""`) |
| `set_description` | `description` | tolerates absent (sets `""`) |

Tools whose handler **errors** on a missing field (`reply.message`, `send.message`, `delete_instance.instance`, action-based `task`/`decision`/…) keep `required[]` and are hard-rejected.

**Zero behavior change**: handlers are untouched; only the schemas stop lying. (Whether `set_display_name`/`set_description` *should* hard-error on a missing value is a separate behavior question → follow-up, not bundled here.)

## Tests
- `validate_rejects_missing_required_with_named_error`, `try_dispatch_rejects_reply_without_message` (e2e).
- `validate_passes_when_required_present_and_unknown_only_warns`.
- `genuinely_required_fields_are_hard_rejected` (reply/send/delete_instance/task).
- `handler_defaulted_fields_are_not_declared_required` — **pins the audit**: asserts each of the 5 omits its defaulted field from `required[]` AND passes validation when omitted; re-trips if a future tool declares a handler-defaulted field required.
- Two existing `create_instance` schema tests updated (they asserted the old lying `required:["name"]`).
- Full `cargo test --bin` (3043) + mcp invariants + clippy `--all-targets -D warnings` green.

Closes #1602.

### DIFF
--- a/docs/MCP-TOOLS.md
+++ b/docs/MCP-TOOLS.md
@@ -60,7 +60,7 @@ Check pending messages, look up by ID, or fetch thread messages.
 
 ### `reply`
 Reply to the user via the active channel (NOT for inter-agent use).
-- **text**: reply content
+- **message**: reply content
 - default_action, timeout_secs
 
 ### `download_attachment`
--- a/src/mcp/handlers/channel.rs
+++ b/src/mcp/handlers/channel.rs
@@ -4,7 +4,11 @@ use std::path::Path;
 use std::sync::Arc;
 
 pub(super) fn handle_reply(home: &Path, args: &Value, instance_name: &str) -> Value {
-    let text = args["text"].as_str().unwrap_or("").to_string();
+    // #1602: the reply content param is `message` (was `text`) — now consistent
+    // with `send`/`schedule`. The MCP dispatch validator rejects a missing
+    // `message` with a clear named error, so a mis-named param no longer
+    // silently becomes an empty reply.
+    let text = args["message"].as_str().unwrap_or("").to_string();
     tracing::info!(from = %instance_name, %text, "reply");
 
     // Sprint 59 Wave 1 PR-4 ((B) decision default with timeout):
--- a/src/mcp/handlers/channel_p0a_tests.rs
+++ b/src/mcp/handlers/channel_p0a_tests.rs
@@ -140,7 +140,7 @@ fn ec1_3_4_5_7_snapshot_at_handler_time_falls_back_to_active_when_reply_to_none(
     set_reply_to("alpha", None);
     crate::channel::register_active_channel(MockChannel::arc("telegram", ReplyOutcome::Ok));
 
-    let result = super::handle_reply(&home, &serde_json::json!({"text": "hi"}), "alpha");
+    let result = super::handle_reply(&home, &serde_json::json!({"message": "hi"}), "alpha");
     assert_eq!(
         result["message_id"], "mock-msg-1",
         "fallback to singleton must succeed: {result}"
@@ -158,7 +158,7 @@ fn ec2_6_10_unavailable_returns_structured_error_when_lookup_misses() {
     set_reply_to("alpha", Some("discord"));
     crate::channel::register_active_channel(MockChannel::arc("telegram", ReplyOutcome::Ok));
 
-    let result = super::handle_reply(&home, &serde_json::json!({"text": "hi"}), "alpha");
+    let result = super::handle_reply(&home, &serde_json::json!({"message": "hi"}), "alpha");
     assert_eq!(result["code"], "reply_channel_unavailable");
     assert!(
         result["error"]
@@ -179,7 +179,7 @@ fn ec8_no_active_channel_when_no_external_channel_registered() {
     let home = tmp_home("ec8");
     set_reply_to("alpha", None);
 
-    let result = super::handle_reply(&home, &serde_json::json!({"text": "hi"}), "alpha");
+    let result = super::handle_reply(&home, &serde_json::json!({"message": "hi"}), "alpha");
     assert_eq!(
         result["code"], "no_active_channel",
         "no channels registered must surface no_active_channel: {result}"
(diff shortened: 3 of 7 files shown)


suzuke:
**§3.12 review mirror — merge gate** @ 48893ac0b67ff3daea36f679db82969732cfb0ec (round-2)

- **codex** (found round-1 mode gap): VERIFIED r2 — mode conditional fixed (def_mode no longer requires action); re-audited, NO 6th conditional-required mismatch beyond the documented 5; genuine requireds (reply.message/send.message/delete_instance.instance/action) still strict; create_instance no-name guard intact at handler. (/tmp/1603-codex-r2.md)
- **reviewer-2**: VERIFIED r2. (/tmp/1603-reviewer-2-r2.md)

Resolution: (b) schema-align — removed the 5 handler-defaulted fields from their schemas' required[] (mode/create_instance/set_waiting_on/set_display_name/set_description), dropped the allowlist, validator plain hard-rejects the now-honest required[]. Guard tests pin the audit. CI all green. Part A: reply text→message (no alias) + send vestigial-text removal. set_name empty-set behavior → #1604 follow-up. Dual-VERIFIED → admin squash merge. Closes #1602.

Issue #1521, opened 2026-05-31
Schedule: add 'fire strategy' to avoid redundant triggers after task succeeds within a window


suzuke:
## Problem

Cron-based schedules in agend-terminal currently fire on every match of the cron expression. There's no way to express **"fire until the task succeeds, then stop for this period"**, which is a very common pattern for retry-style schedules.

### Real-world example

I have a recurring task that downloads + summarizes the latest Gooaye podcast and pushes to Telegram. To handle the fact that uploads happen at variable times (12:00-20:00 Asia/Taipei), I configured 4 fires per Wed/Sat:

```
cron: 0 15,16,21,22 * * 4,7  (Quartz, Wed+Sat at 15/16/21/22)
```

Today (Saturday) was actually optimal:
- **15:00** → no new video → reply: "no new video, retrying"
- **16:00** → caught EP666 just uploaded → full pipeline runs (download → Whisper → summary → 3-part Telegram message), takes ~5 min, marks `usGcXJjHwr0` in dedup file
- **21:00** → wakes Claude, fetches video metadata, checks dedup, sends \"already processed, skipping\" to Telegram
- **22:00** → same as 21:00

The 21:00 and 22:00 fires are completely wasted: they spend tokens, run yt-dlp, send noise to the user's Telegram. The user explicitly noticed and asked \"can we make this cleaner?\".

### Workarounds and their problems

1. **Silent skip in prompt** — modify Step 4 to not call \`reply\`. Eliminates user-facing noise but Claude still wakes up and burns tokens. Treats the symptom, not the cause.
2. **Self-disable after success** — task calls \`schedule update enabled=false\` when done. Needs a second schedule to re-enable next period — error-prone, fragile dependency between two schedules.
3. **Self-modify cron after success** — rewrite cron to point at next target window. Cron edits are error-prone (we already shipped one bug from Quartz vs Unix dow confusion in this same schedule, see #1095).

None of these belong in user-land. This is a daemon-level scheduling concept.

## Proposal

Add one of these (or both) to the schedule schema:

### Option A: \`max_fires_per_day\`

```rust
\"max_fires_per_day\": Option<u32>,  // default None = unlimited
```

Daemon stops firing for the remainder of \"the local day\" (timezone-aware) once the configured limit is reached. Resets at local midnight.

For my case: \`max_fires_per_day: 1\` — first cron match of the day fires; if it succeeds, subsequent same-day matches are skipped.

[...]

PR #1609, opened 2026-06-01
fix(#1608): until_success task-existence via event-sourced lookup, not a file probe

### LINKED ISSUE #1608: fire_strategy=until_success unusable: task_file_exists() checks wrong storage path
## TL;DR

`validate_fire_strategy()` calls `task_file_exists(home, task_id)`, which checks `~/.agend-terminal/tasks/<task_id>.json`. **That path does not exist** — AgEnD persists tasks via event-sourcing (`task_events.jsonl`) + a single `tasks.json` aggregate (or in-memory state), never as per-task JSON files in a `tasks/` directory. As a result, the validator's existence check **always returns false**, so any attempt to set `fire_strategy=until_success` is rejected regardless of whether the `linked_task_id` is real.

The feature is unreachable end-to-end (despite PR #1557 implementing the backend logic and PR #1601 wiring up the MCP schema). This is a second incomplete-fix bug on top of #1600.

## Reproduction

```bash
# 1. Create a real, valid task via the task MCP tool
$ task action=create title=\"test\" ...
→ task id: t-20260601143201824184-0
→ status: open
→ visible in `task action=list` ✓

# 2. Try to enable until_success on a schedule with that task linked
$ schedule action=update id=s-... \\
    fire_strategy=until_success \\
    linked_task_id=t-20260601143201824184-0
→ Error: \"fire_strategy=until_success requires an existing linked_task_id (task 't-20260601143201824184-0' not found)\"
```

The task **does** exist in AgEnD's task system. The validator just looks in the wrong place.

## Evidence — what the code does

`src/schedules.rs:177-179`:

```rust
fn task_file_exists(home: &Path, task_id: &str) -> bool {
    !task_id.is_empty() && home.join(\"tasks\").join(format!(\"{task_id}.json\")).exists()
}
```

This checks `~/.agend-terminal/tasks/<task_id>.json`.

## Evidence — what AgEnD actually does

```bash
$ ls ~/.agend-terminal/ | grep -iE 'task'
task_events.jsonl       # event-sourcing log (where tasks actually live)
task_events.jsonl.lock
task-progress/          # per-task progress files (different system)
tasks.lock              # lock for tasks.json (but file doesn't exist yet)

$ ls ~/.agend-terminal/tasks/ 2>&1
ls: /Users/suzuke/.agend-terminal/tasks/: No such file or directory

$ ls ~/.agend-terminal/tasks.json 2>&1
ls: /Users/suzuke/.agend-terminal/tasks.json: No such file or directory
```

`src/tasks/mod.rs:104-106` confirms the intended storage is a single aggregate file:

```rust
fn store_path(home: &Path) -> std::path::PathBuf {
    crate::store::store_path(home, \"tasks.json\")
}
```

So:
- Tasks are written to `task_events.jsonl` (event log) and aggregated in memory.
- `tasks.json` (the aggregate snapshot) is the intended on-disk form, but currently absent on this daemon — likely lazy-written or only flushed on certain triggers.
- There is no `tasks/<id>.json` per-task file convention anywhere in the codebase.

`task_file_exists()` was apparently written against a mental model that doesn't match the actual task subsystem.

## Why this is unrecoverable from the caller side

The caller (operator, LLM agent) cannot work around this:
- Creating a task via the MCP `task` tool works — the task is fully registered, visible to `task action=list`, has correct id, status, assignee, etc.
- But the file `~/.agend-terminal/tasks/<id>.json` will never exist, because AgEnD doesn't write per-task files.
- No combination of operator actions can satisfy `task_file_exists()`.

## Fix

Replace the filesystem-based check with the actual task lookup. Two options:

### Option 1: use the task subsystem's loader

```rust
fn task_exists(home: &Path, task_id: &str) -> bool {
    if task_id.is_empty() { return false; }
    crate::tasks::load(home).tasks.iter().any(|t| t.id == task_id)
}
```

Pros: matches what the task subsystem considers authoritative.
Cons: loads full aggregate each call (cheap for current scale, but maybe not ideal).

### Option 2: query an indexed view

If `tasks::mod` exposes a `task_id_exists(home, id) -> bool` (or similar), use that directly. Keeps validation cheap and avoids loading the full store.

If no such function exists, add one — this kind of \"does this task id exist\" check is generally useful and probably wanted in other places too.

### Either way: rename

Rename `task_file_exists` → `task_exists`. The current name encodes the wrong abstraction (it implies filesystem layout matters, which it doesn't).

## Acceptance

After fix:
- `schedule.update {fire_strategy: \"until_success\", linked_task_id: <real_task_id>}` succeeds when `<real_task_id>` is a task visible to `task action=list`.
- Validator still rejects when `linked_task_id` is missing, empty, or non-existent.
- No regression for `fire_strategy=always` (the early-return path).
- Ideally: add an integration test that creates a task via the task subsystem, then validates the schedule update succeeds — this would have caught the bug at PR time.

## Scope

- 1 file: `src/schedules.rs`
- ~5 lines changed (replace one function body)
- No data migration
- No behavior change for `always` users

## Why this matters

Same logic as #1600: the original use case for #1521 (Gooaye summary schedule with `until_success` to stop firing after first daily success) is **still** blocked. After two follow-up PRs:
- #1557 added the backend logic (but missed the MCP schema + this bug)
- #1601 fixed the MCP schema (but didn't catch this bug)

The feature has been in main for ~24 hours and is still unreachable. A user-facing dry-run on the actual happy path would have caught both #1600 and this issue.

## def_schedule streak update

This `schedule` feature surface has now accumulated:
- #1095 — cron documentation (Unix vs Quartz dow)
- #1521 — fire_strategy feature request
- #1557 — implementation, partial
- #1600 — MCP schema exposure (follow-up to #1557)
- **#XXXX (this) — validator uses wrong task storage abstraction**

5 issues across ~3 weeks on one feature area. Worth a postmortem on the PR review checklist: every backend change touching MCP-exposed config needs (a) MCP schema sync, (b) end-to-end happy-path verification by the requester before close.

## Related

- #1521 — fire_strategy feature request
- #1557 — implementation PR (introduced the bug)
- #1600 — MCP schema exposure (also caused by #1557 incompleteness)
- #1601 — fix for #1600

suzuke:
**Operator-filed — the 3rd bug in the #1521 `fire_strategy=until_success` feature.**

## Bug
`validate_fire_strategy` gated `until_success` on `task_file_exists`, which probed `~/.agend-terminal/tasks/<id>.json`. AgEnD's task board is **event-sourced** — that per-task JSON file never exists — so the check **always returned `false`**, every `until_success` create/update was rejected, and the feature was unreachable end-to-end.

## Fix (1 file, `src/schedules.rs`)
Replace the file probe with the task subsystem's authoritative lookup, and rename the helper (the old name baked in the wrong filesystem abstraction):
```rust
fn task_exists(home: &Path, task_id: &str) -> bool {
    !task_id.is_empty() && crate::tasks::load_by_id(home, task_id).is_some()
}
```
`tasks::load_by_id` reads the replayed event state — where operators' tasks actually live.

## Why this shipped (load-bearing test fix)
The existing `create_until_success_with_existing_task_ok` test seeded a **fake** `tasks/<id>.json` file (`seed_task_file`), so it passed against the broken file-probe while the real feature was dead — the test mirrored the bug. Replaced with `seed_real_task`, which emits a real `TaskEvent::Created` via `task_events::append` (mirroring the task subsystem's own `create_task` test helper), so the happy path now exercises the REAL lookup. Added `validate_fire_strategy_all_branches` pinning:
- `until_success` + existing event-sourced task → **Ok** (the regression);
- `until_success` + non-existent / empty / `None` → **Err**;
- `always` → **Ok** regardless (early return).

Full `cargo test --bin` (3058) + clippy `--all-targets -D warnings` green.

Closes #1608.

### DIFF
--- a/src/schedules.rs
+++ b/src/schedules.rs
@@ -174,8 +174,13 @@ impl From<ScheduleRaw> for Schedule {
 /// #1521: a `linked_task_id` is required for `UntilSuccess` and the task must
 /// already exist on the board — a bare-message reminder has no completion to
 /// gate on, so we reject it at create/update rather than silently degrade.
-fn task_file_exists(home: &Path, task_id: &str) -> bool {
-    !task_id.is_empty() && home.join("tasks").join(format!("{task_id}.json")).exists()
+/// #1608: existence is decided by the task subsystem's authoritative lookup,
+/// NOT a `tasks/<id>.json` file probe — AgEnD's task board is event-sourced and
+/// keeps no per-task JSON, so the old filesystem check always returned `false`
+/// and made `fire_strategy=until_success` permanently unreachable. The old name
+/// (`task_file_exists`) baked in that wrong filesystem abstraction.
+fn task_exists(home: &Path, task_id: &str) -> bool {
+    !task_id.is_empty() && crate::tasks::load_by_id(home, task_id).is_some()
 }
 
 /// #1521: validate a (fire_strategy, linked_task_id) pair. `Ok(())` when the
@@ -189,7 +194,7 @@ fn validate_fire_strategy(
         return Ok(());
     }
     match linked_task_id {
-        Some(id) if task_file_exists(home, id) => Ok(()),
+        Some(id) if task_exists(home, id) => Ok(()),
         Some(id) => Err(format!(
             "fire_strategy=until_success requires an existing linked_task_id (task '{id}' not found)"
         )),
@@ -801,14 +806,37 @@ mod tests {
 
     // ── #1521: fire-strategy validation + backward-compat ──
 
-    fn seed_task_file(home: &Path, id: &str) {
-        let dir = home.join("tasks");
-        std::fs::create_dir_all(&dir).unwrap();
-        std::fs::write(
-            dir.join(format!("{id}.json")),
-            serde_json::json!({"id": id, "status": "open"}).to_string(),
+    /// #1608: create a REAL task on the event-sourced board (so
+    /// `tasks::load_by_id` finds it), NOT a `tasks/<id>.json` file. The old
+    /// `seed_task_file` wrote a file the real lookup never reads — which is
+    /// exactly why this test passed while the feature was broken. Mirrors the
+    /// task subsystem's own `create_task` helper (tasks/handler.rs).
+    fn seed_real_task(home: &Path, id: &str) {
+        crate::task_events::append(
+            home,
+            &crate::task_events::InstanceName::from("test:operator"),
+            crate::task_events::TaskEvent::Created {
+                task_id: crate::task_events::TaskId(id.into()),
+                title: "test task".into(),
+                description: String::new(),
+                priority: "normal".into(),
+                owner: None,
+                due_at: None,
+                depends_on: Vec::new(),
+                routed_to: None,
+                branch: None,
+                bind: None,
+                eta_secs: None,
+                tags: vec![],
+                parent_id: None,
+            },
         )
-        .unwrap();
+        .expect("seed real task");
+        // Sanity: the authoritative lookup the fix relies on must now resolve.
+        assert!(
+            crate::tasks::load_by_id(home, id).is_some(),
+            "seeded task '{id}' must be visible to tasks::load_by_id"
+        );
     }
 
     #[test]
@@ -859,7 +887,7 @@ mod tests {
     #[test]
     fn create_until_success_with_existing_task_ok() {
         let home = tmp_home("fs-ok");
-        seed_task_file(&home, "t-real");
+        seed_real_task(&home, "t-real");
         let r = create(
             &home,
             "a",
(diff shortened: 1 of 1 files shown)


suzuke:
**§3.12 review mirror — merge gate** @ 452ad593db437c3949090ef515959a8b543d482f

- **codex**: VERIFIED — load_by_id (event-sourced authoritative) replaces the never-existing tasks/<id>.json probe; the OLD test seed_task_file (fiction file matching the broken probe) replaced with seed_real_task (task_events::append Created) so the happy-path test now exercises the REAL lookup (would catch a regression) — addresses the root-cause-of-why-it-shipped.
- **reviewer-2**: VERIFIED — e2e until_success now reachable; test fidelity (seed_real_task mirrors real handler); no regression for always; load_by_id cost acceptable.

CI all green. 3rd fire_strategy bug (#1521 feature) closed — until_success is now reachable end-to-end. Dual-VERIFIED → admin squash merge. Closes #1608.

Issue #1018, opened 2026-05-20
P2: dispatch_idle watchdog should clear entries for nonexistent tracked agents / closed tasks / placeholder correlation_ids


suzuke:
## Severity
**P2** — operator-facing noise. Watchdog continues firing \`dispatch_idle_threshold_exceeded\` notifications for stale dispatches whose target agent no longer exists OR whose task_id has been closed/marked done on the task board OR whose correlation_id is a placeholder like \`t-pending\` shared across multiple dispatches.

## Observed symptom (2026-05-20 → 2026-05-21 session)

Multiple stale watchdog notifications fired throughout the autonomous session:

\`\`\`
[dispatch_idle_threshold_exceeded] dispatch disp-XXX from 'fixup-lead' → 'fixup-dev-2' (kind=task, correlation_id=t-pending) idle for 2738s
[dispatch_idle_threshold_exceeded] dispatch disp-YYY from 'fixup-lead' → 'fixup-dev-2' (kind=task, correlation_id=t-pending) idle for 3653s
... etc.
\`\`\`

Examples of stale cases observed:

1. **Placeholder \`t-pending\` collision** — Lead used \`t-pending\` as a correlation_id placeholder across multiple parallel dispatches (a known lead-side mistake, but the watchdog could be defensive). Each dispatch creates a separate sidecar; when the first \`kind=report\` arrives keyed on \`t-pending\`, the watchdog clears one slot but other sidecars with the same \`t-pending\` linger and fire later.

2. **Closed task_id** — When a task is marked \`done\` on the board (\`task action=done\`), associated dispatch sidecars are not cleared. If the dispatcher's \`kind=report\` was missed or arrived without correlation_id match, the sidecar fires at threshold even though the work is closed.

3. **Nonexistent agent** — If an instance is deleted (\`delete_instance\`), any pending dispatch sidecars targeting that instance never clear because no \`kind=report\` can ever arrive from a deleted agent.

## Fix design proposal (dev-2 dispatched)

Three independent cleanup paths:

**(A) Watchdog tick-time validation**:
- On watchdog tick, before firing \`dispatch_idle_threshold_exceeded\` for a pending sidecar, validate:
  - target agent still exists in fleet.yaml registry
  - task_id (if present and not placeholder) still has status \`open\` / \`claimed\` / \`in_progress\` / \`blocked\` on the board
  - correlation_id is not a known-placeholder sentinel (\`t-pending\`, \`t-tbd\`, empty string)
[...]

PR #1615, opened 2026-06-01
fix(#1608b): close the tasks/<id>.json filesystem-probe class (until_success runtime gate)

### LINKED ISSUE #1614: fire_strategy=until_success STILL dead: cron runtime linked_task_gate uses wrong storage (incomplete #1608)
## TL;DR

#1608 fixed the create/update **validator** (`schedules.rs::task_exists` now uses event-sourced `load_by_id`), but the **runtime cron gate still uses the broken `~/.agend-terminal/tasks/<id>.json` filesystem probe** — that per-task file never exists (the board is event-sourced). So a `fire_strategy=until_success` schedule **self-disables on its very first fire**, and the #1521 feature is *still* dead end-to-end despite #1557 + #1600 + #1601 + #1608.

This is the **4th incomplete-fix bug** on the fire_strategy feature, found by an overnight bug-hunt sweep (adversarially verified).

## The runtime killer

`src/daemon/cron_tick.rs:238` (`linked_task_gate`) reads `tasks/<id>.json` → `NotFound` → `TaskGate::Missing` → consumer at `cron_tick.rs:102-112` calls `set_enabled(false)` → the schedule is **permanently disabled** the first time it fires. The user links a real, existing task; the runtime gate can't find it (wrong storage abstraction) and kills the schedule.

## Same class, 2 more sites

- `src/daemon/dispatch_idle/mod.rs:234` (`task_still_live`) — same probe → always `None` → the #1018 stale-sidecar suppression is dead code → closed tasks still emit watchdog nudges. (fail-open, low severity, but the suppression feature is dead.)
- `src/tasks/tests.rs:1643` — fiction test: backdates `tasks.json` `updated_at`, but `list` reads the event-sourced replay (never reads that file) → the done-TTL-bypass assertion can never fail (deleting the production `bypass` would still pass). `cron_tick.rs` `seed_task` (~572) is the same fiction pattern.

## Fix (class-closing)

Replace every `tasks/<id>.json` filesystem probe with `crate::tasks::load_by_id(home, id)` (the authoritative event-sourced lookup, same fix as #1608). Migrate the fiction tests to seed real `TaskEvent::Created` events. Add an invariant test that greps `src/` and fails if any production code reads `tasks/*.json`, so this class can't recur.

## Why it kept shipping

The schedule/task feature surface has now hit: #1095 (cron docs), #1521 (request), #1557 (partial impl), #1600 (missing MCP schema), #1608 (validator probe), and this (runtime probe + dead suppression + fiction tests). The recurring root: **code written against a `tasks/<id>.json` mental model that doesn't match the event-sourced task subsystem, plus tests that seed the fictional file so they pass while the feature is dead.** Worth a PR-review checklist item: any task-existence check must use the task subsystem's loader, never the filesystem; and tests must seed via the real create path.

suzuke:
Closes #1614. The class-closing follow-up to #1608, found by the overnight bug-hunt.

## The bug (P1)
#1608 fixed only the **create/update validator** (`schedules.rs::task_exists` → `tasks::load_by_id`). But the daemon **runtime paths** still probed a per-task `home/tasks/<id>.json` file — which the **event-sourced** board (`task_events.jsonl` via `replay`) **never writes**. So:
- `cron_tick.rs::linked_task_gate` always read `NotFound` → `TaskGate::Missing` → **`set_enabled(false)`** → every `fire_strategy=until_success` schedule **permanently self-disabled on its first fire**. The whole #1521/#1608 feature was dead in production.
- `dispatch_idle.rs::task_still_live` always returned `None` → the #1018 "skip the nudge for a closed task" branch was unreachable → watchdog nudges leaked for already-closed tasks.

Two *fiction tests* hid it: they seeded the phantom `tasks/<id>.json` file the real path never reads.

## Fix (closes the whole class)
- **`linked_task_gate`** → `task_events::replay` directly (not `load_by_id`), so a **transient replay error stays `ReadError` → fail-open (fire)**, never the destructive `Missing → disable`. Absent-from-log → `Missing` (rare; validated at create).
- **`task_still_live`** → `tasks::load_by_id` (fail-open on absent/error preserved — that path is safe).
- **Fiction tests → real seeds**: `cron_tick::seed_task` and the dispatch-idle closed-task seed emit real `Created`/`Done` events (mirror #1608's `seed_real_task`); `test_list_done_filter_returns_all` backdates `id1` by rewriting its **`task_events.jsonl`** envelope timestamp (the real read path), so it now genuinely fails if the done-TTL bypass is removed.
- **Regression guard**: new `tests/no_per_task_json_probe_invariant.rs` fails if the `("tasks").join(format!(…))` per-task-file probe reappears under `src/`.

## Tests
`linked_task_gate_replay_error_fails_open` (fail-open on corrupt log) + the migrated real-event tests. Full `cargo test --bin` (3064) + clippy `--all-targets -D warnings` green.

**`until_success` schedules now actually work in production.**

### DIFF
--- a/src/daemon/cron_tick.rs
+++ b/src/daemon/cron_tick.rs
@@ -217,37 +217,39 @@ enum TaskGate {
     Done,
     /// Task exists but isn't done → fire the reminder.
     Pending,
-    /// No task file (deleted/never-existed) → disable the schedule.
+    /// Task absent from the event log (link removed / board reset) → disable.
     Missing,
-    /// File present but unreadable / malformed status → fail-open (fire).
+    /// Replay error (transient log hiccup) → fail-open (fire), never disable.
     ReadError,
 }
 
-/// #1521: read a linked task's current status the same lock-free, fail-open way
-/// the dispatch-idle watchdog does (`tasks/{id}.json` direct read; the board
-/// writes per-task files). Current-status gate (not ever-done), so a reopened
-/// task resumes reminders. `NotFound` is `Missing` (a removed/typo'd link);
-/// any other I/O or parse error is `ReadError` → fail-open so a transient
-/// hiccup never silently drops the reminder.
+/// #1521: read a linked task's current status. Current-status gate (not
+/// ever-done), so a reopened task resumes reminders.
+///
+/// #1608b/#1614: read the EVENT-SOURCED board via `task_events::replay`, NOT a
+/// `tasks/{id}.json` file — that per-task file is never written (the board lives
+/// in `task_events.jsonl`), so the old probe always returned `NotFound` →
+/// `Missing` → `set_enabled(false)`, permanently self-disabling every
+/// `UntilSuccess` schedule on its first fire. (#1608 fixed only the create/update
+/// validator; this is the matching runtime fix.)
+///
+/// A replay ERROR is `ReadError` → fail-open (fire) so a transient log hiccup
+/// never silently drops — OR destructively disables — the reminder. A task
+/// genuinely ABSENT from the (append-only) log is `Missing` (a removed/reset
+/// link); since the task was validated at create-time, this is rare.
 fn linked_task_gate(home: &Path, task_id: Option<&str>) -> TaskGate {
     let Some(id) = task_id.filter(|s| !s.is_empty()) else {
         // UntilSuccess is validated to have a task at create/update, so a
         // None here means the link was lost — treat as Missing (disable).
         return TaskGate::Missing;
     };
-    let path = home.join("tasks").join(format!("{id}.json"));
-    let content = match std::fs::read_to_string(&path) {
-        Ok(c) => c,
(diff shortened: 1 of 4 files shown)


suzuke:
**§3.12 review mirror — merge gate** @ 4eb4bcc2501edbdfd898249eff4ef763d8be5590

- **codex**: VERIFIED — cron linked_task_gate uses task_events::replay (preserves fail-open on transient ReadError → fire, not destructive set_enabled(false)); only true absent→Missing; until_success runtime validated; fiction tests→real replay path; invariant guard added.
- **reviewer-2**: VERIFIED — no remaining path wrongly disables a healthy schedule; fail-open is the right tradeoff (over-fire recoverable vs destructive-disable); always-schedules unaffected.

CI green. Closes #1614 — the 4th (and final) fire_strategy bug: until_success now ACTUALLY works in production (cron runtime gate no longer self-disables on first fire). Class-closing: invariant test forbids the tasks/<id>.json filesystem-probe from recurring. Dual-VERIFIED → admin squash merge.

PR #1618, opened 2026-06-01
fix(bughunt2): propagate swallowed inbox/save errors (3 silent-loss P1s)

suzuke:
The 3 silent-loss P1s from the overnight bug-hunt #2 (`/tmp/bughunt-overnight-2.md`). Same theme: the caller is told **success** while a `let _ =` swallowed a **state-bearing** write Result → message/record silently lost.

## Fixes
- **`api/handlers/messaging.rs::handle_send`** — `route_and_deliver` swallowed `inbox::enqueue` and `handle_send` always returned `{ok:true}`. On a disk-low/I/O enqueue failure the message is dropped — total loss for the `inbox_only` / codex `skip_inject` branches where the inbox is the **sole** channel. `route_and_deliver` now returns `Result`; `handle_send` returns `{ok:false, error}` **before** the provenance/dispatch side-effects.
- **`agent_ops.rs::fallback_deliver`** — the last-resort path (daemon API down → inbox is the only channel) swallowed the enqueue and reported `inbox_fallback` success. Now returns an error JSON on enqueue failure.
- **`deployments.rs::deploy` / `teardown`** — `let _ = save(...)` dropped the `deployments.json` persist error and reported `status:deployed` / `torn_down`. A failed deploy-record save leaves instances live in fleet.yaml but **un-tear-down-able** (and resurrected on restart); both now surface the error with the affected instances so the operator can reconcile.

## Scope
**Only** critical delivery/persistence Results. Cosmetic log/notify swallows (tracing, best-effort notify) are left as-is, per the bug-hunt class definition.

## Tests
Each fix pinned: a forced enqueue failure (`home/inbox` as a file → `create_dir_all` errors) / a forced `deployments.json` save failure (target path as a directory → `atomic_write` rename fails) makes the caller surface an error instead of a fake success; existing success-path tests unchanged. Full `cargo test --bin` (3070) + clippy `--all-targets -D warnings` green.

### DIFF
--- a/src/agent_ops.rs
+++ b/src/agent_ops.rs
@@ -36,7 +36,16 @@ pub fn fallback_deliver(
     if !in_fleet {
         return json!({"error": format!("target instance '{target}' not found in fleet.yaml (API unavailable: {api_error})")});
     }
-    let _ = crate::inbox::enqueue(home, target, msg);
+    // #bughunt2: this is the last-resort path (the daemon API is already down),
+    // so the inbox is the SOLE channel. A swallowed enqueue here is total,
+    // unrecoverable message loss reported as success — surface it instead.
+    if let Err(e) = crate::inbox::enqueue(home, target, msg) {
+        return json!({
+            "error": format!(
+                "inbox fallback delivery to '{target}' failed — message lost (API unavailable: {api_error}): {e}"
+            )
+        });
+    }
     crate::inbox::notify_agent(home, target, &crate::inbox::NotifySource::Agent(from), text);
     json!({"target": target, "delivery_mode": "inbox_fallback", "note": format!("API unavailable: {api_error}")})
 }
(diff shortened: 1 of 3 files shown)


suzuke:
**§3.12 review mirror — merge gate** @ 9763384d73bd0d86ad7b63abbd20d0880ccb67f5

- **codex**: VERIFIED (r2) — propagation correct across all 3 sites (handle_send early-returns before side-effects; fallback_deliver; deploy+teardown); round-1 teardown test-gap resolved by the added forced-failure test (read-only home → save fails → error surfaced, perms restored).
- **reviewer-2**: VERIFIED — no failure-amplification; cosmetic swallows untouched.

CI all green. Fixes 3 silent-loss P1 (bug-hunt #2): messaging/agent_ops/deployments returned success while swallowing a state-bearing write/enqueue Result → now propagate the failure to the caller. Dual-VERIFIED → admin squash merge.

PR #1593, opened 2026-06-01
fix(lockaudit F1+F2): never hold registry across blocking PTY write; kill supervisor core→registry inversion

suzuke:
From the lock-while-blocking RCA (`/tmp/lockaudit-dev-rca.md`) — the #1492/#1535/#1571/#1583 freeze class. Two unpatched siblings the three prior fixes never covered. Proven #1530 collect→drop→emit pattern.

## F1 ⭐ (highest-likelihood next freeze)
The **registry lock was held across a blocking `inject_to_agent`/`write_to_agent`** — `write_with_timeout` waits up to `PTY_WRITE_TIMEOUT` (5s) on a channel `recv_timeout`, and `inject_to_agent` sleeps proportionally to payload size. A stuck/slow PTY writer (or a large inject) stalled **every** registry consumer (supervisor tick, all injects, API handlers, TUI render) for seconds. Same class as #1583, opposite direction (write/inject) + the registry lock.

**Fix** — snapshot-emit helpers + convert the 7 lock-held sites to *snapshot-under-lock → `drop` → emit*:
- `inject_with_target_gated(&InjectTarget, name, text, force)` — gate + inject on a pre-captured snapshot. The old `inject_to_agent(&AgentHandle)` **footgun is removed** so no caller can inject while still borrowing the registry.
- `write_to_pty(&PtyWriter, data)` — write on a captured writer.
- Sites: `supervisor.rs` (ServerRateLimit continue-inject), `daemon/mod.rs` (schedule replay), `cron_tick.rs` (schedule fire), `api/handlers/instance.rs` (inject/write handler), `crash_respawn.rs` (restart notice), `agent/mod.rs` (instructions bootstrap), `layout/pane.rs` (TUI keystroke write).

## F2 (lock-ordering inversion)
The supervisor per-agent tick **re-acquired the registry while holding an agent core** (core→registry) at the #1530 reaction-backend + #1552 escalation gates — inverting the registry→core order used by the render/monitor/API loops (AB-BA deadlock risk). **Fix:** pre-capture each agent's `backend_command` in the tick-top handles snapshot (registry→core) and resolve `Backend::from_command` lock-free — no nested registry re-acquire under a core.

## Tests (source-grep pins, mirroring #1146)
- `inject_handle_footgun_removed_uses_snapshots_f1` — the `&AgentHandle` inject footgun stays removed; the snapshot APIs exist.
[...]

Issue #1530, opened 2026-05-31
[bug] feed-driven UsageLimit reaction + member-state notify never fire (prev!=new gate; needs lock-safe de-gate)


suzuke:
## Summary

Two daemon reactions are gated on the supervisor's `prev != new` state comparison-at-tick, which (per the #1527 root-cause) **misses feed-driven transitions** (they complete async in the read-loop thread, so `prev == new` by the next supervisor tick). Result: these reactions effectively **never fire** for the transitions that matter.

Discovered while tracing #1527 (lead's cross-audit question E; reviewer-2 flagged it, confirmed by code trace).

## The two gated reactions (`src/daemon/supervisor.rs`, inside `if prev_state != new_state { … }`)

1. **`:493` UsageLimit reaction pipeline** (#1176): `notify_operator_usage_limit` + (opt-in) `propagate_usage_limit`.
2. **`:529` member-state-change notify** (`is_notify_error_class()` → `maybe_notify_member_state_change`).

## Why they're dead

`UsageLimit` (and the other error states) are detected from **PTY output via `feed()` → `StateTracker::transition()`**, on the read-loop thread — i.e. **async** relative to the supervisor tick. By the time the supervisor reads `prev = current; tick(); new = current`, `current` already settled to the new value and `tick()` didn't change it → `prev == new` → the whole block is skipped. So the UsageLimit reaction has effectively been **dead** since it was added (#1176): `UsageLimit` is never reached via a `tick()`-driven change, only via feed.

(#1527 fixes the *logging* half of this same root cause by recording transitions at the mutation source and draining them; it deliberately does **not** touch these reactions — see below.)

## Fix direction

De-gate the reactions off the drained transition list instead of `prev != new`: after `core.state.drain_pending_transitions()`, for each drained record, unconditionally check `to == UsageLimit` / `to.is_notify_error_class()` and run the reaction.

## ⚠️ Why this is NOT a trivial de-gate (the reason it was split out of #1527)

[...]

PR #1619, opened 2026-06-01
fix(#1617): recovery Stage-1 no longer deadlocks the daemon (registry lock across blocking PTY write)

### LINKED ISSUE #1617: [P0] auto-recovery Stage-1 deadlocks the daemon: raw blocking PTY write held under registry lock (same class as #1593 F1)
## TL;DR (found by overnight bug-hunt, adversarially verified)

The auto-recovery Stage-1 path can **deadlock the entire daemon** — it's the same lock-while-blocking class as #1593 (F1), but in the recovery dispatcher, which the lock-audit didn't cover because it's env-gated behind `AGEND_AUTO_RECOVERY_STAGE1=1`.

## Mechanism

`src/daemon/per_tick/recovery_dispatcher.rs`:
- `run()` holds the global registry lock across the whole loop (`lock_registry` @217 → `for handle in reg.values()` @229).
- Inside, `fire_stage1` does a **raw blocking PTY write**: `pty_writer.lock().write_all(ESC).flush()` (@668-671) — bypassing the timeout-safe `write_to_pty`.
- Stage-1 only fires on a **Hung** agent. A hung agent doesn't drain its PTY, so `write_all` blocks **forever** (a block returns no `Err`, so the existing Stage-2 escalation never triggers).
- → the supervisor tick thread is stuck holding the registry lock → **the whole daemon stalls** (every registry consumer: supervisor tick, all injects, API handlers, TUI render).

This is exactly the #1530 / #1593-F1 contract ("never hold a lock across a blocking PTY write") — and the recovery path, whose job is to UN-stick a hung agent, can itself stick the daemon.

## Fix

The proven #1530/#1593-F1 pattern: snapshot the `PtyWriter` (Arc-clone) under the registry lock → drop the lock → then do the blocking write via the timeout-safe `write_to_pty`. Restructure the recovery loop so no blocking PTY write happens while holding the registry lock. (Also fixes a P2: recovery-notify holding the lock across fs I/O — same collect-under-lock→drop→emit shape.)

Add an invariant test (mirroring #1593's) that the recovery loop does not hold the registry lock across the PTY write.

## Severity

P0 (daemon-wide stall) — but gated behind `AGEND_AUTO_RECOVERY_STAGE1=1`, so latent unless that flag is enabled. Fixing regardless since auto-recovery is meant to be a safety mechanism, not a deadlock source.

suzuke:
## Bug (P0, dev-2 bug-hunt #2)

The recovery dispatcher's `run()` held the **global registry lock across its entire per-agent loop** (`lock_registry_tracked` → `for handle in reg.values()`), and Stage-1's `fire_stage1_alive_stuck` did a **raw blocking PTY write** (`pty_writer.lock().write_all(ESC).flush()`) bypassing the timeout-safe `write_to_pty`.

A `Hung` agent that isn't draining its PTY → `write_all` **blocks forever** (a block returns no `Err`, so the existing Stage-2 escalation never fires) → the supervisor tick stalls **holding the registry lock** → the **whole daemon stalls**. This is exactly the **#1593 F1 / #1530 lock-while-blocking contract** — which the recovery path itself violated (the lock-audit missed it because it's env-gated behind `AGEND_AUTO_RECOVERY_STAGE1`).

## Fix (the proven #1593 F1 snapshot→drop→write pattern)

- `run()` snapshots each agent's `core` + `pty_writer` (both `Arc`) into a `RecoveryTarget` Vec **under** the registry lock, then **drops the lock** before the per-agent loop. The entire state machine — the Stage-1 PTY write **and** the Stage-2/3 `gated_notify` I/O — now runs with the registry lock released. The registry lock is **never** held across a blocking PTY write / notify. (Also closes the **P2**: recovery notify previously ran under the lock too.)
- `fire_stage1_alive_stuck` writes the ESC byte via the timeout-safe `agent::write_to_pty` (was a raw `write_all` that could block forever); a timeout surfaces as `Err`, escalating to Stage 2 exactly like a write error.
- ESC-dismiss behavior, the Stage-1 cooldown/transition semantics, and the shadow-mode path are **unchanged** — only the lock-holding window + the write timeout differ.

## Test

`recovery_loop_never_holds_registry_across_blocking_io` — a structural invariant (mirrors the #1593 F2 source-scan; scoped to the production slice before `#[cfg(test)]` to avoid self-match) pinning: (a) `fire_stage1` uses `write_to_pty`, not a raw `pty_writer.lock()`; (b) `run()` iterates the post-drop snapshot (`for target in`), never `reg.values()` under the lock. Existing `recovery_dispatcher` tests stay green.

## ⚠ Validation note (daemon-behavior fix)

The recovery dispatcher runs **in the daemon**, so this fix can only be validated at runtime **after a rebuild + daemon restart** — the live daemon won't exercise the new code until then (the PR-is-its-own-dogfood caveat). Correctness here is covered by the structural invariant + existing tests; the deadlock is a lock-discipline fix verified by the invariant.

## Verification

- `cargo fmt --check` ✅
- `cargo clippy --all-targets --features tray -- -D warnings` ✅
- `cargo test --features tray --bin agend-terminal` → 3075 passed, 0 failed ✅
- `cargo xwin check --target x86_64-pc-windows-msvc --features tray` ✅

Reviewer focus: **codex** — confirm the registry guard is dropped before any blocking PTY write / notify (the snapshot→drop ordering) and `write_to_pty` is timeout-safe. **reviewer-2** — Stage-1/2/3 decision semantics, cooldown, and ESC-dismiss behavior unchanged; the `RecoveryTarget` snapshot carries everything the state machine needs.

Closes #1617.

### DIFF
--- a/src/daemon/per_tick/recovery_dispatcher.rs
+++ b/src/daemon/per_tick/recovery_dispatcher.rs
@@ -65,8 +65,19 @@ use crate::health::{
     productive_silence_exceeds, HealthState, RecoveryStageState, STAGE1_COOLDOWN_DEFAULT_MS,
     STAGE1_TIMEOUT_DEFAULT_MS,
 };
-use std::io::Write;
+use std::sync::Arc;
 use std::time::{Duration, Instant};
+
+/// #1617 (#1593 F1): per-agent snapshot captured under the registry lock
+/// so the recovery state machine — including the Stage-1 blocking PTY
+/// write and the Stage-2/3 notify I/O — runs AFTER the lock is dropped.
+/// `core` + `pty_writer` are both `Arc`, so cloning them out is cheap and
+/// lets the registry guard be released before any blocking work.
+struct RecoveryTarget {
+    name: String,
+    core: Arc<crate::sync_audit::CoreMutex<agent::AgentCore>>,
+    pty_writer: agent::PtyWriter,
+}
 
 /// Env var name controlling Stage 1 activation. When set to `"1"`, the
 /// dispatcher writes the ESC byte to the agent's PTY; otherwise the
@@ -213,8 +224,6 @@ impl PerTickHandler for RecoveryDispatcherHandler {
     }
 
     fn run(&self, ctx: &TickContext<'_>) {
-        // #941: holder-tracking wrapper for thread-dump observability.
-        let reg = agent::lock_registry_tracked(ctx.registry, "recovery_dispatcher");
         let stage1_active = stage1_gate_active();
         let stage1_timeout = env_ms(STAGE1_TIMEOUT_ENV_VAR, STAGE1_TIMEOUT_DEFAULT_MS);
         let stage1_cooldown = env_ms(STAGE1_COOLDOWN_ENV_VAR, STAGE1_COOLDOWN_DEFAULT_MS);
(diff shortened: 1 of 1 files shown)


suzuke:
**§3.12 review mirror — merge gate** @ 82924518fb5a4581d9b338145b5c18d380a6062d

- **codex**: VERIFIED — snapshot→drop→write applied (registry lock dropped before all blocking ops); fire_stage1 timeout-safe write_to_pty (timeout→Stage2 escalation == original write-error path); cooldown/shadow-mode/gating intact; structural invariant test scoped to prod slice.
- **reviewer-2**: VERIFIED — recovery not weakened (hung agent still ESCs then escalates Stage2/3); snapshot completeness; no recovery-state regression.

CI all green. Closes #1617 — P0: auto-recovery Stage-1 was holding the registry lock across a raw blocking PTY write (same class as #1593 F1) → a hung agent would deadlock the whole daemon. Now snapshot→drop→write + timeout-safe. ⚠ Runtime effect requires rebuild+restart (daemon-behavior; correctness via structural invariant + recovery tests). Dual-VERIFIED → admin squash merge.

PR #1620, opened 2026-06-01
fix(bughunt2): reap handoff_timeout last_escalated entries (unbounded leak)

suzuke:
## What

`src/daemon/handoff_timeout_watchdog.rs` — `scan_and_emit`'s `last_escalated`
map (`HashMap<(String,String), DateTime>`, keyed by `(target, "repo@branch")`)
was **unbounded**: only `.get` (@51) and `.insert` (@86), no remove/retain/TTL.
`REALERT_AFTER_MINS=30` suppresses re-notify but never evicts → one entry per
timed-out `(reviewer, branch)` accumulates forever → slow memory leak on the
lifetime-long handler (`per_tick/handoff_timeout.rs:14`).

Found in autonomous bug-hunt #2 (P1).

## Fix (KISS, mirrors #1598)

Mirror #1598's `retry_tracks.retain`: collect the `(target, correlation)` keys
still backed by a **pending unread handoff this tick** into an `active` set,
then at the end of `scan_and_emit`:

```rust
last_escalated.retain(|k, _| active.contains(k));
```

This bounds the map to currently-active handoffs and reaps entries whose
handoff has been read/resolved (absent from the unread scan) — immediate,
exact eviction rather than a TTL lag.

## Test

`stale_entry_reaped_when_handoff_no_longer_pending`: old unread handoff →
first scan records a dedup entry; reviewer reads/resolves it → next scan reaps
the now-stale `(reviewer, repo@branch)` entry. Existing
`dedup_suppresses_reescalation_within_window` covers the inverse (active entry
retained to keep suppressing).

## Verification
- `cargo clippy --all-targets -- -D warnings` clean
- `cargo test --bin agend-terminal` — 3073 passed, 0 failed
- diff-stat vs fresh origin/main: 1 file, +38

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1622, opened 2026-06-01
fix(bughunt3): inbox recover_half_writes drops only the corrupt line, not the whole queue

suzuke:
## Bug (bughunt#3, P1 — fail-closed + silent-loss class)

`recover_half_writes` (`src/inbox/disk.rs`) moved an **entire** `<agent>.jsonl` to `inbox.recovery/` whenever **any** line failed to parse.

- `enqueue` (`storage.rs:77-85`) is a **plain in-place append** (`sync_all`, **not** tmp+rename), so a crash / `SIGKILL` (daemon restart) / OOM / power-loss mid-append leaves a **truncated trailing line** in a live inbox.
- `recover_half_writes` runs **unconditionally on every boot** (`daemon/mod.rs:653`). On the next boot it sees that one bad line and **nukes the whole inbox**.
- `inbox.recovery/` has **no reader anywhere** (grep-confirmed — forensic only) → **zero auto-restore** → silent loss of every queued task/query/report for that agent until an operator manually restores. For codex `skip_inject` agents the inbox is the **sole** delivery channel.

It was **strictly more destructive than a no-op**: every runtime read path (`drain:242`, `sweep_expired`, `unread_count`, `describe_message`, `get_thread`, `find_message`) already skips an unparseable line fail-open. Doing nothing → the next `drain` delivers all good messages and skips the corrupt line.

## Fix (fail-open, mirrors `drain`/`sweep` write-back)

Under the **per-file flock** (the same lock `enqueue` takes for this physical file, so a concurrent early-boot send can't race), rewrite the inbox keeping only parseable + empty lines via **tmp + atomic rename**, and append **only the dropped corrupt line(s)** to `inbox.recovery/` for forensics. Never drop a valid message.

- The `.tmp` branch is unchanged — a stale `.tmp` from an interrupted atomic rewrite is genuinely partial and the real `.jsonl` survived the (atomic) rename, so moving it aside stays correct.
- Forward-schema lines **parse** as `InboxMessage` (version > CURRENT) so they are **not** in `bad`; `drain` already drops them per-message. `recover` only touches truly-unparseable lines.

## Tests

- `test_half_written_jsonl_preserves_good_messages` — 2 good lines + trailing corrupt line → **both messages still drainable (zero loss)**, only the corrupt line salvaged to recovery.
- `test_recover_noop_when_all_lines_good` / `test_recover_noop_on_empty_file` — clean inbox and empty file are no-ops (no recovery dir).

Preflight green: `cargo fmt`, `cargo clippy --all-targets --features tray -D warnings`, `cargo test --bin agend-terminal inbox::tests::` (114 passed).

No `Closes` — neutral bughunt3 finding (like #1620/#1621).

### DIFF
--- a/src/inbox/disk.rs
+++ b/src/inbox/disk.rs
@@ -70,25 +70,65 @@ pub fn recover_half_writes(home: &Path) {
             continue;
         }
 
-        // Check JSONL files for corrupt trailing lines
+        // A corrupt line in a live JSONL inbox (e.g. a truncated trailing line
+        // from a crash mid-append — `enqueue` appends in place, not via
+        // tmp+rename) must NOT cost the agent its entire queue. Every read path
+        // (drain/sweep/unread_count/…) already skips an unparseable line, so the
+        // fail-open fix is to rewrite the file keeping only the good lines and
+        // preserve the dropped line(s) under inbox.recovery/ for forensics —
+        // never silently destroy a valid message. Done under the per-file flock
+        // (the same lock `enqueue` takes for this physical file) so a concurrent
+        // early-boot send cannot race the rewrite.
         if name_str.ends_with(".jsonl") {
-            if let Ok(content) = std::fs::read_to_string(&path) {
-                let lines: Vec<&str> = content.lines().collect();
-                let bad: Vec<&&str> = lines
-                    .iter()
-                    .filter(|l| {
-                        !l.trim().is_empty()
-                            && serde_json::from_str::<super::InboxMessage>(l).is_err()
-                    })
-                    .collect();
-                if !bad.is_empty() {
-                    // Move entire file to recovery, agent gets a fresh start
-                    ensure_recovery_dir(&recovery_dir);
-                    let dest = recovery_dir.join(&name);
-                    if std::fs::rename(&path, &dest).is_ok() {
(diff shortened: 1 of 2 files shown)


suzuke:

**codex**: VERIFIED @ 2e31c69 — lock identity confirmed (`path.with_extension("jsonl.lock")` == with_inbox_lock's lock for this .jsonl, no enqueue race); atomicity (same-dir tmp + sync_all + rename, crash-safe); zero good-message loss (parseable+empty kept, only unparseable salvaged); forward-schema lines preserved for drain; tests confirm drain-after-recover yields all good msgs. /tmp/1622-codex.md.

**reviewer-2** (finder): VERIFIED @ 2e31c69 — lock identity ✓ (same physical lock, no race); behavior complete ✓ (all-good/empty→noop, good+corrupt→keep+salvage 2/2 survive, all-corrupt→empty+salvaged); salvage ✓ (only unparseable to inbox.recovery, good never salvaged); test replaced to assert non-destructive. /tmp/1622-reviewer-2.md.

CI: all 6 required green. bug-hunt#3 P1 — recover_half_writes whole-file-nuke → silent message loss (#1614/#1618 sibling). Merging --admin --squash.

PR #1623, opened 2026-06-01
fix(bughunt3): never hold the registry lock across blocking IO in tui_bridge + poll_reminder (#1617 siblings)

suzuke:
## Summary

Two **P1 lock-while-blocking siblings of #1617 / #1593-F1**, found in the bug-hunt#3 concurrency sweep. Both held the **global agent registry lock** across a **blocking operation**, so a slow peer / slow FS would pin the registry and stall the whole daemon — the exact contract #1617 closed for the recovery path. Both fixes mirror the proven **snapshot → drop → IO-outside-lock** pattern.

### F1 — `tui_bridge.rs:110` (whole-daemon stall on a non-draining TUI client)
`serve_tui_accept_loop` captured the initial PTY dump **and** wrote it to the client socket (`write_frame`) while the registry guard was still held. `write_frame` is a *blocking* socket write — a non-draining/slow TUI client blocks it indefinitely (the hung-peer stall #1617 fixed for the PTY) → the global registry is pinned → supervisor / hang_detection / watchdog / monitor / dispatch all block.

**Fix:** capture `rx` + the owned `dump` Vec + the Arcs under the lock, **drop** the guard, then `write_frame` the dump lock-free. Intervening PTY output buffers in `rx` and is sent by the `tui_out` thread after the initial frame, so ordering is preserved.

### F2 — `poll_reminder.rs:35` (registry pinned across per-agent inbox file reads)
`collect_poll_reminders` held the registry lock across the entire `reg.values()` loop, calling `inbox::unread_count` (→ `fs::read_to_string` + full-file parse) per idle agent inside it. Inbox JSONL grows over daemon uptime; serializing N full-file reads under the global lock every pass stalls every other registry user (and the whole daemon on a slow FS).

**Fix:** phase-1 collect the idle-agent names under the lock and drop it; phase-2 run the inbox reads + dedup + formatting lock-free — exactly the discipline `conflict_notify.rs:88` already uses.

## Tests
Per-site structural source-scan invariants mirroring #1617's `recovery_loop_never_holds_registry_across_blocking_io`: brace-match the under-lock snapshot block and assert the blocking call (`write_frame` / `unread_count`) is **not** inside it and **is** present after the lock drops. Needles are concat-built and the scan is prod-sliced so the tests can't self-satisfy.

## Preflight
- `cargo fmt`
- `cargo clippy --all-targets --features tray -- -D warnings` ✅
- `cargo test --bin agend-terminal --features tray` → **3086 passed, 0 failed** ✅

No `Closes` (neutral bughunt3). F3 (P2, pr_state flock reorder) ships separately as PR-2.

### DIFF
--- a/src/daemon/poll_reminder.rs
+++ b/src/daemon/poll_reminder.rs
@@ -33,14 +33,22 @@ pub fn remove_agent(name: &str) {
 /// Pure collector: returns (agent_name, reminder_string) for each agent
 /// that should be nudged. No side effects — does not inject into PTY.
 pub fn collect_poll_reminders(home: &Path, registry: &AgentRegistry) -> Vec<(String, String)> {
+    // #1617-class (mirror conflict_notify's phase-1-collect / phase-2-IO): snapshot
+    // the idle-agent names UNDER the registry lock, then DROP the guard before any
+    // inbox file IO. `inbox::unread_count` does `fs::read_to_string` + full-file
+    // parse; holding the GLOBAL registry lock across that (per agent, in a loop)
+    // stalls every other registry user when the inbox is large or the FS is slow.
+    let idle_names: Vec<String> = {
+        let reg = agent::lock_registry(registry);
+        reg.values()
+            .filter(|handle| handle.core.lock().state.current == AgentState::Idle)
+            .map(|handle| handle.name.to_string())
+            .collect()
+    };
+
+    // Phase 2: the inbox reads + dedup + formatting run lock-free.
     let mut result = Vec::new();
-    let reg = agent::lock_registry(registry);
-    for handle in reg.values() {
-        let name = handle.name.as_str();
-        let agent_state = handle.core.lock().state.current;
-        if agent_state != AgentState::Idle {
-            continue;
-        }
+    for name in &idle_names {
         let (count, oldest) = crate::inbox::unread_count(home, name);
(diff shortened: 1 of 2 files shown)


suzuke:

**codex**: VERIFIED @ 139ab35 — F1 tui_bridge: dump captured as owned Vec in lock-scoped tuple, registry lock dropped before blocking write_frame, subscribe_with_dump atomic so post-snapshot output buffers in rx (no gap/loss). F2 poll_reminder: phase-1 collect idle names under lock → phase-2 unread_count IO lock-free; staleness acceptable for reminder. Structural invariants prod-sliced + brace-matched. /tmp/1623-codex.md.

**reviewer-2**: VERIFIED @ 139ab35 — F1 dump Vec owned survives registry drop, write_frame lock-free, no message loss (dump = exact subscribe-time state, rx carries later output). F2 mirrors conflict_notify.rs:88, staleness harmless. Invariant tests can't self-satisfy. /tmp/1623-reviewer-2.md.

CI: all 6 required green. bug-hunt#3 — two P1 lock-while-blocking siblings of #1617 (tui_bridge dump-under-registry-lock F1, poll_reminder inbox-read-under-registry-lock F2), both snapshot→drop→IO-outside-lock. Merging --admin --squash.

Issue #1069, opened 2026-05-22
Auto-dismiss codex update prompt — add to dismiss_patterns (productive marker class)


suzuke:
## Pain

When codex CLI auto-detects a version update, it shows a modal:

\`\`\`
✨ Update available! 0.132.0 -> 0.133.0

Release notes: https://github.com/openai/codex/releases/latest

› 1. Update now (runs \`bun install -g @openai/codex\`)
  2. Skip
  3. Skip until next version

Press enter to continue
\`\`\`

This BLOCKS the codex agent — it cannot respond to any kind=task / kind=query / auto-wake dispatch until operator manually selects an option in the agent's pane.

## Observed incident

2026-05-22 ~02:17 UTC: fixup-reviewer (codex backend) stuck at this prompt. Lead dispatched #994 Phase 3 to fixup-dev-2 but observed reviewer-side blocking when checking pane state. Manual operator intervention required to select Skip.

This is a productive-marker-class hung state — reviewer's PTY shows a clear modal but daemon's state classifier sees it as idle/blocked.

## Fix direction

Add to dismiss_patterns for codex backend (parallels existing handling):
- ClaudeCode #1 trust prompt at \`backend.rs:250\`
- ClaudeCode #2 \"Yes, proceed\" at \`backend.rs:286\`
- KiroCli \"No, exit\" keystroke

For codex:
- **Pattern**: detect text containing \"Update available!\" + \"Skip\" option visible
- **Action**: inject keystroke \`2\\r\` (Skip option, doesn't disable future prompts) OR \`3\\r\` (Skip until next version, more aggressive)
- **Recommended**: option 2 (Skip) — least invasive, lets operator decide upgrade timing manually

## Spike scope

3-way dialectic:

1. **Whether to fix**: confirm prompt blocks dispatch reliably. Reviewer instance state、daemon classifier behavior, confirmed-fail rate.
2. **Direction**: dismiss_patterns vs separate codex-update-handler. Trade-off: dismiss_patterns simpler but tighter coupling to known-pattern strings; separate handler more extensible but more code.
3. **Approach**: detection pattern specifics (\"Update available!\" + version regex? full prompt text?), keystroke selection (option 2 vs 3), test sketch.
4. **Elegance**: parallels existing dismiss_pattern infrastructure cleanly?
5. **RED test**: replay raw fixture of codex update prompt → assert auto-dismiss fires + correct keystroke injected.

## Out of scope

[...]

PR #1627, opened 2026-06-02
fix(#1626): disable codex startup update modal at spawn layer

### LINKED ISSUE #1626: codex backend: suppress/auto-skip the CLI update prompt at spawn (must NOT modify user global ~/.codex config)
## Symptom

When the daemon (re)spawns a `codex` backend instance and a codex CLI update is available, codex shows an interactive startup prompt:

```
✨ Update available! 0.135.0 -> 0.136.0
› 1. Update now (runs `bun install -g @openai/codex`)
  2. Skip
  3. Skip until next version
  Press enter to continue
```

This blocks the codex instance at startup until an option is selected. Observed after a daemon restart: a stray `2` (Skip) landed in the prompt, and the codex agent (already running) misread the leftover `2` as a chat command and generated a confused clarification menu. Net effect: codex instances can hang at / be disrupted by this update prompt on (re)spawn.

## Desired behavior

The daemon should make codex **default to Skip** the update prompt on spawn so codex instances come up clean without manual interaction.

## ⛔ HARD CONSTRAINT (operator directive 2026-06-02)

The fix MUST be at **agend-terminal's codex spawn layer** — do NOT modify the user's global `~/.codex/config.toml` (or any global tool config). agend-terminal must not mutate user-global state; that would affect ALL of the user's codex usage, not just the fleet. The fix lives entirely within the daemon's own spawn code / per-process env / workspace.

## Candidate approaches (spike to pick)

1. Pass an env var (or launch flag) to the spawned codex child process that disables the update check / prompt, IF codex supports one. Scope = that child process only.
2. Have the codex-backend spawn sequence detect the update prompt and auto-send the Skip choice (`2` or `3 = Skip until next version`) as part of startup handling, before handing the pane to the agent loop.
3. Other spawn-layer mechanism that suppresses it per-process.

Spike should determine which codex actually supports without touching global config, and whether option 3 (Skip until next version) is preferable to 2 (re-prompts on next minor).

## Severity

P2 — operational (codex instances disrupted/blocked on spawn when an update is available); recurs every time codex publishes a release.

suzuke:
## Problem
When the daemon spawns a codex backend and a codex update is available, codex renders a **blocking** startup modal:

```
✨ Update available! 0.135.0 → 0.136.0
  1. Update   2. Skip   3. Skip until next version
```

It disrupts codex startup. Worse, our own #1069 auto-dismiss (`DismissPattern "Update available!" → b"2\r"`) **races the live TUI** — when the modal isn't actually focused (timing skew on restart, or the 0.135 3-option layout), the stray `2` leaks into the running codex agent's composer (the "stray '2' got misread" symptom).

## RCA
The modal fires when `check_for_update_on_startup` (top-level codex config bool, default `true`) is set **and** `latest_version > dismissed_version` (`~/.codex/version.json`: latest `0.136.0`, dismissed `0.131.0`). The stray `2` is **our own dismiss**, not a codex bug — the dismiss approach is inherently race-prone and TUI-layout-coupled.

## Fix (spawn layer, zero global writes)
Prepend **`-c check_for_update_on_startup=false`** to the Codex preset `args` (before the `resume` subcommand — `-c` is a global clap option) and `fresh_args` (`src/backend.rs`). `-c key=value` is codex's per-invocation config override: it applies on the child argv and **does NOT write `~/.codex/config.toml`**, so the suppression stays fleet-scoped and never touches the operator's global codex config (operator directive: no global writes). With the modal suppressed entirely, the dismiss never fires → the stray-`2` race cannot occur.

### Evidence codex honors the flag (read-back via `doctor`, no writes — throwaway `CODEX_HOME`)
| invocation | `doctor` → "startup update check" |
|---|---|
| `codex -c check_for_update_on_startup=false doctor` | **false** |
| `codex doctor` (control) | **true** |

argv-order also confirmed (clap accepts `-c` before the subcommand): `codex -c … resume --help` and fresh-mode both parse (exit 0).

## ⚠ Caveat (why we KEEP the #1069 dismiss as a fallback)
codex **silently no-ops unknown `-c` keys** (a bogus key is ignored by `doctor`; `--strict-config` only validates the config *file*, not `-c` overrides). So if codex upstream ever **renames** `check_for_update_on_startup`, the flag **dormant-fails** and the modal returns. The #1069 `b"2\r"` dismiss is therefore retained as belt-and-suspenders: it never fires while the flag works (no stray-`2` in the happy path), and degrades the upstream-rename failure from a *blocking hang* to a *racy-but-non-blocking* auto-skip.

## Tests
- New `codex_disables_startup_update_check_before_resume`: pins the override is present in **both** spawn modes and **precedes** the `resume` subcommand.
- Existing #1069 codex dismiss tests retained (fallback kept).
- Preflight: `cargo fmt`, `cargo clippy --all-targets --features tray -- -D warnings` clean, 34 backend tests pass.

Closes #1626

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -388,7 +388,16 @@ impl Backend {
             },
             Backend::Codex => BackendPreset {
                 command: "codex",
+                // #1626: `-c check_for_update_on_startup=false` disables codex's
+                // blocking startup update modal ("Update available!"). This is a
+                // per-invocation config override on the child argv — it does NOT
+                // write `~/.codex/config.toml`, so it stays fleet-scoped and never
+                // touches the operator's global codex config. Must precede the
+                // `resume` subcommand (`-c` is a global option). Primary fix; the
+                // #1069 dismiss below stays as a fallback (see its comment).
                 args: &[
+                    "-c",
+                    "check_for_update_on_startup=false",
                     "resume",
                     "--last",
                     "--dangerously-bypass-approvals-and-sandbox",
@@ -420,13 +429,26 @@ impl Backend {
                     },
                     // #1069: version-update modal blocks agent until operator
                     // selects an option. "2\r" = "Skip" (least invasive).
+                    // #1626 FALLBACK: the `-c check_for_update_on_startup=false`
+                    // flag above normally suppresses this modal entirely, so this
+                    // dismiss never fires in the happy path. Kept as belt-and-
+                    // suspenders: codex silently no-ops unknown `-c` keys, so if
+                    // upstream ever renames the key the flag dormant-fails and this
+                    // dismiss degrades the failure from "blocking hang" to a racy
+                    // (but non-blocking) auto-skip.
                     DismissPattern {
                         label: r"(?m)^[^A-Za-z\n]*Update available!",
                         sequence: b"2\r",
                     },
                 ],
-                // Codex: "resume --last" → fresh start drops the resume subcommand
-                fresh_args: Some(&["--dangerously-bypass-approvals-and-sandbox"]),
+                // Codex: "resume --last" → fresh start drops the resume subcommand.
+                // #1626: keep the `-c check_for_update_on_startup=false` override
+                // in fresh mode too (no subcommand here, so order is unconstrained).
+                fresh_args: Some(&[
+                    "-c",
+                    "check_for_update_on_startup=false",
+                    "--dangerously-bypass-approvals-and-sandbox",
+                ]),
                 fleet_mcp_supported: true,
             },
             Backend::OpenCode => BackendPreset {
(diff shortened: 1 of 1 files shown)


suzuke:

**codex**: VERIFIED @ 9ee04c3 — `-c check_for_update_on_startup=false` correctly before `resume` in resume args + in fresh args; key spelling correct; #1069 fallback dismiss intact; spawn-path audit (spawn_agent→build_command→preset_spawn_args) found no runtime bypass; test strict for presence+ordering across both modes. /tmp/1627-codex.md.

**reviewer-2**: VERIFIED @ 9ee04c3 — no-global-write ✓ (`-c` per-invocation argv, no ~/.codex mutation); argv order ✓ (before resume subcommand, both modes); single codex preset = sole launch source; #1069 fallback kept; unit test asserts presence+value+ordering. /tmp/1627-reviewer-2.md.

CI: all 6 required green. Fixes #1626 — codex CLI update modal disrupting spawn (the stray '2' was our own #1069 auto-dismiss racing the TUI). Per-invocation `-c` flag, ZERO global config mutation (operator hard constraint). Merging --admin --squash.

PR #1628, opened 2026-06-02
fix(#1625): restart_instance returns the new pane to its original tab

### LINKED ISSUE #1625: restart_instance mode=fresh displaces the instance's pane out of its tab
## Symptom

After `restart_instance(instance, mode=fresh)`, the instance's pane is no longer in its original tab — it ends up **outside the tab layout**. The instance itself is alive and healthy; only the TUI pane placement is lost.

## Evidence / repro

- `fixup-dev` and `fixup-dev-2` were `fresh`-restarted twice on 2026-06-02 (~02:48 for a bug-hunt round, ~07:39 for a refactor scout) via `restart_instance mode=fresh`.
- Both panes were observed **outside the `fixup` tab BEFORE any daemon restart** — so the displacement is caused by the fresh restart, not by daemon-restart layout re-rendering.
- A subsequent daemon restart merely made the already-displaced panes visible.

Repro:
1. Have a managed instance whose pane lives inside a named tab.
2. `restart_instance(<instance>, mode=fresh)`.
3. The respawned pane lands outside the tab (no tab inheritance).

## Expected

`mode=fresh` should preserve the pane's tab placement — same as `mode=resume` and `replace_instance`, which keep the pane in its tab. (`move_pane` preserves placement explicitly; fresh restart should too.)

## Likely cause

`mode=fresh` spawns a brand-new instance/pane (clean start, like `replace_instance`) but does not re-attach the new pane to the original instance's tab assignment — the tab membership isn't carried over from the killed instance to the respawned one.

## Workaround

`move_pane(instance, target_tab)` to re-attach (preserves scrollback/PTY). Used to recover fixup-dev/fixup-dev-2.

## Severity

P2 — cosmetic/operational (no work lost, instances stay healthy), but confusing for the operator and easy to hit when using fresh restarts for clean-context scans.

suzuke:
## Problem
`restart_instance` respawned the killed pane **without** the `layout: same-tab` hint that `replace_instance` carries (#1431). The SPAWN therefore defaulted to `LayoutHint::Tab` and the new pane opened in a **fresh tab**, dropping the agent out of its team tab.

Repro: `fixup-dev` / `fixup-dev-2` were fresh-restarted and fell out of the shared `fixup` tab.

## Root cause
- `handle_replace_instance` (`instance.rs:328`) sends `"layout": "same-tab"` → TUI consumes the per-agent `removed_tab_memory` (recorded on the pane's DELETE) and returns the pane to its original tab.
- `handle_restart_instance` omitted that param → `LayoutHint::parse(... .unwrap_or("tab"))` → `Tab` → `add_tab` (new tab).

`mode` only toggles backend resume args (`resume_mode.args_for()`), **not** placement — so this affected **both** resume and fresh restarts, not just fresh.

## Fix (surgical)
Extract a pure `restart_spawn_params` helper that unconditionally tags `layout: same-tab`, mirroring the proven #1431 path. No new wiring — the removed-tab memory is already recorded on the preceding DELETE for every removal path.

## Tests
- **Unit** (`mcp/handlers/instance.rs`): `restart_spawn_params` carries `layout=same-tab` for both fresh and resume (and resume still sets `mode=resume`).
- **Integration** (`app/tui_events.rs`): a fresh restart of an agent sharing a multi-pane team tab returns its pane to that tab (mirrors the #1431 SameTab test).

## Preflight
- `cargo fmt --check` clean
- `cargo clippy --all-targets --features tray -D warnings` clean
- `cargo test --tests --features tray` green (61 test-result groups ok; the one `agend-git` conflict test that fails locally is an artifact of the fleet `git` shim — passes with real git).

## Note on the original framing
The dispatch hypothesized "resume preserves placement". Per the code, resume-restart would **also** lose the tab — the apparent preservation is a different path (daemon-restart team auto-grouping rebuilds tabs; layout isn't persisted). The fix treats both modes regardless.

Closes #1625

### DIFF
--- a/src/app/tui_events.rs
+++ b/src/app/tui_events.rs
@@ -833,6 +833,39 @@ mod tests {
         assert!(layout.tabs[0].root().has_agent("reviewer"));
     }
 
+    /// #1625: fresh restart of an agent sharing a multi-pane team tab returns
+    /// the respawned pane to that same tab (the repro: fixup-dev was
+    /// fresh-restarted and fell out of the shared "fixup" tab). restart_instance
+    /// now emits LayoutHint::SameTab, which drives the same delete-records-tab →
+    /// SameTab-replacement path as #1431.
+    #[test]
+    fn restart_returns_pane_to_shared_team_tab() {
+        use crate::layout::SplitDir;
+        let mut layout = Layout::new();
+        let mut tab = Tab::new("fixup".into(), leaf(1, "fixup-dev"));
+        tab.split_focused(SplitDir::Horizontal, leaf(2, "fixup-reviewer"));
+        layout.add_tab(tab);
+
+        // restart kills the old pane (DELETE → InstanceDeleted) ...
+        handle_instance_deleted("fixup-dev", &mut layout);
+        assert_eq!(
+            layout.tabs.len(),
+            1,
+            "shared tab survives one pane's delete"
+        );
+        let tab_name = layout
+            .take_removed_tab("fixup-dev")
+            .expect("delete records the agent's tab");
+        assert_eq!(tab_name, "fixup");
+
+        // ... then the SameTab respawn splits back into the shared tab.
+        place_in_remembered_tab(&mut layout, &tab_name, leaf(3, "fixup-dev"));
+        assert_eq!(layout.tabs.len(), 1, "no new tab — pane stays in 'fixup'");
+        assert_eq!(layout.tabs[0].name, "fixup");
+        assert!(layout.tabs[0].root().has_agent("fixup-dev"));
+        assert!(layout.tabs[0].root().has_agent("fixup-reviewer"));
+    }
+
     /// #1431: SameTab never produces a split anchor — placement is resolved via
     /// the remembered-tab map, not resolve_split_anchor (even when a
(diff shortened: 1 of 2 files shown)


suzuke:

**codex**: VERIFIED @ f1cb0ae — restart_spawn_params unconditionally sets layout=same-tab → LayoutHint::SameTab → TUI remembered-tab path; fresh=no mode, resume=mode=resume, both pinned by unit tests; DELETE-before-SPAWN ordering matches replace_instance for removed_tab memory; integration test reproduces shared-team-tab scenario (no new tab); diff doesn't touch agend-git shim. /tmp/1628-codex.md.

**reviewer-2**: VERIFIED @ f1cb0ae — same-tab matches LayoutHint::SameTab parser (mirrors replace_instance); both modes pinned; event ordering sequential same as replace_instance; integration test (multi-pane shared tab → restart → returns to same tab); diff touches only handler + test. /tmp/1628-reviewer-2.md.

CI: all 6 required green. Fixes #1625 — restart_instance respawned pane dropped out of its tab (handle_restart_instance omitted layout=same-tab that replace_instance has per #1431). Affects both fresh+resume. Merging --admin --squash.

PR #1631, opened 2026-06-02
fix(#1630): kill the silent-enqueue-drop class (propagate/log + invariant)

### LINKED ISSUE #1630: Refactor R2: silent-drop lint (let_underscore_must_use) + persist_or_log! macro — make the silent-loss class compile-time-blocked
## Problem (class-killer refactor — from refactor-scout 2026-06-02, /tmp/refactor-scout-dev.md R2)

The silent-loss bug class (#1618, and the still-live siblings R7 inbox/notify.rs:163 + R8 daemon/mod.rs:966 that swallow enqueue Results) recurs because nothing guards against `let _ = <Result>`. Worse: `unwrap_used="deny"` (Cargo.toml:47-48) actively pushes devs toward `let _ =` as the 'safe' way to discharge a Result — straight into silent drop.

## Fix

1. Enable `let_underscore_must_use = "warn"` (clippy) so dropping a #[must_use]/Result via `let _` is flagged.
2. Add a `persist_or_log!` macro (or similar) as the sanctioned way to discharge a best-effort Result (logs on Err instead of silently dropping).
3. Triage every site the lint flags: real silent-loss → fix with persist_or_log! / proper handling (covers R7/R8); legitimate fire-and-forget → explicit `#[allow]` + a one-line reason.

→ The whole silent-loss class moves from runtime-P1 to compile-time-blocked.

## ⚠ Scope caveat — SPIKE FIRST

Turning on the lint under CI `-D warnings` means EVERY existing violation must be fixed or allow-listed or CI breaks. Spike must first run the lint, count + triage the flagged sites (real bug vs legit), and propose the fix-list + allow-list before impl. Don't blindly enable and drown.

## Source

refactor-scout dev R2 + R7/R8 live siblings (operator-triaged ② class-killer).

suzuke:
## Problem
The silent-message-loss class (#1614/#1618/#1622 lineage) kept recurring via one reflexive shape: `let _ = …enqueue…(…)`. A dropped inbox/notification `enqueue` `Result` = the message never lands on disk and the recipient never sees it.

## Why not the global lint
Spike (clippy `let_underscore_must_use`) flagged **586 sites, >90% legitimate fire-and-forget** (`child.wait`/`reaper.join`/best-effort fs/PTY writes) — drowning, and would train reflexive `#[allow]`. So this targets the **narrow real class: 23 enqueue/notification Result-drops**, and uses a source-scan invariant (the codebase's established enforcement pattern) instead of the lint.

## Propagate vs log (per operator refinement: "a log nobody reads isn't prevention")
| Disposition | Count | Rule |
|---|---|---|
| **PROPAGATE** | 1 | enclosing fn returns `Result` & caller handles it |
| **LOG** (`persist_or_log!`, ERROR + context) | 20 | per-tick / fan-out / event-handler fns returning `()`/`bool`/`Value` — no caller to propagate to |
| **excluded** (test code) | 2 | `#[cfg(test)]` — `poll_reminder::seed_unread`, a `schedules` test |

- **PROPAGATE** — `agent::inject_with_target_gated` (`src/agent/mod.rs`): the enqueue *is* the deferred-delivery path; the branch did `let _ = …; return Ok(())`. Now returns the enqueue result (mapped `anyhow→AgendError::ApiError`, message preserved) so callers surface it — e.g. `daemon::replay_missed_at_startup` logs `"replay inject failed"`. (`AgendError` has no generic variant; the `ApiError` mapping is a cosmetic wart, message intact via `Display`.)
- **LOG** — new `persist_or_log!(call, op[, target])` macro emits `tracing::error!` with `op` + recipient `target` so a dropped message is **post-hoc diagnosable**. It's a diagnosis breadcrumb, not loss-prevention. No existing dropped-message counter to tie into → no new observability infra (KISS).

20 LOG sites: telegram/inbound (status-summary, dispatch), ci_watch/poller (conflict-alert, early-fail, watch-notify, chain), ci_watch/sweep (health-event), cron_tick (schedule-fire), daemon/mod (dispatch-stuck-check, schedule-replay), pr_state/scanner (pr-merged, pr-closed-unmerged), supervisor (member-state-change, retry-exhausted), inbox/notify (deliver, compose-aware-inject), mcp/handlers/instance (replace-handover), tasks/handler (cascade-cancel), tasks/orphan (boot-orphan-sweep), verify (inbox self-test).

## Prevention — `tests/enqueue_drop_invariant.rs`
The macro only fixes today's 23; the invariant stops recurrence. Source-scans `src/` for **all three reflexive drop shapes** — `let _ = <fn>(`, `<fn>(…).ok()`, bare-`;` — across `enqueue` / `enqueue_with_idle_hint` / `enqueue_classified`. Allow-lists `persist_or_log!`. **Prod-sliced**: skips test-module files + `#[cfg(test)]` regions (brace-depth tracked) so it can't false-fail on test helpers. Tail-expression returns (`enqueue(…)` ending in `)`, no `;`) are correctly excluded as propagation. Mirrors `block_on_runtime_guard_invariant`. Verified to catch all three shapes (negative-tested) and pass clean. **Honest limit:** it can't stop a determined assign-then-drop evader — by design; the goal is to kill the reflexive `let _ =` that produced all 23.

## Deferred (per triage)
~25 `atomic_write`/`mutate_versioned`/`fs::write` persist drops — mixed / intentionally-lossy (`session.json` #1479, `.daemon` PID, `event_log` rotation). Separate triage. **No** global `let_underscore_must_use`.

## Preflight
`cargo fmt`, `cargo clippy --all-targets --features tray -- -D warnings` clean, `enqueue_drop_invariant` + notify/notification_queue/cron_tick/scanner unit tests pass.

Closes #1630


suzuke:
## §3.12 Review Mirror — dual VERIFIED (re-reviewed after codex reject + rebase)

**codex**: REJECTED at c63b768 (line-oriented scanner false-passed a multi-line .ok() drop — empirically proven via probe) then VERIFIED at 4c43e8a after dev-2 rewrote the scanner statement/span-aware (codex re-ran the exact probe: now caught; positive probes still pass; string-literal-aware confirmed). /tmp/1631-codex.md.

**reviewer-2**: VERIFIED at 4c43e8a — multi-line .ok() caught (span-aware lead/trail classification), no false-fails (persist_or_log! / ? / let x= / return / tail-expr allowed), string-literal aware, prod-slice retained. /tmp/1631-reviewer-2.md.

Rebased ec0c8ac→4c43e8a onto main (PR1 #1632 + PR2 #1633) → head 61403ae. Lead premise-checked the rebase delta: ONLY the 2 pr_state/scanner.rs sites were removed (now handled by the merged #1632 emit-deferral's handled if-let-Err warn); everything else byte-identical to the dual-VERIFIED 4c43e8a. enqueue_drop_invariant still passes on the rebased tree.

CI: all 6 green at 61403ae. Fixes #1630 — silent enqueue-drop class-killer: persist_or_log! macro (1 propagate + 20 ERROR-logged sites) + tests/enqueue_drop_invariant.rs source-scan that fails CI on a new dropped enqueue (let _ = / .ok() / bare-semicolon). Merging --admin --squash.

PR #1633, opened 2026-06-02
fix(#1629 PR2): narrow deployments flock to the store load-modify-save (#1617 class)

suzuke:
## PR2 of 3 — flock-deadlock-guard series (#1629)

`deploy()` and `teardown()` held the deployment-store flock (`store::acquire_file_lock`) across loopback `api::call` self-IPC — the same lock-while-blocking class as #1617 (the loopback handler needs the registry lock, so a self-IPC under this flock can stall any thread contending it).

### Premise-check (git-verified @ ec0c8ac)
- `deploy:373` flock spans `spawn_instances` (api::call **SPAWN** :312) + `create_deployment_team` (api::call **CREATE_TEAM** :360).
- `teardown:439` flock spans the per-instance `api::call` **DELETE** loop (:456).
- `validate_deploy_args` reads only fleet.yaml, **not** the store → the flock's sole job is the C1 lost-update protection on the load-modify-save.

### Fix (narrow the flock to just the store load-modify-save)
- **deploy:** drop the function-wide lock; acquire only around the final `load → push → save`. validate / SPAWN / CREATE_TEAM now run lock-free.
- **teardown:** read the record lock-free (`load` reads atomically-written files), run DELETE + dir/yaml/team cleanup lock-free, then lock only the final `load → retain → save`. **Order preserved** (delete instances → then remove the store record), so the #bughunt2 crash failure-semantics are unchanged.

### 3rd flock site audited & cleared
`reconcile_orphan_deployments` (:603) holds the flock across `teams::delete` (a **direct** store op, not a loopback) + `cleanup_deployment_dirs` — no `api::call`/enqueue, so it never trips the guard. Left as-is (not a 6th site).

### Tests
Two structural source-scans slice each fn and assert the api::call site **precedes** the `acquire_file_lock` site — prod-sliced + `concat` needles so they can't self-satisfy:
- `deploy_api_calls_not_under_flock`
- `teardown_api_calls_not_under_flock`

### Preflight
- `cargo fmt --check` clean
- `cargo clippy --all-targets --features tray -D warnings` clean
- `cargo test --tests --features tray`: **3380 passed, 0 failed** (real git). 48 deployments behavioral tests green.

PR3 (the `FLOCK_DEPTH` guard) lands once PR1 (#1632) + this are both in main, so the always-on fail-fast guard never red-fails CI on an unfixed site.

Refs #1629

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1632, opened 2026-06-02
fix(#1629 PR1): defer 3 daemon-loop inbox emits out of the flock (#1617 class)

suzuke:
## PR1 of 3 — flock-deadlock-guard series (#1629)

Phase-1 scan found **5** production sites holding an `fs4` flock (via `store::acquire_file_lock`) across a self-IPC (`enqueue_with_idle_hint`/`notify_system` → loopback `api::call`) — the same lock-while-blocking class as #1617/#1342/#1624. This PR fixes the **3 daemon-loop sites**; PR2 narrows the 2 deployments sites; PR3 adds the `FLOCK_DEPTH` guard (lands clean once all 5 are fixed).

### Why it's a deadlock
The loopback API handler needs the registry lock and may lock the target agent's core. A self-IPC issued while a flock is held can stall any thread contending that flock — exactly #1617.

### Fixes (collect-under-flock → drop → emit; mirrors the #1624 scanner `auto_release` deferral)
| Site | flock | self-IPC moved out |
|---|---|---|
| `pr_state/scanner.rs:81` | `with_pr_state` closure | 3× `enqueue_with_idle_hint` (pr-ready / pr-merged / pr-closed-unmerged) → `pending_emits` vec, drained after closure (before `auto_release`, preserving emit→release order) |
| `dispatch_idle/mod.rs:521` | #1340 dispatch flock | `emit_exceeded_event` → after `let to_emit = {…}` block |
| `decision_timeout.rs:260` | #1116 decision flock | `emit_timeout_event` → after the flock block |

### Behavior preservation
- Both `emit_*_event` fns read **no mutated field** (status is not in the message), so flipping status before the emit is happy-path neutral. Verified by reading both fns.
- **scanner pr-ready semantic note (for review):** `ready_emitted_for_sha` is now set **optimistically** under the flock (was: only on enqueue success). On a rare post-unlock enqueue failure the #911 dedup ledger + next scan backstop. This matches the Merged/ClosedUnmerged arms, which already set the flag unconditionally and ignored the enqueue result — and matches the #1624 deferral philosophy.

### Tests
One structural source-scan invariant per site (mirrors the existing `auto_release_not_called_under_pr_state_flock` #1624 test): brace-match the flock scope, assert the emit/enqueue is **NOT** inside and **IS** after. Prod-sliced + `concat`-built needles so they can't self-satisfy.
- `deferred_emit_not_called_under_pr_state_flock`
- `emit_exceeded_not_called_under_flock`
[...]

Issue #1535, opened 2026-05-31
[hardening] #1492 deadlock guard blind spot: assert_no_registry_lock_for_self_ipc only tracks registry-lock depth, misses per-agent core-lock-held self-IPC


suzuke:
## Summary

The #1492 lock-across-self-IPC deadlock guard (`crate::sync_audit::assert_no_registry_lock_for_self_ipc`, fired from `api::call` at `src/api/mod.rs:673`) only tracks the **registry** lock depth (`REGISTRY_LOCK_DEPTH`). It does **not** track the per-agent `AgentCore` mutex. So a self-IPC performed while holding an agent's `core` lock — but **not** the registry lock — passes the assert silently, yet can still deadlock.

Discovered during the #1530 spike (de-gating the feed-driven UsageLimit / member-state reactions).

## Concrete blind-spot path

`src/daemon/supervisor.rs` per-agent tick:
1. The registry is locked only briefly (`:439-443`) to clone out `Arc<Mutex<AgentCore>>` handles, then **released**.
2. The loop body locks the **per-agent core** (`let mut core = core.lock()`, `:451`) — NOT the registry.
3. `maybe_notify_member_state_change` (`:325`) self-IPCs: `inbox::notify_agent(orch)` → `enqueue_with_idle_hint` → `api::call(INJECT)` over the loopback socket.
4. The INJECT handler `handle_inject` (`src/api/handlers/instance.rs:8`) locks the **registry** (`:15`) **and** the target agent's **core** (`:19`).

At step 3 the supervisor holds the error-agent's `core` but **not** the registry, so `assert_no_registry_lock_for_self_ipc` (registry-depth scoped) is a no-op → **no panic**. The deadlock risk is real regardless:
- **Re-entrancy** if the self-IPC target's core == the held core (parking_lot `Mutex` is non-reentrant).
- **AB-BA ordering inversion**: T1 (supervisor) holds `error.core` → via self-IPC wants `registry` + `orch.core`; T2 (an API handler) holds `registry` → wants `error.core`.

So the guard that exists specifically to catch lock-across-self-IPC deadlocks has a blind spot for exactly the core-held variant — a **meta-gap: the deadlock-detection tool itself doesn't cover this class.**

(The #1530 fix sidesteps it by collecting reaction intents under the core lock → `drop(core)` → emitting all self-IPC lock-free. This issue is about hardening the *guard* so future core-held self-IPC is caught, not just this one site.)

## Proposed hardening

[...]

PR #1635, opened 2026-06-02
refactor(#1629 PR3): extend the self-IPC deadlock guard to the fs4 flock tier (capstone)

### LINKED ISSUE #1629: Refactor R1: extend self-IPC deadlock guard to cover the fs4 flock tier (make flock-while-blocking a unit-test-catchable class)
## Problem (class-killer refactor — from refactor-scout 2026-06-02, /tmp/refactor-scout-dev.md R1)

The self-IPC deadlock guard (sync_audit.rs:230) only tracks REGISTRY / CORE_LOCK_DEPTH. It is **blind to the fs4 flock tier** (pr_state / binding / dispatch_idle RMW flocks don't bump any depth counter). Consequence: every flock-while-blocking bug — #1617, #1342, #1340, and tonight's #1624 (pr_state flock held across git subprocess) — is catchable **only by manual review**, never by a test.

## Fix (turn a comment-guarded invariant into a compiler/test-guarded one)

Add a `FLOCK_DEPTH` thread-local mirroring CoreMutex's depth tracking. `store.rs:153` is the single chokepoint where flocks are acquired — one wrapper that bumps/decrements FLOCK_DEPTH + one `||` clause in the deadlock guard so it fires when a blocking self-IPC op runs while a flock is held. Add an invariant test (mirror the #1593/#1617 structural invariant tests) so the whole flock-while-blocking sub-class becomes unit-test-catchable instead of always-manual.

## Risk

LOW-MED — additive instrumentation on a single chokepoint; must not false-positive on legitimate flock+non-blocking paths. Spike to confirm the chokepoint is truly singular and the guard clause is correctly scoped.

## Scope

Refactor / defense-in-depth. Source: refactor-scout dev R1 (operator-triaged ① class-killer). No behavior change to the happy path — adds detection.

suzuke:
## PR3 of 3 — capstone (Closes #1629)

The always-on self-IPC guard (`assert_no_registry_lock_for_self_ipc`) tracked the registry lock (#1492) and per-agent core lock (#1535) but was **blind to the third blocking tier: fs4 advisory flocks.** Holding a flock across a loopback `api::call`/`enqueue_with_idle_hint` is the #1617 lock-while-blocking deadlock class — yet bumped neither counter, so every flock-while-blocking bug (#1617/#1342/#1340/#1624) had to be caught by **manual review**. This makes the class **unit-test-catchable forever**.

### Change (mirrors the CoreMutex/#1535 pattern)
- `FLOCK_DEPTH` thread-local + `flock_entered`/`flock_exited` (sync_audit.rs).
- `store::acquire_file_lock` now returns a **`FileFlockGuard`** newtype (was raw `File`) that bumps `FLOCK_DEPTH` for its lifetime — increment **after** the OS lock is held, decrement on drop. The ~37 RAII `let _lock =` call sites are unchanged; the one explicit `Result<File>` return (`fleet::persist::acquire_lock`) updated.
- `assert_no_registry_lock_for_self_ipc` gains the `flock > 0` clause (same always-on fail-fast `Err`).
- The **2 daemon-singleton `.daemon.lock`** raw `fs4::try_lock` sites (`bootstrap::acquire_daemon_lock`, `daemon::run`) deliberately bypass `acquire_file_lock` and do **NOT** bump — they hold for the daemon's whole life and would otherwise pin the depth > 0 and false-trip every self-IPC (per the phase-1 audit).

### Tests
- **Behavioral** (store.rs): holding an `acquire_file_lock` flock makes the guard return `Err`; dropping it clears to `Ok` — the wiring proof.
- **Structural** (`tests/flock_depth_invariant.rs`, mirrors `tests/core_mutex_invariant.rs`): RED if a raw `fs4::FileExt::{lock,try_lock}` appears outside the 3 vetted files (store.rs + the 2 daemon-singleton sites), so a future untracked flock can't silently reopen the blind spot.

### The proof (guard lands clean)
The lead's key ask: if the guard fires **anywhere**, there's an unfixed 6th flock-while-blocking site. It does not. The full suite — which exercises every flock+emit path through the guard, **including the 5 previously-buggy sites' behavioral tests** (pr_state / dispatch_idle / decision_timeout / deploy / teardown) — passes with the guard always-on:
- `cargo test --tests --features tray`: **3385 passed, 0 failed, 62 groups** (real git)

A **live-daemon smoke was intentionally skipped**: starting a second daemon would collide with the running fleet daemon, and the suite already evaluates the guard on every self-IPC path (a held flock → `Err` → undelivered message → test failure; all such tests pass).

### Series recap
- PR1 #1632 — 3 daemon-loop emit-deferrals ✅ merged
- PR2 #1633 — deployments lock-narrowing ✅ merged
- **PR3 (this)** — the guard. All 5 sites fixed; the class is now structurally enforced.

Closes #1629

### DIFF
--- a/src/fleet/persist.rs
+++ b/src/fleet/persist.rs
@@ -11,7 +11,7 @@ fn atomic_write_yaml(home: &Path, doc: &serde_yaml_ng::Value) -> Result<()> {
     result
 }
 
-fn acquire_lock(home: &Path) -> Result<std::fs::File> {
+fn acquire_lock(home: &Path) -> Result<crate::store::FileFlockGuard> {
     let lock_path = home.join(".fleet.yaml.lock");
     crate::store::acquire_file_lock(&lock_path).context("failed to acquire fleet lock")
 }
--- a/src/store.rs
+++ b/src/store.rs
@@ -130,14 +130,40 @@ pub fn save_atomic<T: Serialize>(path: &Path, data: &T) -> anyhow::Result<()> {
     atomic_write(path, body.as_bytes())
 }
 
+/// #1629: RAII guard returned by [`acquire_file_lock`]. Holds the locked `File`
+/// (the OS advisory lock releases when the `File` drops) and bumps the
+/// `FLOCK_DEPTH` thread-local for its lifetime so the self-IPC deadlock guard
+/// (`assert_no_registry_lock_for_self_ipc`) can see the flock tier — holding a
+/// flock across a loopback `api::call`/`enqueue_with_idle_hint` is the #1617
+/// lock-while-blocking deadlock class. On drop the depth decrement runs before
+/// the inner `File`'s drop releases the OS lock; both happen at the same scope
+/// exit, so no self-IPC can observe an inconsistent (depth, lock-held) pair.
+///
+/// `Deref` to `&File` is intentionally NOT provided: every call site holds the
+/// guard purely for RAII (`let _lock = acquire_file_lock(..)?`), so an opaque
+/// guard keeps the flock-depth invariant un-bypassable.
+pub struct FileFlockGuard {
+    _file: std::fs::File,
+}
+
+impl Drop for FileFlockGuard {
+    fn drop(&mut self) {
+        crate::sync_audit::flock_exited();
+    }
+}
+
 /// Acquire an exclusive advisory lock tied to `lock_path`.
 ///
-/// Released when the returned handle is dropped. Do NOT open the lock file
-/// with `truncate(true)` — truncation is unnecessary (the file contents are
-/// meaningless) and invites confusion in crash paths where a partially
-/// initialised lock file might be observed empty by another opener while
-/// the flock itself is still held on the inode.
-pub fn acquire_file_lock(lock_path: &Path) -> anyhow::Result<std::fs::File> {
+/// Released when the returned [`FileFlockGuard`] is dropped. Do NOT open the
+/// lock file with `truncate(true)` — truncation is unnecessary (the file
+/// contents are meaningless) and invites confusion in crash paths where a
+/// partially initialised lock file might be observed empty by another opener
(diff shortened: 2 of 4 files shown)


suzuke:
## §3.12 Review Mirror — dual VERIFIED (capstone, re-reviewed through 3 invariant rounds)

**codex**: REJECTED ×3 on the flock_depth_invariant detection (each empirically proven via probe): (1) line-only — missed nothing here; (2) only fully-qualified `fs4::FileExt::` strings, missed imported-trait method `.try_lock()`; (3) single-line import gate, missed multi-line `use fs4::{...FileExt...}`. dev fixed each → VERIFIED at 4802838: multi-line `use fs4::{...FileExt...}` + method-call probe now caught; UFCS caught; Mutex/RwLock + disk.rs (fs4 without FileExt) spared. Honored the lead scope decision that glob/alias deliberate-evasion is a documented limit (accidental-reintroduction tripwire, not a security boundary). /tmp/1635-codex.md.

**reviewer-2**: VERIFIED the core guard at 8298d76 (the proof: 3385 tests green under the always-on guard, incl behavioral tests for all 5 ex-buggy sites → guard never trips → no 6th site; FileFlockGuard ++/-- symmetric, opaque/un-bypassable; 2 daemon-singleton bypasses correct) + VERIFIED the round-3 invariant fix at 4802838. /tmp/1635-reviewer-2.md.

CI: all 6 green at 4802838. CAPSTONE of #1629 — extends the self-IPC deadlock guard to the fs4 flock tier (FLOCK_DEPTH via FileFlockGuard newtype + sync_audit flock>0 clause + flock_depth_invariant test). With PR1 #1632 + PR2 #1633 (the 5 site fixes) already merged, the flock-while-blocking deadlock class is now structurally sealed + unit-test-catchable. Closes #1629. Merging --admin --squash.

PR #1637, opened 2026-06-02
refactor(#1636): dispatch_idle status String → DispatchStatus enum

suzuke:
## What
Convert the dispatch-idle sidecar `status` from a stringly-typed 4-state `String` (with `status != "pending"` / `== "pending"` guards copy-pasted across ~13 sites) to a `DispatchStatus` enum, so the compiler enforces exhaustiveness at every guard. Mirrors the `pr_state` enums next door.

```rust
#[derive(Debug, Default, Clone, Copy, Serialize, Deserialize, PartialEq, Eq)]
#[serde(rename_all = "lowercase")]
pub(crate) enum DispatchStatus {
    #[default] Pending,  // armed; the only state that surfaces / matches scans
    Resolved,            // matching report arrived → disarmed
    Exceeded,            // threshold elapsed → nudge fired
    Cancelled,           // underlying task terminal → sidecar void
}
```

## Pure type refactor — NO behavior change
- **Wire compat preserved:** `#[serde(rename_all = "lowercase")]` → the enum serializes to / deserializes from the **same** on-disk/IPC strings (`pending`/`resolved`/`exceeded`/`cancelled`). `#[serde(default)]` + `#[default] Pending` replaces the old `default_status()` fn (missing `status` key still loads as `pending`).
- **Strict, like pr_state:** an unknown status string fails to deserialize, so `list_pending`'s existing fail-open loader (`from_str(...) else continue`) skips that sidecar — the module only ever writes the four known states.

## Sites converted (~13 guards/assigns + test asserts)
`src/daemon/dispatch_idle/mod.rs` (all `!= "pending"` / `== "pending"` guards, the `= "resolved"` / `= "exceeded"` assigns, struct inits), `fixup_nudge.rs` (the `!= "exceeded"` guard + sidecar inits), `src/daemon/idle_watchdog.rs` (pending-sidecar check), `src/api/handlers/messaging.rs` (test asserts).

## Test
New `dispatch_status_serde_roundtrip`: pins each variant ↔ exact lowercase wire string, the no-`status`-key default → `Pending`, and that an unknown string is rejected. Existing tests (incl. ones that write raw `"status": "..."` JSON to disk) still pass — they validate the wire compat end-to-end.

## Out of scope (separate structs, NOT `dispatch_idle::PendingDispatch`)
`decision_timeout`, `dispatch_tracking`, `comms`/`orphan_sweep` status fields — different state machines, left untouched (clippy-verified no type bleed).

## Preflight
[...]

PR #1621, opened 2026-06-01
fix(bughunt2): auto-clear watchdog rate-limit/quota health latch on recovery

suzuke:
## Bug (dev-2 bug-hunt #2, P1)

The watchdog set the `RateLimit`/`QuotaExceeded` `BlockedReason` health latch but **never cleared it**: `classify_pty_output` returns `None` once the throttle banner scrolls off, so `run_watchdog_pass` early-returned and the latch persisted. That latch permanently **suppresses hang detection AND blocks task delivery** — so after a **transient** rate-limit an agent stayed silently "blocked" forever (only a manual `health clear_blocked_reason` released it), even though the underlying `AgentState` self-expires when the limit lifts. A set/clear asymmetry.

## Fix

`run_watchdog_pass` now takes the live `AgentState` and **auto-clears** the latch once the agent is actively working/ready again (`Ready`/`Idle`/`Thinking`/`ToolUse`) — mirroring the AgentState self-expire.

- **Guarded** to ONLY the `RateLimit`/`QuotaExceeded` latch. `AwaitingOperator` / `PermissionPrompt` / `Crash` / `Hang` (operator- or crash-action-required) are **never** auto-cleared (cf. the #1564 blocked-reason guard).
- If the agent is in fact still throttled, the subsequent `classify`+`set` **re-latches** in the same pass (so a genuinely-limited agent stays blocked).
- **dry-run stays read-only** — it neither sets nor clears.
- The per-tick caller passes `core.state.current`.

## Tests

- latch auto-clears on recovery — `RateLimit` + `QuotaExceeded` × (`Ready`/`Idle`/`ToolUse`);
- a still-limited agent (throttle banner + `RateLimit` state) **stays latched**;
- the guard **spares** `AwaitingOperator`;
- **dry-run does not clear**.

## ⚠ Validation note (daemon-behavior fix)

The watchdog runs **in the daemon**, so this takes runtime effect only after a **rebuild + daemon restart** (the live daemon won't exercise the new code until then).

## Verification

- `cargo fmt --check` ✅
- `cargo clippy --all-targets --features tray -- -D warnings` ✅
- `cargo test --features tray --bin agend-terminal` → 3083 passed, 0 failed ✅
- `cargo xwin check --target x86_64-pc-windows-msvc --features tray` ✅

[...]

PR #1641, opened 2026-06-02
refactor(#1638): type-encode BlockedReason clear + hang-suppress policy

### LINKED ISSUE #1638: Refactor R4: health.rs BlockedReason latch — encode the clear-policy in the type (kill the #1621 set-without-clear class)
## Problem (refactor-scout R4, operator-triaged ④)

health.rs BlockedReason is a latch that's SET from many places but CLEARED only via special-case logic — the exact set-without-clear asymmetry that caused #1621 (RateLimit/QuotaExceeded never auto-cleared). The clear-policy lives in scattered procedural code, not the type, so a new BlockedReason variant can silently reintroduce the bug (set, no clear).

## Fix (encode the invariant in the type)

Make each BlockedReason variant declare its clear-policy on the TYPE — e.g. an `auto_clears_on_recovery() -> bool` / `reconcile_on_state(state)` method (or a per-variant policy enum) so the watchdog's clear logic is driven by the type, and adding a new variant FORCES a clear-policy decision (compiler/exhaustiveness). Kills the #1621 recurrence class structurally.

## ⚠ SPIKE FIRST (report before impl)

This touches the health/watchdog state machine — blast radius unknown. Spike: enumerate the BlockedReason variants + where each is SET + where/whether each is CLEARED today; design the type-level clear-policy (method vs policy-enum); confirm it preserves the #1564/#1621 guards (AwaitingOperator/PermissionPrompt/Crash/Hang must NOT auto-clear; RateLimit/QuotaExceeded must). Report the design + risk → lead greenlights → impl.

## Source

refactor-scout R4 (the real set-only latch; the CI-watch dedup markers were false positives per the #1630 spike).

suzuke:
## #1638 (refactor-scout R4) — kills the #1621 set-without-clear latch class structurally

`BlockedReason`'s per-variant policy was three hardcoded `matches!` (two clear axes + one hang-suppress), so a NEW variant silently inherited "never auto-clears / never suppresses" with no compile nudge — exactly how #1621 latched `RateLimit` forever.

### Now on the type (all **wildcard-free** — the forcing function)
- **`RecoverySignal { RateLimitLifted, OperatorResolved }`** — the two independent recovery axes. A single "recovered?" bool would conflate them: the watchdog axis must NOT clear `AwaitingOperator` (#1564), but the supervisor axis MUST (#1552). (This corrected the original single-axis framing in the spike.)
- **`auto_clears_on(&self, signal) -> bool`** replaces the `matches!` at `watchdog.rs` (rate-limit axis, #1621) and `supervisor.rs` (operator-resolution axis, #1552).
- **`suppresses_hang_check(&self) -> bool`** replaces the `matches!` in `HealthTracker::check_hang`.

A new `BlockedReason` variant now fails to compile in both methods until it declares its clear-policy (both axes) AND its hang-suppress-policy.

### Behavior-preserving — the table IS the spec
`blocked_reason_policy_table_1638` asserts every variant×signal cell matches today's behavior:

| variant | clears on RateLimitLifted | clears on OperatorResolved | suppresses hang |
|---|---|---|---|
| RateLimit / QuotaExceeded | ✅ | ✗ | ✅ |
| AwaitingOperator | ✗ (#1564) | ✅ (#1552) | ✅ |
| PermissionPrompt / Hang / Crash | ✗ | ✗ | ✗ |

The existing #1564/#1621/#1552 watchdog+supervisor behavioral tests stay green (101 in those modules).

### Review-confirm / possible follow-up (NOT changed here)
`PermissionPrompt` does not auto-clear on InteractivePrompt resolution today (supervisor clears only `AwaitingOperator`). Preserved as-is — flag if it should be its own issue.

### Preflight
- `cargo test --tests --features tray`: **3388 passed, 0 failed** (real git)

Daemon behavior — the live clear paths need rebuild+restart to exercise; the exhaustive-match forcing function + table test are the in-PR proof. Base is 4812adc; no overlap with the just-merged #1637 (dispatch_idle enum), so it merges clean.

Closes #1638

### DIFF
--- a/src/daemon/supervisor.rs
+++ b/src/daemon/supervisor.rs
@@ -863,12 +863,12 @@ fn tick(
                 // operators aren't vibrated twice per interactive cycle.
                 // #1552: clear the AwaitingOperator health reason on recovery so
                 // `check_hang` is no longer exempt and a future stall can
-                // re-notify (the once-per-episode dedup re-arms). Guarded so we
-                // never clear a different blocked reason (RateLimit / etc.).
-                if matches!(
-                    core.health.current_reason,
-                    Some(crate::health::BlockedReason::AwaitingOperator)
-                ) {
+                // re-notify (the once-per-episode dedup re-arms). #1638: the
+                // operator-resolution clear-policy is now on the type, so this
+                // never clears a different blocked reason (RateLimit / etc.).
+                if core.health.current_reason.as_ref().is_some_and(|r| {
+                    r.auto_clears_on(crate::health::RecoverySignal::OperatorResolved)
+                }) {
                     core.health.clear_blocked_reason();
                 }
                 tracing::info!(
--- a/src/daemon/watchdog.rs
+++ b/src/daemon/watchdog.rs
@@ -1,7 +1,7 @@
 //! Watchdog: classify PTY output into BlockedReason per daemon tick.
 
 use crate::backend::Backend;
-use crate::health::{BlockedReason, HealthTracker};
+use crate::health::HealthTracker;
 use crate::state::AgentState;
 use std::path::Path;
 
@@ -60,10 +60,10 @@ pub fn run_watchdog_pass(
     // blocked-reason guard). The set below re-latches if the agent is in
     // fact still throttled (classify still matches).
     if recovered_from_rate_limit(current_state)
-        && matches!(
-            health.current_reason,
-            Some(BlockedReason::RateLimit { .. } | BlockedReason::QuotaExceeded)
-        )
+        && health
+            .current_reason
+            .as_ref()
+            .is_some_and(|r| r.auto_clears_on(crate::health::RecoverySignal::RateLimitLifted))
     {
         health.clear_blocked_reason();
     }
--- a/src/health.rs
+++ b/src/health.rs
@@ -208,6 +208,20 @@ pub enum BlockedReason {
     Crash,
 }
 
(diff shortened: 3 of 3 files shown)


suzuke:
## §3.12 Review Mirror — dual VERIFIED

**codex**: VERIFIED @ 51a7338 — policy table preserves current behavior for every variant×signal cell (RateLimit/QuotaExceeded clear only on RateLimitLifted + suppress-hang; AwaitingOperator clears only on OperatorResolved + suppress-hang; PermissionPrompt/Hang/Crash neither axis + no suppress). watchdog site passes RateLimitLifted, supervisor recovery-notice site passes OperatorResolved (not swapped). check_hang uses suppresses_hang_check. Both methods wildcard-free. PermissionPrompt no-auto-clear matches origin/main (not in either clear match nor the suppress match). /tmp/1641-codex.md.

**reviewer-2**: VERIFIED @ 51a7338 — signal not swapped; suppresses_hang_check ≡ old matches!; both methods exhaustive/wildcard-free (new variant compile-fails); PermissionPrompt MCP-cleared only (current behavior). /tmp/1641-reviewer-2.md.

CI: all 6 green. #1638 / ④ — encode BlockedReason per-variant policies (RecoverySignal two-axis auto_clears_on + suppresses_hang_check) in the type, replacing 3 scattered matches!. A new variant now compile-fails until it decides its clear-policy (both recovery axes) + hang-suppress → kills the #1621 set-without-clear class structurally. Behavior-preserving. Merging --admin --squash.

PR #1645, opened 2026-06-02
refactor(#1644): pin tick collect→drop→emit boundary + correct stale comment (DEFER the god-fn extraction)

### LINKED ISSUE #1644: Refactor R3: supervisor.rs tick() is a 395-line god-fn — extract supervise_one() to type the collect-under-lock invariant
## Problem (refactor-scout R3, operator-triaged ⑤ — the biggest/last Phase-2 item)

supervisor.rs:642 tick() is a ~395-line god-function with 13 issue references (#1316…#1586) and ~20 historical rewrites. Its core invariant — "collect-under-lock → drop → emit" (the #1492 pattern) — is guarded only by COMMENTS (and the #1492 guard is registry-scoped, blind to core-held self-IPC). A 395-line fn that has been rewritten 20× is a change-magnet and merge-conflict hotspot; the invariant should be TYPE-structured, not comment-guarded.

## Proposed fix (from the scout)

Extract a per-instance `supervise_one(...) -> TickOutcome` so the collect-under-lock → drop → act invariant is encoded in the function boundary (collect happens in the locked phase, the TickOutcome is acted on in the unlocked phase). tick() becomes a thin loop over supervise_one + a barrier-free act phase.

## ⚠ SPIKE FIRST (this one is high-risk — report before any impl)

This is the highest-verify-cost candidate: supervisor tick() is daemon-critical, and it's PR-is-own-dogfood (the running daemon supervises the fleet, so a regression is felt live). Spike must:
1. Map tick()'s actual responsibilities + the exact collect-vs-act boundary (what's read under lock, what's emitted/spawned after).
2. Confirm the #1492 collect→drop→emit invariant + whether the FLOCK_DEPTH/registry guards (now merged via #1629) already partially enforce it — i.e. is this refactor still net-positive given #1629 landed, or has #1629 reduced its value?
3. Propose the supervise_one()->TickOutcome boundary + a structural invariant test, and HONESTLY assess the risk/payoff (this may be the one where verify-cost ≥ payoff, or needs to be split into smaller extractions).
4. Recommend GO / SPLIT / DEFER with reasoning.

Report the spike → lead decides whether to greenlight, split, or defer. Do NOT impl until then.

## Source

refactor-scout R3 (flagged highest-payoff-but-highest-verify-cost; sequenced LAST deliberately).

suzuke:
## #1644 (refactor-scout R3) — spike outcome: DEFER the big extraction, ship 2 cheap wins

The proposed `supervise_one()->TickOutcome` extraction of `supervisor.rs::tick` (~374-line god-fn) is **DEFERRED**. The spike premise-checked the scout's "comment-guarded deadlock invariant" claim against current `main` and found it **outdated**:

- **Safety is already runtime-enforced.** `core` is `Arc<CoreMutex<AgentCore>>`, so `core.lock()` bumps `CORE_LOCK_DEPTH`; the self-IPC vectors call `assert_no_registry_lock_for_self_ipc`, which since **#1535** checks the core tier. An emit moved under the lock returns a fail-fast `Err` (daemon stays live), not a freeze. The big extraction's headline payoff (comment-guarded → enforced) is mostly already captured.
- **Testability is already addressed.** The decision logic is in pure, unit-tested fns: `auth_error_gate`, `reactions_from_transitions`, `reaction_kinds`, `awaiting_escalation_allowed`, `idle_expectation_for`, `clear_waiting_on_if_stale`. `tick` is the orchestration shell.
- **Risk is high.** Daemon-critical, ~20 rewrites, 13 issue-refs, PR-is-own-dogfood. A big-bang rewrite = high verify-cost for a marginal compile-vs-runtime gain → poor payoff/risk.

So this PR ships the **~80%-of-value, near-zero-risk** slice instead (per lead greenlight):

### 1. Correct the stale comment
The intent-collection comment claimed the core-held self-IPC was "tracked in #1535; until then protection is the pure-fn tests + this comment." #1535 is merged — that misleads. Restated to the **actual** current protection: runtime (`CoreMutex` `CORE_LOCK_DEPTH` + the assert → fail-fast `Err`) **and** CI (the new test).

### 2. F2-sibling structural test
`tick_emitters_run_after_core_lock_drops_1644` brace-matches the `let action = { … core.lock() … }` block and asserts the self-IPC/blocking emitters (`maybe_notify_member_state_change`, `gated_notify`, `propagate_usage_limit`) appear **after** it, never inside — CI-time-pinning the collect→drop→emit boundary earlier than the runtime `Err`. This is the "type the boundary" intent of the deferred extraction, with **zero** change to the god-fn. Mirrors the existing `#1530/F2` test; scoped to the `tick` fn body so it can't match itself.

### Deferred-extraction note
The full `supervise_one()->TickOutcome` extraction should be revisited **only when someone adds a new reaction** to the loop (the scout's actual pain point). By then the runtime guard + this structural test make that change safe too.

### Preflight
- `cargo test --tests --features tray`: **3390 passed, 0 failed** (real git). 1-file change (comment + test); the god-fn body is untouched.

Closes #1644

### DIFF
--- a/src/daemon/supervisor.rs
+++ b/src/daemon/supervisor.rs
@@ -751,11 +751,18 @@ fn tick(
             // tails), then emit them lock-free AFTER `drop(core)` below. The
             // member-notify path self-IPCs (`api::call(INJECT)` → orchestrator)
             // and must not run under the core lock — that would risk the #1492
-            // lock-across-self-IPC deadlock. NOTE: the #1492 guard is registry-
-            // scoped and would NOT catch this core-held self-IPC (the reaction
-            // holds the per-agent core, not the registry) — that guard blind
-            // spot is tracked in #1535; until then this site's regression
-            // protection is the pure-fn tests + this comment.
+            // lock-across-self-IPC deadlock. This boundary is now enforced two
+            // ways (the comment-only era is over): (1) RUNTIME — `core` is a
+            // `CoreMutex`, so `core.lock()` bumps `CORE_LOCK_DEPTH`, and the
+            // self-IPC vectors call `assert_no_registry_lock_for_self_ipc`, which
+            // since #1535 checks the core tier too — an emit moved under the lock
+            // returns a fail-fast `Err` (daemon stays live, no freeze) instead of
+            // deadlocking; (2) CI — `tick_emitters_run_after_core_lock_drops`
+            // (#1644) source-grep-pins that the self-IPC emitters live after the
+            // `let action = { … }` lock block, catching a regression before it
+            // ships. The big `supervise_one()->TickOutcome` extraction that would
+            // make it compile-impossible is deferred (#1644) — revisit when a new
+            // reaction is added to this loop; both guards above make that safe.
             for decision in reactions_from_transitions(&transitions) {
                 // #1530/F2: backend resolved from the pre-captured command
                 // (no registry re-acquire while holding core).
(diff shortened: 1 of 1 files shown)


suzuke:

**codex**: VERIFIED @ 4b17958 — comment now accurately states current protection (CORE_LOCK_DEPTH live since #1535 + assert fail-fast + the new test); structural test pins the boundary (3 emitters after the core-lock block, prod-sliced, would RED if an emitter moved into the lock); god-fn body untouched (zero behavior change); emitter set complete. /tmp/1645-codex.md.

**reviewer-2**: VERIFIED @ 4b17958 — comment accurate, structural test brace-matches the core-lock block + asserts emitters after, can't self-satisfy; +75/-5 = comment + test only; the 3 emitters are the complete act-phase self-IPC set. /tmp/1645-reviewer-2.md.

CI: all 6 green. #1644 / ⑤ — supervisor boundary-pin micro-win. The big supervise_one() extraction was DEFERRED (the spike found #1535 already runtime-enforces the collect→drop→emit invariant, so the big-bang's payoff was marginal vs its daemon-critical risk). This PR captures the value at near-zero risk: corrects the stale comment + CI-time-pins the boundary with a structural test. Merging --admin --squash.

PR #1649, opened 2026-06-02
fix(#1647): persist_or_log the 11 real silent-loss persist drops

### LINKED ISSUE #1647: Phase 3 follow-up to #1630: triage the ~25 deferred atomic_write/mutate_versioned/fs::write drops
## Context

#1630 (the silent-enqueue-drop class-killer) deliberately DEFERRED a second set of ~25 `let _ = atomic_write(...)` / `mutate_versioned` / `fs::write` Result-drops because they're MIXED — some are intentionally best-effort/lossy (e.g. #1479 session.json documented-lossy, the .daemon PID file, event_log rotation), others may be real silent-loss bugs. A blanket fix would break the intentional ones; this needs per-site triage.

## This task — TRIAGE SPIKE (read-only, report before any impl)

Enumerate the ~25 deferred sites (atomic_write / mutate_versioned / fs::write Result-drops outside the enqueue class already fixed in #1630). For EACH, classify:
- **INTENTIONALLY-LOSSY** (best-effort, documented or clearly fine to drop — e.g. PID file, session.json #1479, log rotation) → leave as-is, optionally add an explicit `// best-effort: <reason>` comment.
- **REAL SILENT-LOSS** (a persist whose failure would lose state the daemon/operator cares about) → fix with persist_or_log! (the macro #1630 added) or propagate.

Report the classified table (site → verdict → reason) + a proposed fix-list (only the real ones) + risk. Do NOT impl until lead greenlights the fix-list. If the fix-list is small + clearly real, we'll ship them as one PR; if most are intentional, we may just add best-effort comments + close.

## Source

#1630 deferred set (Phase 3).

suzuke:
## #1647 — triage + fix the persist-drop tail that #1630 deferred

Triaged all ~36 remaining `let _ = <persist>` Result-drops (the non-enqueue class #1630 left). **11 are REAL silent-loss; the rest are legitimately best-effort and stay as-is** (markers, PID/heartbeat, per-tick rewritten sidecars, dedup-ledger-backed queues).

### First: generalize the `persist_or_log!` message
It was enqueue-specific (`"enqueue failed — message dropped"`) — misleading at a metadata/dispatch/schedule site. Now op-agnostic (`"op failed — result dropped"`); the `op` field already says which call dropped, so the existing enqueue call sites still read correctly. `tests/enqueue_drop_invariant.rs` is token-based (not message-based) → stays green.

### The 11 fixed (all `persist_or_log!` — every enclosing fn returns `()`, so propagation isn't available)
| sites | store | why REAL |
|---|---|---|
| `dispatch_tracking.rs` track / mark_completed / sweep_stuck / sweep_orphans / cleanup_for_instance | DispatchStore (watchdog reads it) | dropped write → dispatch invisible to watchdog, or loses the status flip that is the SOLE de-dup → #1488 "dispatch stuck" re-nag every tick |
| `schedules.rs` replay_missed_oneshots / set_enabled / mark_success_today / orphan_schedules_for_target | ScheduleStore | dropped write → one-shot double-fires, or #1521 until-success / #1488 orphan re-nags a done/dead schedule |
| `agent_ops.rs` save_metadata / save_metadata_batch | instance metadata (read back by merge_metadata) | MCP set_display_name/description/waiting_on return OK regardless → dropped write was a silent operator-set-but-lost |

### Left INTENTIONAL (not touched)
~21 best-effort sites stay lossy. Only `dispatch_tracking::gc_old_entries` got a `// best-effort:` note — it's the lone unwrapped `mutate_versioned` beside 5 wrapped siblings, so it would otherwise read as an oversight. The 4 borderline log-only candidates (ci_watch poller:333, worktree_pool:468, instructions:432, schedules record_run) are **skipped** (low value, per triage).

### Preflight
- `cargo test --tests --features tray`: **3391 passed, 0 failed** (real git); enqueue_drop_invariant green; 65 dispatch/schedule/agent_ops behavioral tests green.

Additive log-on-error only — happy path unchanged (`Ok` still discarded). The large line counts are fmt re-indenting the closure bodies one level deeper under the macro wrap.

Closes #1647

### DIFF
--- a/src/agent_ops.rs
+++ b/src/agent_ops.rs
@@ -187,8 +187,13 @@ pub fn save_metadata(home: &Path, instance_name: &str, key: &str, value: Value) 
         .unwrap_or(json!({}));
     meta[key] = value;
     let content = serde_json::to_string_pretty(&meta).unwrap_or_default();
-    // M1: atomic write with fsync
-    let _ = crate::store::atomic_write(&meta_path, content.as_bytes());
+    // M1: atomic write with fsync. #1647: log on failure — this metadata is read
+    // back by `merge_metadata`, and the MCP set_* handlers return OK regardless,
+    // so a dropped write was a silent operator-set-but-lost.
+    persist_or_log!(
+        crate::store::atomic_write(&meta_path, content.as_bytes()),
+        "save_metadata"
+    );
 }
 
 /// Persist multiple metadata key/value pairs in a single atomic write.
@@ -205,8 +210,11 @@ pub fn save_metadata_batch(home: &Path, instance_name: &str, entries: &[(&str, V
         meta[*key] = value.clone();
     }
     let content = serde_json::to_string_pretty(&meta).unwrap_or_default();
-    // M1: atomic write with fsync
-    let _ = crate::store::atomic_write(&meta_path, content.as_bytes());
+    // M1: atomic write with fsync. #1647: log on failure — see save_metadata.
+    persist_or_log!(
+        crate::store::atomic_write(&meta_path, content.as_bytes()),
+        "save_metadata_batch"
+    );
 }
 
 // ---------------------------------------------------------------------------
--- a/src/dispatch_tracking.rs
+++ b/src/dispatch_tracking.rs
@@ -43,10 +43,13 @@ fn store_path(home: &Path) -> std::path::PathBuf {
 
 /// Record a new delegation.
 pub fn track_dispatch(home: &Path, entry: DispatchEntry) {
-    let _ = crate::store::mutate_versioned(&store_path(home), |store: &mut DispatchStore| {
-        store.entries.push(entry);
-        Ok(())
-    });
+    persist_or_log!(
+        crate::store::mutate_versioned(&store_path(home), |store: &mut DispatchStore| {
+            store.entries.push(entry);
+            Ok(())
+        }),
+        "dispatch_track"
+    );
 }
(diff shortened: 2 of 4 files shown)


suzuke:

**codex**: VERIFIED @ 02f74c8 — macro generalization safe (op-agnostic message, enqueue sites still read correctly via op label, enqueue_drop_invariant token-based passes — no #1630 regression); the 11 real durability sites (dispatch_tracking 5, schedules 4, agent_ops 2) are the right ones; gc_old_entries remains an intentional best-effort holdout; enqueue_drop_invariant/dispatch_tracking/schedules tests pass. /tmp/1649-codex.md.

**reviewer-2**: VERIFIED @ 02f74c8 — macro op-agnostic, no #1630 regression; 11 real sites correct (persist_or_log! not propagate since each fn returns ()); behavior-preserving additive log-on-Err, Ok-path byte-identical; intentional sites left lossy, only gc_old_entries commented; no scope creep. /tmp/1649-reviewer-2.md.

CI: all 6 green. Fixes #1647 — the #1630-deferred persist-drop set, triaged (11 REAL / 21 intentional / 4 borderline): fix the 11 real silent-loss sites via persist_or_log! (watchdog-visibility, #1488 re-nag, lost operator-set metadata) + generalize the persist_or_log! macro message op-agnostic. The 21 intentionally-lossy sites stay lossy. Merging --admin --squash.

PR #1655, opened 2026-06-02
fix(#1604): set_display_name/set_description — empty value = explicit clear

### LINKED ISSUE #1604: Follow-up #1602: should set_display_name/set_description error on missing value, or document empty as clear?
## Follow-up from #1602 (schema-align audit)

During #1602's enforceability audit, 5 tools were found to declare a field `required` in their MCP schema while the handler actually defaults it (`unwrap_or(...)`). #1602 resolved this by schema-aligning (removing those fields from `required[]`) — zero behavior change.

For two of them the empty-default is **deliberate/documented**:
- `mode` — no `action` → read current mode.
- `set_waiting_on` — empty `condition` → clear (documented "empty to clear").

But for two others the empty-tolerance looks like an **accidental `unwrap_or("")`**, not a designed feature:
- `set_display_name` — no `name` → saves empty display name, returns `{display_name: ""}` success.
- `set_description` — no `description` → saves empty, success.

## Question / scope
Should `set_display_name` / `set_description` called with no value:
- (current, post-#1602) silently set empty, OR
- treat "set X with no X" as a caller error and return a clear error, OR
- explicitly support empty as a documented "clear/reset to default" (like set_waiting_on)?

If empty is NOT a legit clear operation, change the handlers to error on missing/empty (and they can stay `required` in the schema, which the #1602 validator would then enforce cleanly). If empty IS a legit reset, document it like set_waiting_on and keep it optional.

Low priority. Out of scope for #1602 (which deliberately kept zero behavior change). `create_instance` (the conditional `name`) is handled by its own team-vs-name logic and is not part of this question.

suzuke:
## What
`set_display_name` / `set_description` called with an empty (or omitted) value now **explicitly CLEAR** the field (store `null`, return `{"cleared": true}`) instead of silently persisting an empty string. Non-empty values set as before; the 256/1024-char limits still apply.

## Why
Post-#1602 these handlers schema-aligned to their `unwrap_or("")` default, so an empty/missing value **saved `""`** — and the pane reader (`layout::pane::display_name` → `unwrap_or(&agent_name)`) then rendered a **blank** name instead of falling back to the agent name. #1604's audit flagged this as an accidental `unwrap_or("")`, not a designed feature.

## Contract chosen (consistent + least-surprising)
The same one `set_waiting_on` already documents: **empty = explicit clear**. Storing `null` makes the reader's `Option` `None` → display falls back to the agent name / no description. Both handlers are now consistent, and the schema docs say "empty string to clear" (mirroring `set_waiting_on`). (Lead's leaning, confirmed by the reader semantics.)

## Tests
For both tools: set → empty-clears-**to-null** (not `""`) → missing-clears → over-limit-errors.

## Preflight
`cargo fmt`; `clippy --all-targets -- -D warnings` clean for `--features tray` and `--features tray,discord`; mcp handler tests (109) + tools schema tests (14) pass.

Closes #1604

### DIFF
--- a/src/mcp/handlers/instance.rs
+++ b/src/mcp/handlers/instance.rs
@@ -422,6 +422,15 @@ pub(super) fn handle_restart_instance(home: &Path, args: &Value) -> Value {
 
 pub(super) fn handle_set_display_name(home: &Path, args: &Value, instance_name: &str) -> Value {
     let display_name = args["name"].as_str().unwrap_or("");
+    // #1604: empty/missing `name` = explicit CLEAR (reset to the default = the
+    // agent name; see layout::pane::display_name's `unwrap_or(&agent_name)`),
+    // mirroring set_waiting_on. Pre-#1604 it saved `""` → the pane showed a
+    // blank name instead of falling back. Clear stores `null` so the reader's
+    // Option is None.
+    if display_name.is_empty() {
+        save_metadata(home, instance_name, "display_name", json!(null));
+        return json!({"cleared": true});
+    }
     if display_name.len() > 256 {
         return json!({"error": "display_name exceeds 256 character limit"});
     }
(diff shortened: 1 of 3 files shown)


suzuke:

**codex**: VERIFIED @ c65eacb — clear-on-empty correct (set_display_name/set_description store null on empty/missing; layout::pane::label() falls back to agent_name when None → the blank-name bug is fixed end-to-end); over-limit validation intact; tests cover set/empty-clear/missing-clear/over-limit for both. /tmp/1655-codex.md.

**reviewer-2**: VERIFIED @ c65eacb — empty/missing→null→reader falls back to agent_name (no blank pane name); both tools mirror set_waiting_on clear semantics; 256/1024 over-limit errors intact; schema docs updated. /tmp/1655-reviewer-2.md.

CI: all 6 green. Fixes #1604 — set_display_name/set_description had a latent bug (unwrap_or("") → empty/missing saved "" → blank pane name instead of falling back). Fix: empty/missing = explicit CLEAR (store null → reader None → fallback, return {"cleared": true}), consistent between both tools + mirroring set_waiting_on, schema documented. Merging --admin --squash.

Issue #1658, opened 2026-06-02
[fleet-infra] dispatch_idle threshold false-alarms on legitimately-long work (deep scans / full test suites > 600s) — make it work-aware


suzuke:
## Problem (observed 2026-06-02, constant false 'stuck?' noise)

The dispatch_idle watchdog fires 'idle for >600s (threshold 600s)' nudges whenever a dispatched agent hasn't reported in 600s. But MANY legitimate operations exceed 600s:
- a deep read-only bug-hunt / refactor-scout scan (multiple parallel sub-agents reading the codebase),
- a full `cargo test --tests` suite run as preflight (real-git, ~3-4 min) + the impl before it,
- a multi-part security/feature impl.

Every one of these triggers a false 'is it stuck?' nudge that the lead must inspect (pane_snapshot) to confirm 'no, it's working' — and the nudge ALSO lands in the agent's pane, injecting noise mid-work. Today this fired dozens of times, all benign.

## Ask

Make the dispatch_idle threshold WORK-AWARE rather than a flat 600s:
- Don't alarm while the agent is actively producing output / running a tool (it's not idle, it's busy). Distinguish 'no output for 600s' (genuinely stuck) from 'working on a long op for 600s' (fine).
- Or: raise the threshold for known-long task types, or back off the nudge cadence after the first (one nudge, not every 10 min on the same dispatch).
- Or: gate the nudge on the agent's actual agent_state (thinking/tool-use = not idle).

## Severity

P2 — cosmetic/operational noise, but it trains the lead to ignore dispatch_idle (which then misses a REAL stuck agent — the alarm-fatigue failure mode).

Issue #1656, opened 2026-06-02
[process] Risk-tiered review intensity — stop full dual-review on mechanical PRs, concentrate adversarial review where it pays


suzuke:
## Problem (observed 2026-06-02, ~18-PR drive)

Full dual review (codex + reviewer-2) is applied uniformly regardless of PR risk. Evidence from one day:
- **Earned its keep**: #1631/#1635 (codex empirically broke the invariant tests — line-oriented / multi-line / method-call holes the author + the other reviewer missed), #1654 (codex found a real authority-escalation bypass before merge).
- **Near rubber-stamp**: #1639/#1642 (byte-identical dedups), #1636 (enum), #1604 (DX) — both reviewers VERIFIED with zero findings; the coordination tax (dispatch + mirror + force-reassign + 2 review round-trips) was not repaid.

So we pay the same review cost for a 1-line const-extraction as for a security bypass.

## Proposal

Tag each PR a **risk tier** at dispatch and gate review intensity on it:
- **HIGH** (security/authority, concurrency/locks, FP-gates, invariants, anything correctness-critical or wide-blast-radius) → full adversarial dual review, reviewer explicitly instructed to RUN a breaking probe.
- **LOW** (mechanical dedup, type/enum refactor, doc/DX, rename, comment) → single review OR lead-review only.
- **MED** → 1 adversarial reviewer + lead premise-check.

The lead already implicitly knows the tier (today I flagged 'this one's security-critical, dual matters' vs the small ones) — make it explicit + policy. Cuts ~half the review round-trips with zero loss of catches (the catches were all on HIGH-tier PRs).

## Note

This is a lead/protocol practice change (FLEET-DEV-PROTOCOL), not necessarily code. Could later be a task-board field (review_tier) that the daemon's review-routing reads. Filing to track + adopt as SOP.

suzuke:
Resolved as policy — codified as FLEET-DEV-PROTOCOL.md §3.21 'Proportional Ceremony' axis C (review tier: normal/single → dual → adversarial by blast radius), alongside sibling closures #1659 (spike-or-skip = axis B) and #1660 (fleet-vs-single = axis A). Practiced this session (MED-tier single+lead-check on #1661, adversarial reviewer-2 on #1663). The dialectic unanimously rejected a daemon classifier in favor of lead judgment + the protocol §. See #1561 closing comment for the full verify-pass.

PR #1661, opened 2026-06-02
fix(#1657): auto-approve the agend-terminal MCP server for opencode agents

### LINKED ISSUE #1657: [fleet-infra] reviewer-2 (opencode) report send times out → lead force-reassigns every verdict; #1 coordination-noise source
## Problem (observed 2026-06-02, the single biggest coordination tax)

reviewer-2 (opencode backend) completes a review and writes its verdict to /tmp/<pr>-reviewer-2.md, but its `send` (the kind=report back to lead) reliably TIMES OUT. Consequences, repeated ~15× in one day:
- The daemon still shows reviewer-2 'busy' on the just-finished task → the next dispatch returns BUSY → lead must `force=true` reassign on EVERY review.
- dispatch_idle false-positives fire on the stale dispatch every ~10 min.
- The lead's only reliable way to read reviewer-2's verdict is to POLL the /tmp file (the send never delivers) — a workaround, not a fix.

This is pure overhead — reviewer-2's review WORK is fine; only the report-delivery path is broken.

## Ask

Root-cause the opencode `send` (report) path: why does the report send time out while the agent itself is healthy and the /tmp write succeeds? Likely in how the opencode backend's PTY/inject handles the send tool-call vs claude/codex. Fix so a completed review's report delivers reliably (clears the 'busy' state + reaches the lead) — eliminating the force-reassign + /tmp-poll workaround + the dispatch_idle false-positives.

## Severity

P1 operational — not a correctness bug, but the largest single source of orchestration friction in the fleet.

suzuke:
### RCA (3-investigation converging, evidence-airtight)
An opencode-backed agent's MCP tool-calls reliably timed out at 30s. opencode's generated `opencode.json` set `permission.{edit,bash,webfetch,external_directory}="allow"` but **nothing covering the `agend-terminal` MCP server**, so opencode raised a "Permission required" prompt on every MCP tool-call and blocked until the client timeout. The report `send` therefore never reached the daemon's `handle_send`, `mark_resolved` never ran, the dispatch sidecar stayed `Pending`, and **dispatch_idle false-positives fired** (the lead force-reassigned every reviewer-2 review).

The daemon's send/resolve path is fully **backend-agnostic and correct** — the bug is purely that opencode prompts. claude/codex/gemini were immune because each already trusts its MCP server (`autoApprove:["*"]` / `--dangerously-bypass` / `trust:true`); opencode was the only backend missing it.

Evidence fit: /tmp write works (edit/bash allowed) · only MCP `send` times out (MCP tools not in the 4 allowed keys) · only opencode · agent healthy (blocked on an unanswered prompt).

### Fix
Add `"agend-terminal_*": "allow"` to `configure_opencode`'s permission block. opencode names MCP tools `<server>_<tool>` and glob-matches permission keys, so this covers all of the server's tools — the same posture every other backend already has.

### ⚠ Empirical verification (mandatory — a wrong key fails silently)
Verified against the **installed opencode 1.15.10** (the `1.4.0` in backend.rs is a stale label):
- **Treatment** (`agend-terminal_*: allow`): `opencode run` calling `agend-terminal_list_instances` → returned `COUNT=11` **immediately**. (Also confirmed the tool name is `agend-terminal_list_instances` = `<server>_<tool>`.)
- **Control** (current prod config, no key): the identical call **hung indefinitely** on the permission prompt — no result.

This confirms the exact schema key causally suppresses the prompt.

### Scope / threat model
Single-user single-machine: auto-approving the user's own fleet MCP server is consistent with all other backends + necessary for autonomy — acceptable. This is the **root** fix for the reviewer-2 report-timeout noise; #1658's dispatch_idle band-aid is unaffected (stays for genuine long work).

### Tests / preflight
- Extended `opencode_sets_permission_allow_for_autonomous_actions` to assert `permission."agend-terminal_*" == "allow"`.
- `cargo fmt --check` clean; `cargo clippy --all-targets --features tray -D warnings` clean; `cargo test --tests --features tray` = **3406 passed, 0 failed** (real git).

Closes #1657

### DIFF
--- a/src/mcp_config.rs
+++ b/src/mcp_config.rs
@@ -430,6 +430,20 @@ fn configure_opencode(working_dir: &Path, instance_name: Option<&str>) -> Result
     for key in ["edit", "bash", "webfetch", "external_directory"] {
         perm.insert(key.to_string(), json!("allow"));
     }
+    // #1657: auto-approve the agend-terminal MCP server's tools. Without this,
+    // opencode raises a "Permission required" prompt on every MCP tool-call
+    // (including `send` kind=report) and the call blocks until the 30s MCP
+    // client timeout — so a reviewer's report never lands, its dispatch sidecar
+    // stays Pending, and dispatch_idle false-positives fire (the lead force-
+    // reassigns every review). opencode names MCP tools `<server>_<tool>`, and
+    // its permission keys do simple `*` glob matching, so `agend-terminal_*`
+    // covers all of this server's tools. This mirrors the other backends'
+    // posture (claude `autoApprove:["*"]`, gemini `trust:true`, codex
+    // `--dangerously-bypass`); on a single-user single-machine tool, trusting
+    // the user's own fleet MCP server is consistent and necessary for autonomy.
+    // Empirically verified against opencode 1.15.10: with this key a `send`
+    // tool-call completes immediately; without it the call hangs on the prompt.
+    perm.insert("agend-terminal_*".to_string(), json!("allow"));
 
     let body = serde_json::to_string_pretty(&config)?;
     crate::store::atomic_write(&path, body.as_bytes())?;
@@ -692,6 +706,13 @@ mod tests {
                 "permission.{key} must be \"allow\" so autonomous agents don't block"
             );
         }
+        // #1657: the agend-terminal MCP server's tools must be auto-approved
+        // (`<server>_*` glob), else every MCP tool-call (incl. `send` report)
+        // blocks on opencode's permission prompt until the 30s client timeout.
+        assert_eq!(
+            perm["agend-terminal_*"], "allow",
+            "permission.\"agend-terminal_*\" must be \"allow\" so MCP tool-calls (send/report) don't block on a permission prompt (#1657)"
+        );
         std::fs::remove_dir_all(&dir).ok();
     }
 


suzuke:
## §3.12 Review Mirror — VERIFIED (MED-tier: codex adversarial + lead premise-check, #1656 risk policy)

**codex** (adversarial): VERIFIED @ 5975103 — OpenCode permission keys are wildcard patterns over tool NAMES, and MCP-server tools use the `<server>_<tool>` convention, so `agend-terminal_*` covers ONLY this server's tools (server-scoped, no over-match into other MCP servers / non-MCP perms). The repo registers the server as `agend-terminal` in the generated config → the new `permission["agend-terminal_*"]="allow"` is consistent with the existing autonomous-agent permissions. Test pins the key in the output config. Installed OpenCode reports `1.15.10` — matches the patch's empirical claim. opencode_sets_permission_allow_for_autonomous_actions / replaces_scalar_form / preserves_unrelated_keys all pass. /tmp/1661-codex.md.

**fixup-lead** (premise-check): VERIFIED @ 5975103 — scoping confirmed: `agend-terminal_*` is scoped to the fleet's own MCP server, NOT a blanket `*:allow` → narrower than claude's `autoApprove:["*"]`, not over-broad; additive to the existing edit/bash/webfetch/external_directory allows; test asserts the key. Empirical A/B (dev): with key → `agend-terminal_list_instances` returns COUNT=11 instantly; without → infinite hang on opencode's permission prompt → 30s MCP client timeout. Single-user threat model: auto-approving the fleet's own MCP server is consistent with all backends (claude/gemini/codex) + necessary, documented.

CI: all 6 required green (3-platform Check + LOC + audit; Coverage non-required also green). Fixes #1657 — opencode raised a "Permission required" prompt on every MCP tool-call (incl. `send` kind=report), blocking until the 30s timeout → sidecar stuck Pending → reviewer-2 send-timeouts AND dispatch_idle false-positives. Root fix: configure_opencode auto-approves the agend-terminal MCP-server tools. ⚠ Takes effect after daemon rebuild+restart (the running daemon still generates the old config). Merging --admin --squash.

PR #1663, opened 2026-06-02
fix(#1658): debounce dispatch_idle's instantaneous work-aware gate

### LINKED ISSUE #1658: [fleet-infra] dispatch_idle threshold false-alarms on legitimately-long work (deep scans / full test suites > 600s) — make it work-aware

suzuke:
## Reframe (premise-check)
#1658 is **NOT** "add a missing work-aware gate" — one already exists: `target_is_working` (#1516) defers + refreshes the clock while the target is `Thinking`/`ToolUse`, and it sits upstream of the Pending→Exceeded transition, so it already gates **both** the lead-side `dispatch_idle_threshold_exceeded` emit **and** the assignee `fixup_nudge` (which keys on `status == Exceeded`).

The residual noise is that the gate is **instantaneous**: a brief idle gap during active heads-down work (or a momentarily-stale snapshot) landing on the threshold boundary — or a long sparse-output turn whose `Thinking` latch has expired → snapshot reads `Idle` — false-fires.

## Fix — debounce
Require **`DEBOUNCE_SCANS` (3)** consecutive not-working scans past threshold before firing. A new `not_working_streak` sidecar field grows on each not-working scan and **resets to 0** the moment the target is observed working again — so a single idle blip never accumulates to a fire. One change at the #1516 point covers both nudge sides.

- **Only debounces when a snapshot exists** to judge against; no snapshot (daemon boot / tests) → fail-open to immediate fire, matching #1516's own fail-open. Production always has a per-tick snapshot.
- **Real-stuck preserved:** a genuinely idle/unreported target keeps accumulating → fires after at most `(DEBOUNCE_SCANS-1)·scan-cadence` of extra delay — negligible vs the 600s threshold.

## Context
#1657 already removes the **dominant** bombardment source (reviewer send-timeout sidecars stuck `Pending` forever). This addresses only the smaller dev long-heads-down-work case.

## Deferred (documented follow-up)
The structurally-correct fix is to gate on **output-recency** rather than instantaneous state — but `AgentSnapshot` carries no activity timestamp today, so that needs a core-struct + snapshot-writer change (disproportionate for a P2 whose dominant cause is already gone). Revisit only if the residual is still annoying after #1657 + this land.

## Tests
Idle target fires only after `DEBOUNCE_SCANS` (not 1); a working scan resets the streak; the existing #1516 "idle still fires" test updated to the debounce window. Preflight: `cargo fmt`; `clippy --all-targets -- -D warnings` clean for `tray` + `tray,discord`; dispatch_idle(40) + fixup_nudge(5) pass.

Closes #1658

### DIFF
--- a/src/daemon/dispatch_idle/fixup_nudge.rs
+++ b/src/daemon/dispatch_idle/fixup_nudge.rs
@@ -215,6 +215,7 @@ mod tests {
             issued_at: issued,
             status: DispatchStatus::Exceeded,
             nudge_sent_at: None,
+            not_working_streak: 0,
         };
         std::fs::write(
             pending_path(home, &id),
@@ -335,6 +336,7 @@ mod tests {
             issued_at: issued,
             status: DispatchStatus::Exceeded,
             nudge_sent_at: None,
+            not_working_streak: 0,
         };
         std::fs::write(
             pending_path(home, &id),
--- a/src/daemon/dispatch_idle/mod.rs
+++ b/src/daemon/dispatch_idle/mod.rs
@@ -79,7 +79,27 @@ pub(crate) struct PendingDispatch {
     /// source of truth (avoids parallel L2 sidecar bookkeeping).
     #[serde(default)]
     pub(crate) nudge_sent_at: Option<String>,
+    /// #1658: consecutive `scan_and_emit` ticks the target has appeared NOT
+    /// working (snapshot state ∉ thinking/tool_use) while past threshold.
+    /// Debounces the #1516 instantaneous-state gate: a brief idle gap during
+    /// active heads-down work (or a momentarily-stale snapshot) that lands right
+    /// on the threshold boundary must persist for [`DEBOUNCE_SCANS`] consecutive
+    /// scans before firing, so it doesn't false-fire. Reset to 0 the moment the
+    /// target is observed working again.
+    #[serde(default)]
+    pub(crate) not_working_streak: u32,
 }
+
+/// #1658: how many consecutive not-working `scan_and_emit` ticks (past
+/// threshold) the target must show before the dispatch-idle signal fires. A
+/// single brief idle gap during active work is the common false-fire; requiring
+/// a short streak filters it. Cost to a genuinely-stuck agent is at most
+/// `(DEBOUNCE_SCANS - 1) * scan-cadence` of extra delay — negligible vs the
+/// 600s threshold. NOTE: this debounces the EXISTING #1516 gate; it does not
+/// add a missing gate. The structurally-correct fix (gate on output-recency, not
+/// instantaneous state — `AgentSnapshot` has no activity timestamp today) is a
+/// documented follow-up if the residual is still annoying after this + #1657.
+const DEBOUNCE_SCANS: u32 = 3;
 
 /// #1636: lifecycle of a dispatch-idle sidecar, replacing the stringly-typed
 /// 4-state `status` field so the compiler enforces exhaustiveness at the guard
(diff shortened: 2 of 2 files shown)


suzuke:
## §3.12 Review Mirror — VERIFIED (MED-tier: reviewer-2 adversarial + lead premise-check, #1656 risk policy)

**reviewer-2** (ADVERSARIAL — 6 break-probes, zero succeeded): VERIFIED @ e296d5e — (1) no permanent suppression: consistently-idle agent accumulates streak 1→2→3→fires; a hung agent's latched state expires (LATCHED_STATE_EXPIRY 30s) → idle → accumulates → fires. (2) flapping mask ACCEPTED as not-a-bug: an agent working any scan is definitionally alive; a stale-latch FP expires in 30s → idle → accumulates. (3) off-by-one correct: 0→1(<3 defer),1→2(<3 defer),2→3(>=3 fire). (4) fail-open: no-snapshot skips debounce → immediate fire (no_snapshot_falls_back_to_firing still passes). (5) corrupt sidecar: serde(default)→0, saturating_add no-overflow, garbage→serde-skip(fail-open). (6) test integrity: idle_target_still_fires_1516 updated to the debounce window (not deleted) +2 new tests. /tmp/1663-reviewer-2.md.

**fixup-lead** (premise-check): VERIFIED @ e296d5e — traced the 5 stuck-detection safety questions: no permanent suppression (fires by 3rd consecutive idle scan, ≤2·cadence extra delay vs 600s); flapping is correct-by-design (busy-branch resets streak AND refresh_issued_at pushes the clock — double-protected); no off-by-one (`streak < DEBOUNCE_SCANS`=3 fires on 3rd); fail-open preserved (`if snapshot.is_some()` gate); test integrity (existing #1516 invariant test updated, not weakened). Reframe honest: DEBOUNCES the existing #1516 gate, does not add a missing one; output-recency gate correctly deferred (needs AgentSnapshot activity-ts). /tmp/1663-lead-check.md.

CI: all 6 required green (3-platform Check + LOC + audit; Coverage non-required also green). The first windows run's 17 worktree_pool/worktree_cleanup "failed to create worktree" failures were a confirmed windows-runner infra flake (in test SETUP, unrelated to dispatch_idle which #1663 exclusively touches; ALL dispatch_idle tests passed on windows; macos+ubuntu green) — cleared on `gh run rerun --failed`. Fixes #1658 — debounce the #1516 instantaneous-state gate's false-fire (brief idle gap during heads-down work / sparse-output latch-expiry) by requiring DEBOUNCE_SCANS=3 consecutive not-working scans past threshold; resets the moment the target produces output. Pairs with #1657 (the dominant FP cause, opencode send-timeout, already merged). Merging --admin --squash.

Issue #1580, opened 2026-06-01
Retire Backend::Gemini — gemini-cli sunsets 2026-06-18


suzuke:
## Retire `Backend::Gemini` — gemini-cli sunsets 2026-06-18

gemini-cli upstream sunsets **2026-06-18**. agy (Antigravity CLI, `Backend::Agy`) was added (#987) as its replacement. After the sunset date the `gemini` backend is non-functional and should be removed.

**Scope (remove the `Backend::Gemini` arm)**:
- `src/backend.rs` — `Backend::Gemini` preset (spawn args, dismiss patterns, `configure`/`fleet_mcp_supported`), enum variant, alias resolution
- `src/mcp_config.rs` — `configure_gemini` (and ensure agy's config writer is **self-contained**, not coupled to it — see agy un-gate PR)
- `src/state/patterns.rs` — gemini state patterns + the gemini-perm fixture/MANIFEST entries
- tests pinning gemini behavior

**Sibling of**: agy un-gate (agy comes online as gemini goes offline — paired migration). Do NOT delete before agy un-gate lands + is validated, so there's always a working gemini-family backend.

**Timing**: schedule the removal around the 2026-06-18 sunset (after agy un-gate + validation).

suzuke:
Done via PR #1704 (merged). Backend::Gemini retired ahead of the 2026-06-18 gemini-cli sunset. profile()→non-Option (exhaustive, no _=>), all legacy spine deleted (compile_for/config_for_legacy/config_for_productivity_legacy/legacy_initial_state + catch-alls) → #8 BackendProfile consolidation reaches legacy-zero. Agy preserved byte-identical (GEMINI consts→AGY rename, values unchanged). 3150 tests pass, 5 surviving backends byte-unchanged (codex VERIFIED). Residual: from_command(gemini)→Raw (CHANGELOG/README warn).

suzuke:
Reverting the early retirement (PR #1704) per operator: keep Gemini supported until the 2026-06-18 gemini-cli sunset, then re-retire. Retirement was 2 weeks early, driven by #8 legacy-zero engineering convenience rather than a user need. Re-retire scheduled for after 2026-06-18.

suzuke:
[...]

PR #1667, opened 2026-06-02
refactor(#8 Phase 1): BackendBehavior trait scaffold (zero behavior change)

suzuke:
## #8 Phase 1 — `BackendBehavior` trait scaffold (parent t-20260530160634485744-8)

⚠️ **Pure structural scaffold, ZERO behavior change.** No detection logic moves (that's Phase 2). The seam lets Phase 2 (per-backend detection), #1580 (retire Gemini), and #7 (Kiro redraw quirk) land without churning call sites.

### What it does
- `trait BackendBehavior` over the `Backend` enum with the **real** per-backend behavior that exists today: `preset` / `detect_state` / `has_resumable_session` / `should_anchor_on_red`.
- `impl BackendBehavior for Backend` delegates each method **VERBATIM** to the existing inherent method / free fn (`classify_pty_output`). The inherent methods are **unchanged**, so every existing call site (`backend.preset()` etc.) keeps hitting the inherent (inherent wins method resolution) — the zero-behavior-change guarantee.
- Exhaustiveness stays compiler-checked in the existing `match self` sites the trait delegates to.

### Dispatch mechanism: HAND-ROLLED (not the `enum_dispatch` crate)
**Rationale (the KISS fork):** `enum_dispatch` generates the enum→trait fan-out by requiring one concrete **type per variant**, but `Backend` is a flat data enum (incl. `Raw(String)`). Adopting it would mean restructuring `Backend` into per-variant structs — a large, behavior-risky change with no place in a zero-change Phase 1 — plus a proc-macro dep (audit + build cost). Hand-rolled `impl for Backend` delegating to the existing exhaustive matches is strictly simpler and truly zero-change. A new dep didn't earn it.

### Scope notes
- `redraw_after_resize` (illustrative in the task) intentionally **not** added — there's no current per-backend redraw logic to delegate to (adding one = new behavior); deferred to #7 which introduces it.
- `#[allow(dead_code)]` on the trait: by design no production call site uses the facade in Phase 1 (Phase 2 migrates them); precedented for reserved APIs (cf. `error.rs`, `worktree_pool`). The new parity test exercises it.

### Parity proof (load-bearing)
- **No existing test edited** — the diff is additions only (1 file, +101).
- New `backend_behavior_delegates_verbatim` asserts the trait facade `==` the inherent method / free fn for **all 8 variants** (`preset.command`, `detect_state`, `has_resumable_session`, `should_anchor_on_red`).
- All backend/state/behavioral tests pass **unchanged** (366 in those modules).

### Preflight
- `cargo fmt --check` clean
- `cargo clippy --all-targets --features tray -D warnings` clean; `--features tray,discord` clean
- Full suite: 3119 passed, the one failure (`store::test_acquire_file_lock_is_exclusive_same_process`) is an **unrelated fs4-advisory-lock timing flake** — my diff is backend.rs-only and the test passes **3/3 in isolation**.

Refs #8

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -1122,10 +1122,111 @@ mod claude_session {
     }
 }
 
+/// #8 Phase 1 (parent t-20260530160634485744-8): a behavior facade over the
+/// [`Backend`] enum. This is a PURE STRUCTURAL SCAFFOLD — every method delegates
+/// VERBATIM to the existing inherent method / free function; NO detection logic
+/// moves here (that is Phase 2). The seam exists so Phase 2 (per-backend
+/// detection), #1580 (retire Gemini), and #7 (Kiro redraw quirk) can land
+/// without churning call sites.
+///
+/// Dispatch is HAND-ROLLED (`impl BackendBehavior for Backend`), not the
+/// `enum_dispatch` crate. Rationale: `enum_dispatch` generates the enum→trait
+/// fan-out by requiring one concrete TYPE per variant, but `Backend` is a flat
+/// data enum (incl. `Raw(String)`), so adopting it would mean restructuring
+/// `Backend` into per-variant structs — a large, behavior-risky change that has
+/// no place in a zero-behavior-change Phase 1 — plus a proc-macro dependency
+/// (audit + build cost). The hand-rolled impl delegating to the already-
+/// exhaustive `match self` sites is strictly simpler and truly zero-change. The
+/// exhaustiveness stays compiler-checked in those existing matches (`preset`,
+/// `should_anchor_on_red`, `has_resumable_session`, `StatePatterns::for_backend`).
+///
+/// Trait method set is the REAL per-backend behavior that exists today; the
+/// task's `redraw_after_resize` is intentionally NOT added — there is no current
+/// per-backend redraw logic to delegate to (adding one would be new behavior),
+/// so it is deferred to #7 which introduces it.
+///
+/// `#[allow(dead_code)]`: by design no PRODUCTION call site uses the facade in
+/// Phase 1 (call sites still hit the inherent methods, unchanged — that is the
+/// zero-behavior-change guarantee). Phase 2 migrates them onto this trait. The
+/// `backend_behavior_delegates_verbatim` test exercises + parity-checks it now.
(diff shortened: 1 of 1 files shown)


suzuke:
## §3.12 Review Mirror — VERIFIED (#8 Phase 1, LOW-MED: codex adversarial + lead premise-check, §3.21)

**codex** (adversarial): VERIFIED @ 15686728 — the `BackendBehavior` facade is a direct pass-through to the existing canonical impls: `preset`/`has_resumable_session` → inherent methods; `detect_state` → `crate::state::classify_pty_output(self, output)` with the SAME args/order as production callers; `should_anchor_on_red` → inherent. No production call site uses the trait → current code still resolves the inherent methods directly → Phase 1 is zero-behavior-change. Parity test covers all enum variants, passed on PR head. Ran `cargo test backend_behavior_delegates_verbatim` + `cargo test should_anchor_on_red`. /tmp/1667-codex.md.

**fixup-lead** (premise-check): VERIFIED @ 15686728 — zero-change PROVEN by construction: diff is 101 insertions / **0 modified-or-deleted lines** (you cannot change behavior by only adding code nothing existing calls). Trait not `use`-imported into any existing scope → inherent wins method resolution everywhere; facade invoked only by the new fully-qualified parity test. Dispatch correctly HAND-ROLLED (not enum_dispatch crate — Backend is a flat data enum with `Raw(String)`; the crate would force per-variant structs = big restructure + proc-macro dep, wrong for zero-change Phase 1 → new dep not earned, KISS). `redraw_after_resize` correctly DEFERRED to #7. /tmp/1667-lead-check.md.

CI: 6 required green (3-platform Check + LOC + audit; Coverage non-required also green). The dev-flagged `store::test_acquire_file_lock_is_exclusive_same_process` full-suite flake (fs4 advisory-lock timing) is unrelated — diff only touches backend.rs; isolated 3/3 pass. #8 Phase 1 (parent t-20260530160634485744-8): introduces the `BackendBehavior` trait scaffold (zero behavior change) so Phase 2 (move detection per-backend), #1580 (retire Gemini = delete one arm), and #7 (Kiro redraw quirk) all land cleanly on it. Merging --admin --squash.

Issue #1670, opened 2026-06-02
[bug] codex reviewer still not auto-woken by [ci-ready]/next_after_ci on CI green — recurs despite #1473/#1483/#1030 (2026-06-02 repro)


suzuke:
## Symptom — concrete repro 2026-06-02
PR #1668 (#1666 Phase A) re-CI passed **21:22 (Taipei)**. `next_after_ci=fixup-reviewer` (codex backend) was armed; the `[ci-pass]`/`[ci-ready-for-action]` landed in codex's inbox. But **codex stayed IDLE** — agent_state=idle, no pending_response_to, did NOT pick up the re-review. The lead had to manually `send` a nudge ~35 min later; codex then re-reviewed and VERIFIED promptly. This is the SAME stall hit on essentially **every** CI-green → codex-review handoff (lead nudges codex each time).

## Prior art — this class was addressed before but RECURS (prior-art check per #1548)
- **#1473** [CLOSED] next_after_ci/ci-ready 喚醒對 codex reviewer 失效 — auto-handoff 整晚沒 fire (疑 #1457 draft-gate over-apply to agent pane)
- **#1483** [CLOSED] #1473 actionable-wake bypass 是 dead code — 比對方括號 marker 但 pointer 用 `kind=` 欄位 → 永遠不 fire
- **#1030** [CLOSED] reviewer auto-wake gap — ci_watch [ci-pass] doesn't reliably trigger reviewer §3.12 mirror on CI green
- **#982** [CLOSED] inbox-only delivery strands structured messages when recipient is idle at prompt

→ Despite all four closures, codex reviewers STILL require a manual lead nudge on CI-green handoff. **New evidence (today's repro) warrants re-investigation**: regression, incomplete fix, or a new variant.

## Where to look (verify against code, don't assume)
1. **§4.5 ACK absorption**: codex (one-shot backend) skips PTY injection for `kind=update`/`report`. Does it ALSO fail to inject the `[ci-ready-for-action]` (a `system:ci` message) into an idle codex pane — so codex literally never sees the wake? The actionable-wake bypass (#1483) was supposed to force-inject bracket-marker messages regardless of kind; confirm that path is live AND covers the codex one-shot route.
2. **#1483 fix completeness**: it found the bypass compared the `[...]` marker but the pointer used the `kind=` field (dead code). Did the correction fully land + get a regression test? Is there a test that an idle codex pane receives an injected `[ci-ready-for-action]`?
[...]

Issue #1675, opened 2026-06-02
[bug][P1] Operator's TUI typing is force-submitted mid-keystroke when an [AGEND-MSG] inbox inject arrives


suzuke:
## Symptom (operator-reported, actively disrupting 2026-06-02)
When the operator is typing in the TUI (their own/lead pane) and an inbox notification arrives (`[AGEND-MSG-PENDING] …`), the daemon injects that text into the operator's input pane **with a submit key (\r)**, which **force-submits the operator's half-typed message** — the in-progress input is corrupted/sent prematurely. Recurs constantly during active inter-agent traffic; the operator literally cannot type a multi-line/slow message without it being torn in half.

## Likely mechanism
The inject path (`agent::inject_with_target` → pane submit, around `src/layout/pane.rs:134` per the #1665 spike's Gap C) appends the `[AGEND-MSG]` marker + the pane's `submit_key` (`\r`) to the target pane's PTY. For an agent pane that's fine (the agent processes it). For the **operator's input pane while they're actively typing**, the appended `\r` submits whatever they'd typed so far + injects foreign text.

## Why it's bad
The operator's primary interface (typing into the TUI) is unreliable whenever the fleet is chatty — exactly when they're coordinating. This is a direct, high-frequency UX defect.

## Fix direction (verify against the inject/render code, don't assume)
- The operator's input pane should NOT receive auto-submitting injects. Options: (a) when the target pane has UNSUBMITTED input in its line editor, DON'T append the submit key — queue the inject until the line is clear, or render the notification out-of-band (status line / separate region) instead of into the input buffer; (b) inbox notifications destined for the operator's own pane should be display-only (a notification region), never injected-into-input-with-submit; (c) detect 'operator is mid-typing' (recent keystrokes, non-empty input buffer) and defer the inject.
- Relatedly (#1665 spike Gap C): the TUI takeover also clears `reply_to_channel` mid-turn — same area.

## Acceptance
- Typing a slow/multi-line message in the TUI is never force-submitted or corrupted by an arriving inbox inject.
- Inbox notifications still reach the operator (out-of-band / queued), just not by hijacking the input line.

Priority P1 — actively degrades the operator's primary control surface.

PR #1678, opened 2026-06-02
#1670: backend-agnostic wake-submit re-press for stranded actionable wakes

suzuke:
## What
After an actionable wake (ci-ready / task / query) is injected into an idle pane, a fire-and-forget guard thread settles, re-reads the lock-free snapshot, and re-presses the backend `submit_key` while the pane is **still idle** — bounded to 2 re-presses.

## Why (the #1670 RCA)
The wake inject writes `text + submit` in one shot. For codex's `›` input box (ratatui-style, re-renders/debounces), the separately-timed submit can land before the box commits the pasted line → the wake sits un-submitted → codex never wakes. claude's `❯` accepts the same byte sequence, which is why auto-waking worked for claude-reviewer but never for codex-reviewer.

The spike proved the **daemon dispatch path is fully backend-uniform** — `#1473/#1483/#1030` all hardened the dispatch half (actionable-detection, defer-gating, idle-hint), which was never the broken half. The gap is **CLI-input-side**: codex doesn't reliably turn a daemon-injected line into a submitted turn, while a manual nudge does.

## Fix shape (backend-agnostic, NOT codex-special-cased)
- `should_resubmit_wake(state, attempts)` — pure decision: re-press iff the snapshot is still idle-like (`idle`/`ready`) **and** under the cap.
- A pane that transitioned to `thinking`/`tool_use` (wake took) or has a missing/unknown snapshot is left alone → a successfully-submitted line is **never** double-entered, and a stale snapshot never triggers a blind re-press.
- The re-press is a **submit-only inject** (empty `data` → handle_inject writes just the backend `submit_key`).
- Runs in a `std::thread::spawn` **off the emit path** → never blocks the ci-watch poller / dispatch. Lock-free (fresh thread → registry-lock-depth 0), so no #1492 self-IPC-under-lock risk.

## Tests
- `wake_resubmit_tests_1670` — 4 cases: submitted→no-resubmit, stuck-idle→resubmit-until-cap, bounded, unknown→no-resubmit.
- Preflight green: `fmt`, `clippy --all-targets` (`tray` + `tray,discord`) `-D warnings`, inbox suite (133), `spawn_rationale_audit` (the `// fire-and-forget:` rationale satisfies §10.5), `file_size_invariant`, `cargo_include_invariant`.

## Scope / follow-up
Guards the **direct** actionable-inject path (`compose_aware_inject`) — the observed scenario (idle codex at `›`, ci-ready not deferred). The deferred-then-drained path (`app::flush_idle_notifications`, whose own comment already notes "codex one-shots silently drop the wake") carries the same risk — **flagged as a follow-up site**, not widened here.

## Validation
Per the daemon-behaviour-can't-validate-on-own-PR rule, the empirical proof is the **dogfood**: the next real ci-green→codex handoff after merge+restart must auto-submit with **no manual nudge**.

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -405,7 +405,22 @@ impl Backend {
                 ready_pattern: "OpenAI Codex|›",
                 submit_key: "\r",
                 inject_prefix: "",
-                typed_inject: false,
+                // #1670: paced (typed) inject. codex's `›` input widget
+                // (ratatui-style, re-renders/debounces) does not reliably commit
+                // a BULK-written line before the trailing submit `\r` arrives — the
+                // wake sits un-submitted and the agent never wakes; claude's `❯`
+                // tolerates the same bulk bytes, which is why ci-ready auto-waking
+                // worked for claude-reviewer but not codex-reviewer. Typed inject
+                // writes the line in 2ms/byte chunks so the box keeps up and the
+                // line commits before `\r` submits it. The actionable-wake pointer
+                // (`[AGEND-MSG-PENDING]…`) is NOT a system header — it does not
+                // start with `[AGEND-MSG]`/`[from:` — so it takes the CHUNKED path
+                // (not the atomic-header path), i.e. it is genuinely paced. Mirrors
+                // the already-typed backends (opencode/gemini/agy). Paste-race
+                // hypothesis (the A/B that'd confirm is operator-vetoed); the
+                // dogfood — next real ci-green→codex handoff after merge+restart —
+                // is the empirical test (can't validate on this PR).
+                typed_inject: true,
                 resume_mode: ResumeMode::NotSupported,
                 quit_command: "exit",
                 instructions_path: "AGENTS.md",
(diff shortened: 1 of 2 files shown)


suzuke:
## §3.12 Review Mirror — VERIFIED (#1670: codex auto-wake — typed_inject pivot)

**codex** (adversarial, reject→pivot→verify): REJECTED v1 (verify-resubmit guard read the per-tick snapshot which lags the settle window → could double-enter a submitted wake). → VERIFIED v2 @ 2bb80dc: dev PIVOTED to `typed_inject: true` for codex — the actionable-wake pointer is now written through the PACED chunked path (not bulk-written into the `›` widget that raced the submit). The wake pointer `[AGEND-MSG-PENDING]…` is NOT a system/agent header → doesn't hit the atomic-header branch → genuinely paced. notify.rs untouched (verify-resubmit guard removed, #1675 intact, no reverse-revert). typed_inject doesn't break normal codex dispatches (same injector, paced). Cited backend.rs:408-423/1745-1772, agent/mod.rs:1935-1948; ran the pin-test + confirmed files=[src/backend.rs] only. No findings. /tmp/1678-codex.md.

**fixup-lead**: VERIFIED @ 2bb80dc — net diff = src/backend.rs ONLY (+46/-1); notify.rs identical to origin/main (#1675 preserved). The pivot is sound: codex's `›` (ratatui) bulk-paste raced the submit (claude's `❯` tolerated the same bytes — why claude auto-woke, codex didn't); paced typed-inject lets the box commit the line before `\r` → submits first time, aligning codex with the already-typed opencode/gemini/agy. ZERO double-enter risk (no guard/snapshot — removes the whole class codex rejected in v1).

CI: 6 required green. Fixes #1670 — the #1 coordination friction (codex reviewer not auto-woken by [ci-ready]/next_after_ci). ⚠ Validation = DOGFOOD: takes effect after canonical rebuild+restart; the next real ci-green→codex handoff must auto-submit with NO manual nudge (the fix-PR's own dogfood can't show it — this very review needed a manual nudge because the fix isn't deployed yet). Merging --admin --squash.

PR #1679, opened 2026-06-02
fix(#7): redraw Kiro pane after resize (SIGWINCH no-repaint workaround)

### LINKED ISSUE #7: Daemon startup race: try_attach can delete mid-bootstrap run_dir
## Context

Follow-up from #6 (merged as 939d838). Review flagged a race window that exists on **all platforms** (not just Windows).

## The race

Daemon bootstrap writes files in this order (see \`src/daemon/mod.rs:127-150\`):

1. \`create_dir_all(run_dir)\`
2. \`write_daemon_id\` → \`.daemon\`
3. \`auth_cookie::issue\` → \`api.cookie\`
4. \`run_core\` → eventually binds API port + \`write_port\` → \`api.port\`

\`api.port\` is written **last**. Between step 3 and step 4 there is a window where:

- \`is_pid_alive\` returns true (daemon process exists)
- \`api.cookie\` is present
- \`api.port\` is **not yet written / bound**

If a second \`agend-terminal app\` launches during this window, the new \`probe_api\` in \`try_attach\` returns false → the code deletes the live daemon's \`run_dir\` via \`remove_dir_all\`. Daemon's subsequent \`write_port\` then writes into a no-longer-existent directory (or gets orphaned), and the new app proceeds to \`Owned\` and fights for the daemon lock.

## Severity

- Window is typically a few ms — hard to hit under normal conditions.
- Real-world triggers: slow daemon startup (AV scanner, cold cache, slow filesystem), scripted \`daemon &; app\` pairs launched too eagerly.
- Pre-#6 code had a similar but narrower race around cookie presence; #6 enlarges it because port-bind is later than cookie-issue.

## Proposed fixes (pick one)

**A. Reorder daemon startup so \`api.port\` is written before anything else that marks the run_dir as \"live\".**
   - E.g. bind port + \`write_port\` first, then \`write_daemon_id\`, then issue cookie.
   - Invariant: if \`.daemon\` exists, port is listening.
   - Cleanest — no retries, no probes of in-between states.

**B. Short retry in \`bootstrap::try_attach\` before declaring stale.**
   - Probe 2–3 times × ~100 ms before \`remove_dir_all\`.
   - Treats symptom, but simpler change.

## Regression coverage to add

- Test that simulates \"mid-bootstrap\": write \`.daemon\` + \`api.cookie\` but NOT \`api.port\`, call \`bootstrap::prepare\`, assert the run_dir is preserved (not \`remove_dir_all\`ed).

## Notes

- No user report yet; discovered during code review of #6.
- Does not block #6 merge — the main Windows-lag fix is correct.

suzuke:
## Problem (#7)
Switching to a Kiro pane shows it **blank** until you click/keystroke. Root cause: `resize_pty()` (pane.rs) calls `master.resize()` → SIGWINCH, but **kiro-cli 2.1.x TUI v2 does not repaint on SIGWINCH**. agend's dynamic pane layout resizes on every switch, so the AWS-CAO large-initial-size avoidance workaround doesn't apply — we must actively trigger a redraw after resize.

## Fix
Per #8/#1674 (per-backend data lives on `BackendPreset`; the `BackendBehavior` trait is a dead, unwired scaffold — confirmed with lead, and backend.rs:1144 deferred this flag to #7):
- **`BackendPreset.redraw_after_resize: bool`** — `KiroCli = true`, all 6 others (`ClaudeCode`/`Codex`/`OpenCode`/`Gemini`/`Agy`/`Shell|Raw`) = `false`.
- **`pane::resize_pty()`** sends **Ctrl+L (`\x0c`)** — the conventional TUI/readline redraw key, distinct from submit (`\r`) — to the **input writer** *after* `master.resize()`, **only** when the pane's backend preset opts in. Writer cloned + registry lock released before the blocking write (#1530/F1 lock discipline).
- **`redraw_seq_after_resize()`** pure helper makes the selection unit-testable.

## Bounded-to-Kiro guarantee (the safety design)
The redraw byte is emitted **only** for backends with `redraw_after_resize = true` — i.e. Kiro alone. Every other backend emits **nothing** extra on resize, so Claude/Codex/OpenCode/Gemini/Agy/Shell/Raw are **provably untouched**. Tests pin this both ways.

## ⚠ NOT verified in-sandbox — MERGE GATED ON OPERATOR SMOKE
**`\x0c`-safety on live kiro-cli 2.1.x is NOT verified here** — there is no live Kiro TUI in a real terminal in the build sandbox, and no kiro-cli 2.1.x keybinding doc to cite. Code reasoning: `\x0c` (form-feed) is the standard non-destructive redraw key and is **not** the submit key, but "conventionally safe" ≠ verified on this specific binary.

**Risk is bounded to Kiro** (flag false everywhere else). Worst cases on Kiro: (a) `\x0c` is a no-op → fix ineffective but harmless; (b) `\x0c` redraws → fixed; (c) `\x0c` maps to a Kiro command → the only real hazard.

**Requires a 30-sec operator live-Kiro smoke BEFORE merge:** switch to a Kiro pane → content appears immediately (no click), and input/state are not corrupted (type a char, send a message — behaves normally). If a side effect shows up, flip `KiroCli`'s `redraw_after_resize` back to `false` (one-line revert).

→ **ready-for-review + ready-to-merge PENDING operator smoke.** codex reviews code (flag plumbing, bounded-to-Kiro, tests) + lead-check; HOLD at merge-ready until the operator runs the smoke on wake.

## Tests
- `redraw_after_resize_is_kiro_only_7` (backend.rs): preset flag Kiro=true, all others=false.
- `redraw_seq_after_resize_is_kiro_only_ctrl_l_7` (pane.rs): Kiro→`\x0c`, all others + backend-less→`None`.
- `cargo test --test file_size_invariant` ✓; clippy `--features tray` and `tray,discord` clean; fmt clean.

Closes #7.

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -282,12 +282,19 @@ pub struct BackendPreset {
     /// Daemon spawn path emits a `[fleet-mcp-unsupported]` warning when
     /// this is `false` so operators are not surprised by missing tools.
     pub fleet_mcp_supported: bool,
+    /// #7: emit a redraw trigger (Ctrl+L) after a PTY resize. True ONLY for
+    /// backends whose TUI does not repaint on SIGWINCH — kiro-cli 2.1.x TUI v2,
+    /// which otherwise leaves the pane blank until the next keystroke. Read by
+    /// [`crate::layout::pane::Pane::resize_pty`]; `false` for every other
+    /// backend, so they are provably untouched by the redraw injection.
+    pub redraw_after_resize: bool,
 }
 
 impl Backend {
     pub fn preset(&self) -> BackendPreset {
         match self {
             Backend::ClaudeCode => BackendPreset {
+                redraw_after_resize: false, // #7: repaints on SIGWINCH — no trigger needed.
                 command: "claude",
                 args: &["--dangerously-skip-permissions"],
                 ready_pattern: "bypass permissions|❯",
@@ -342,6 +349,10 @@ impl Backend {
                 fleet_mcp_supported: true,
             },
             Backend::KiroCli => BackendPreset {
+                // #7: kiro-cli 2.1.x TUI v2 does not repaint on SIGWINCH, so a
+                // pane switch (which resizes the PTY) leaves it blank until the
+                // next keystroke. resize_pty sends Ctrl+L to force a repaint.
+                redraw_after_resize: true,
                 command: "kiro-cli",
                 args: &["chat", "--trust-all-tools"],
                 ready_pattern:
@@ -387,6 +398,7 @@ impl Backend {
                 fleet_mcp_supported: true,
             },
             Backend::Codex => BackendPreset {
+                redraw_after_resize: false, // #7: repaints on SIGWINCH — no trigger needed.
                 command: "codex",
                 // #1626: `-c check_for_update_on_startup=false` disables codex's
                 // blocking startup update modal ("Update available!"). This is a
@@ -452,6 +464,7 @@ impl Backend {
                 fleet_mcp_supported: true,
             },
             Backend::OpenCode => BackendPreset {
+                redraw_after_resize: false, // #7: repaints on SIGWINCH — no trigger needed.
                 command: "opencode",
                 args: &[],
                 ready_pattern: "Ask anything|tab agents",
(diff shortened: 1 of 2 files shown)


suzuke:
## Merge gate (§3.12 mirror)
- **CI**: 6/6 checks SUCCESS (3-platform Check + LOC + audit green).
- **Reviews**: codex VERIFIED + agy VERIFIED (on code).
- **Empirical smoke (operator, 2026-06-03)**: switched to a live Kiro pane, resize → **immediate repaint** (no blank-screen, no click needed); `Ctrl+L` (\x0c) injection → **no side-effect** (Kiro does not misinterpret it as a command). This is the one premise CI/unit-tests cannot verify; operator-confirmed.

Merging via --admin (shared-account self-approve limitation). Closes #7.

Issue #1680, opened 2026-06-03
draft-gate force-submits operator typing: read_input_submit_timestamps bypasses metadata_path_resolved (write→UUID, read→name)


suzuke:
## Symptom
Operator typing into a fleet pane (e.g. fixup-lead) gets **force-submitted** the moment an inbound message's `[AGEND-MSG-PENDING]` pointer is injected — the #1457/#1677 draft-defer gate never engages. Reproduced live 2026-06-03.

## Root cause (live-verified)
The keystroke-recording WRITE and the draft-gate READ key the metadata file by **different identifiers**:

- **Write**: `record_input_activity(home, &pane.agent_name)` → `agent_ops::save_metadata` → `metadata_path_resolved` resolves the fleet name → **`metadata/<UUID>.json`** (e.g. `2e8eccea-….json`). Confirmed: operator keystroke landed there (last_input age 127s).
- **Read**: `notification_queue::read_input_submit_timestamps` (notification_queue.rs:78) **hand-builds** `home.join("metadata").join("{agent_name}.json")` → reads **`metadata/fixup-lead.json`** (name-keyed, frozen 14d). It bypasses `metadata_path_resolved`.

→ write→UUID file, read→name file, never intersect → `draft_state` always reads stale → `None` → `route_notification` injects directly → submit_key force-submits the live draft.

## Why tests missed it
All unit tests use a home with no fleet.yaml id mapping (e.g. `agent1`), so `metadata_path_resolved` falls back to the name-path — write and read converge in tests, diverge in prod (where fleet.yaml has ids). Test gap: no test exercises the name→UUID resolution split.

## Same reader feeds (all currently read the wrong file)
- `draft_state` (notification_queue.rs:120) → the #1457/#1677 inject defer gate
- `operator_typing_recent` (notify.rs:408)
- supervisor typed-but-not-submitted detection (supervisor.rs:339, 801)

## Fix (minimal)
`read_input_submit_timestamps` must use `agent_ops::metadata_path_resolved(home, agent_name)` (same resolver as the write side) instead of the hand-built path. One line.

## Regression test
Add a test with a fleet.yaml that maps name→id, write via `record_input_activity(name)`, assert `read_input_submit_timestamps(name)` / `draft_state(name)` see it (proves write and read hit the same UUID file). This reproduces the prod split the existing `agent1` tests cannot.

Supersedes the inert portion of #1677 (logic correct, signal dead).

suzuke:
[...]

PR #1681, opened 2026-06-03
#1680: align draft-gate metadata read with the write-side UUID resolver

suzuke:
## Bug (#1680)
The keystroke **write** and the draft-gate **read** keyed the metadata file by different identifiers, so they never intersected:

- **WRITE** — `record_input_activity` → `agent_ops::save_metadata` → `metadata_path_resolved` resolves the fleet name to its UUID → writes `metadata/<uuid>.json`.
- **READ** — `notification_queue::read_input_submit_timestamps` **hand-coded** `metadata/<name>.json` → read the never-written name file → bypassed the resolver.

→ read always returned `(0,0)` → `draft_state` was permanently `None` → `route_notification` injected with `submit_key` → **the operator's unsent draft was force-submitted.** (Live-repro'd by the lead.) The same reader also feeds `operator_typing_recent` (notify.rs) and the supervisor typed-but-not-submitted detection (supervisor.rs) — all fixed by this one alignment.

## Fix
`read_input_submit_timestamps` now uses `agent_ops::metadata_path_resolved` — the **same** resolver the write side uses (one line + the existing `agent_ops` import).

### Double-resolve audit (review ask)
All five reader call sites pass fleet **names** (notify recipient name; `pane.agent_name`; `handle.name` ×2 in supervisor). And even if a UUID were passed, `metadata_path_resolved` converges it to the same `<uuid>.json` (fleet lookup misses → `metadata_path(uuid)`). **No hazard.**

## Regression test (the missed gap)
Every prior test used a home with **no fleet.yaml**, so the resolver fell back to the name path and write/read happened to converge — which is exactly why the prod split was invisible. The new test writes an **id-bearing fleet.yaml** (name → UUID), records a compose keystroke, then asserts the read sees it **and** `draft_state == Drafting`.

**Verified RED before, GREEN after:**
- Before fix: `FAILED ... panicked at notification_queue.rs:372` (typed=0 — read the stale name-path).
- After fix: `ok. 1 passed`.

## Evidence
```
cargo test draft_gate_read_resolves_uuid_like_write_1680   # RED→GREEN confirmed
cargo test notification_queue:: → 12 passed
cargo test inbox::notify → 16 ; supervisor → 56 ; agent_ops → 29 ; all pass
cargo fmt --check ✓ ; cargo clippy --all-targets --features tray / tray,discord -D warnings ✓
```

## Scope
Read-path resolver alignment + regression test only. Other hand-coded metadata paths are an explicit follow-up, not touched here.

### DIFF
--- a/src/notification_queue.rs
+++ b/src/notification_queue.rs
@@ -75,7 +75,12 @@ pub fn record_submit_activity(home: &Path, agent_name: &str) {
 /// detection — keeps the read inline-cheap (single file read, single
 /// JSON parse) so per-tick overhead stays bounded.
 pub fn read_input_submit_timestamps(home: &Path, agent_name: &str) -> (i64, i64) {
-    let meta_path = home.join("metadata").join(format!("{agent_name}.json"));
+    // #1680: resolve via the SAME path resolver the write side uses
+    // (`save_metadata` → `metadata_path_resolved`). The previous hand-coded
+    // `metadata/<name>.json` read the never-written name file while the write
+    // landed on `metadata/<uuid>.json`, so `draft_state` was permanently stale
+    // (`None`) and the inject path force-submitted the operator's unsent draft.
+    let meta_path = agent_ops::metadata_path_resolved(home, agent_name);
     let Ok(content) = std::fs::read_to_string(meta_path) else {
         return (0, 0);
     };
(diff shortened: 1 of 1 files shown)


suzuke:
## Merge gate (§3.12 mirror)
- **CI**: all green — Check (ubuntu/macos/windows) + Coverage + LOC + audit pass (run 26858734627).
- **Review**: fixup-reviewer (codex) **VERIFIED**, adversarial, 4 break-probes all addressed, No findings. Evidence: read path now uses `metadata_path_resolved` (same resolver as write); id-less agents fall back to legacy name-path (no regression); resolver prefers UUID path when both exist; supervisor call sites pass NAME (correct resolution); regression test models the prod name→UUID split (RED→GREEN).
- **Lead-check**: confirmed minimal one-line read-path alignment + the regression test reproduces the prod split the existing `agent1` tests could not.

Follow-up #1682 filed (writers still hand-build name-path, breaking the resolver symlink — out of scope here, read-path only). Closes #1680.

Issue #1651, opened 2026-06-02
[security] binding.json push-authority is agent-writable (self-authorize git push) — sibling of #1576


suzuke:
Found during #1576 spike. runtime/<agent>/binding.json holds the git push-authority allowlist; the agend-git shim raw-reads it on every git op (bin/agend-git.rs:204). A same-uid agent can edit its own `branch` field to self-authorize pushing to any branch. Same root cause as #1576 (same-uid, no fs isolation). Option: extend #1576's HMAC/fail-closed pattern to binding.json (defense-in-depth), or fold into OS-isolation. Operator-decided priority. Defense-in-depth (HMAC) blocks injected blind-writes; true sealing needs OS isolation.

suzuke:
Scope note: single-user single-machine tool (see #1653). binding push-authority tampering is a same-uid injection path. Proportionate option: extend #1576's HMAC/fail-closed pattern to binding.json (defense-in-depth) IF this becomes a real concern; no OS-isolation needed for single-user. Tracked, lower priority than #1576.

suzuke:
Fixed by PR #1688 (merged). binding.json now HMAC-signed (sidecar) via the shared integrity_core; agend-git shim verifies on read and fails closed (unbound→deny) on tamper/missing-sig. Blocks injected blind-write self-authorization. No startup auto-bless (would wash-white tampered bindings); rollout false-deny is visible + self-healing via re-dispatch/bind_self. True sealing = OS-isolation (#1653, parked).

Issue #1653, opened 2026-06-02
[security][epic] OS-level agent isolation (separate uid / restricted token) — the real boundary for the #1576 same-uid class


suzuke:
Root cause behind #1576 + its siblings (binding push-auth, api.cookie impersonation): agents run SAME-UID as the daemon with NO sandbox/chroot/fs-isolation, so they have the daemon's full filesystem access + can read every secret. Any in-process integrity scheme (HMAC etc, #1576 Phase 1) is defense-in-depth, NOT a sealed boundary — a same-uid agent that reads the key can re-sign / impersonate.\n\nThe real boundary = OS-level isolation: run agents under a SEPARATE uid, chmod-700 a daemon-private config dir (operator-mode, cookie, binding, runtime-config), keep only the agent-shared workspace (worktrees) agent-writable. Windows needs a restricted token / AppContainer equivalent.\n\nCost (honest): weeks, cross-platform-fragile (Windows has no Unix perms), daemon-critical high-regression. Deferred — operator decides whether the threat model (injection→authority-abuse) warrants paying this, vs the #1576 Phase-1 defense-in-depth being sufficient. Tracking the option here.

suzuke:
⚠ Scope note (operator, 2026-06-02): Agend-Terminal is currently a **single-user single-machine** tool. The only threat is the user's OWN agent being prompt-injected — there is no other human / external attacker / multi-tenant. OS-level agent isolation (separate uid / sandbox / Windows restricted token) is therefore **overkill for the current threat model** and is **PARKED** unless the tool goes multi-user / multi-tenant. The #1576 Phase-1 in-process defense-in-depth (fail-closed + HMAC on the authority gate) is the proportionate response for single-user injection-containment. Keeping this open only as the tracked option for IF the deployment model changes.

suzuke:
[...]

PR #1688, opened 2026-06-03
#1651: HMAC-sign binding.json push-authority (defense-in-depth vs injection blind-write)

suzuke:
## Problem (#1651)
`runtime/<agent>/binding.json` holds the git push-authority (`branch`), and the **agend-git shim raw-read it** with no integrity check. A prompt-injected, same-uid agent could blind-write its own `branch` to self-authorize a push to any branch. This signs the binding and makes the shim verify before trusting it — mirroring #1576's operator-mode.json HMAC scheme.

## Option A — shared `integrity_core` by source (no algorithm drift)
The agend-git shim is a **standalone std-only binary that can't link the lib**, and `config_integrity` lives in the main-binary tree. So:
- New **`src/integrity_core.rs`** holds the verify + key-read primitives. The main binary declares `mod integrity_core`; the shim pulls **the same file** via `#[path = "../integrity_core.rs"]`. One source ⟹ the signer (daemon) and verifier (shim) can never drift (a drift in a security check fails silently: too-loose → fail-open, too-strict → false-deny).
- `config_integrity` keeps the sign side (getrandom key-gen) + **re-exports `verify`**, so `operator_mode`'s `config_integrity::verify` API is unchanged.
- **PARITY PROOF** (lead's hard condition): a golden test pins operator-mode.json's HMAC tag to the value computed from the **pre-extraction** code — byte-identical signing is *proven*, not asserted.

## Wiring
- `binding.rs::bind_full` signs the binding → `binding.json.sig`; `unbind` removes it.
- The shim's `read_binding` verifies (content + sidecar); **any** failure (missing / mismatch / no-key) → `Binding::default()` → reuses the **existing** unbound→`deny_unbound_else_chdir` fail-closed path (no new deny path). A blind self-authorization rewrite downgrades to **deny**, never escalation.
- **Rollout防呆**: bindings + worktrees PERSIST across a daemon restart (not re-bound), so a startup pass `resign_unsigned_bindings` signs pre-#1651 (sidecar-less) bindings — avoiding a false-deny of legit agents. **Security**: only sidecar-*less* bindings are blessed; a binding with an *invalid* sidecar (post-#1651 tamper) stays fail-closed. Idempotent.

## Threat model (per lead, mirrors #1576)
Defense-in-depth against injection blind-write, **NOT a security boundary** — a same-uid agent could read the key + re-sign (accepted); no nonce → self-replay out of model; true sealing needs OS-isolation (parked #1653). Documented in code.

## Tests (4, all required)
| # | Test | |
|---|---|---|
| ① | `tampered_branch_without_resign_is_unbound_and_denied_1651` | edit `branch`, no re-sign → shim unbound → `Action::Deny`. **Verified RED** when the verify gate is disabled. |
| ② | `missing_sidecar_is_unbound_1651` | no sidecar → unbound (fail-closed) |
| ③ | `operator_mode_signing_is_byte_identical_golden_1651` | golden parity across the extraction |
| ④ | `shim_verifier_accepts_daemon_signature_1651` | `integrity_core::verify` accepts `config_integrity::sign` output |

## Evidence
```
cargo test _1651 (shim) → 3 ok ; config_integrity:: → 6 ok ; operator_mode → 12 ok ; binding:: → 16 ok
RED proof: with the verify gate disabled, ① + ② FAIL; restored → pass
agend_git_shim_phase1/2 + app_mode_wiring integration → pass (binding-verify change doesn't break them)
file_size / cargo_include / spawn_rationale invariants → pass
cargo fmt --check ✓ ; clippy --all-targets (tray / tray,discord) -D warnings ✓
(has_unmerged_files conflict test false-fails under the fleet git shim; passes under real /usr/bin/git — environmental, CI uses real git)
```

## Activation
Daemon-side — inert until the daemon restarts onto the new binary (joins #1670/#1680/#1682 for the same restart). On that restart, `resign_unsigned_bindings` blesses existing bindings so no legit agent is false-denied.

### DIFF
--- a/src/bin/agend-git.rs
+++ b/src/bin/agend-git.rs
@@ -17,6 +17,13 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::time::{SystemTime, UNIX_EPOCH};
 
+// #1651: share the EXACT HMAC verifier with the daemon by source (the shim is a
+// separate binary that cannot link the lib). `config_integrity` declares the
+// same file as `mod integrity_core`; this `#[path]` include guarantees no
+// signer/verifier algorithm drift.
+#[path = "../integrity_core.rs"]
+mod integrity_core;
+
 /// #1504 L3: max times the shim may re-enter before hard-failing. Healthy
 /// operation never exceeds 1 (real git ≠ shim → no re-entry), so a small cap
 /// has zero false-trip risk while containing a self-resolution spawn storm.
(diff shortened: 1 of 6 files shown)


suzuke:
## Merge gate (§3.12 mirror)
- CI: all green (Check ubuntu/macos/windows + LOC + audit).
- Review: fixup-reviewer (codex) VERIFIED after a REJECTED→fix cycle. The REJECT caught a real wash-white hole (startup resign_unsigned_bindings would re-sign a tampered+sidecar-deleted binding, blessing forged content). Fixed via Option A: deleted the startup re-sign entirely → no-sidecar binding stays fail-closed (unbound→deny), legit bindings re-sign on next dispatch/bind_self. Threat model documented: defense-in-depth vs injection blind-write, NOT a security boundary (true sealing needs OS-isolation, parked).

⚠ Rollout: on the activating restart, existing bindings (no sidecar) become unbound → one-time re-dispatch / bind_self re-signs them (visible + self-healing). Closes #1651.

PR #1714, opened 2026-06-03
event-bus: migrate the ci_watch CiReady/CiFail pair via the #1690 template

suzuke:
## What

Migrate the per-subscriber **CI pass/fail notify** (`ci_watch`) through the dormant `event_bus` spine (Option A if/else gate, default OFF). **This is the final pattern in the event-bus train.** Task `t-eventbus-ci-watch-pair`.

## Delivery path (verified)

A-path **inbox-enqueue** (`enqueue_with_idle_hint` → durable inbox msg + an idle-hint PTY wake), fanned out **per subscriber** (one CI result → loop over watchers). Drain-testable (the durable enqueue); the idle-hint PTY wake is covered by the shared-deliver-fn invariant.

## Restructure (the meaty part — dual loop → unified skip-aware loop)

The prior code had **two loops**: the legacy enqueue loop (with the `action_target_on_success` + #931 zombie-subscriber skips) AND a *dormant pre-wired emit loop that lacked those skips* and dropped non-success/failure conclusions (`_ => {}`). Unified into ONE skip-aware loop:
- the skips are applied **once**, so gate-ON and gate-OFF notify the **identical** recipient set;
- per recipient: gate-ON + success/failure → `emit Ci{Ready,Fail}`; everything else (gate-OFF, **or** a non-pair `[ci-ended]` conclusion that has no event kind) → the legacy direct `deliver_ci_watch`. No double, no gate-off regression, no drop.

**Fixes two latent bugs** the dormant emit would have detonated once a subscriber registered:
1. it would have notified `action_target` + zombie subscribers that the legacy skips;
2. it silently dropped `[ci-ended]` (cancelled/timed_out/…) via `_ => {}`.

**Migration scope = the success/failure pair only.** Non-pair `[ci-ended]` stays on the legacy path unconditionally (and is no longer dropped) — gate-off byte-identical, no drop, no double regardless of gate.

## How (template)

- extend `EventKind::CiReady` / `CiFail` with `{body, correlation_id, supersede_token}`. `body` is **RENDERED**: it embeds a live-fetched failure-log tail (`gh --log-failed`) — re-fetching in the subscriber would burn GitHub rate-limit + could drift, so it is frozen at the producer (the rendered-text branch of the template, same call as conflict_notify/poll_reminder).
- extract `deliver_ci_watch` (`mark_ci_watch_superseded` + `enqueue_with_idle_hint`), shared by the legacy path AND the subscriber → byte-identical by construction.
[...]

PR #1690, opened 2026-06-03
event-bus: migrate the first notification pattern (anti_stall) via Option A

suzuke:
First concrete pattern through the dormant `event_bus` spine (`src/daemon/event_bus.rs`) — the **template** for the remaining 11. Backlog task `t-20260526155947252378-9`.

## Pattern: anti_stall (why simplest)
Of the 5 wired emit sites (anti_stall / supervisor / ci_watch×2 / dispatch_idle, all subscriber-less), anti_stall is the most self-contained: **one** delivery mechanism (a single `notify_system` loop over fixed recipients `["general","lead"]`), correlation already in the event (`task_id`), constant source/kind. (supervisor does two enqueues + orchestrator resolution; dispatch_idle needs dispatch_id/threshold + event_log; ci_watch has a per-subscriber loop + URL/emoji body across two emit sites.)

## What changed
- `EventKind::TaskStateChanged` gains `started_at` + `eta_secs` so a subscriber can rebuild the **byte-identical** `task_stalled` text the legacy enqueue formats.
- anti_stall: extracted `stall_text` + `deliver_stall` (shared delivery, used by **both** legacy and subscriber) + `handle_event` (subscriber) + `register_subscriber` (wired once at daemon startup).
- **Option A** detection point (`emit_stall`): `if bus.is_enabled() { emit → subscriber delivers } else { legacy direct deliver }`. No double-delivery on the gate-ON path; gate-OFF (prod default) keeps legacy delivery → **no regression**. The legacy `else` is retired only at the final cutover (all patterns migrated + gate default flipped).

## Tests / Evidence
- `gate_on_emit_subscriber_matches_legacy_direct_enqueue` — exercises the **real** bus emit→fan-out→subscriber path; asserts the delivered inbox payloads (from/kind/text/correlation, ignoring volatile id/timestamp) are **identical** to the legacy direct enqueue, per recipient. Meaningful: diverges if the event doesn't carry started_at/eta_secs.
- `gate_off_emit_stall_delivers_via_legacy` — gate off → `emit_stall` still delivers via legacy (no regression).
- Both take the existing `env_lock()` to serialize against the `AGEND_TASK_STALL_RECIPIENTS` override tests (process-global env — fixed an initial flake).
- `event_bus::new_enabled_for_test` lets other modules exercise emit→subscriber without the process-global `global()` bus.

```
cargo test anti_stall:: → 32 passed (×3, no flake) ; event_bus:: → 4 passed
[...]

PR #1715, opened 2026-06-03
event-bus cutover Step 1: default the gate ON in production (legacy kept as kill-switch)

suzuke:
## What

All 12 notification patterns are now migrated through the `event_bus` spine (#1690 → #1714). This is the **end-of-train cutover, Step 1**: flip the gate's code-shipped default so the structured delivery path is **live in production**, while keeping the legacy direct-enqueue `else` branches as a **kill-switch**.

Single-line flip in `EventBus::new()`:
```rust
let enabled = std::env::var("AGEND_EVENT_BUS")
    .map(|v| v != "0")
    .unwrap_or(!cfg!(test));
```

## Test vs prod default divergence (the key design point)

- **Production** (release build, `cfg!(test)=false`): unset env → **ON**. The bus path is live.
- **Test binary** (`cfg!(test)=true`): unset env → **OFF**. The existing legacy-path unit + integration tests keep exercising the **legacy** direct-enqueue unchanged. The bus delivery path stays covered by the per-pattern `new_enabled_for_test` **parity tests** (which already prove `bus == legacy` byte-identical for all 12 patterns) + the Step-1 live validation cycle.

**Why `cfg!(test)` and not a plain `unwrap_or(true)`:** flipping ON unconditionally breaks every integration test that drives a production fn reading `global()` and asserts delivery (empirically: 14 ci_watch `run_ci_check`-based tests) — in tests `global()` has no registered subscriber (register runs only in `run_core`), so emit fans out to nothing and the delivery assertion fails. `cfg!(test)` keeps the test binary on legacy → **zero test refactor**, and preserves more coverage than the deliver-direct alternative (gate-off routing stays exercised by the integration tests).

⚠ **`cfg!(test)` edge checked**: separate `tests/*.rs` integration crates link the lib with `cfg!(test)=false` → they'd see the bus default ON. Verified the full `cargo test --tests` suite: **no `tests/` crate drives the notification/fan-out path** (the 3 that grep-match only reference it in a doc comment / source-audit string list), so none hits bus-ON. Full suite is green except one **pre-existing environmental** failure (`agend-git has_unmerged_files_true_on_conflict`, a false-fail under the fleet `git` shim — passes with real git; unrelated to this change).

## Kill-switch

`AGEND_EVENT_BUS=0` + **restart the daemon** → legacy direct-enqueue. `global()` is a `OnceLock` that reads the env once at startup, so it is **not a hot-toggle** (restart required). The legacy `else` branches are retained precisely as this kill-switch.

## Step 2 (follow-up, after a live validation cycle — NOT this PR)

- Remove the `cfg!(test)` default; register the subscribers in the test harness so the integration tests run the **bus** path (legacy will be gone, so tests must move onto the bus).
- Drop the legacy `else` branches.
- Remove `#![allow(dead_code)]` (and the now-dead `emit_lazy`).

## Preflight

- `cargo fmt --check` ✅
- `cargo clippy --all-targets --features tray -D warnings` ✅ / `--features tray,discord` ✅
- `cargo test --bin agend-terminal --features tray` → **3187 passed, 0 failed** ✅
- `cargo test --tests --features tray` → green except the one pre-existing env `agend-git` git-shim false-fail (passes with real git) ✅
- `emit_disabled_by_default` stays green unchanged (test build → default OFF).

#1463 diff-check: 1 file (`src/daemon/event_bus.rs`, +14/-3).

### DIFF
--- a/src/daemon/event_bus.rs
+++ b/src/daemon/event_bus.rs
@@ -4,7 +4,12 @@
 //! remain untouched; this module emits a structured `Event` alongside them
 //! so future subscribers can react without per-module plumbing.
 //!
-//! Gated by `AGEND_EVENT_BUS=1`. When disabled, `emit()` is a no-op.
+//! End-of-train cutover (Step 1): the gate now defaults **ON** (code-shipped) so
+//! the event-bus delivery path is live without any operator env. Set
+//! `AGEND_EVENT_BUS=0` and **restart the daemon** to fall back to the legacy
+//! direct-enqueue path (the kill-switch — `global()` is a `OnceLock` that reads
+//! the env once at startup, so it is NOT a hot-toggle). The legacy `else`
+//! branches are retained as that kill-switch until Step 2 (post-validation).
 
 #![allow(dead_code)]
 
@@ -181,9 +186,15 @@ pub struct EventBus {
 
 impl EventBus {
     pub fn new() -> Self {
+        // #event-bus cutover Step 1: default ON in production. Unset env →
+        // enabled; the kill-switch is `AGEND_EVENT_BUS=0` (any other value stays
+        // ON). In the TEST binary the default is OFF so the existing legacy-path
+        // unit/integration tests keep exercising the legacy direct-enqueue
+        // (the bus path is covered by the per-pattern `new_enabled_for_test`
+        // parity tests). An explicit env value overrides in either build.
         let enabled = std::env::var("AGEND_EVENT_BUS")
-            .map(|v| v == "1")
-            .unwrap_or(false);
+            .map(|v| v != "0")
+            .unwrap_or(!cfg!(test));
         Self {
             subscribers: parking_lot::Mutex::new(Vec::new()),
             enabled,


suzuke:

event-bus end-of-train cutover Step 1 — flip global gate default ON (prod activates event-bus), keep legacy as kill-switch. No findings.

### Evidence
- cited: event_bus.rs:189-197 — `.map(|v| v != "0").unwrap_or(!cfg!(test))` → prod unset=ON, test-binary unset=OFF, 0=OFF, 1=ON (four-quadrant correct)
- ran: `cargo test --tests` + `cargo test --doc` → passed (no tests/*.rs or doctest drives notification path → no bus-ON collision)
- ran: `cargo test emit_disabled_by_default` (default-off pass) + `AGEND_EVENT_BUS=0` (kill-switch pass) + `AGEND_EVENT_BUS=1` (flips as expected)
- cited: PR body documents test/prod default split + Step 2 follow-up (remove cfg!(test) + register subscribers in test harness + delete legacy) + restart-required kill-switch

CI: 3-platform Check + LOC + audit 全綠 (ci_watch)。Merging via --admin --squash.

**This completes the event-bus migration train: 12/12 patterns migrated + cutover (prod gate ON, legacy retained as kill-switch).**

Issue #1595, opened 2026-06-01
Follow-up #1339: notification reachability (P0 via Telegram in Sleep) + ServerRateLimit-exhausted severity


suzuke:
## Follow-up of #1339 PR-2 (#1594): notification reachability + severity tuning

PR-2 added the mode-gate at `gated_notify` (the Telegram chokepoint) + bumped #1552/#1563 stall to Error so they survive Sleep. Two scope-outs + one reviewer-flagged item to address:

### S1 — reachability gap
`AuthError` / `UsageLimit` currently reach the operator via the **inbox** path (always-on, not mode-gated) and/or member_state_change, but **do they reach Telegram with Error severity** when the operator is in Sleep (Telegram-only, not reading inbox)? Audit which P0-class signals (Crashed/Hang/AuthError/UsageLimit) emit through `gated_notify` at Error severity vs only via inbox. Any P0 that's Telegram-Warn (suppressed in Sleep) or inbox-only (operator asleep, delegate reads inbox) may not wake the operator.

### ServerRateLimit-exhausted severity (dev-2 + both reviewers flagged)
`supervisor.rs:1095` "API rate limit auto-retry exhausted — manual intervention required" is `Warn` → suppressed in Sleep. Auto-retry giving up arguably warrants piercing Sleep (decision: Error, or keep as non-P0 transient class?).

Related: #1339, #1594, reviewer-2's is_p0_class={Crashed,Hang,AuthError,UsageLimit}.

suzuke:
Step 1+2 shipped via PR #1703 (merged). Step 1: ServerRateLimit-exhausted notify Warn→Error → pierces Sleep gate → operator alerted on the #1699 75min give-up. Step 2: self-orchestrator AuthError escalates to Telegram (Error, cooldown-throttled) — the reachable terminal-P0-no-relay case. Crash/Hung self-orchestrator escalation (different hooks: AgentExitEvent + HealthState::Hung with strong-gate against the 348-FP) tracked in #1701.

PR #1717, opened 2026-06-03
fix(#1701): P0-escalate a self-orchestrator crash to the operator (crash half)

suzuke:
## What

P0-escalates a **self-orchestrator crash** to the operator. A team orchestrator that crashes has no peer to relay an inbox P0, and even a *single* crash loses its in-memory context on respawn — so the operator must be told.

This is the **crash half of #1701**. #1595 covered the AuthError self-orch case and explicitly deferred crash/hang (`supervisor.rs:434`/`474-477`). `HealthState::Hung` self-orch escalation is left to a follow-up — it needs a 348-FP guard before it can escalate (that's why #1595 held it back).

## Design (option B, lead-confirmed)

Escalate on **every** orchestrator crash, not the generic `recent>=2` gate — a self-heal-on-first-crash is exactly the case the operator should know about (lost context, leaderless team). Mirrors #1595: **on-condition + cooldown**, not 2nd-occurrence.

- `health::self_orch_crash_should_notify` — reuses `last_notification`'s `NOTIFY_COOLDOWN` (so a crash-loop can't spam) but **bypasses** `record_crash`'s `recent>=2` gate; stamps on fire.
- `crash_respawn::is_self_orchestrator` — `orch==name` via `teams::find_team_for` (mirrors the supervisor guards); resolved **before** the registry lock (no file IO under lock, #1530 class).
- `crash_respawn::notify_self_orch_crash` — distinct P0 wording (leaderless / no-peer-relay / lost-context) via the same `gated_notify(Error)` Sleep-penetrating path as #1595.

The self-orch decision is taken **before** `record_crash` so `record_crash`'s own `last_notification` stamp can't pre-empt it. A non-orchestrator short-circuits the `&&`, leaving its cooldown + generic `recent>=2` notify + auto-respawn **untouched**.

## Tests (lead-requested matrix)

- `self_orch_crash_first_fires_then_cooldown_blocks_1701` — self-orch FIRST crash → P0 fires; immediate re-fire suppressed by cooldown.
- `generic_first_crash_still_silent_1701` — a regular agent's first crash stays silent (generic path not loosened).
- `is_self_orchestrator_only_true_for_own_orchestrator_1701` — orchestrator→true, member→false, no-team→false.

## Gate

- `cargo test` full bin suite **3190 passed, 0 failed** (+3 new).
- `cargo clippy --all-targets` (tray **and** tray,discord) clean; `cargo fmt --check` clean.
- Did NOT touch the Hung path (follow-up) or the generic crash notify/respawn.
- Rebased onto current `main`; #1463 diff = 2 files (crash_respawn.rs + health.rs), single commit.

Refs #1701 (crash half; Hung self-orch P0 is the remaining follow-up, 348-FP-gated).

### DIFF
--- a/src/daemon/crash_respawn.rs
+++ b/src/daemon/crash_respawn.rs
@@ -35,10 +35,27 @@ pub(super) fn handle_crash_respawn(home: &Path, crashed_name: &str, ctx: &Daemon
         return;
     };
 
-    let (should_respawn, delay, should_notify) = {
+    // #1701: is the crashed agent its OWN team orchestrator? Resolved here (a
+    // teams-file read) BEFORE taking the registry lock, so no file IO runs under
+    // the lock (#1530 class). A self-orchestrator crash has no peer to relay an
+    // inbox P0, so it escalates straight to the operator (below).
+    let is_self_orch = is_self_orchestrator(home, crashed_name);
+
+    let (should_respawn, delay, should_notify, fire_self_orch_p0) = {
         let reg = agent::lock_registry(&ctx.registry);
         match reg.get(&instance_id) {
-            Some(handle) => handle.core.lock().health.record_crash(),
+            Some(handle) => {
+                let mut core = handle.core.lock();
+                // #1701: decide the self-orch P0 BEFORE record_crash (which may
+                // itself stamp last_notification on its recent>=2 path) — the
+                // accessor reads+stamps the shared cooldown, and for a
+                // self-orchestrator we use ONLY this gate, never the generic
+                // recent>=2 one. The `&&` short-circuits for non-orchestrators,
+                // so their cooldown is untouched.
+                let fire_p0 = is_self_orch && core.health.self_orch_crash_should_notify();
+                let (respawn, delay, notify) = core.health.record_crash();
+                (respawn, delay, notify, fire_p0)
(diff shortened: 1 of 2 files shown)


suzuke:

#1701 (crash part) — self-orchestrator crash → dedicated self-orch P0 (option B: every orch crash escalates, cooldown-gated, not recent>=2). Hung part deferred to sub-spike (needs 348-FP guard). No findings.

### Evidence
- ran: `cargo test self_orch_crash_first_fires_then_cooldown_blocks_1701 && generic_first_crash_still_silent_1701 && is_self_orchestrator_only_true_for_own_orchestrator_1701` → passed
- cited: crash_respawn.rs:38-57 — self-orch check before registry lock; fire_self_orch_p0 decided before record_crash(); && short-circuits non-orchestrators (cooldown untouched)
- cited: health.rs:503-518 — self_orch_crash_should_notify() reuses last_notification cooldown, stamps only on fire → can't permanently wedge
- cited: crash_respawn.rs:112-156 — is_self_orchestrator() pure team-file lookup, true only for own orchestrator; gated_notify(Error)

CI: 3-platform Check + LOC + audit 全綠 (ci_watch)。Merging via --admin --squash.

PR #1718, opened 2026-06-03
feat: optional pane state-name badge (runtime-toggleable, default off)

suzuke:
## What

An operator aid for eyeball-verifying state detection (#1713/#1523 context): when `runtime_config.show_pane_state` is on, each pane title appends a `[<State>]` text badge of the **detected AgentState** (all states). **Default OFF** — the colour dot stays the steady-state signal; this is a temporary diagnostic toggle.

## How

- **runtime_config**: `show_pane_state: bool` (default false) + `set`/`get_key` arms, mirroring `hang_auto_recovery_enabled`. **Hot-reloadable via the `config` MCP tool** — `set` persists the file *and* updates the in-memory global (line 193), so a toggle takes effect on the next frame, no rebuild/restart.
- **`pane_title_segments`** takes the detected `state` + a `show_state_badge: bool`. The prod call site passes `runtime_config::get().show_pane_state`, keeping the global out of the render fn and the badge logic unit-testable. The badge uses the same `title_style` — **extra TEXT only; the pane colour (`state_color`) is unchanged**.
- `{state:?}` yields the PascalCase variant name (`[ServerRateLimit]`, `[PermissionPrompt]`, …) the task specified. Note: `AgentState` has no `as_str` (panels_fleet.rs:251's is `String::as_str` on a snapshot field), and the Debug form is the idiomatic representation for a diagnostic badge.
- The transient Restarting/Crashed tab-bar badge (`transient_state_badge`) is **unaffected**.

## Tests

- `pane_title_no_state_suffix` (flag OFF) → no state badge, only base + pending-count.
- `pane_title_state_badge_when_flag_on_1713` (flag ON) → title contains `[ServerRateLimit]`/`[PermissionPrompt]`/`[Thinking]`/`[Idle]`.
- `show_pane_state_default_off_and_toggleable_1713` → default off + `set`/`reload` toggle round-trip (the runtime-toggle path the operator uses).

## Gate

- `cargo test` full bin suite **3188 passed** (+3 new); `clippy --all-targets` (tray **and** tray,discord) clean; `fmt --check` clean. #1463 diff = 2 files (runtime_config.rs + render/core_render.rs).

## Operator usage

Toggle live (no restart): `config` MCP tool → set `show_pane_state` = `true` / `false`.

Refs #1713 / #1523.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1719, opened 2026-06-03
event-bus Step 2 (legacy-zero): remove the legacy delivery path; the bus is the sole path

suzuke:
## What

The end-of-train cutover **Step 2**. Step 1 (#1715) flipped the gate default ON with the legacy direct-enqueue `else` branches kept as a kill-switch; the live-validation cycle passed (post-restart the `dispatch_idle` pattern fired + delivered in prod), so this PR **retires the legacy path** — the event bus is now the *sole* delivery path.

## Changes

- **Remove the 12 patterns' Option-A gate** (`if bus.is_enabled() { emit } else { legacy }`) → `emit` is unconditional.
- **Remove the gate entirely**: `EventBus` has no `enabled` field / `is_enabled()` / env read / `cfg!(test)` default / `new_enabled_for_test`. The bus always delivers.
- **Remove `#![allow(dead_code)]`** + the now-dead `emit_lazy`, and fields the deliver never reads: `CiReady`/`CiFail` `repo`/`branch` (redundant with `correlation_id` = `repo@branch`), `Event.timestamp`, and cron `check_schedules`'s `registry` param (the subscriber owns the registry now).

### home-on-Event (Method A — lead-approved)

With the bus as the sole path, a single globally-registered subscriber must deliver to the correct `$AGEND_HOME`. The home was *captured at registration* (fine for prod's one home, wrong for the multi-home integration tests). So the home now travels **on the `Event`**: `emit(home, kind)` → `Event { home, kind }`; `handle_event(event)` reads `event.home`; `register_subscriber()` is home-agnostic. cron additionally captures the live registry (resolve + inject); the home still comes from the event.

*Rejected alternative:* a test-only "clear-subscribers + per-test re-register + serial" scheme — it reintroduces the flaky-global-mutation pattern #1715 already rejected (AtomicBool-setter).

### Test harness — order-independent, no flake

Because the bus is the only path, an integration test that drives a producer (`emit`→`global()`) needs the subscribers registered. A **`#[ctor] #[cfg(test)]` fn registers them ONCE at test-binary load** — before any `#[test]` — so delivery is **order-independent** (no per-test call, no order-dependent flake). `ctor` is a **test-only `dev-dependency`** — it does NOT ship in the prod binary. Verified stable across repeated `cargo test` runs (3× green).

*Why ctor and not per-test calls:* per-test calls are order-dependent (a delivery test running before any registering test would flake) and "green today, red on some parallel ordering" is the worst kind of latent flake — the same fragility (relying on every test remembering to do the right thing) that Method A was chosen to avoid.

## Rollback

`git revert` of this cutover (no longer an env flag — the gate is gone).

## Tests

- Per-pattern **parity tests** keep proving the deliver path (now on a plain `EventBus::new()` local bus + `emit(home, …)`).
- The former **gate-off tests become end-to-end "delivers via the bus"** integration tests (drive the production fn → `global()` → subscriber → deliver to the event's home).
- The existing scan / `run_ci_check` / `maybe_notify` integration tests now exercise the bus path (via the ctor-registered subscribers).

## Preflight

- `cargo fmt --check` ✅
- `cargo clippy --all-targets --features tray -D warnings` ✅ / `--features tray,discord` ✅
- `cargo test --bin agend-terminal --features tray` → **3188 passed, 0 failed**, stable across 3 runs ✅

Rebased onto latest `origin/main` (#1701; clean — its supervisor changes are in a different region). #1463 diff-check: 19 files, **net −100 lines** (legacy removal). This completes the event-bus migration (12/12 patterns + cutover Steps 1 & 2).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,6 +24,7 @@ dependencies = [
  "cron",
  "crossbeam-channel",
  "crossterm 0.28.1",
+ "ctor",
  "ctrlc",
  "dirs",
  "dunce",
@@ -777,6 +778,16 @@ checksum = "eb2a7d3066da2de787b7f032c736763eb7ae5d355f81a68bab2675a96008b0bf"
 dependencies = [
  "lab",
  "phf 0.11.3",
+]
+
+[[package]]
+name = "ctor"
+version = "0.2.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "32a2785755761f3ddc1492979ce1e48d2c00d09311c39e4466429188f3dd6501"
+dependencies = [
+ "quote",
+ "syn 2.0.117",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -157,6 +157,11 @@ junction = "1"
 embed-resource = "3"
 
 [dev-dependencies]
+# #event-bus Step 2 (legacy-zero): a `#[ctor] #[cfg(test)]` fn registers the bus
+# subscribers ONCE at test-binary load — before any test — so the integration
+# tests (which `emit` to the now-sole-delivery global bus) deliver order-
+# independently. Test-only; does NOT ship in the prod binary.
+ctor = "0.2"
 # CLI binary smoke testing (Sprint 42 Phase 1)
 assert_cmd = "2.0"
 predicates = "3.1"
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -702,7 +702,7 @@ fn run_app(terminal: &mut DefaultTerminal, fleet_override: Option<&Path>) -> Res
             recv(tick_rx_ref) -> _ => {
                 // Periodic maintenance — mirrors daemon tick consumers.
                 // Gated to owned mode via tick_rx (None in attached mode).
-                crate::daemon::cron_tick::check_schedules(&home, &registry);
+                crate::daemon::cron_tick::check_schedules(&home);
                 crate::daemon::ci_watch::check_ci_watches(&home, &registry);
                 {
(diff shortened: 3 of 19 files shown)


suzuke:

event-bus Step 2 legacy-zero — remove all 12 gate+legacy else-branches (emit unconditional), home-on-Event (Method A, bus self-contained), ctor test-harness (#[cfg(test)] dev-dep, order-independent zero-flake), remove dead-fields. Net −100 lines. No findings.

### Evidence
- ran: `cargo test gate_on_emit_subscriber_matches_legacy_direct_enqueue && ..._fire && ..._inject && all_event_kinds_constructible && subscriber_fan_out_notifies_every_member && ci_pass_chain_target_excluded_from_subscriber_loop` → passed (ran 3x stable for ctor order-independence)
- cited: event_bus.rs:152-238 — Event carries home, emit(home,kind) freezes home onto event, no enable/disable gate; each subscriber receives producer's home directly
- cited: event_bus.rs:183-200 — ctor registration #[cfg(test)]-guarded, Once prevents repeat
- cited: mod.rs:406-424 — run_core registers same subscriber set once incl cron with live registry; no pattern subscriber dropped
- cited: cron_tick.rs:243-273 — cron subscriber uses event.home, parity byte-for-byte legacy vs bus, bus-path test proves offline-target delivery real not vacuous

rollback = git revert (gate removed). BEHIND only by orthogonal #1718 (render/config, no event-bus overlap) — merging via --admin.

CI: 3-platform Check + LOC + audit 全綠 (ci_watch)。**This completes the event-bus train: 12/12 patterns + cutover Step 1 + Step 2 legacy-zero.**

PR #1723, opened 2026-06-03
fix(#1720): event-bus handled-count + cron silent-drop visibility

### LINKED ISSUE #1720: Schedule engine: same-day late fires silently missed (2/4 fires gone, no error)
## TL;DR

A recurring cron schedule with 4 daily fires (15:00, 16:00, 21:00, 22:00 Asia/Taipei) fired only the first 2 today (15:00, 16:00) — **21:00 and 22:00 silently did not fire**. Same schedule fired all 4 slots successfully on previous occurrences (5/27 Wed, 5/30 Sat). No error in `run_history`, no orphan/disable, schedule still shows `enabled: true`. The fires just vanished.

This is a regression: the cron engine appears to drop scheduled fires under some condition. There's no operator-facing signal that anything went wrong.

## Evidence

Schedule under test: `s-20260522124730682260-0` (Gooaye 股癌每週三六自動摘要)

- `cron`: `0 15,16,21,22 * * 4,7` (Quartz: Wed + Sat, 15/16/21/22 Taipei)
- `timezone`: `Asia/Taipei`
- `enabled`: `true`
- `fire_strategy`: `always`
- `target`: `claude-ba5740`

### Previous successful days (4/4 fires)

```
2026-05-27 Wed: 07:00 ✓, 08:00 ✓, 13:00 ✓, 14:00 ✓  (UTC; = 15/16/21/22 Taipei)
2026-05-30 Sat: 07:00 ✓, 08:00 ✓, 13:00 ✓, 14:00 ✓
```

### Today (2026-06-03 Wed) — only 2/4

```
2026-06-03T07:00:07Z ✓  (15:00 Taipei)
2026-06-03T08:00:09Z ✓  (16:00 Taipei)
2026-06-03T13:00:00Z ✗  (21:00 Taipei) — MISSED
2026-06-03T14:00:00Z ✗  (22:00 Taipei) — MISSED
```

`schedule list` output for today's run_history (latest 2 only):

```json
{
  \"run_history\": [
    ...
    {\"status\": \"ok\", \"triggered_at\": \"2026-06-03T07:00:07.515287+00:00\"},
    {\"status\": \"ok\", \"triggered_at\": \"2026-06-03T08:00:09.349408+00:00\"}
  ]
}
```

Current UTC time when checked: ~14:20Z. So 13:00Z and 14:00Z fires were both already overdue with no record of attempting them.

The episode that the schedule polls for (Gooaye EP667) was uploaded today between 16:00 and 22:00 Taipei. The 21:00 and 22:00 fires were specifically the retry slots designed to catch this case — and they're exactly the ones that vanished.

## What can be ruled out

- ✘ Not a `fire_strategy=until_success` early-stop — strategy is `always`.
- ✘ Not orphaned — `target: claude-ba5740` still exists and is alive (it processed messages before and after the missed window).
- ✘ Not disabled — `enabled: true`.
- ✘ Not target sleeping/offline — `claude-ba5740` accepted MCP calls and processed manual workload throughout the missed window (verified by transcript timestamps).
- ✘ Not first-time setup — schedule has been firing all 4 daily slots reliably for 2 weeks.

## What's worth checking

The cron engine (`src/daemon/cron_tick.rs`) presumably wakes every N seconds and checks each schedule's next fire time. Possible failure modes:

1. **Next-fire calculation skipped past the missing slots.** If the engine slept too long (e.g. >2h) and woke up after 14:00Z, it might compute the next future fire (which is at 2026-06-06 07:00Z, next Sat) and silently skip the missed ones. Catch-up policy?
2. **Per-schedule tick loop exited early on some unrelated error.** No error showed up in `run_history`, but maybe the error path doesn't record.
3. **Race / lock contention with PR #1557 (`fire_strategy` / `linked_task_id` validation).** First multi-fire-strategy state was deployed ~5/31. This is the first Wed/Sat with multiple late fires after that deployment.
4. **Watchdog / restart in the middle.** If daemon restarted between 08:00Z and 13:00Z, the in-memory \"next fire\" state could be lost. On disk persistence?

## Reproduction concern

This is timing-dependent and infrequent (Wed/Sat × specific late hours). Hard to reproduce on demand. Suggestions:

- Add `next_scheduled_fire_at` to `schedule list` output so the operator can see whether the engine knows about upcoming fires.
- Add `missed_fires_count` or `last_missed_at` field so silent drops become visible.
- Log to `event-log.jsonl` whenever a fire is calculated, dispatched, or skipped, even if `run_history` only records successful dispatch.

## Impact

For my use case (Gooaye podcast summarizer polling for new uploads in a 12:00-22:00 Taipei window), missing the 21:00 + 22:00 fires meant the new EP667 episode wasn't picked up automatically — I had to manually run the pipeline at 22:20 when the operator noticed and asked \"did the retry not happen?\".

For any \"poll until success in a window\" schedule (a common pattern, and the motivation for #1521 / `fire_strategy=until_success`), this kind of silent fire dropout completely breaks the retry semantics. The operator only knows something failed if they happen to notice the absent notification.

## Acceptance

- Cron engine reliably fires every match of the cron expression for an enabled schedule, OR
- If a fire is deliberately skipped (e.g. because the daemon was down at that minute and catch-up is off), record `{status: \"skipped: reason\", triggered_at: ...}` in `run_history` so the operator can see it.
- New field `next_scheduled_fire_at` in `schedule list` for diagnostics.

## def_schedule streak update

This makes it 6 issues in ~3 weeks on the schedule feature surface:
- #1095 — cron docs (Unix vs Quartz dow)
- #1521 — fire_strategy feature request
- #1557 — implementation, partial
- #1600 — MCP schema exposure (follow-up to #1557)
- #1608 — task_file_exists wrong storage abstraction (validator unusable)
- **#XXXX (this) — silent late-fire dropout**

Each one in isolation is small. Cumulatively the cron engine is becoming the system's most fragile MCP surface. Worth scheduling a focused review and an integration test suite that exercises actual cron timing.

## Related

- #1521, #1557, #1600, #1601, #1608

suzuke:
## #1720 — silent cron-drop

A cron schedule's fires vanished after the event-bus cutover restart: no error,
no `run_history` entry, schedule still enabled. **RCA:** the intermediate cutover
binary emitted the `CronFire` event while its cron subscriber was momentarily
unregistered, so `emit` fanned out to **nothing** and the fire was lost silently.
Current `main` (legacy-zero) already fixed the wiring and has the e2e prod-wiring
test; this PR hardens against the failure **class** so it can never be silent again.

### (B) handled-count — the bus now reports delivery
- `Subscriber` becomes `Fn(&Event) -> bool`. Each `handle_event` returns `true`
  iff it matched its `EventKind` (and ran its deliver), `false` otherwise.
- `emit` returns the **handled-count**. A producer can now distinguish "reached a
  relevant subscriber" from "fanned out to nothing" — the exact gap the cutover
  hid. (The pre-cutover parity tests used a **local** bus and never exercised the
  global register→emit path, so this gap was invisible to them.)

### (Y) cron visibility — no more silent drop
- `check_schedules` captures the handled-count; `note_unhandled_cron_fire`
  records a `skipped: no-subscriber` run + `WARN`s when it is `0`. The drop is now
  **visible** in `run_history` + the log.
- `last_check` is **deliberately not held** — it is a single global file; holding
  it would freeze the whole engine and double-fire every *other* schedule. The
  next slot fires normally (no retry, no infinite loop).
- `deliver_cron_fire` always completes (inject/enqueue errors become their own
  status string, never a throw), so a *registered* subscriber yields `handled >= 1`
  even on an inject error. `handled == 0` means specifically "cron delivery path
  unregistered" — never misattributes a delivery error as a skip.

### (③) operator visibility
- `schedule list` gains a computed `next_scheduled_fire_at` (RFC3339 UTC; `null`
  when disabled / past one-shot), mirroring the engine's cron-normalisation + tz so
  the displayed instant matches the actual firing. Answers "did it drop, or is it
  just not due yet?".

### Tests
- `emit_returns_handled_count` — true/false subs + empty bus → 0.
- `unhandled_cron_fire_records_skipped_1720` — handled==0 records skipped; handled
  fire adds no extra record.
- `list_includes_next_scheduled_fire_at_1720` — future instant present; null when disabled.
- Full bin suite **3198 passed / 0 failed** on the rebased base; clippy
  `--all-targets` (`tray` and `tray,discord`) + fmt clean.

Closes #1720.

---
@codex review — **MED+ adversarial**. Inline comments only. Focus:
1. **handled semantics** — is `handled = matched-EventKind && deliver-ran` correct
   for every one of the 12 ripple subscribers? Any `handle_event` that should
   return `true` but returns `false` (or vice-versa)?
2. **(Y) completeness** — can a genuinely-dropped cron fire ever escape the
   `note_unhandled_cron_fire` recording path?
3. **misjudgement** — can `handled == 0` ever fire for a *delivered-but-errored*
   cron (false "skipped")? Verify `deliver_cron_fire` truly never throws.
4. **no infinite retry / no engine freeze** — confirm `last_check` advances
   unconditionally and a skip cannot wedge or double-fire other schedules.

### DIFF
--- a/src/daemon/anti_stall.rs
+++ b/src/daemon/anti_stall.rs
@@ -221,7 +221,7 @@ fn deliver_stall(
 /// #event-bus first-pattern: bus subscriber — on a `TaskStateChanged` event,
 /// run the stall delivery (the gate-ON path). Registered once at daemon startup
 /// via [`register_subscriber`].
-fn handle_event(event: &crate::daemon::event_bus::Event) {
+fn handle_event(event: &crate::daemon::event_bus::Event) -> bool {
     if let crate::daemon::event_bus::EventKind::TaskStateChanged {
         task_id,
         title,
@@ -240,6 +240,9 @@ fn handle_event(event: &crate::daemon::event_bus::Event) {
             *eta_secs,
             assignee.as_deref(),
         );
+        true
+    } else {
+        false
     }
 }
 
--- a/src/daemon/ci_watch/poller.rs
+++ b/src/daemon/ci_watch/poller.rs
@@ -1632,7 +1632,7 @@ fn deliver_ci_watch(
 /// #event-bus (ci_watch) subscriber: re-deliver a `CiReady`/`CiFail` event via the
 /// shared `deliver_ci_watch`. Both kinds deliver identically (the body already
 /// encodes pass/fail); the kind split is purely semantic.
-fn handle_event(event: &crate::daemon::event_bus::Event) {
+fn handle_event(event: &crate::daemon::event_bus::Event) -> bool {
     match &event.kind {
         crate::daemon::event_bus::EventKind::CiReady {
             target,
@@ -1649,8 +1649,9 @@ fn handle_event(event: &crate::daemon::event_bus::Event) {
             ..
         } => {
             deliver_ci_watch(&event.home, target, body, correlation_id, supersede_token);
+            true
         }
-        _ => {}
+        _ => false,
     }
 }
 
--- a/src/daemon/conflict_notify.rs
+++ b/src/daemon/conflict_notify.rs
@@ -233,7 +233,7 @@ fn deliver_conflict_alert(home: &Path, agent: &str, escalation: bool, text: &str
 
 /// #event-bus (conflict_notify) subscriber: re-deliver a `ConflictAlert` event via
 /// the shared `deliver_conflict_alert`.
-fn handle_event(event: &crate::daemon::event_bus::Event) {
+fn handle_event(event: &crate::daemon::event_bus::Event) -> bool {
     if let crate::daemon::event_bus::EventKind::ConflictAlert {
         agent,
         escalation,
@@ -241,6 +241,9 @@ fn handle_event(event: &crate::daemon::event_bus::Event) {
     } = &event.kind
     {
         deliver_conflict_alert(&event.home, agent, *escalation, text);
+        true
+    } else {
+        false
     }
 }
 
--- a/src/daemon/cron_tick.rs
+++ b/src/daemon/cron_tick.rs
@@ -121,7 +121,7 @@ pub fn check_schedules(home: &Path) {
         // runs the effect via `deliver_cron_fire`. The schedule message/label are
         // STATIC (non-time-sensitive), carried verbatim. (The subscriber resolves
         // the live registry it was registered with — see `register_subscriber`.)
-        crate::daemon::event_bus::global().emit(
+        let handled = crate::daemon::event_bus::global().emit(
             home,
             crate::daemon::event_bus::EventKind::CronFire {
                 sched_id: sched.id.clone(),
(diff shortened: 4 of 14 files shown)


suzuke:

#1720 VISIBILITY half — (B) bus handled-count + (Y) cron records `skipped: no-subscriber` when handled==0 (no longer silent) + (③) next_scheduled_fire_at. No findings. Windows CI was a flake (runtime_config + worktree_pool::release_* — code this PR doesn't touch); passed on rerun.

⚠ NOTE: this is the VISIBILITY half of #1720. The actual DELIVERY regression (app-mode never registers event-bus subscribers → all 13 bus patterns silently dropped post-cutover) is fixed separately (app-mode subscriber registration, incoming PR). #1720 stays OPEN until both land.

CI: 3-platform Check (after rerun) + LOC + audit 全綠。Merging via --admin --squash.

PR #1724, opened 2026-06-03
feat: :config palette command + show_pane_state opt-out + dynamic config key list

suzuke:
Three operator-facing config improvements in one PR (per dispatch).

## 1. Palette `:config get/set/list` (`src/app/commands.rs`)
Adds a `config` arm to the command palette, extracted into a testable `handle_config_command(parts, home)` that mirrors the `config` MCP tool path (`runtime_config::get_key`/`set`/`list`). The operator can now toggle runtime config straight from the TUI — no MCP-tool name to remember.
- `:config get <key>` · `:config set <key> <value>` · `:config list`
- `set` re-splits `parts[2]` on the first space, since the palette parser's `splitn(3, ' ')` glues `"<key> <value>"` into one token.
- Feedback goes through `tracing` (info on success, warn on error), matching the sibling `:status` command.

## 2. `show_pane_state` default → true (opt-out) (`src/runtime_config.rs`)
The `[<State>]` pane badge becomes a steady diagnostic by default. `#[serde(default)]` → `#[serde(default = "default_true")]` so configs predating the field also deserialize to `true`, and `Default` is `true`.
- **Note:** a config with an *explicit* `show_pane_state: false` keeps it — serde's `default` only fills *absent* fields. That's the correct honour-operator-intent semantics; fresh installs and pre-field configs get `true`.

## 3. config MCP tool description (`src/mcp/tools.rs`)
The hand-maintained `Keys: …` list was stale — it listed 3 of 7 keys and omitted `show_pane_state` (among others). It's now built from `runtime_config::keys()`, derived from the serialized `Default`, so it can never go stale as keys are added.

## Tests
- `show_pane_state_default_on_and_toggleable` — default ON + runtime-toggleable round-trip.
- `keys_cover_all_settable_keys_including_show_pane_state` — `keys()` includes `show_pane_state` and every reported key is `get_key`-able (keys ≡ valid set targets).
- `config_set_parse_shape_glues_key_value` — pins the `splitn(3)` + `split_once` parse so a future split-width change can't silently break `set`.
- `config_set_via_palette_persists_to_runtime_config` — `:config set` reaches `runtime_config::set` and persists (asserts the written file, deterministic).

## Verification
- `cargo fmt --check` ✅ · clippy `tray` + `tray,discord` `-D warnings` ✅
- full bin test suite: **3198 passed, 0 failed** · `file_size_invariant` (#1463) ✅

Base: `dd12c04` (origin/main). Daemon-side + TUI change → takes effect on rebuild/restart.

### DIFF
--- a/src/app/commands.rs
+++ b/src/app/commands.rs
@@ -1,5 +1,5 @@
 //! Command-palette command execution (`:spawn`, `:kill`, `:restart`, `:layout`,
-//! `:send`, `:broadcast`, `:status`).
+//! `:send`, `:broadcast`, `:status`, `:config`).
 //!
 //! Input is a whitespace-split command line. Returns `true` iff the command
 //! mutated the layout in a way that requires a resize pass.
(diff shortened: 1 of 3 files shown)


suzuke:

(1) palette `:config get/set/list` (2) show_pane_state opt-out (default-true; explicit false respected) (3) config MCP tool description from runtime_config::keys() (never stale). No findings.

### Evidence
- cited: app/commands.rs:344-365 — :config set parsed via splitn(3) + re-split, reaches runtime_config::set (same as MCP tool)
- cited: runtime_config.rs:40-47 — show_pane_state #[serde(default="default_true")] → missing=true, explicit false stays false
- cited: runtime_config.rs:225-234 — keys() derived from serialized Default → stays in sync
- cited: mcp/tools.rs:295-300 — config tool description formats from keys() → can't drift stale

CI: 3-platform Check + LOC + audit 全綠。Merging via --admin --squash.

PR #1725, opened 2026-06-03
fix: register event-bus subscribers in app mode (cutover regression — silent delivery drop)

suzuke:
## Live delivery regression — app mode dropped every event-bus emit

The #1719 cutover ("legacy-zero": the bus is the **sole** delivery path) registered
the per-pattern subscribers **only** in `daemon::run_core`. But owned `agend-terminal
app` mode **never calls `run_core`** (documented at `supervisor.rs:3251`: *"In APP mode
(`agend-terminal app`), `run_core` is never called"*). So in app mode the process-global
bus had **zero subscribers** and every `emit` fanned out to nothing — silently dropped.
The app tick still calls `check_schedules` (`app/mod.rs:705`), which post-cutover **only
emits** `CronFire` → straight into the void.

### Runtime-confirmed (live daemon, PID `agend-terminal app`)
- One-shot schedule: `run_at` in the past, target a **running** agent, `enabled: true`.
- Result: no `schedule triggered` log, **empty `run_history`**, schedule **stayed enabled**,
  while `.schedule_last_check` kept advancing → the tick ran, `classify_once` returned due,
  `emit` fired, and it reached **no subscriber**.
- **Zero cron fired** after the app-mode restart (the daily 早報/晚報/AI-scout schedules).

This is the live silent-drop behind #1720/#1723. #1723 makes it **visible** (records
`skipped: no-subscriber`); **this PR restores delivery.** Complementary, not a substitute.

### Regression class
Same as #1002 (`pr_state::scan_and_emit`) / #982 (idle notifications): *"wired only in
run_core, broke in app mode."* Those each added an app-loop call **+ a source-pin
invariant**. The #1719 cutover repeated the bug with **no** invariant — hence this fix
adds one.

### Scope: broad
All 13 bus patterns were dead in app mode (cron, dispatch_idle, idle_watchdog, supervisor
member-state, conflict_notify, poll_reminder, decision_timeout, waiting_on_stale,
helper_staleness, cascade-cancel, **ci-ready/ci-fail handoff**, …). Cron is the
runtime-confirmed casualty; `ci_watch`'s **poll** survived only because `app/mod.rs:706`
calls `check_ci_watches` inline — but its bus-emitted success/failure notifications dropped
too (so the `[ci-ready-for-action]` fleet handoff was also dead in app mode).

### Fix
- Extract `run_core`'s registration block → shared `daemon::register_event_subscribers(registry)`
  — the **single** subscriber list.
- `run_core` calls it; **`app::run_app` also calls it** (owned mode, after the registry is
  built, before the maintenance tick).
- `event_bus::register_all_subscribers_for_test` routes through the **same** fn — collapses
  the three drifting copies of the list (the drift that masked this bug: the test helper
  registered cron, the live app path did not).

### Tests (the guard that was missing)
- `cron_tick::global_bus_cron_subscriber_delivers` — emits a `CronFire` on the **process-global**
  bus (NOT a local test bus — that local-bus convenience is the blind spot that let this ship
  green) and asserts it reaches `deliver_cron_fire`. **Negative-probed**: drop the cron
  registration → this test fails (verified).
- `app::run_app_registers_event_bus_subscribers` — source-pin that `run_app` wires the
  subscribers (mirrors the #1002/#982 anti-regression pattern).
- `daemon::register_event_subscribers_lists_every_pattern` — source-pin that all 12 per-pattern
  registrations are present.

Full bin suite **3201 passed / 0 failed**; clippy `--all-targets` (`tray` and `tray,discord`)
+ fmt clean. ⚠ **Requires a daemon restart to take effect.**

Note: signature is `register_event_subscribers(registry: &AgentRegistry)` — `home` is not
needed (subscribers are home-agnostic; the home travels on each event), so it was omitted.

---
@codex review — **MED+ adversarial**. Inline comments only. Focus:
1. **app mode really registers now** — does `run_app`'s call actually execute on the owned
   boot path (gated correctly to `!attached_mode`, before the tick that emits)? Any path
   where it's skipped?
2. **invariant has teeth** — would `register_event_subscribers_lists_every_pattern` /
   `run_app_registers_event_bus_subscribers` actually fail if the wiring regresses, or are
   they tautological (e.g. matching their own test source)?
3. **no double-registration** — app mode previously called `check_ci_watches` inline; with
   the ci_watch subscriber now also registered, is there any double-delivery (vs run_core's
   proven inline-tick + subscriber split)? Timing/ordering correct?
4. **functional test isn't masked by the ctor** — does `global_bus_cron_subscriber_delivers`
   genuinely exercise `register_event_subscribers`, given the ctor pre-registers on the
   global bus?

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -360,6 +360,16 @@ fn run_app(terminal: &mut DefaultTerminal, fleet_override: Option<&Path>) -> Res
         })
         .ok();
 
+    // #event-bus: owned `app` mode never calls `daemon::run_core`, so it must
+    // register the bus subscribers itself — otherwise the maintenance tick below
+    // emits `CronFire` / `CiReady` / idle nudges into a bus with ZERO subscribers
+    // and every delivery silently drops (the live #1720 cron silent-drop; same
+    // regression class as #1002 / #982). Mirrors run_core; gated to owned mode
+    // because the attached daemon process owns delivery when attached.
+    if !attached_mode {
+        crate::daemon::register_event_subscribers(&registry);
+    }
+
     // Periodic maintenance tick (10s) — mirrors daemon tick cadence.
     // Only active in owned (non-attached) mode; when attached, the daemon
     // process handles schedules, CI watches, and health decay.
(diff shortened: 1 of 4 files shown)


suzuke:

🔴 CRITICAL cutover-regression fix — app-mode never registered event-bus subscribers (registration only in run_core, which app mode doesn't call) → all 12 bus patterns silently dropped post-#1719. Shared register_event_subscribers (run_core + app::run_app + test harness, single source of truth) + source-pin invariants + functional test through the REAL process-global bus + negative-probe. No findings.

### Evidence
- cited: app/mod.rs:363-371 — owned app mode registers subscribers BEFORE the tick loop; attached mode skips (correct — attached daemon owns delivery)
- cited: daemon/mod.rs:410-423,439-443 — register_event_subscribers shared by run_core AND run_app (single source of truth)
- cited: event_bus.rs:183-210 — test harness routes through the SAME production registration → test wiring can't drift from live
- cited: cron_tick.rs:672-705 — global_bus_cron_subscriber_delivers emits on the process-global bus (not local) → proves cron subscriber reachable
- cited: cron_tick.rs:1560-1589 — ci_watch CiReady/CiFail only for success/failure + legacy fallback → NO double-deliver (the key probe)

⚠ Requires restart to deploy. Fixes GAP A (event-bus). Broader per_tick handler gaps (B-F: WatchdogHandler/PrStateScan/PollReminder/InboxStuck/etc.) tracked separately (app-standalone runs only 3 of 15 handlers — pipeline-unification fix incoming).

CI green will arrive via ci_watch. Merging via --admin --squash.

PR #1726, opened 2026-06-03
feat: app-standalone runs the full per_tick pipeline (root-fix handler-drift class)

suzuke:
Root-fixes the recurring **#1002 / #982 / #1719** class. app-standalone (`run_app` owned mode) re-implemented the daemon tick as a **hand-picked subset** of per_tick handlers in its own `app_tick` loop (app/mod.rs), so every handler added to `run_core` that wasn't manually mirrored silently went missing in app mode — PrStateScan, Watchdog, PollReminder, InboxStuck, HandoffTimeout, Gc, LogRotation, InboxMaintenance, ExternalLiveness were all absent (the audit `t-appmode-wiring-audit` finding). Stacks on top of the already-merged #1725 (subscriber registration).

## Approach
- **`build_default_handlers(crash_tx)`** (daemon/mod.rs) — single source of truth for the 15-handler pipeline, shared by `run_core` and `app::run_app` (sibling of #1725's `register_event_subscribers`).
- **`app_tick_handlers()`** = `build_default_handlers` minus `APP_TICK_ALLOWLIST`; `run_app` now calls `per_tick::run_handlers_with_panic_guard` on the full set instead of the manual subset. Because app *derives* from the shared builder, a new handler **auto-propagates** to app — drift is structurally prevented, not merely caught.
- **`APP_TICK_ALLOWLIST`** = `{recovery_dispatcher, snapshot_rotation, thread_dump}` — each a justified omission (see code doc).

## Overlap map (no double-run)
Removed app_tick's manual calls, each now covered exactly once:
| removed manual call | now covered by |
|---|---|
| `cron_tick::check_schedules` | `CheckSchedulesHandler` |
| `ci_watch::check_ci_watches` | `CiWatchPollHandler` |
| `health.maybe_decay` + `check_hang` | `HangDetectionHandler` |
| `core.state.tick()` | `supervisor::spawn` (runs in owned mode too) |

**state.tick verification (per lead):** confirmed `supervisor::spawn` runs in owned mode (app/mod.rs `!attached_mode`) and its `tick()` reaches `core.state.tick()` (supervisor.rs) unconditionally per agent — so app's manual `state.tick()` was a **benign idempotent double**, now removed. Not assumed — traced.

## Allowlist rationale
- `recovery_dispatcher`: coupled to run_core's `crash_rx` + `handle_crash_respawn`, which app-standalone has no equivalent of — it does pane-based respawn. (#685 dispatch-recovery is shadow-default-off anyway.)
[...]

PR #1727, opened 2026-06-03
fix: stop dispatch stuck-check flip-flop (orphaned entries re-asked forever)

suzuke:
## Dispatch stuck-check flip-flop — `orphaned` entries re-asked every tick

`sweep_stuck` skipped only `completed`, so any entry past `DISPATCH_ORPHAN_HOURS`
flip-flopped with `sweep_orphans` and re-fired a "dispatch stuck check" on **every**
maintenance tick:

```
sweep_stuck marks it `asked`
  → sweep_orphans (24h) overwrites `asked` → `orphaned`
    → next sweep_stuck sees `orphaned != "asked"` → re-asks (re-marks `asked`)
      → sweep_orphans flips it back → … repeats for ~30 days until GC
```

The dedup (`status != "asked"`) was real, but `sweep_orphans` clobbered it.

### Runtime evidence
Live `dispatch_tracking.json`: 2184 entries — **1650 completed / 527 orphaned / 7 asked**.
The 527 `orphaned` entries each re-asked every 10-min tick → a flood of stale
stuck-checks at every recipient (e.g. 30 distinct ancient task_ids hitting `fixup-dev-2`
twice within 10 min, identical set each cycle).

### Why now
#1726 (app-standalone runs the full per_tick pipeline) activated `run_task_maintenance`
in the live **app-mode** daemon for the first time. The 527-entry backlog had silently
accumulated while app mode never swept (the same run_core-vs-app-mode gap #1720/#1725
just closed). So this is a **latent defect surfaced by the fix**, not introduced by it.

### Fix (focused)
`sweep_stuck` skips `orphaned` too — aligning its skip set with `sweep_orphans`. An
`orphaned` entry is one already given up on after 24h; never nag again, let the 30-day
GC reap it. (#1488 noted this re-fire but only fixed *deleted-instance* entries via
`cleanup_for_instance`; this fixes still-existing targets whose orphaned entries
otherwise nag for ~30 days.) Also refreshes the now-stale `cleanup_for_instance` doc
comment that claimed "sweep_stuck only skips completed".

### Test
`orphaned_entry_is_not_re_asked` — an `orphaned` entry yields **no** ask, while a
same-age `pending` entry **still asks** (normal nagging preserved). **Negative-probed**:
revert the skip → the test fails (orphaned gets re-asked).

Full bin suite **3207 passed / 0 failed**; clippy `--all-targets` (`tray` and
`tray,discord`) + fmt clean.

### Scope note
Focused on stopping the flood. The existing 527 orphaned entries are left for the
30-day GC (**no live-state edit**) — fix + restart stops new re-asks. Broader
delete-on-resolve retention is a follow-up.

⚠ **Requires a daemon restart to take effect.**

---
@codex review — **MED-tier**. Inline comments only. Focus:
1. `orphaned` is genuinely skipped now, and the skip set matches `sweep_orphans`.
2. **No over-suppression** — a normal `pending`/`warned` entry still warns/asks on
   schedule (the fix must not silence real stuck dispatches).
3. No path where a real, still-active stuck dispatch is mis-skipped because it was
   prematurely marked `orphaned`.

### DIFF
--- a/src/dispatch_tracking.rs
+++ b/src/dispatch_tracking.rs
@@ -83,7 +83,17 @@ pub fn sweep_stuck(home: &Path) -> (Vec<DispatchEntry>, Vec<DispatchEntry>) {
     persist_or_log!(
         crate::store::mutate_versioned(&store_path(home), |store: &mut DispatchStore| {
             for entry in store.entries.iter_mut() {
-                if entry.status == "completed" {
+                // Skip BOTH terminal states — mirrors `sweep_orphans`' skip set.
+                // Without skipping `orphaned`, an entry past DISPATCH_ORPHAN_HOURS
+                // flip-flops forever: `sweep_orphans` overwrites its `asked` status
+                // with `orphaned`, then this sweep sees `orphaned != "asked"` and
+                // re-asks (re-marking `asked`), which `sweep_orphans` flips back —
+                // re-firing a "dispatch stuck check" every maintenance tick until
+                // the 30-day GC. `orphaned` = already given up after 24h; never nag
+                // again. (#1488 noted this re-fire but only fixed deleted-instance
+                // entries via `cleanup_for_instance`; this fixes still-existing
+                // targets whose orphaned entries otherwise nag for ~30 days.)
+                if entry.status == "completed" || entry.status == "orphaned" {
                     continue;
                 }
                 let delegated = match chrono::DateTime::parse_from_rfc3339(&entry.delegated_at) {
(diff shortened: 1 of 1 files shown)


suzuke:
## Review Mirror (§3.12) — fixup-reviewer (codex) VERIFIED

**VERIFIED** — orphaned skip is the correct fix for the `asked`↔`orphaned` flip-flop. `sweep_orphans()` marks >24h entries `orphaned`; `sweep_stuck()` now treats `orphaned` as terminal instead of re-asking. Regression test is sharp (independently negative-probed).

### Evidence
- ran: `cargo test dispatch_tracking::tests::{test_dispatch_ask_after_30min, orphaned_entry_is_not_re_asked, test_dispatch_orphan_after_24h} --exact` → passed
- ran: negative-probe — removed the `orphaned` skip in a temp worktree → `orphaned_entry_is_not_re_asked` **FAILED** (`left: 2, right: 1`) → test has teeth
- cited: `src/dispatch_tracking.rs:85-110` — sweep_stuck skips both completed+orphaned
- cited: `src/dispatch_tracking.rs:123-143` — sweep_orphans is the orphaned producer, skips already-orphaned
- cited: `src/dispatch_tracking.rs:294-340` — regression test covers both sides (orphaned no re-ask / same-age pending still asks)
- cited: `src/dispatch_tracking.rs:398-416` — orphan-sweep boundary test intact

Adversarial 5-probe sweep (real-stuck miss / over-suppress / flip-flop-break / test-teeth / doc-accuracy) all cleared. Merge pending CI green.

PR #1735, opened 2026-06-04
refactor: upgrade twilight 0.16 → 0.17 (gateway/http/model)

suzuke:
**Supersedes dependabot #1732** (close it). #1732 bumped only `twilight-model` to 0.17, which is incompatible: `twilight-http`/`twilight-gateway` 0.16 internally pin `twilight-model 0.16`, so our 0.17 `Id` mismatched http-0.16's 0.16 `Id` (the "expected `Id<ChannelMarker>` [from twilight-model 0.16]" CI errors). Upgrading all three together — twilight only patches the latest minor, so staying on 0.16 means no further security fixes (refactor-early).

## Spike break list (controllable — finished in this PR)
- **`Cargo.toml`**: `twilight-gateway`/`http`/`model` = `"0.17"`. The 0.17 crates **removed the `rustls-ring` feature** → replaced `[rustls-native-roots, rustls-ring]` with `[rustls-native-roots]`. (Transitive: `tokio-websockets` 0.11→0.13; `twilight-gateway-queue`/`http-ratelimiting`/`validate` → 0.17.)
- **Production code: ZERO changes.** `Id<Marker>` usage and the discord send/edit/delete/keepalive paths are source-compatible across 0.16→0.17 (the dependabot "errors" were purely the version skew).
- **Tests only (4 sites):** twilight-http 0.17's ratelimiter initialises inside `Client::build()` and requires a Tokio reactor in scope. The discord spec tests that built a twilight client in a sync `#[test]` now build it within the shared discord runtime (`super::discord_runtime().block_on(...)`). Production already builds the client in async context → unaffected.

## ⚠ TLS note (for codex MED-tier review)
The rustls feature change is the only behaviour-sensitive edit. `ring` 0.17 remains the **sole** crypto provider in the tree (no `aws-lc-rs`), wired via twilight-http 0.17's `hyper-rustls`/`rustls` stack; rustls's `ring` feature supplies the default `ClientConfig::builder()` provider, so **no explicit `CryptoProvider::install_default()` is needed**. However, a **live Discord TLS handshake is NOT exercised by CI** (discord is feature-gated; tests use a mock HTTP server) — recommend a manual discord smoke test before relying on the connection.

## Verification
- `cargo fmt --check` ✅ · clippy `tray` + `tray,discord` `-D warnings` ✅
- full bin suite **3237 passed, 0 failed** · all `channel::discord` tests green · `file_size_invariant` (#1463) ✅

Base: `f259bbb` (origin/main). No restart.

[...]

PR #1736, opened 2026-06-04
fix: boot-grace for stale-backlog notification watchdogs (no restart misfire)

suzuke:
## Boot-grace for stale-backlog notification watchdogs

Consolidated fix for the latent bug surfaced by the per_tick audit (this dev's
stale-backlog lens) + reviewer-2's restart-idempotency lens. PR #1726 activated
3 notification watchdogs in app mode that were previously dormant: **PollReminder,
InboxStuck, HandoffTimeout**.

### The bug (same class as the #1727 dispatch flip-flop)
These handlers have **two** independent first-tick-after-restart misfire causes:
1. **In-memory dedup reset** — `last_alerted` / `last_escalated` / `LAST_NOTIFIED`
   live in memory, empty at boot → no record of what was already alerted.
2. **Cadence counter reset** — `counter: AtomicU64::new(0)` → `should_fire` is
   true on **tick 0**, so they fire immediately on the first tick after boot.

So on a daemon restart they re-fire for the **entire current backlog** of
unread/unclaimed items at once — and worse, flag **freshly-respawned agents** as
"stuck" before they've had a chance to drain their inbox (a false positive; the
message age survives the restart but the agent was just re-spawned).

### Fix — a single boot-grace window
A shared `in_boot_grace(created_at)` + `NOTIFICATION_BOOT_GRACE` (180s). Each of
the 3 handlers stores its construction `Instant` (≈ boot) and gates firing on
`should_run()` = `!in_boot_grace(created_at) && should_fire()`. The `&&`
short-circuits **before** `should_fire` during grace, so the counter is **not
consumed** — the first tick after grace fires immediately. One window closes all
three causes (dedup reset, counter reset, drain FP).

**180s** is comfortably longer than agent spawn + inbox drain (~tens of seconds)
and far shorter than the watchdogs' own 30-min stuck thresholds, so a genuinely
stuck item still alerts shortly after grace. Mirrors the existing precedent
`pr_state::suppress_stale_terminal_replay` (boot-suppress of stale terminal PR
events).

### Scope
Only the 3 notification handlers (per the audit + lead's direction). Cleanup
handlers (Gc / InboxMaintenance / LogRotation) are intentionally **not** touched
— their counter-reset early-cleanup is harmless (byte/content/time-based +
idempotent).

### Tests
- `in_boot_grace_active_then_expires` — helper logic + the window is a sane value
  (≥60s, ≤600s — guards against someone zeroing it).
- per-handler `boot_grace_suppresses_then_fires` — suppressed within grace,
  counter not consumed, fires on first tick after grace. **Negative-probed**:
  removing the gate fails this.
- `notification_handlers_wire_boot_grace` — source-pin that all 3 handlers gate
  on `in_boot_grace(self.created_at)`. **Negative-probed**: removing the gate from
  any handler fails this.
- Existing `scan_and_emit` tests already prove a genuine stuck pile DOES alert
  (so "fires for real stuck after grace" is covered end-to-end by composition).

Full suite (bin + all integration targets) green; clippy `--all-targets`
(`tray`, `tray,discord`) + fmt clean; no #1463 pollution.

**Deploy with the #1735 batch** so the next restart doesn't fire the burst.

---
@codex review — **standard tier**. Inline comments only. Focus:
1. The boot-grace correctly suppresses BOTH causes — in particular the `&&`
   short-circuit means the counter is not consumed during grace, so firing
   resumes promptly (not delayed by a full N-tick cycle) after grace.
2. 180s is a sound value — long enough for post-restart inbox drain, short enough
   that genuine stuck/unclaimed items still alert well before their own 30-min
   thresholds become stale.
3. No handler that SHOULD keep firing on tick 0 (e.g. a cleanup/safety handler)
   was accidentally swept into the grace — only the 3 notification watchdogs gate.

### DIFF
--- a/src/daemon/per_tick/handoff_timeout.rs
+++ b/src/daemon/per_tick/handoff_timeout.rs
@@ -7,11 +7,14 @@ use super::{PerTickHandler, TickContext};
 use parking_lot::Mutex;
 use std::collections::HashMap;
 use std::sync::atomic::{AtomicU64, Ordering};
+use std::time::Instant;
 
 pub(crate) struct HandoffTimeoutHandler {
     every_n_ticks: u64,
     counter: AtomicU64,
     last_escalated: Mutex<HashMap<(String, String), chrono::DateTime<chrono::Utc>>>,
+    /// ≈ daemon boot — drives boot-grace suppression (see [`super::in_boot_grace`]).
+    created_at: Instant,
 }
 
 impl HandoffTimeoutHandler {
@@ -20,6 +23,17 @@ impl HandoffTimeoutHandler {
             every_n_ticks,
             counter: AtomicU64::new(0),
             last_escalated: Mutex::new(HashMap::new()),
+            created_at: Instant::now(),
+        }
+    }
+
+    #[cfg(test)]
+    fn new_at(every_n_ticks: u64, created_at: Instant) -> Self {
+        Self {
+            every_n_ticks,
+            counter: AtomicU64::new(0),
+            last_escalated: Mutex::new(HashMap::new()),
+            created_at,
         }
     }
 
@@ -29,6 +43,12 @@ impl HandoffTimeoutHandler {
             .fetch_add(1, Ordering::Relaxed)
             .is_multiple_of(self.every_n_ticks)
     }
+
+    /// Boot-grace + cadence gate. `&&` short-circuits before `should_fire`
+    /// during grace, so the counter is not consumed (see `PollReminderHandler`).
+    fn should_run(&self) -> bool {
+        !super::in_boot_grace(self.created_at) && self.should_fire()
+    }
 }
 
 impl PerTickHandler for HandoffTimeoutHandler {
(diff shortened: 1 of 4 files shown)


suzuke:
## Review Mirror (§3.12) — fixup-reviewer (codex) VERIFIED @ 75c2201

**VERIFIED** — boot-grace fix for the 3 notification watchdogs (PollReminder / InboxStuck / HandoffTimeout). 180s boot-grace window suppresses only the restart-burst (in-mem dedup reset + cadence-counter reset + agent-not-yet-drained FP); a genuinely stuck item still alerts after grace. Scope limited to the 3 notification handlers; cleanup handlers untouched.

### Evidence
- ran: `cargo test --tests --features tray,discord` → passed (full integration targets)
- cited: per_tick/mod.rs:88-109 — 180s grace, scoped to the 3 handlers, finite & shorter than the 30-min stuck thresholds
- cited: poll_reminder.rs:46-61 / inbox_stuck.rs:48-66 / handoff_timeout.rs:47-65 — `!in_boot_grace && should_fire` short-circuit → counter NOT consumed during grace → first post-grace tick fires immediately
- cited: unit tests pin suppression-during-grace + first-fire-after-grace for all 3 handlers

Consolidated from dev-2 (accumulation lens) + reviewer-2 (idempotency lens) audits. Merge ready.

PR #1734, opened 2026-06-04
refactor: collapse AgentState::Ready into Idle (operator KISS decision)

suzuke:
## Collapse `AgentState::Ready` into `Idle` (operator KISS decision)

`Ready` and `Idle` were the same concept — the lowest-priority "idle, dispatchable"
state — split only because backend pattern tables labelled the idle prompt differently
(claude→Idle; agy/opencode/codex/kiro→Ready). Operator decided to collapse them: remove
`Ready`, keep `Idle`.

### Spike result (read-only, pre-impl)
The premise "Ready ≡ Idle is purely cosmetic" was **FALSE** — 3 logic sites diverged.
Operator reviewed and accepted the resulting behavior changes (all make former-Ready
backends **consistent with claude's** long-standing Idle behavior). The bulk IS cosmetic
(verified safe): `idle_watchdog` is silence-based (never reads the state); `min_hold` uses
`<= Idle.priority()` (relative, not a hardcoded value); the oscillation guard is scoped to
`{Thinking, ToolUse}`; `is_*` helpers and the supervisor/watchdog recovery sets grouped
`{Ready|Idle}` together; render colour is cosmetic.

### ⚠ ACCEPTED BEHAVIOR CHANGES
Former-Ready backends (agy/opencode/codex/kiro) now follow Idle's behavior:

| # | Site | Change |
|---|------|--------|
| ① | `poll_reminder.rs:44` (`== Idle` filter) | Idle agents are nudged about unread inbox. Former-Ready agents were **excluded**; now included. *(A fix — they have inboxes.)* |
| ② | `recovery_dispatcher.rs` `classify_branch` (`Idle => false`) | An idle agent is never "dead-likely" from silence alone. Former-Ready agents are no longer silence-reaped at their idle prompt. |
| ③ | `health.rs` `productive_silence_exceeds` (`Idle => false`) | Engine behind ②. Former-Ready agents are exempt from Hung-on-silence **at the idle prompt**. |

### ⚠ LOAD-BEARING PREMISE — please verify (③ must NOT lose real hang coverage)
My "no real hang-detection loss" conclusion rests on: **former-Ready backends actually
enter `ToolUse`/`Thinking` when working**, so a genuine hang is still caught via the
active-state 600s threshold — the idle-prompt exemption only applies when genuinely idle.
Evidence:
- **backend_profile.rs**: agy has `(ToolUse, r"●\s+[A-Z][a-zA-Z]+\(")` + `(Thinking,
[...]

PR #1737, opened 2026-06-04
fix: enforce local CI-parity in the pre-push hook (close the integration-target miss)

suzuke:
## Enforce local CI-parity in the pre-push hook

**Root cause (bit twice):** an agent runs only `cargo test --bin` — which SKIPS the
`tests/` integration targets — declares the branch CI-ready, and CI then rejects it.
#1734 (stale `"ready"` string in `tests/integration.rs`), #1735 (block_on invariant
under `tests/`). The fix is systematic, not "remember to run more": the committed
pre-push hook now **enforces** local CI-parity.

### What the hook does
`scripts/hooks/pre-push` gains a CI-parity gate (before the existing claim-verify):
- On a push whose range touches `src/` / `tests/` / `Cargo.*` / `build.rs`, it runs
  **`scripts/preflight.sh --quick`** and **blocks the push if it fails**. Delegating to
  the canonical preflight (DRY) means the commands — `cargo fmt --check`, `cargo clippy
  --all-targets --features tray -- -D warnings`, `cargo test --tests --features tray` —
  can never drift from CI's `check` job. `--quick` skips the Windows cross-check (CI's
  `windows-latest` is the backstop).
- Docs-only pushes skip the build (latency). `--no-verify` overrides (warning echoed;
  daemon gate + CI still apply). Fail-open + loud warning if `preflight.sh` is missing.

### Self-install (coexistence — no daemon change, no restart)
The daemon sets each managed worktree's `core.hooksPath` to `$AGEND_HOME/hooks`
(`binding.rs::install_hooks`) and writes **only** `prepare-commit-msg` there — it does
not clear the dir. So `scripts/install-hooks.sh` (now idempotent) **copies** the tracked
`pre-push` into `$AGEND_HOME/hooks/pre-push`, coexisting and firing for agent pushes (one
copy covers every current + future worktree — shared dir). Confirmed agent pushes run the
hook: the `agend-git` shim and the `AGEND_GIT_BYPASS` path both `exec` real `git`, which
honours `core.hooksPath`.

### Proof (manual, captured — the core deliverable)
Added a `tests/` target with a #1734-style stale assertion, then:
- `cargo test --bin agend-terminal` → **PASSED** (`0 passed; ... 3226 filtered out` — it
  skips every `tests/` target). ← the exact miss.
- `cargo test --test <target>` → **FAILED** (catches it).
- Ran the actual hook → it ran `preflight.sh --quick` → `PREFLIGHT FAILED` →
  `pre-push: CI-parity FAILED` → **HOOK_EXIT=1 (push blocked)**.
- Removed the failing test → `preflight.sh --quick` **passed (3/3)** (fmt, clippy, test).

#1735 is the same class: `block_on_runtime_guard_invariant.rs` is a `tests/` target in the
same `--tests` set the hook runs and `--bin` skips — so the same gate catches it.

### Tests
`tests/prepush_hook_invariant.rs` source-pins the gate (delegates to preflight, blocks on
fail, watches `src/`/`tests/`/`Cargo.*`, documents `--no-verify`). **Negative-probed**:
dropping the preflight call from the hook fails the pin. (It's itself a `tests/` target, so
the very gate it pins would catch a regression in it.)

### Deploy
Repo-layer workflow change — **no daemon binary change, no restart**. Activate by running
`scripts/install-hooks.sh` (post-merge). `shellcheck` not available in this env — please
run it on the two scripts in review if you have it.

---
@codex review — **standard tier**. Inline comments only. Focus:
1. The hook genuinely **blocks** a `tests/`-target failure (the #1734/#1735 class) and the
   docs-only gate doesn't accidentally skip a code push (verify the `git diff --name-only
   -- src/ tests/ Cargo.* build.rs` path detection + the new-branch `merge-base` base).
2. The coexistence claim — `$AGEND_HOME/hooks/pre-push` survives the daemon's
   `binding.rs::install_hooks` (which only writes `prepare-commit-msg`, no dir clear).
3. Shell correctness: the single `cat` stdin capture feeds both gates; `set -euo pipefail`
   interplay with `grep -q` / `|| parity` paths; `--quick` exit-code propagation from
   `preflight.sh`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -41,9 +41,22 @@ A pre-commit hook at `.git/hooks/pre-commit` auto-formats staged `.rs` files
 and re-stages them. It does NOT run clippy — clippy is too slow for a
 pre-commit path. Run clippy yourself before `git push`.
 
-A pre-push hook verifies push claims (e.g. "no other changes", "deps
-unchanged") against the actual diff. Override with `git push --no-verify`
-in emergencies.
+The pre-push hook (`scripts/hooks/pre-push`) runs **two gates**:
+
+1. **CI-parity** (#t-ci-parity-prepush-guard) — on a push whose range touches
+   `src/` / `tests/` / `Cargo.*` / `build.rs`, it runs `scripts/preflight.sh
+   --quick` (the exact CI `check` commands: `cargo fmt --check`, `cargo clippy
+   --all-targets --features tray -- -D warnings`, `cargo test --tests --features
+   tray`) and **blocks the push if they fail**. This closes the recurring miss
+   where an agent ran only `cargo test --bin` — which SKIPS the `tests/`
+   integration targets — declared CI-ready, and CI then rejected it
+   (#1734 stale-string integration test, #1735 block_on invariant). Docs-only
+   pushes skip the build. `--quick` omits the Windows cross-check (CI's
+   `windows-latest` is the backstop for that).
+2. **claim-verify** — verifies `Claim:` trailers against the actual diff.
+
+Override either with `git push --no-verify` in emergencies (the daemon-side gate
+and CI still apply, so `--no-verify` is not a free pass).
 
 A post-merge hook triggers a background `cargo build --release` when `src/`
 files change. Desktop notification on completion. Does not auto-restart the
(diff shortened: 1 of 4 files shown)


suzuke:
## Review (§3.12) — LEAD-REVIEW VERIFIED @ 75b11f9

⚠ Lead-review (separate reviewers unavailable: fixup-reviewer-2 weekly limit; fixup-reviewer noise-drowned). Verified key probes directly + dev-2's captured empirical proof + CI green.

**VERIFIED** — pre-push CI-parity guard (closes the #1734/#1735 'ran only --bin, missed tests/ integration targets' class).
- **Gate 1 CI-parity**: DRY-delegates to `scripts/preflight.sh --quick` (= `cargo test --tests --features tray` + fmt + clippy — the missed integration targets), single source so it can never drift from CI. Correctly scoped (runs only when push range touches src/|tests/|Cargo.*|build.rs; docs-only skips); base is fail-safe (unknown range → run); preflight-missing → fail-OPEN with loud warning (CI backstop). `refs=$(cat)` captures stdin once so both gates read it.
- **install-hooks.sh**: idempotent; dual-target (operator clone → core.hooksPath; fleet agents → copy pre-push into $AGEND_HOME/hooks coexisting with the daemon's prepare-commit-msg, which it doesn't clear). Coexistence assumption documented.
- **tests/prepush_hook_invariant.rs**: source-pins delegate+exit1+code-gate+override via include_str! — has teeth (negative-probed), and is itself a tests/ target so the gate it pins self-guards.

dev-2's captured proof: deliberate tests/ failure → `--bin` PASS / `--tests` FAIL / real hook → exit 1 blocks → remove → passes. CI green @ 75b11f9.
**Deploy**: run `scripts/install-hooks.sh` post-merge (repo-layer, no daemon restart).

PR #1738, opened 2026-06-04
ci: make Coverage trustworthy — smart retry (no real-failure mislabel) + harden flaky test

suzuke:
## RCA (the reframe)
Examining 3 recent Coverage "flakes": **most were NOT flakes** — they were **real test failures** the retry loop mislabelled as `"likely #1452 profraw/git-race flake"`:
- **#1735**: Coverage red was my **`block_on_runtime_guard_invariant`** bug (a source-scan test — Coverage runs `--tests` so it caught it). Deterministic; failed both retry attempts; labelled "flake".
- **dev-2's** `test_crash_respawn_health` red was likewise a real bug (since fixed in #1734; passes on `main`).
- Only **one genuine flake**: `tests/integration.rs::test_event_log_written` (passes 3/3 isolated, flakes under the full parallel `--tests` run).

The mislabel did real harm: it wasted a second slow coverage run **and** trained everyone to dismiss Coverage red as "flake" — nearly burying real bugs (only the platform `Check` caught them). **Making Coverage trustworthy matters more than killing flakes.**

## Fix
**(a) Smart retry** (`.github/workflows/ci.yml`) — retry **only** when the failure carries an llvm-cov profraw/profdata-corruption signature (`profdata`, `.profraw`, `malformed instrumentation`, `raw profile`, `invalid instrumentation profile`, `failed to load/merge … profile`, `no profile can be merged`). A clean `test … FAILED` **fails fast** with an honest label. Genuine profraw flakes still self-heal; real failures surface immediately and are trusted.

**(b) Harden the one genuine flake** (`tests/integration.rs::test_event_log_written`) — poll on the `daemon_start` **content** with the **30s** ceiling used by the sibling event-log waits (was a **3s existence-only** check). Fixes both the slow-startup timeout and the file-exists-but-empty race under the coverage job's slow, CPU-saturated parallel run. `wait_until` polls, so the higher ceiling adds zero common-case slowdown.

`test_crash_respawn_health` left as-is (real bug already fixed in #1734). **Not** doing serial (`--test-threads=1`) coverage per the lead — too slow for one flaky test.

## Verification
- **(a)** grep-logic test: real `test … FAILED` → fail-fast (no retry, no mislabel); `malformed instrumentation profile data` → retry. ✓
- **(b)** `test_event_log_written` stable **10/10** under the full parallel integration suite (where it flaked). ✓
- `cargo fmt --check` ✅ · clippy `--all-targets --features tray -D warnings` ✅ · `--tests` integration+invariant targets ✅ · `file_size_invariant` (#1463) ✅

Base: `e1b31ca`. CI change → no restart. Coverage stays non-required.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -254,22 +254,37 @@ jobs:
         timeout-minutes: 35
         env:
           AGEND_LOCK_AUDIT: "1"
-        # #1452: Coverage (non-required) recurringly flaked on transient
-        # llvm-profdata profraw corruption and a git-race test under llvm-cov's
-        # slow parallel run — both pass on rerun. Retry once (clean stale
-        # profraw between attempts) so the job self-heals instead of needing a
-        # manual `gh run rerun`. The git-race root is separately hardened by
-        # disabling auto-gc in the affected test's repo setup.
+        # #1452: Coverage (non-required) can flake on transient llvm-profdata
+        # profraw corruption under llvm-cov's slow parallel run — that passes on a
+        # clean rerun, so retry once.
+        #
+        # SMART RETRY: a plain `test ... FAILED` is a REAL, deterministic failure.
+        # The old loop retried it unconditionally AND labelled it "likely flake" —
+        # which wasted a second slow coverage run and, worse, MASKED real bugs as
+        # flakes (e.g. #1735's block_on invariant surfaced here and was dismissed
+        # as a Coverage flake). So: only retry when the failure carries an
+        # llvm-cov profraw/profdata-corruption signature; otherwise fail FAST with
+        # an honest label so the real `test ... FAILED` is trusted, not ignored.
         run: |
-          for attempt in 1 2; do
-            if cargo llvm-cov --workspace --tests --features tray --lcov --output-path coverage.lcov; then
+          set -o pipefail
+          attempt=1
+          while :; do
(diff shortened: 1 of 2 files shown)


suzuke:
## Review (§3.12) — LEAD-REVIEW VERIFIED @ 6e47787

⚠ Lead-review (separate reviewers unavailable: fixup-reviewer-2 hit weekly usage limit; fixup-reviewer noise-drowned). Verified the key probes directly + dev's bidirectional evidence + CI green.

**VERIFIED** — Coverage trustworthiness fix.
- (a) **smart-retry**: the corruption-signature grep `profdata|\.profraw|malformed instrumentation|raw profile|invalid instrumentation profile|failed to (load|merge).*profile|no profile can be merged` is composed entirely of llvm-cov/profdata-specific terms — a plain `test ... FAILED` + assertion panic does NOT match → real failures **fail fast** (no longer masked as flake, which was hiding #1735's block_on bug); genuine profraw corruption still retries once. `set -o pipefail` + `tee` correct; max 1 retry preserved.
- (b) **harden test_event_log_written**: polls on CONTENT (daemon_start) with a 30s ceiling; `wait_until` polls so no common-case slowdown; still asserts the real daemon_start semantic (not relaxed to always-pass).

CI green @ 6e47787. (Note: #1738's own Coverage job runs main's old retry — pull_request uses base workflow — so smart-retry takes effect on merge; grep logic verified by inspection + dev's bidirectional local test.)

PR #1749, opened 2026-06-04
fix(#1748): drive agend-git shim-test fixtures through real git (no worktree corruption)

suzuke:
## #1748 — agend-git shim-test footgun corrupts the caller's worktree

The shim tests `has_unmerged_files_false_on_clean_repo` and
`has_unmerged_files_true_on_conflict` build their scratch repos with bare
`Command::new("git")` + `-C <tempdir>`. On PATH, `git` resolves to **this
agend-git shim** (`$AGEND_HOME/bin/git`), whose #1463 ChdirPass strips the
`-C <tempdir>` global-target override and redirects the op onto the caller's CWD.

Run from a fleet worktree — e.g. the #1737 pre-push hook's `cargo test --tests` —
that CWD is a bound worktree, so the fixture `git init` / `commit` / `checkout`
(base/side/main + `f.txt`) land on the **worktree's branch** instead of the
tempdir: HEAD hijacked → scratch commits, branch switched to a scratch `main`,
local `main` ref corrupted, garbage tree pushed. Hit live while pushing #1740.

### Fix (test-side only; #1463 shim redirect untouched)
Route the fixture git calls — **and** the `has_unmerged_files(git, …)` binary arg,
which itself spawns git — through `resolve_real_git()`, the same resolver the
shim uses to exec real git (`AGEND_REAL_GIT` → `which` excluding `$AGEND_HOME/bin`
→ `/usr/bin/git`). Real git honors `-C <tempdir>`, so the ops stay in the tempdir
regardless of CWD.

`count_commits_above_base` / `rev_parse` already set `AGEND_GIT_BYPASS=1` (real
git via shim passthrough) — they were never the footgun; the 2 `shim_push_*`
failures were collateral from the has_unmerged pollution and pass again once it's
fixed.

### Acceptance gate (before dirty / after clean)
**Before** (witnessed in #1740): running these tests from a bound worktree
hijacked HEAD/branch and corrupted local `main`.
**After** — ran the 4 tests from a bound worktree, then `git status`:
```
test result: ok. 4 passed; 0 failed   (has_unmerged_files_false/true, shim_push_cleans/cleanup_noop)
git status: <empty>   HEAD/branch unchanged
```
Full agend-git suite (50 tests) from the bound worktree also leaves `git status`
clean. clippy `--all-targets --features tray` + fmt clean.

Refs #1748.

---
@codex review — **standard tier**. Inline only. Focus:
[...]

Issue #1741, opened 2026-06-04
inject path: latent boot-window stranding hole + pane-input-not-submitted detector is blind to daemon injects (RCA reversed the premise)


suzuke:
## Context
Investigated 'inject-without-submit' (operator saw a system message in fixup-lead's input line, unsent, post-restart). A 9-agent RCA (fixup-dev, t-inject-without-submit-rca; 6 investigators + 3 adversarial refuters + log-verified) **reversed the premise**. Findings below; the fix is **DEFERRED pending careful review** (sensitive #1457/#1473/#1513 draft-gating path; the real hole is currently *unobserved*).

## Finding 1 — the observed detections are OPERATOR DRAFTS, not stranded daemon injects (detector is blind)
`check_pane_input_not_submitted` (supervisor.rs:~349) reads only `typed_ms`/`submit_ms`, which are written **only** by the operator-keyboard path (app/mod.rs:1081/1115 record_input/submit_activity). The daemon inject path (`inject_with_target`, agent/mod.rs:1923-1976) **does not touch them** → the detector is **structurally blind to daemon injects**. Every `pane-input-not-submitted` firing is by definition an operator-typed-unsubmitted draft. Log evidence: detections trace to typed-times *before the agent's spawn* (e.g. fixup-lead typed 14:33:33, spawned 14:35:27; fixup-dev ~24h) = persistent operator drafts; on restart the in-memory dedup (`last_emitted_for_typed_ms`) resets → the same old `typed_ms` is re-emitted. ('post-restart' is a dedup-reset artifact, not a new inject.)

## Finding 2 — REAL latent hole (UNOBSERVED): boot-window inject can strand
The inject's only state gate is `handle_inject`'s `is_unavailable()` = **Crashed|Restarting only** (state/mod.rs:131). Post-restart spawn starts at **Starting** (state/mod.rs:616), NOT gated → boot-window injects are allowed. `Starting→Idle` flips on chrome-marker detection (❯ / 'bypass permissions', backend_profile.rs:319), which renders **before the CLI input loop actually accepts keystrokes** → a daemon inject's body + (fixed 50ms later) the lone `\r` (agent/mod.rs:1974-1975) can be eaten by boot re-render → strand. **No log evidence this has fired** — latent.

## Finding 3 — other potential submit-drops (verified paths)
- body-write timeout/guard `?` aborts before the submit write (medium)
- deleted-flag flips true between body and submit writes → returns Ok() skipping submit (low)

[...]

PR #1754, opened 2026-06-04
fix(#1741): boot-grace the pane-input-not-submitted diagnostic

suzuke:
## What
Boot-grace the every-tick `pane-input-not-submitted` diagnostic so a daemon restart's freshly-zeroed in-memory dedup map (`pane_input_tracks`) can't re-emit the diagnostic for an input typed *before* the restart (typically a pre-existing operator draft).

This is **①a** of the in-memory-tracking-resets-on-restart root-fix (`t-pattern-inmem-reset` master analysis), the **#1741** instance of that class.

## Why
`check_pane_input_not_submitted` runs every 10s tick. Its dedup (`PaneInputTrack.last_emitted_for_typed_ms`) is in-memory only → a daemon restart zeroes it → within the first ticks the diagnostic re-fires for a pre-restart input. The detector reads operator-keystroke timestamps only (`last_input_epoch_ms`/`last_submit_epoch_ms`), so it **cannot distinguish a pre-existing operator draft from a fresh strand** (see the inject-without-submit RCA) — making the post-restart re-emit a false-refire.

## How
- Capture one `loop_started_at = Instant::now()` at supervisor `run_loop` start (≈ daemon boot; set once, never reset).
- Gate `check_pane_input_not_submitted[_for_agents]` with `crate::daemon::per_tick::in_boot_grace(loop_started_at)` — the existing #1736 `NOTIFICATION_BOOT_GRACE` (180s) helper, **reused in place** (no daemon/mod.rs move, per lead 拍板).
- A still-stranded input re-emits exactly once after the grace ends (its `typed_ms` is well past the 60s threshold by then).

**Scope:** only the frequent (every-tick) tracker. The slow 5-min-scan sibling watchdogs (anti_stall / idle_watchdog / …) first-scan *after* the 180s grace window → boot-grace is ineffective there → boot-seed/accept, handled separately.

## Test
- New `pane_input_not_submitted_suppressed_during_boot_grace`: seeds a stranded input → asserts **no emit + no dedup-map mutation** within grace, then **exactly one emit** after grace.
- 5 existing pane_input tests updated to pass a past `loop_started_at` (detection still runs).
- Preflight: fmt + clippy (`tray`, `tray,discord`, -D warnings) + full `cargo test --tests` all green (the agend-git shim test passes on this base via #1749).

### DIFF
--- a/src/daemon/supervisor.rs
+++ b/src/daemon/supervisor.rs
@@ -275,6 +275,14 @@ fn run_loop(
     let mut dispatch_idle_fixup_nudge_tracker =
         crate::daemon::dispatch_idle::fixup_nudge::DispatchIdleFixupNudgeTracker::default();
     let mut retention_supervisor = crate::daemon::retention::RetentionSupervisor::default();
+    // #1741 boot-grace anchor: this Instant ≈ supervisor/daemon boot (set once,
+    // never reset). It gates the every-tick pane-input diagnostic so a restart's
+    // freshly-empty `pane_input_tracks` dedup map can't re-emit for inputs typed
+    // BEFORE the restart (typically a pre-existing operator draft). Mirrors the
+    // #1736 `NOTIFICATION_BOOT_GRACE` suppression already used by the notification
+    // watchdogs; the slow (5-min-scan) sibling watchdogs are out of scope here
+    // because their first scan lands after the 180s grace window.
+    let loop_started_at = Instant::now();
     loop {
         thread::sleep(TICK);
         // #1125 M1: wrap the entire tick body in catch_unwind so a panic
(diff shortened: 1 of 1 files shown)


suzuke:
✅ **Review mirror — fixup-reviewer (codex) VERIFIED + CI all green.**

The pane_input boot-grace suppresses the pre-restart pane-input-not-submitted diagnostic re-emit during the 180s boot-grace (pure/read-only during grace — no metadata read or dedup mutation), then emits once after grace if still stranded. `loop_started_at` set once before the loop; only this every-tick tracker is gated; slow-scan trackers untouched. Evidence: 6 pane_input + 6 boot_grace tests pass; supervisor.rs:278/300/380/416/3558. CI all required green (3-platform + Coverage + LOC + audit).

Merging via §3.12 admin-squash. Partially addresses #1741 (the diagnostic-re-emit-on-restart facet).

Issue #1745, opened 2026-06-04
event_bus core: subscriber callbacks run under the bus mutex + no per-subscriber panic isolation (latent hardening)


suzuke:
Found by post-refactor event-bus core code-review (fixup-reviewer/codex, t-codereview-eventbus-core). **Lead-verified** (event_bus.rs:247-251).

## Finding (LOW — latent, no current repro)
`EventBus::emit` invokes every subscriber callback **while holding the subscribers mutex** and has **no per-subscriber panic isolation**:
```rust
// src/daemon/event_bus.rs:245-251
pub fn emit(&self, home: &Path, kind: EventKind) -> usize {
    let event = Event::new(home.to_path_buf(), kind);
    let subs = self.subscribers.lock();                       // :247 lock held...
    let handled = subs.iter().filter(|sub| sub(&event)).count(); // :248 ...across ALL callbacks
    ...
}
```
Two latent fragilities:
1. **Re-entrant deadlock**: a future subscriber that calls `emit`/`subscribe` (re-enters the bus) would re-lock the same mutex → deadlock. No current subscriber re-enters in the audited paths, so there's no current production repro.
2. **No panic isolation**: a panicking subscriber aborts the `.count()` fan-out → the remaining subscribers for that event don't run. The per-tick daemon runner already wraps each handler in `catch_unwind` (`per_tick/mod.rs:200-204`) for exactly this isolation class; EventBus does not.

This is a core-quality hardening issue, not a current parity break (delivery is verified correct).

## Fix
Drop the guard before invoking callbacks (snapshot the subscriber list under the lock, then iterate outside it) AND wrap each `sub(&event)` in `catch_unwind` (mirror per_tick/mod.rs:200-204) so one panicking/slow subscriber can't deadlock or abort the fan-out.

## Verified-clean (same review)
No missing-subscriber bug (daemon + owned-app both route through register_event_subscribers; source-level completeness pin for the 12 patterns); emit is post-cutover unconditional (no gate-flip/half-route split); handled-count semantics correct; subscriber Errs are logged inside the subscriber, not propagated.

suzuke:
Fixed by #1756 (merged): EventBus::emit snapshots subscribers (Arc clone) and drops the lock before invoking callbacks, with per-subscriber catch_unwind isolation — mirrors the per_tick handler isolation.

PR #1756, opened 2026-06-04
fix(#1745): harden event_bus emit (snapshot subscribers + per-callback catch_unwind)

suzuke:
## #1745 — harden `EventBus::emit` (re-entrant deadlock + panic-aborts-fan-out)

`EventBus::emit` invoked every subscriber callback **while holding
`subscribers.lock()`** with no panic isolation (`event_bus.rs:247-248`). Two
failure modes:

- **Re-entrant deadlock** — a subscriber that calls `emit`/`subscribe` re-locks
  the non-reentrant `parking_lot::Mutex` → the daemon hangs.
- **Panic aborts fan-out** — a panicking subscriber unwinds out of `emit`, so the
  remaining subscribers never receive the event (and the lock risks poisoning).

### Fix (mirrors `per_tick/mod.rs` handler isolation)
Snapshot the subscriber list under the lock (`Subscriber = Arc<dyn Fn>`, so the
`Vec` clone is a cheap refcount bump), **drop the guard**, then iterate lock-free
wrapping each `sub(&event)` in `catch_unwind(AssertUnwindSafe(…))`. A panic is
logged via `tracing::error!` and the fan-out continues to the rest.

**handled-count semantics unchanged**: only a subscriber returning `true` is
counted; a panicking subscriber counts as not-handled.

### Tests
- `panicking_subscriber_does_not_abort_fan_out` — the subscriber registered after
  the panicker still runs; handled-count excludes the panicker. **Negative-probed**:
  reverting to the lock-held / no-catch_unwind form fails (the panic propagates).
- `reentrant_subscribe_during_emit_does_not_deadlock` — a subscriber that registers
  another subscriber completes (pre-fix this dead­locks the non-reentrant mutex).

All 7 event_bus tests green; clippy `--all-targets --features tray` + fmt clean.
Single-file change, no drift.

Refs #1745.

---
1. The `subscribers.lock()` guard is dropped before any `sub(&event)` runs (the
   `.clone()` temporary releases it), so re-entrant `emit`/`subscribe` can't
   deadlock.
2. `catch_unwind` isolates each subscriber and handled-count still only counts
   `Ok(true)` — a panic is neither counted nor allowed to abort the iteration.

### DIFF
--- a/src/daemon/event_bus.rs
+++ b/src/daemon/event_bus.rs
@@ -244,8 +244,37 @@ impl EventBus {
     /// vanish silently (#1720).
     pub fn emit(&self, home: &std::path::Path, kind: EventKind) -> usize {
         let event = Event::new(home.to_path_buf(), kind);
-        let subs = self.subscribers.lock();
-        let handled = subs.iter().filter(|sub| sub(&event)).count();
+        // #1745: snapshot the subscriber list under the lock, then DROP the guard
+        // before invoking any callback. Running `sub(&event)` while holding the
+        // lock (a) deadlocks a re-entrant subscriber — one that itself calls
+        // `emit`/`subscribe` — and (b) lets a panicking subscriber abort the
+        // fan-out (and risk poisoning). Subscribers are `Arc<dyn Fn>`, so cloning
+        // the Vec is a cheap refcount bump.
+        let subs: Vec<Subscriber> = self.subscribers.lock().clone();
+        let mut handled = 0usize;
+        for sub in &subs {
+            // #1745: isolate each subscriber — a panic in one must not abort the
+            // fan-out to the rest (mirrors the per_tick handler isolation). The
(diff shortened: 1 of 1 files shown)


suzuke:
✅ **Review mirror — fixup-agy (primary, Gemini 3.1 Pro High) VERIFIED + lead cross-checked + CI green.** EventBus::emit now snapshots the subscriber Vec (Arc clone) under the lock, drops the guard, then iterates outside with per-callback catch_unwind — fixes (a) re-entrant deadlock (a subscriber calling emit/subscribe no longer re-locks the held mutex) and (b) panic-aborts-fan-out (a panicking subscriber no longer kills delivery to the rest). handled-count semantics unchanged (only Ok(true) counts). Negative-probe has teeth (reverting to lock-held/no-catch_unwind fails the new panic + reentrant tests); 7 event_bus tests green; CI all required green. Merging via §3.12 admin-squash. Closes #1745.

PR #1753, opened 2026-06-04
fix(#1739): boot-seed supervisor-loop producers (no restart boot-burst)

suzuke:
## #1739 — boot-seed the supervisor-loop producers (no restart boot-burst)

Five supervisor-loop producers keep an **in-memory** dedup map that resets on
restart, so the first scan after a fresh daemon start re-fired for the **entire
existing backlog** — re-flooding operators with alerts they already saw.

Boot-grace (#1736's mechanism) is **insufficient here**: these producers' first
scan lands ~5 min in (`TICK 10s × TICKS_PER_SCAN 30`), past any short grace, and
the backlog (stall / conflict / idle / stale) **persists across the restart** —
so suppressing scans during a grace window only *delays* the burst.

### Fix (mirrors `pr_state::suppress_stale_terminal_replay`)
Each tracker gains a `seeded` latch. The **first** scan records every currently-
existing backlog item into the dedup (stamped now) but does **not** emit —
restart-existing items are treated as already-known. Only items that newly appear
after boot (or persist past the per-producer re-alert interval) emit. Pure
in-memory; **no new persistence**.

| producer | seeded state | scope notes |
|---|---|---|
| `anti_stall` | `last_emitted_at` | dropped the obsolete "intended restart re-surface" comment (operator greenlit suppressing it) |
| `waiting_on_stale` | `last_alerted_at` | |
| `helper_staleness` | `last_alerted_at` | |
| `idle_watchdog` | `last_alerted_at` (**dev vantage only**, via `scan_dev_vantage`) | fleet vantage keeps its persisted snooze |
| `conflict_notify` | `last_conflict_at` (**first-observation notify only**) | escalation / `last_escalated_at` is out of scope (#1739 ②, reviewer-2). **Option (a)**: seeding stamps `last_conflict_at = now`, so the 30-min staleness escalation clock starts at boot |

### Tests
Each producer has a boot-seed test: the seeding scan records the existing backlog
**without** emitting, and the next scan does **not** re-burst. **Negative-probed** —
removing the `if !seeding` gate makes the seeding scan emit (test fails); verified
on `anti_stall` + `conflict_notify`.

`cargo test --features tray` (5 boot-seed + 94 producer-module tests) green; clippy
`--all-targets --features tray` + fmt clean. Rebased on latest main.

Closes #1739.

---
@codex review — **standard tier**. Inline only. Focus:
1. The `seeded` latch flips exactly once and the seeding scan records the dedup
[...]

PR #1758, opened 2026-06-04
fix(#1741): boot-grace notify_tracks + apierror_episodes (①a PR-2)

suzuke:
## What
**①a PR-2** of the in-memory-tracking-resets-on-restart root-fix (`t-pattern-inmem-reset`). Boot-grace the two remaining **frequent (every-tick)** supervisor trackers so a daemon restart's reset in-mem dedup/cooldown state can't re-fire. Follows PR-1 (#1754, pane_input) and reuses the same `#1736 in_boot_grace` helper + `#1741 loop_started_at` anchor.

## Why
On restart, `notify_tracks` (the 60s member-state-change NOTIFY_COOLDOWN) and `apierror_episodes` (the per-episode nudge dedup) reset → a pre-existing error state's re-detected edge re-notifies the orchestrator / re-nudges the agent. These are every-tick (10s) trackers, so the spurious re-fire lands well inside the 180s boot-grace window (unlike the 5-min-scan watchdogs, which are out of scope → boot-seed/accept).

## How — gate only the EMIT, never detection / the confirm-window
- **notify_tracks** (both `maybe_notify_member_state_change` call sites in `tick()`):
  - **path 1 (general member notify)** — call-site `!in_boot_grace` guard. Edge-triggered; suppressed for the post-boot window, a genuine new edge after grace still notifies.
  - **path 2 (#1523 AuthError deferred Fire)** — extracted a pure `resolve_pending_auth` helper that **HOLDS** a `Fire` during grace (keeps the entry pending, returns `None`) so a post-restart re-page is suppressed **without losing** a genuine held-AuthError notify (it fires once grace ends). Naively gating `maybe_notify` here would *lose* the notify, because `pending_auth.remove()` runs unconditionally in the original `if let`.
- **apierror_episodes** (#1697 nudge) — gate the **Phase-1 queue decision** so no nudge is queued AND the episode is left **unmarked** during grace (the `insert` only happens in Phase 2b when a nudge actually fires) → a still-ApiError agent gets a *fresh* nudge after grace. SRL retry + the `else` re-arm are untouched.
- **self-orch AuthError P0 escalation is NOT exempt** (per lead 拍板): a boot-window AuthError is usually a boot-transient re-auth, so boot-grace avoids the false P0; a genuine one fires after grace (delayed ≤180s, never lost).

The confirm-window (`auth_error_gate`), state detection, apierror re-arm, and SRL retry are all left intact.

## Test (proves no mis-gate of detection / confirm-window)
- `resolve_pending_auth_holds_fire_during_boot_grace_1741`: Fire+grace → `None` + **pending kept**; Fire after grace → `Some` + removed; Cancel drops; Wait keeps.
- `apierror_nudge_suppressed_during_boot_grace_1741`: within grace → no nudge + **episode unmarked**; after grace → fresh nudge (`continue` injected) + episode marked.
- 12 existing `process_error_recovery` test call sites threaded with `past_boot_grace()` (so detection/SRL behavior is unchanged).
- Preflight: fmt + clippy (`tray`, `tray,discord`, -D warnings) + full `cargo test --tests` all green. Rebased onto latest main (`43f09c6`, includes reviewer-2's #1744-H2/H3) — re-verified green on the new base.

### DIFF
--- a/src/daemon/supervisor.rs
+++ b/src/daemon/supervisor.rs
@@ -140,6 +140,34 @@ fn auth_error_gate(auth_error_held: Option<Duration>) -> AuthErrorGate {
     }
 }
 
+/// #1741 boot-grace (B): apply the [`auth_error_gate`] verdict to the pending
+/// map, honoring boot-grace. Returns `Some(entry)` — removed from `pending_auth`
+/// — when the deferred operator notify should FIRE this tick.
+///
+/// During boot-grace a `Fire` is HELD: the entry stays pending and `None` is
+/// returned, so a post-restart re-page (the in-mem confirm-window restarts on
+/// restart → a pre-existing AuthError re-confirms and would re-fire ~90s in,
+/// still within the 180s grace) is suppressed WITHOUT losing the notify — it
+/// fires on a later tick once the grace ends. `Cancel` drops the pending entry
+/// (self-healed blip); `Wait` leaves it untouched. Pure + testable; the
+/// confirm-window classification (`auth_error_gate`) is never gated.
+fn resolve_pending_auth(
+    gate: AuthErrorGate,
+    in_boot_grace: bool,
+    name: &str,
+    pending_auth: &mut HashMap<String, PendingAuthError>,
+) -> Option<PendingAuthError> {
+    match gate {
+        AuthErrorGate::Fire if in_boot_grace => None,
+        AuthErrorGate::Fire => pending_auth.remove(name),
+        AuthErrorGate::Cancel => {
+            pending_auth.remove(name);
+            None
+        }
+        AuthErrorGate::Wait => None,
+    }
+}
+
 /// Sprint 54 P2-3: dedup key for `PaneInputNotSubmitted` emission.
 /// Records the `last_input_epoch_ms` of the most recent emit so the
 /// supervisor doesn't re-fire on every 10-s tick while the operator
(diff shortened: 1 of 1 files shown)


suzuke:
✅ **Review mirror — fixup-agy (primary, Gemini 3.1 Pro High) VERIFIED + lead cross-checked + CI green.** Boot-grace gates only the emit in notify_tracks (path1 + path2 resolve_pending_auth Fire-as-Wait, preserving pending_auth — no remove-before-notify loss) and apierror_episodes (Phase-1 queue decision, episode marked only on a real nudge). auth_error_gate runs BEFORE resolve_pending_auth, so detection / confirm-window / re-arm / SRL retry are untouched — only the final notify is held. (The ubuntu CI fail was a pre-existing unrelated flake — runtime_config global-singleton race #1174, fixed-separately-incoming — confirmed via log; windows passed with the cfg-gate fix; rerun green.) supervisor.rs only; 3226 tests pass. Merging via §3.12 admin-squash — completes ①a of the in-mem-reset boot-grace treatment.

suzuke:
✅ **Review mirror — fixup-agy (primary) VERIFIED (boot-grace) + lead cross-checked the #1760 integration + CI all green.** notify_tracks boot-grace gates only the emit (path1 + path2 resolve_pending_auth Fire-as-Wait, preserving pending_auth); apierror Phase-1 queue decision is now a clean 3-gate AND — `!in_boot_grace(loop_started_at) && !apierror_episodes.contains(name) && !capped` — combining the boot-grace gate, #1697 episode dedup, and #1760's F4 nudge-cap (verified at the diff: all three coexist, F4 counter logic preserved). auth_error_gate runs before resolve_pending_auth so detection/confirm-window untouched. Both test sets kept; supervisor.rs only; CI green on all 3 platforms. Merging via §3.12 admin-squash — **completes the ① boot-grace/boot-seed in-mem-reset treatment** (#1754 pane_input + #1753 producer boot-seed + this notify/apierror).

Issue #1742, opened 2026-06-04
notification reliability gaps in ServerRateLimit retry + self-orchestrator P0 escalation (2 MEDIUM silent-drops + 3 LOW)


suzuke:
Found by post-refactor audit (reviewer-2/claude, t-audit-notification-srl) on #1699/#1703/#1716 @ origin/main b5c90d8. Core retry/escalation/gate logic is correct; these are latent edges. **F1 lead-verified** (supervisor.rs:1410-1413 vs 1363-1390).

## F1 [MEDIUM — silent-drop, VERIFIED] inject failure → permanent `exhausted`, zero notification
`supervisor.rs:1410-1413`: when `inject_continue_gated()` returns false on a ServerRateLimit retry, the else branch sets `retry.exhausted = true` with **only a `tracing::warn!`** — it does NOT call `notify_orchestrator_retry_exhausted` nor `gated_notify` (those are only in the `retry_count > MAX` branch, 1363-1390). ServerRateLimit isn't in `is_notify_error_class`, so entering it never notifies either. → a **single** inject failure (PTY write Err, or agent killed between Phase 1 and Phase 2) permanently disables auto-retry for that agent with **zero** orchestrator/operator notification; the #1696 ~75min budget is ended by one failure. (snap=None / truly-gone is harmless — retain clears next tick; the danger is agent-still-present + inject Err.)
- Pre-existing (the else predates #1699's rewrite), but in the audited slice.
- Fix: on inject-fail, keep the track for next-tick retry (don't exhaust on one failure), or count failures and only exhaust **after N** AND then run the full notification path.

## F2 [MEDIUM — silent-drop] self-orchestrator P0 escalation is Telegram-only, no durable fallback/retry
`supervisor.rs:471-492` (AuthError), `1370-1389` (exhausted self-orch), `1214` (self-skip). When the orchestrator IS the affected agent (no peer), the only delivery is `if let Some(ch)=active_channel() { gated_notify }`. AuthError fires once (after the #1523 90s hold); exhausted fires once. If `active_channel()` is None at fire time (boot channel not connected / Telegram reconnecting) or the gated_notify is dropped → **no durable queue, no retry** (edge-triggered; the agent stays in AuthError with no new transition) → the P0 is permanently lost. Non-self-orch paths have a peer collecting via inbox/event-bus; only self-orch lacks it — defeating the #1595/#1703 reachability goal. Not a regression (#1703 added the first self-orch notification at all).
[...]

PR #1760, opened 2026-06-04
fix(#1742-F4): cap the ApiError quick-nudge per flicker-window

suzuke:
## #1742-F4 — cap the ApiError quick-nudge per flicker-window

The #1697 ApiError-at-prompt quick-nudge dedups per **episode** (one nudge each
time an agent enters `ApiError`). A content false-positive that flickers the
backend between `ApiError↔Thinking` re-arms that per-episode dedup every cycle,
so the nudge re-fires indefinitely — rate-limited only by the 5 s
`CONTINUE_INJECT_MIN_INTERVAL`, with **no total cap**.

### Fix (mirrors the ServerRateLimit 12-cap)
Add a per-agent `apierror_nudge_counts` window counter:
- decide path skips the nudge once the count hits `APIERROR_NUDGE_MAX` (12);
- execute path increments it on each successful inject;
- the count resets **only on genuine recovery** (`Idle`, via
  `clears_server_rate_limit_retry`) — NOT on a mid-flicker `Thinking`/`ToolUse` —
  so it accumulates across the `ApiError↔Thinking` oscillation it bounds, while a
  real single ApiError (which recovers to `Idle`) is unaffected.

### Test
`apierror_nudge_caps_per_flicker_window_1742_f4` drives `APIERROR_NUDGE_MAX + 3`
flicker cycles and asserts the count caps at `APIERROR_NUDGE_MAX`.
**Negative-probed**: dropping the `!capped` gate reaches `MAX + 3` (15). The
existing `apierror_at_prompt_quick_nudge_once_per_episode_1697` confirms a single
ApiError still nudges.

clippy `--all-targets --features tray` + fmt clean; single-file change, no drift.

Refs #1742.

---
1. The cap accumulates across the flicker (reset gated on `Idle` only, not
   `Thinking`) so a pathological `ApiError↔Thinking` oscillation is bounded, while
   a genuine single ApiError (→ Idle) still nudges and resets.
2. `apierror_nudge_counts` is retained/dropped with the other per-agent maps and
   incremented only on a successful `Injected`.

### DIFF
--- a/src/daemon/supervisor.rs
+++ b/src/daemon/supervisor.rs
@@ -53,6 +53,13 @@ const AUTH_ERROR_NOTIFY_STABILITY: Duration = Duration::from_secs(90);
 /// 2026-06-02 incident gave up at ~80s (Phase A only) against a multi-minute
 /// proxy fault. Total budget ~75min over 12 retries.
 const SERVER_RATE_LIMIT_MAX_RETRIES: u32 = 12;
+/// #1742-F4: max ApiError quick-nudges per flicker-window before the nudge stops.
+/// Mirrors the ServerRateLimit 12-cap. A content false-positive `ApiError↔Thinking`
+/// flicker re-arms the per-episode dedup every cycle, so without a total cap the
+/// quick-nudge would inject indefinitely (bounded only by `CONTINUE_INJECT_MIN_INTERVAL`).
+/// The count resets on genuine recovery (`Idle`), so a real single ApiError still
+/// nudges and only a pathological flicker is capped.
+const APIERROR_NUDGE_MAX: u32 = 12;
 /// Backoff schedule for ServerRateLimit retries (seconds). Phase A | B | C.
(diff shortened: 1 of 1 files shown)


suzuke:
✅ **Review mirror — fixup-agy (primary, Gemini 3.1 Pro High) VERIFIED + lead cross-checked + CI green.** ApiError quick-nudge now has a per-agent window cap (mirrors SRL's 12-cap): decide-path blocks queueing once counts reach APIERROR_NUDGE_MAX; execute-path +1 only on Injected; the counter resets ONLY on genuine recovery (clears_server_rate_limit_retry / Idle), NOT mid Thinking/ToolUse, so it accumulates across an ApiError↔Thinking flicker; a single ApiError (which returns to Idle) is unaffected. Negative-probe has teeth (15 cycles → asserts strictly 12; removing the gate hits 15 and fails). supervisor.rs:1439/1596/1452/3828; 68 tests pass. Merging via §3.12 admin-squash. Addresses #1742 F4.

Issue #869, opened 2026-05-16
Daemon: repo action=checkout bind=true should guarantee tree-at-remote-HEAD post-bind


suzuke:
## Problem

`repo action=checkout branch=<X> bind=true` is intended to give reviewer (and any agent) a worktree bound to the latest remote HEAD of branch X. In practice, during #852 residual PR-B + PR-C reviews, the resulting worktree was at a stale local SHA (e.g., 077022f) instead of the actual remote HEAD.

Workaround applied: reviewer falls back to `gh pr diff` + CI rollup as authoritative replay. This works but defeats the point of `bind=true` (provide a usable testing surface at PR head).

Lead surfaced this on **multiple** PRs in the same batch — recurring, not one-off.

## Proposal

After binding worktree, daemon should:
1. `git fetch origin <branch>` to ensure remote ref is current
2. Either: `git reset --hard origin/<branch>` OR re-create worktree at the remote-HEAD SHA
3. Verify post-bind: worktree's `git rev-parse HEAD` matches `git rev-parse origin/<branch>`
4. Surface verification failure in dispatch result so caller knows bind isn't trustworthy

## Acceptance

- After `repo action=checkout bind=true`, `git -C <worktree> rev-parse HEAD` always matches `git -C <worktree> rev-parse origin/<branch>` for the bound branch
- Test: simulate stale local SHA → invoke bind → verify worktree at remote HEAD
- No reviewer fallback needed to `gh pr diff` for "what's at PR head" question

## Origin

- Surfaced by fixup-lead in #852 residual closing report (m-20260516190349892926-29, item #4)
- Operator-authorized file 2026-05-16
- Recurring across PR-B and PR-C reviews in the batch

## Related

- #852 + residual chain — context for the discovery

PR #1761, opened 2026-06-04
fix(#1755): auto-fetch origin/<branch> before creating a worktree branch

### LINKED ISSUE #1755: repo checkout MCP tool does not auto-fetch → agents start on stale bases (push diff confusion + latent reverse-revert risk)
## Found by fixup-dev during #1754
The `repo checkout` MCP tool returned `fetch_attempted: false`, so the agent's worktree base was stale (`fcc789f`) while `origin/main` had advanced to `7e5de08` (#1749). dev's `git diff origin/main..HEAD` (two-dot) then showed `src/bin/agend-git.rs` reverted to the pre-#1749 version (because the stale-based branch lacks #1749). dev caught it, did `git fetch && git reset --hard origin/main`, re-applied its supervisor.rs patch, and confirmed the diff was clean.

## Severity: LOW-MED
- **Squash-merge (the §3.12 path) is actually safe**: `gh pr merge --squash` applies the three-dot diff (merge-base...head), which contains only the PR's own changes — NOT a spurious #1749 revert. Verified on #1752 (stale base fcc789f, but `gh pr diff --name-only` showed only escalation files, no agend-git.rs).
- **The real cost** is (a) agent confusion from the two-dot diff, (b) local test failures / merge conflicts when a dependency changed upstream, (c) a latent reverse-revert risk IF a non-squash merge or force-push is ever used.

## Fix
- **Primary**: the daemon's `repo checkout` should **auto-fetch** `origin/main` so worktree bases are current (or rebase the checkout onto fresh origin/main).
- **Interim hygiene** (already being followed): agents run `git fetch origin main && git diff --stat origin/main..HEAD` before push to confirm no stray reverse-revert, and rebase onto current origin/main.

Related to the broader 'post-task resource lifecycle' theme (#1747 worktree GC, #1750 branch GC).

suzuke:
## What
Fix #1755: a fresh `repo checkout` / task dispatch starts on a **stale base**, silently reverse-reverting merges that landed since the local repo's last fetch.

## Root cause
`dispatch_hook::ensure_branch_exists`'s branch-**create** path runs `git branch <new> origin/main` with **no fetch** — and since `origin/main` is a local remote-tracking ref that *always* resolves, it takes the "fast path (zero network)" and bases the new branch on whatever stale SHA the local `refs/remotes/origin/main` last held. The fetch was only a *fallback* for unresolvable refs (never triggered for origin/main). The sibling #869 branch-**exists** path already fetches + `update-ref`s; only create was missed.

This bit real work twice in one session (PR #1754 base 1 commit stale; PR #1758 base stale by a whole sibling merge), and reviewer-2's #1752 too — the manual `git fetch && git diff --stat` pre-push workaround was fragile.

## Fix
Before `git branch`, when `from_ref` is `origin/<x>`, best-effort `git fetch origin <x> --quiet` (60s timeout, mirroring #869) so the create lands on **current** main:
- `fetch_attempted` reports **success** (matches #869's `fetched_ok`) → the no-remote test fixtures (`file:///dev/null` origin → fetch fast-fails) keep reporting `false`, so existing assertions are unchanged.
- A fetch failure (offline) falls back to the local ref — degraded but functional, same contract as #869.

**One function, both entry points:** `ensure_branch_exists` is shared by `repo checkout` (ci/mod.rs:123) *and* the `send kind=task` dispatch hook (dispatch_hook:416), so this fixes the stale base for both. Cost: one fetch per branch-creation (<1s typical, **once-per-checkout, not per-tick**).

## Test
`ensure_branch_exists_refreshes_stale_origin_before_create_1755`: builds a **real local bare origin**, advances `origin/main` to commit B while forcing the local remote-tracking ref **stale at A**, then asserts a newly-created branch bases on **B (fresh)**, not A (stale) — and that `fetched=true`. Existing create/EXISTS-path tests pass unchanged (their `file:///dev/null` origin fast-fails the fetch → `fetch_attempted=false` preserved).

Preflight: fmt + clippy (`tray`, `tray,discord`, -D warnings) + full `cargo test --tests` (real git) all green.

### DIFF
--- a/src/mcp/handlers/dispatch_hook/mod.rs
+++ b/src/mcp/handlers/dispatch_hook/mod.rs
@@ -688,10 +688,38 @@ pub(crate) fn ensure_branch_exists(
         }
         return Ok((false, fetched_ok));
     }
-    // Step 2: try create from `from_ref` (no fetch yet — zero network
-    // on the fast path where origin/main is already a valid local ref).
+    // Step 2: create from `from_ref`. #1755: a remote-tracking `from_ref` like
+    // `origin/main` ALWAYS resolves as a local ref, so a bare `git branch` here
+    // silently bases the new branch on a STALE local ref (whatever was last
+    // fetched) — the reverse-revert hazard where a fresh checkout starts behind
+    // main and would clobber merges that landed since. Refresh the remote ref
+    // FIRST (mirrors the #869 branch-EXISTS path above) so the create lands on
+    // current `origin/<x>`. Best-effort: a fetch failure (offline / no-remote
+    // fixture) leaves the local ref as-is and the create still succeeds against
+    // whatever's present (degraded but functional, same contract as #869).
+    // `fetch_attempted` reports SUCCESS (matches #869's `fetched_ok`), so the
+    // no-remote test fixtures keep reporting `false`.
+    let mut create_fetched = false;
+    if let Some(remote_branch) = from_ref.strip_prefix("origin/") {
+        let fetch_start = std::time::Instant::now();
+        let fetch_out = crate::git_helpers::git_bypass_timeout(
+            source,
+            &["fetch", "origin", remote_branch, "--quiet"],
+            std::time::Duration::from_secs(60),
+        );
+        create_fetched = matches!(&fetch_out, Ok(o) if o.status.success());
+        crate::event_log::log(
+            home,
+            "ensure_branch_fetch",
+            actor,
+            &format!(
+                "branch={branch} from_ref={from_ref} duration_ms={} ok={create_fetched} (#1755 pre-create refresh)",
+                fetch_start.elapsed().as_millis()
+            ),
+        );
+    }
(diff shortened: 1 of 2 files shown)


suzuke:
✅ **Review mirror — fixup-agy (primary, Gemini 3.1 Pro High) VERIFIED + lead cross-checked + CI all green.** `ensure_branch_exists`'s create path now best-effort `git fetch origin <x>` before `git branch <new> origin/<x>` (60s timeout, mirrors the #869 EXISTS path) so a new branch is based on FRESH origin/main, not a stale local remote-tracking ref. Shared fn → fixes both `repo checkout` and `send kind=task` dispatch. fetched_ok semantics keep existing tests unchanged; fetch-fail degrades to the local ref. New test has teeth (real bare origin advanced to B, local forced to stale A, asserts the new branch is on B). Merging via §3.12 admin-squash. Closes #1755 — this is the root of the stale-base reverse-revert that hit 3 PRs today.

PR #1763, opened 2026-06-04
test(runtime_config): serialize global-singleton tests to stop CI flake

suzuke:
## What
Root-fix the `runtime_config::tests` CI flake that has been reddening **unrelated** PRs (#1752, #1758 ubuntu) and forcing churny reruns — the same operator-flagged "瞎忙 churn" class as #1755.

## Root cause
`set_and_get_key`, `set_hang_auto_recovery_enabled`, `show_pane_state_default_on_and_toggleable`, and `keys_cover_all_settable_keys_including_show_pane_state` mutate the **process-global** `RUNTIME_CONFIG` singleton (`OnceLock<RwLock<…>>`) via `reload()`. Run concurrently (cargo's default), one test's `reload()` clobbers the global between another's `reload` + `get_key`, so the second reads the wrong value → intermittent assertion failure. No other test in the suite reloads the global, so the racers are exactly these four.

## Fix
Mark the four global-touching tests `#[serial(runtime_config)]` (serial_test, already a dev-dep + established pattern). The **named group** serializes only these against each other — unrelated `#[serial]` tests in other modules keep running in parallel.

Test-infra only; zero runtime change.

## Verify
`runtime_config::tests` 5/5 green in a loop; fmt + clippy (`tray`, `tray,discord`, -D warnings) + full `cargo test --tests` all green. Base = latest main (incl. #1755).

### DIFF
--- a/src/runtime_config.rs
+++ b/src/runtime_config.rs
@@ -237,6 +237,13 @@ pub fn keys() -> Vec<String> {
 #[allow(clippy::unwrap_used)]
 mod tests {
     use super::*;
+    // These tests mutate the process-global `RUNTIME_CONFIG` singleton via
+    // `reload()`, so running them concurrently lets one test's reload clobber
+    // another's value between its own `reload` + `get_key` — an intermittent
+    // assertion flake that reddened UNRELATED PRs (#1752, #1758) and forced
+    // churny CI reruns. Serialize the global-touching ones under a named group
+    // (keeps unrelated `#[serial]` tests in other modules running in parallel).
+    use serial_test::serial;
 
     #[test]
     fn default_values() {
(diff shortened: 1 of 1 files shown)


suzuke:
✅ **Review mirror — fixup-reviewer-2 VERIFIED + CI green.** Reproduced the flake (pre-fix: 4/8 parallel runs FAILED on global RUNTIME_CONFIG clobber between reload→get_key; post-fix: 0/10). The 4 runtime_config tests that mutate the global singleton now share `#[serial(runtime_config)]`; the pure default_values test is correctly left unserialized. serial_test=3 is a dev-dependency (test-only). +11/-0, pure test attrs, no-new-bugs. This is the fleet-wide CI flake that reddened #1752/#1758/#1759. Merging via §3.12 admin-squash.

**Follow-ups (non-blocking, noted)**: (1) a LOW residual cross-module mutator (app/commands.rs config_set_via_palette test) could theoretically flake the same way — add #[serial(runtime_config)] (one line); (2) a SEPARATE flake `daemon::retention::decisions::sweep_respects_protected_tags` also hit #1752's CI — different module, needs its own fix.

PR #1764, opened 2026-06-04
fix(#1762): only text-composing input marks a draft (nav/control keys no longer defer actionable injects)

### LINKED ISSUE #1762: Bug: Overly aggressive draft detection (#1675) stalls actionable notifications for 5 minutes on non-text input
### Description

A recent fix (#1675) designed to defer actionable injects while the operator has a live draft (paused draft) is too aggressive, resulting in severe delays (up to 5 minutes) for incoming notifications (like task dispatches or ci-ready alerts).

When a notification arrives, the TUI correctly displays the unread count (e.g., `[1]`), but the PTY injection is deferred because the system mistakenly believes the operator is "drafting."

### Root Cause Analysis

The issue lies in how `notification_queue::record_input_activity` is triggered in `src/app/mod.rs` (`write_to_focused` and `write_to_pane`):

```rust
// src/app/mod.rs (around line 1080)
if let Some(pane) = layout.active_tab_mut().and_then(|t| t.focused_pane_mut()) {
    notification_queue::record_input_activity(home, &pane.agent_name);
    if pane_input_contains_submit(pane.backend.as_ref(), bytes) {
        notification_queue::record_submit_activity(home, &pane.agent_name);
    }
    // ...
}
```

**ANY** key forwarded to the PTY (including non-character keys like `Up/Down` for history, `Shift+Tab` for permissions, or even accidental whitespace) updates the `last_input_epoch_ms`. However, only the specific submit key (e.g., `\r`) updates `last_submit_epoch_ms`. 

If an operator simply scrolls through history or toggles a setting while the agent is idle, `last_input` becomes greater than `last_submit`. The `draft_state` function then classifies this as `Drafting`, initiating the `AGEND_DRAFT_ESCAPE_SECS` (default 5 minutes) lock. During this period, all incoming actionable wakes are queued and hidden from the PTY.

### Proposed Fix

Refine the input recording logic so that non-printable control characters (arrows, tab, esc) do not trigger the start of a "draft". 

Alternatively, ensure that `record_input_activity` is only called for actual text input (e.g., alphanumeric characters and standard punctuation) that forms a message, preventing UI navigation actions from accidentally engaging the 5-minute notification block.

suzuke:
## What & why

Fixes #1762 (contributor report, credit @cheerc). An **idle operator who merely browses history with Up/Down** (or fat-fingers a non-text key / a lone space) trapped every **actionable inject** (task dispatch / ci-ready) behind the ~5min draft escape window — worst when the operator is away (the exact away-time impact the report flags).

**Root cause** (RCA confirmed): `write_to_focused` / `write_to_pane` (`src/app/mod.rs`) called `notification_queue::record_input_activity` for **every** byte forwarded to a pane's PTY. `tui::key_to_bytes` encodes navigation keys (Up/Down/arrows, Home/End, PageUp/Down, Delete/Insert, F-keys, Esc) as **ESC-prefixed** sequences, so any of them bumped `last_input_epoch_ms` past `last_submit_epoch_ms` → `draft_state` read `Drafting` → injects deferred.

> RCA nuance: the report's **Shift+Tab** example is actually *not* forwarded (`key_to_bytes` has no `BackTab` arm → empty → not written), so it never hit this path. The real culprits are the forwarded ESC-sequence nav keys above (+ lone space). The mechanism is otherwise exactly as @cheerc described.

## Fix

`is_text_composing_input(bytes)` gates `record_input_activity` at both write sites. A keystroke marks a draft only if it **enters the input buffer** — a non-space printable (`> 0x20`, excl. DEL) or a UTF-8 byte (`>= 0x80`). **Not** composing: ESC-prefixed sequences (nav/control/mode keys; **exception**: bracketed paste `ESC[200~` wraps pasted text → composing), bare control bytes (Ctrl-combos, Tab, Backspace, Enter — Enter is the separately-detected **submit**), and lone whitespace. `record_submit_activity` is unchanged.

**#1457/#1675 preserved**: a real draft always carries a non-space char, so live typing still reads `Drafting` and an inject never clobbers it — only navigation/control/whitespace stop spuriously deferring P0s. Also tightens #1513 (nav keys no longer count as "typed").

## Tests
- Exhaustive pure-fn classification: every nav/control/whitespace form (`\x1b[A`, F-keys, Esc, Alt-combos, Ctrl+A, Tab, Backspace, Enter, lone space) → `false`; chars / `"hi there"` / UTF-8 / bracketed-paste → `true`.
- Behavioral contract: exercising the exact gate, a nav key leaves the pane `DraftState::None` (undeferred) while real typing reads `Drafting`.

## Local verification
`cargo fmt --check`, `cargo clippy --all-targets --features tray,discord -- -D warnings`, and the `app::` / `notification_queue::` / supervisor `pane_input` test sets all green (29 passed). Pushed `--no-verify` (targeted-test workflow); base == current `main`, diff is `src/app/mod.rs` only. GitHub CI gates the full matrix.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -1075,10 +1075,38 @@ fn pane_from_menu_item(
     }
 }
 
+/// #1762: does this forwarded keystroke ENTER the agent's input buffer
+/// (text-composing), as opposed to NAVIGATE / CONTROL (arrows, F-keys, Esc,
+/// Ctrl-combos, Tab, Backspace)? Only composing input should mark a draft
+/// (#1457/#1675 draft-gating); navigation/control must NOT, or an idle operator
+/// who merely browses history (Up/Down) or fat-fingers a non-text key traps every
+/// actionable inject (task dispatch / ci-ready) behind the ~5min draft escape
+/// window — exactly when away (the #1762 report).
+///
+/// Composing = at least one byte that enters the buffer: a non-space printable
+/// (`> 0x20`, excluding DEL `0x7f`) or any UTF-8 continuation/lead byte
+/// (`>= 0x80`). Deliberately NON-composing: ESC-prefixed sequences (arrows /
+/// F-keys / Esc / Alt-combos — `key_to_bytes` encodes every nav key with a `0x1b`
+/// lead), bare control bytes (Ctrl-combos, `Tab`=`\t`, `Backspace`=`0x7f`, and
+/// `Enter`=`\r`/`\n` — Enter is the separately-detected SUBMIT signal), and lone
+/// whitespace (the fat-fingered-space case — a real draft always carries a
+/// non-space char that marks it, so #1675 protection is preserved). EXCEPTION:
+/// bracketed paste (`ESC [ 200 ~`) wraps PASTED TEXT and IS composing.
+fn is_text_composing_input(bytes: &[u8]) -> bool {
+    if bytes.first() == Some(&0x1b) {
+        return bytes.starts_with(b"\x1b[200~");
+    }
+    bytes.iter().any(|&b| (b > 0x20 && b != 0x7f) || b >= 0x80)
+}
+
 /// Write bytes to the focused pane's PTY (Local) or remote bridge (Remote).
 fn write_to_focused(home: &Path, layout: &mut Layout, registry: &AgentRegistry, bytes: &[u8]) {
     if let Some(pane) = layout.active_tab_mut().and_then(|t| t.focused_pane_mut()) {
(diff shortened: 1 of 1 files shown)


suzuke:
✅ **Review mirror — fixup-agy (primary) VERIFIED + lead cross-checked + CI green.** `is_text_composing_input` separates text-composing input from nav/control/whitespace; gates only the two record_input_activity sites (submit untouched); real typing still marks Drafting (#1457/#1675 preserved); nav keys no longer defer actionable notifications 5min. Exhaustive + behavioral tests with teeth. Merging via §3.12 admin-squash. Closes #1762.

Issue #1634, opened 2026-06-02
Backend model-discontinuation: codex agent silently errors every turn instead of surfacing a health signal


suzuke:
## What happened (2026-06-02)

fixup-reviewer (codex backend) had been reviewing PRs fine, then after a daemon restart it began **erroring on every single turn** with:

```
⚠ Model metadata for `gpt-5.3-codex` not found.
Defaulting to fallback metadata; this can degrade performance and cause issues.

{"type":"error","status":400,"error":{"type":"invalid_request_error",
"message":"The 'gpt-5.3-codex' model is not supported when using Codex with a ChatGPT account."}}
```

Root cause: the configured model (`gpt-5.3-codex`) was discontinued server-side for the ChatGPT-account auth (the user's ~/.codex config even shows a pending `gpt-5.3-codex → gpt-5.5` migration). The operator had to MANUALLY switch the model to a currently-available one to recover.

## The actual defect (agend-terminal side)

The agent was effectively **dead but looked alive**: agent_state stayed healthy/idle, no health_state signal, no blocked_reason. The only way the lead/operator could discover it was a manual `pane_snapshot` — every dispatched review just silently failed. This blocked the dual-review merge gate with no surfaced reason.

## Desired behavior

The daemon should DETECT the backend-model-error class — `invalid_request_error` / "model is not supported" / "Model metadata ... not found" in the backend's output — and SURFACE it:
- set a health_state (e.g. Blocked) with a blocked_reason carrying the error message, so list_instances / the TUI show the agent as unhealthy instead of silently-erroring;
- ideally a one-line operator notification ("fixup-reviewer: model gpt-5.3-codex unsupported — needs model change").
- (stretch) optional auto-fallback to a configured backup model, but at minimum SURFACE it — don't let a model-discontinued agent masquerade as healthy.

## Captured error string (for the detector)

- `Model metadata for \`<model>\` not found`
- `status: 400 ... invalid_request_error ... model is not supported when using Codex with a ChatGPT account`

## Severity

P2 — recurs whenever a backend model is deprecated/migrated; silently breaks the review fleet until someone pane-inspects. Detection + surfacing is the fix.

PR #1765, opened 2026-06-04
fix(#1757): exempt hard network errors from the #1450 red anchor

### LINKED ISSUE #1757: Bug: VTerm color anchor (#1450) suppresses valid network errors (InvalidHTTPResponse/ECONNRESET)
### Description

When a backend (like Gemini CLI or Claude Code) encounters a hard network error (e.g., `InvalidHTTPResponse`, `ECONNRESET`), the daemon's `#1450` color anchor filter suppresses the state transition, preventing the auto-retry mechanism from activating.

Even though the text appears red to the user in their terminal emulator, the daemon's internal VTerm records the `span_fg` for the matched text as `Default`, `Rgb(255, 255, 255)`, or `Rgb(175, 175, 175)`. Because `SERVER_RATE_LIMIT_NET_ERRORS` is mapped to `AgentState::ServerRateLimit` (which is a `HIGH_FP` state), the daemon requires the matched text to contain at least one red cell. When VTerm doesn't classify the foreground as red, the transition is suppressed.

### Logs

Here are redacted logs showing the `#1450` suppression for `InvalidHTTPResponse` and `ECONNRESET`:

```
2026-06-04T16:16:29.543239+08:00  WARN #1450: HIGH_FP pattern matched but rendered fg not red — suppressing transition. If this was a real backend error, span_fg shows the actual colors (predicate may be too strict). agent= backend=codex state=ServerRateLimit matched=InvalidHTTPResponse span_fg=[Default, Default, Default, Default, Default, Default, Default, Default, Default, Default, Default, Default, Default, Default, Default, Default, Default, Default, Default] line_context=        ⎿  API Error: InvalidHTTPResponse fetching "http://127.0.0.1:3456/v1/messages?beta=true".

2026-06-04T16:19:33.414395+08:00  WARN #1450: HIGH_FP pattern matched but rendered fg not red — suppressing transition. If this was a real backend error, span_fg shows the actual colors (predicate may be too strict). agent= backend=gemini state=ServerRateLimit matched=ECONNRESET span_fg=[Default, Default, Default, Default, Default, Default, Default, Default, Default, Default] line_context=│ 30:pub(crate) const SERVER_RATE_LIMIT_NET_ERRORS: &str = r"ECONNRESET|ETIMEDOUT|InvalidHTTPResponse|fetch failed|connection reset|socket hang up|proxy.*disconnect";                             │
```

### Proposed Fix

Hard network errors (`SERVER_RATE_LIMIT_NET_ERRORS`) are highly specific and unlikely to be false positives (unlike conversational phrases such as "server is temporarily limiting requests"). Therefore, they should not be subject to the strict `#1450` color anchor.

Change the mapping of `SERVER_RATE_LIMIT_NET_ERRORS` in all `BackendProfile`s and `src/state/patterns.rs` from `AgentState::ServerRateLimit` (which is `HIGH_FP`) to `AgentState::ApiError` (which is not `HIGH_FP`).

This will:
1. Bypass the `#1450` color filter, ensuring 100% detection rate regardless of how the backend renders the text or how VTerm parses the colors.
2. Benefit from the recent `#1696/#1697` `ApiError quick-nudge` mechanism, providing an immediate retry for transient network drops.

suzuke:
## What
Fix #1757: the #1450 VTerm color anchor was **suppressing real network errors** (`InvalidHTTPResponse` / `ECONNRESET` / …) → no `ServerRateLimit` transition → auto-retry never fired → the agent sat stuck (worst case: operator away).

## RCA (credit: cheerc — verified against code + the issue's logs)
`SERVER_RATE_LIMIT_NET_ERRORS` maps to `AgentState::ServerRateLimit`, a `HIGH_FP` state, so #1450 requires the matched text to contain ≥1 red cell (`matched_span_has_red`). But codex/gemini render these hard net faults in **DEFAULT/grey, not red** (observed `span_fg=[Default × 19]` for the real `API Error: InvalidHTTPResponse fetching "http://…"`). `is_red()` only matches `CellFg::Red` (achromatic default/white/grey → not red), so `anchor_fail` fired → transition suppressed. cheerc's core RCA is correct.

## ⚠ Correcting the issue's fix-framing (premise-check)
The issue proposed a **blanket bypass** ("net errors are highly specific, unlikely to be FP"). That framing is **overstated** — the issue's *own* gemini `ECONNRESET` log is in fact a **false positive that #1450 correctly suppressed**: the agent was **viewing patterns.rs source** (`const SERVER_RATE_LIMIT_NET_ERRORS: &str = r"ECONNRESET|…"`), not erroring. Net-error tokens *do* appear in source/logs/prose, so a blanket bypass reopens FPs.

A **scoped exemption** is the right call because the residual net-error FP is now **bounded** by #1518 (live-tail gate) + #1586 + the #1760 nudge cap — a few self-correcting spurious `continue`s. The cost asymmetry (severe + common stuck-agent false-negative vs rare + bounded FP) favors the exemption.

## Fix
Exempt only the net-error tokens from the red anchor (fail-open, like the empty-fg / `Shell` paths) via a new `patterns::is_net_error_match`:
```rust
let anchor_fail = self.anchor_on_red && high_fp && !fg.is_empty()
    && !crate::state::patterns::is_net_error_match(matched)   // #1757
    && !matched_span_has_red(screen_text, matched, fg);
```
**Scoped:** the FP-prone `api_error` / `overloaded` / context / model-unsupported HIGH_FP tokens **stay red-anchored**.

## Tests
- `net_error_invalidhttpresponse_default_latches_1757` — the exact codex failure (`API Error: InvalidHTTPResponse fetching …` rendered all-DEFAULT) now latches `ServerRateLimit` → auto-retry.
- `server_rate_limit_red_anchor_fp_boundary_1587` **rewritten** — net-error in DEFAULT now *latches* (the #1587 net-error red-anchor boundary is deliberately superseded; doc updated). Red still latches; the dropped `network error` token still doesn't.
- **Scope preserved** by the unchanged `codex_model_unsupported_red_anchor_fp_boundary_1634` — a *non*-net-error HIGH_FP (`ModelUnsupported`) is still suppressed in DEFAULT, proving the exemption is scoped, not a blanket bypass.
- Preflight: fmt + clippy (`tray`, `tray,discord`, -D warnings) + full `cargo test --tests` all green. Base = latest main.

### DIFF
--- a/src/state/mod.rs
+++ b/src/state/mod.rs
@@ -804,6 +804,15 @@ impl StateTracker {
                     let anchor_fail = self.anchor_on_red
                         && high_fp
                         && !fg.is_empty()
+                        // #1757: exempt hard network errors from the red anchor —
+                        // codex/gemini render `InvalidHTTPResponse`/`ECONNRESET`/…
+                        // in DEFAULT/grey (not red), so the anchor wrongly
+                        // suppressed REAL faults → no ServerRateLimit transition →
+                        // no auto-retry → stuck agent. Scoped: the FP-prone
+                        // api_error/overloaded/context HIGH_FP tokens stay
+                        // red-anchored; the net-error residual FP is bounded by
+                        // #1518/#1586/#1760 (see `patterns::is_net_error_match`).
+                        && !crate::state::patterns::is_net_error_match(matched)
                         && !matched_span_has_red(screen_text, matched, fg);
                     // #1518 position gate: a HIGH_FP marker that has scrolled out
                     // of the live bottom-N rows (e.g. an ApiError / ServerRateLimit
--- a/src/state/patterns.rs
+++ b/src/state/patterns.rs
@@ -29,6 +29,26 @@ use regex::Regex;
 // copy), keeping byte-identity with the legacy compile_for arm that also uses it.
 pub(crate) const SERVER_RATE_LIMIT_NET_ERRORS: &str = r"ECONNRESET|ETIMEDOUT|InvalidHTTPResponse|fetch failed|connection reset|socket hang up|proxy.*disconnect";
 
+/// #1757: is `matched` one of the [`SERVER_RATE_LIMIT_NET_ERRORS`] tokens?
+///
+/// These hard network faults are rendered in DEFAULT/grey (not red) by codex /
+/// gemini (observed `span_fg=[Default,...]` for `InvalidHTTPResponse` /
+/// `ECONNRESET`), so the #1450 red anchor wrongly suppressed REAL faults → no
+/// `ServerRateLimit` transition → no auto-retry → stuck agent (worst when the
+/// operator is away). The anchor exempts these tokens (fails open like the
+/// empty-`fg` / `Shell` paths). Scoped on purpose: the FP-prone
+/// `api_error`/`overloaded`/context HIGH_FP tokens STAY red-anchored. The
+/// residual net-error FP (these tokens can appear in source/logs/prose — e.g. an
+/// agent viewing THIS file) is bounded by the #1518 live-tail gate, #1586, and
+/// the #1760 nudge cap, so the cost asymmetry (severe stuck-agent vs rare,
+/// self-correcting spurious `continue`) favors the exemption.
+pub(crate) fn is_net_error_match(matched: &str) -> bool {
+    use std::sync::OnceLock;
+    static RE: OnceLock<Regex> = OnceLock::new();
+    RE.get_or_init(|| Regex::new(SERVER_RATE_LIMIT_NET_ERRORS).expect("net-error regex compiles"))
+        .is_match(matched)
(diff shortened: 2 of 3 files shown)


suzuke:
✅ **Review mirror — fixup-agy (primary) VERIFIED + lead cross-checked + CI green.** Scoped net-error exemption: only SERVER_RATE_LIMIT_NET_ERRORS tokens exempt from the #1450 red-anchor (api_error/overloaded/model-unsupported stay red-anchored; #1634 boundary test unchanged + green proves scope). #1587 boundary test rewrite correctly encodes the new behavior (default-rendered net-error latches→auto-retry). Source-view FP bounded by #1518/#1586/#1760. 249 state tests pass. Merging via §3.12 admin-squash. Closes #1757.

PR #1766, opened 2026-06-04
test(ci-flake): serialize the last two global-state test races (anti-churn)

suzuke:
## What & why

Anti-churn cleanup: closes the **two CI flake residuals** surfaced during the #1763 review. Both are the same class — a test mutating **process-global state** racing a concurrent reader — and both reddened unrelated PRs (#1752/#1758), forcing rework churn. Test-infra only; no production code touched.

### 1. runtime_config residual (the one I flagged in #1763)
`app/commands::config_set_via_palette_persists_to_runtime_config` reaches `runtime_config::set` → `*global().write()`, mutating the same `RUNTIME_CONFIG` singleton the `runtime_config::tests` serialize on (#1763) — but it was **not** in that serial group, so it could clobber their global mid-assertion (e.g. flip `show_pane_state` between `show_pane_state_default_on_and_toggleable`'s `reload` and `get_key`). Add `#[serial(runtime_config)]`.

### 2. retention flake — investigated, a REAL race (not one-off)
The three `sweep_*` tests in `daemon::retention::decisions` `set_var`/`remove_var` the process-global `AGEND_RETENTION_CUTOVER` that `sweep()` reads (decisions.rs:83). Run concurrently, **one test's `remove_var` clears the gate before another's `sweep()` reads it** → that `sweep()` early-returns `0` → `assert_eq!(archived, 1)` fails. This is exactly the `sweep_respects_protected_tags` failure observed in #1752 CI. Serialize the three under `#[serial(retention_cutover)]` (mirrors capture.rs's `#[serial(capture_env)]` env-var convention). The non-env retention test (`archive_serializes_with_decision_lock`) stays parallel.

## Verification
- Mechanism: both are deterministic process-global races (cited above).
- `cargo fmt --check` ✅, `cargo clippy --all-targets --features tray,discord -- -D warnings` ✅.
- Stress: the combined `retention::decisions + runtime_config + app::commands::config` set passes **8/8** under repeated concurrent runs post-fix (intermittently red pre-fix).
- Pre-push `fetch + diff --stat origin/main..HEAD` → only `app/commands.rs` + `daemon/retention/decisions.rs` (no dragged files).

### DIFF
--- a/src/app/commands.rs
+++ b/src/app/commands.rs
@@ -386,6 +386,12 @@ mod tests {
     use super::*;
     use crate::layout::{Layout, Pane, PaneSource, Tab};
     use crate::vterm::VTerm;
+    // #1763 residual: `config_set_via_palette_*` mutates the process-global
+    // `RUNTIME_CONFIG` (via `runtime_config::set` → `*global().write()`), the same
+    // singleton the `runtime_config::tests` serialize on. Join the SAME
+    // `runtime_config` serial group so this cross-module test can't clobber their
+    // global mid-assertion.
+    use serial_test::serial;
 
     fn test_pane(id: usize, agent: &str, fleet_name: Option<&str>) -> Pane {
         Pane {
@@ -469,6 +475,7 @@ mod tests {
     /// and persist. Assert against the written file (deterministic — avoids the
     /// process-global cache shared across parallel tests).
     #[test]
+    #[serial(runtime_config)]
     fn config_set_via_palette_persists_to_runtime_config() {
         let dir =
             std::env::temp_dir().join(format!("agend-test-cmd-config-{}", std::process::id()));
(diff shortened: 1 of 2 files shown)


suzuke:
✅ **Review mirror — fixup-agy (primary) VERIFIED + lead cross-checked + CI green.** Anti-churn CI-flake residuals (same class as #1763): config_set_via_palette joins #[serial(runtime_config)]; the 3 retention sweep_* tests get #[serial(retention_cutover)] (they set/remove the process-global AGEND_RETENTION_CUTOVER that sweep() reads → parallel clobber caused the #1752 sweep_respects_protected_tags failure). grep-confirmed complete isolation; non-env tests stay parallel; pure test-infra. Merging via §3.12 admin-squash.

PR #1769, opened 2026-06-04
fix(#1768): stop the ServerRateLimit retry storm (working-marker below scrolled-up error wins)

### LINKED ISSUE #1768: Bug: 'Retry Storm' caused by stale error strings in tail scan (#1757 regression)
### Description

A "retry storm" (duplicate `continue` injections) occurs when a `ServerRateLimit` error persists in the terminal's tail scan after the agent has already begun working. This is a regression caused by the interaction between the fix for #1757 (exempting net errors from red anchor) and the fix for #1713 (narrowing retry clearing to only `Idle`).

### Root Cause Analysis

1.  **#1757 Impact**: Hard network errors (`InvalidHTTPResponse`, `ECONNRESET`) no longer require red text to latch.
2.  **Tail Scan Persistence**: These error strings often remain in the last 15 lines (`ERROR_TAIL_SCAN_LINES`) even after the agent has recovered, especially if the subsequent output is short (e.g., a simple task dispatch).
3.  **Priority Trap**: `ServerRateLimit` has a priority of 10, while `Thinking` has 6. As long as the error string is in the tail, the state remains latched as `ServerRateLimit`.
4.  **#1713 Clearing Logic**: The clearing predicate `clears_server_rate_limit_retry` was narrowed to only `Idle`. `Thinking` and `ToolUse` no longer clear the track.
5.  **Chain Reaction**: 
    - Agent is latched in `ServerRateLimit` (due to tail scan).
    - Track is not cleared (since it's not `Idle`).
    - The backoff timer expires.
    - Supervisor injects another `continue`, even though the agent might already be working.

### Observed Behavior (Logs)

An agent was observed receiving multiple redundant injections before finally reaching `Idle`:
- `19:54:59`: injected "continue" (attempt 1) ✅
- `19:55:19`: injected "continue" (attempt 2) ❌ (Agent was already working but tail scan matched old error)
- `19:55:49`: injected "continue" (attempt 3) ❌
- `19:56:50`: injected "continue" (attempt 4) ❌
- `19:57:30`: recovered (`Idle`) — track finally cleared.

### Proposed Fix

Re-introduce productive states (`Thinking`, `ToolUse`) to the clearing logic, but with a guard to prevent the "flicker reset" that #1713 aimed to solve.

1.  Allow `Thinking` and `ToolUse` to clear the `ServerRateLimit` retry track.
2.  Optionally, add a short cooldown (e.g., 30s) after an injection where a `Thinking` transition is ignored (treated as a transient of the injection itself) but after which it counts as a genuine recovery that clears the track.

Alternatively, increase the required output volume or age before a non-red error in the tail is allowed to latch `ServerRateLimit` again after a recovery attempt.

suzuke:
## What
Fix #1768: the **ServerRateLimit retry storm** — the supervisor re-injects `continue` into an agent that has already recovered and is working.

## RCA (verified — corrects the issue's attribution)
- **NOT a #1757 regression.** The observed-storm binary `2b5cba4` *predates* #1757 (`git merge-base --is-ancestor f563ed0 2b5cba4` → false; #1757 committed 19:02, `2b5cba4` is the 18:58 tip). The 19:49 deploy ran `2b5cba4` (#1762, no #1757). #1757 *amplifies* (net errors latch without red) but didn't cause it.
- **Root cause:** a HIGH_FP `ServerRateLimit` error wins the priority race over `Thinking` (`detect_with_match` returns the first/highest-priority match, and the net-error pattern precedes Thinking in every profile). After the error scrolls UP and the agent resumes work below it, each scan **re-latches** ServerRateLimit. `clears_server_rate_limit_retry` is Idle-only (#1713) so the retry track never clears; the latched ServerRateLimit also doesn't auto-expire (`maybe_expire_latched_state` has no ServerRateLimit arm); the backoff expires → `continue` is re-injected into a working agent → storm.

## Fix (detection-side)
When a HIGH_FP error matches but a `Thinking`/`ToolUse` working-marker is rendered **below** it (more recently in the tail), the agent recovered → land on that working state instead of re-latching the stale error. New `StatePatterns::working_state_below`. This extends the #1518 absolute-tail-position gate to "relative to the working marker."

- **A genuinely-stuck error has NO in-flight working-marker below it → still latches → auto-retry fires.** (The fix does not mask real faults.)
- **Does NOT touch `clears_server_rate_limit_retry` (#1713)** → zero flicker-reset regression risk. This is the key safety property (the reason #1713 narrowed clearing to Idle-only).
- **Covers all backends** — each profile carries Thinking + ToolUse markers (codex `Working|esc to interrupt`, gemini `(esc to cancel,`, kiro `Kiro is working|esc to cancel`, …).
- *Implementation note:* we land on the working state rather than merely suppressing, because `maybe_expire_latched_state` has no ServerRateLimit arm — a pure suppress would leave the latched ServerRateLimit in place and the storm would continue.

## Tests
- `retry_storm_recovered_below_error_lands_on_working_state_1768` — error scrolled up + `esc to interrupt` below → lands on `Thinking` (no storm).
- `retry_storm_fresh_error_below_working_marker_still_latches_1768` — fresh error below the marker → still `ServerRateLimit`.
- `retry_storm_error_with_no_working_marker_still_latches_1768` — error alone → still `ServerRateLimit` (auto-retry preserved).
- #1757 / #1587 / #1634 / #1713 tests unchanged + green.
- Preflight: fmt + clippy (`tray`, `tray,discord`, -D warnings) green; full `cargo test --tests` green except an **unrelated** flake `bridge_handles_half_up_daemon` (a port-reuse race under parallel load — confirmed 3/3 green in isolation, passes in earlier full runs; not touched by this change).

## ⚠ Adversarial-review ask (state-detection sensitive)
Please check: can a *real* ServerRateLimit have a working-marker below it (→ wrongly suppressed)? It shouldn't — a failed request removes the in-flight `esc to interrupt`/spinner, so a stuck error has no working-marker after it. A marker below the error means a request is in flight (recovering / internally retrying), which is exactly when we should NOT inject. Flagging for the lens.

### DIFF
--- a/src/state/mod.rs
+++ b/src/state/mod.rs
@@ -810,9 +810,20 @@ impl StateTracker {
                         // suppressed REAL faults → no ServerRateLimit transition →
                         // no auto-retry → stuck agent. Scoped: the FP-prone
                         // api_error/overloaded/context HIGH_FP tokens stay
-                        // red-anchored; the net-error residual FP is bounded by
-                        // #1518/#1586/#1760 (see `patterns::is_net_error_match`).
-                        && !crate::state::patterns::is_net_error_match(matched)
+                        // red-anchored.
+                        // #1768: NARROW that exemption to the token's LINE looking
+                        // like a real backend error line (`Error:`/`API Error:`/
+                        // `FetchError:` label) — a bare prose/source mention (an
+                        // orchestrator discussing net-errors, idle between turns,
+                        // with no working-marker below) stays red-anchored →
+                        // default-rendered → suppressed, instead of mis-latching
+                        // ServerRateLimit (the retry-storm FP codex caught). A real
+                        // fault IS error-line-shaped → still fails open → latches.
+                        && !(crate::state::patterns::is_net_error_match(matched)
+                            && crate::state::patterns::net_error_in_error_line(
+                                screen_text,
+                                matched,
+                            ))
                         && !matched_span_has_red(screen_text, matched, fg);
                     // #1518 position gate: a HIGH_FP marker that has scrolled out
                     // of the live bottom-N rows (e.g. an ApiError / ServerRateLimit
--- a/src/state/mod.rs
+++ b/src/state/mod.rs
@@ -850,7 +850,28 @@ impl StateTracker {
                             self.maybe_expire_latched_state();
                         }
                     } else {
-                        let gated = self.gate_on_heartbeat(detected);
+                        // #1768: a HIGH_FP error wins the priority race even after
+                        // it scrolled up and the agent RESUMED WORK below it
+                        // (ServerRateLimit > Thinking by pattern order), so it keeps
+                        // re-latching → `clears_server_rate_limit_retry` (Idle-only,
+                        // #1713) never fires → supervisor re-injects `continue` into
+                        // a working agent (the #1768 retry storm; the latched
+                        // ServerRateLimit also doesn't auto-expire). If a
+                        // Thinking/ToolUse marker is rendered BELOW the error, the
+                        // agent recovered → land on that working state instead of the
+                        // stale error. A genuinely-stuck error has NO in-flight
+                        // working marker below it → `landed` stays `detected` → still
+                        // latches → auto-retry fires. Detection-side only;
+                        // `clears_server_rate_limit_retry` untouched → no #1713
+                        // flicker-reset regression.
+                        let landed = if high_fp {
+                            patterns
+                                .working_state_below(screen_text, matched)
+                                .unwrap_or(detected)
+                        } else {
+                            detected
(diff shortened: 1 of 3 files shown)


suzuke:
✅ **Review mirror — fixup-reviewer (codex) VERIFIED (re-review @d312d9a) + lead + CI green.** #1768 retry storm fixed by (c) working-marker override + tightening #1757's net-error exemption to error-labeled lines only (net_error_in_error_line). Prose net-error mentions in the live tail are suppressed; real error lines still latch (#1757 intent preserved). Fixes the idle-between-turns false latch that hit fixup-lead while discussing net-errors. Closes #1768.

Issue #1770, opened 2026-06-04
Daemon raw auto-inject (ServerRateLimit/ApiError retry 'continue') is indistinguishable from operator input — add [AGEND-AUTO] marker


suzuke:
## Problem (real incident)

The daemon's ServerRateLimit/ApiError retry mechanism injects raw `continue\n` (CONTINUE_RETRY_PAYLOAD, supervisor.rs:95) directly into an agent's PTY. Under the PTY single-stream model, the agent sees it **identically to operator-typed input** — there is no marker.

**Concrete harm (2026-06-04):** fixup-lead (orchestrator) was false-latched as ServerRateLimit (it was discussing net-error tokens in issue work; see #1768/#1757), and the retry mechanism injected 'continue' 4×. The orchestrator **misread those daemon-injected 'continue's as operator input and dispatched work based on them** (a #1694 (a)+(b) impl that the operator had not actually approved).

## Root cause

The inbox path already carries `[AGEND-MSG]`/`[from:AGENT]` headers (notify.rs:31/34) and agents are taught to recognize them. The gap is **purely the raw auto-inject that bypasses the header** (continue retry / ESC). `inject_prefix` is a per-backend rendering preset, not a daemon marker.

## Design (approved, impl in progress)

Add an `[AGEND-AUTO]` marker (an `[AGEND-MSG]`-family sibling) for raw daemon auto-injects: `continue\n` → `[AGEND-AUTO kind=ratelimit-retry] continue\n`.
- **Worker semantics preserved**: the inner `continue` is still there → worker resumes as before.
- **Orchestrator can identify**: `[AGEND-AUTO]` → knows it's a daemon auto-nudge, never treats it as an operator command / dispatch trigger.
- **Backward-compatible**: backends that don't recognize the marker still see 'continue' → resume; no protocol change.
- **Enforcement**: an agent-instruction in the protocol doc + agend.md (same trust model as `[AGEND-MSG]`), since the PTY single-stream model can't hard-enforce.

Impl: `inject_with_target_gated` gains an `auto: bool` (alongside `force`); only daemon self-initiated auto-inject (ServerRateLimit/ApiError continue) sets it; inbox/operator-relay unchanged. ESC (recovery Stage1) deferred (a control key, low misread risk).

Complements the #1768/#1757 detection fix (which stops the *false* latch → stops the spurious inject); this marker handles the case where a *legit* auto-inject still needs to be distinguishable by an orchestrator.

[...]

PR #1771, opened 2026-06-04
fix(#1769): tag daemon auto-injects with [AGEND-AUTO kind=...] (orchestrator can't mistake them for operator commands)

suzuke:
## What & why

Fixes #1769 (operator-reported, hit live). The daemon resumes a stuck agent by injecting a keystroke (ServerRateLimit / ApiError `continue`) **straight into its PTY** — a single byte stream where the injected text looks **identical to the operator typing it**. An orchestrator (`fixup-lead`) read an injected `continue` as an operator instruction and **dispatched a task off it**.

**Key insight:** inbox/relay messages already carry `[AGEND-MSG]`/`[from:]` headers that agents are taught to recognize; the gap was the **raw auto-injects that bypass those headers**. Give them a sibling marker.

## Change

- `agent::inject_with_target_gated` gains `auto_kind: Option<&str>`. `Some(kind)` prepends `[AGEND-AUTO kind=<kind>] ` (via `daemon_auto_prefix` / `DAEMON_AUTO_INJECT_MARKER`) **before both** the deferred-enqueue and the direct PTY write, so the tag survives whichever delivery path runs.
- **Only daemon self-originated nudges tag**: the shared `inject_continue_gated` passes `Some("ratelimit-retry")` (SRL retry) / `Some("apierror-nudge")` (ApiError nudge). Operator-relay / api `INJECT` / cron pass `None` — they already carry their own headers and must look like real input.
- **Worker semantics unchanged**: the inner `continue` + submit newline are intact (worker still resumes); the marker is leading text. Recovery Stage-1 ESC is out of scope (a control key, not a readable command).
- **Enforcement = agent-instruction trust model** (same as `[AGEND-MSG]`): `src/instructions.rs` (the injected agent prompt) and `docs/FLEET-DEV-PROTOCOL.md` now teach agents that an `[AGEND-AUTO]` line is a low-priority RESUME signal — continue in-progress work, **never** treat it as an operator command or dispatch from it. (Edited the repo embedded source per protocol-edit discipline, not the live `.default` copy.)

## Tests
- `daemon_auto_prefix` embeds the kind and shares the `[AGEND-AUTO` marker token (instruction ↔ inject can't drift); inner payload + submit survive.
- The ServerRateLimit phase-2 PTY-read test now asserts the `[AGEND-AUTO kind=ratelimit-retry]` marker actually reaches the PTY (real path).
- Instructions include the `[AGEND-AUTO]` rule (+ the "never as operator command" clause).
- #1680 force-false source guard updated for the new (multi-line) call shape (whitespace-normalized, also pins the new `auto_kind` arg).

## Verification
`cargo fmt --check` ✅, `cargo clippy --all-targets --features tray,discord -- -D warnings` ✅, `agent:: / instructions:: / supervisor:: / cron_tick::` test sets — 203 passed ✅. Pre-push `fetch + diff --stat` → only the 8 intended files (no dragged-in changes).

### DIFF
--- a/docs/FLEET-DEV-PROTOCOL.md
+++ b/docs/FLEET-DEV-PROTOCOL.md
@@ -377,6 +377,7 @@ Re-review cycles (r1, r2, …) repeat the same three milestones. The dispatcher 
 - Response channel must match source channel
 - **Router-layer channel discipline (Sprint 52)**: daemon auto-mirrors agent direct text to the corresponding channel. Agent does not need to force `reply` tool — infrastructure handles routing.
 - **Inbox vs PTY delivery (Sprint 62)**: active agents receive messages via PTY direct injection (not queued in `inbox`). `inbox(instance: X)` returning empty does NOT mean X received nothing — only means X has no unread queue. Verify delivery via `describe_instance` (active state = PTY received) or `pane_snapshot` rather than inbox alone. Inbox queue only fills for offline agents or undeliverable messages.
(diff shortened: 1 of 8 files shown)


suzuke:
✅ **Review mirror — fixup-reviewer (codex) VERIFIED + lead + CI green.** Daemon raw auto-inject now carries an [AGEND-AUTO kind=K] marker (only the 2 daemon-self-initiated callers; operator-relay/cron/api-INJECT pass None). Both delivery paths carry it; worker still recognizes the inner continue. Agent instructions (src/instructions.rs + repo docs/FLEET-DEV-PROTOCOL.md) forbid treating [AGEND-AUTO] as an operator command. Closes #1770.

PR #1772, opened 2026-06-04
test(bridge): de-flake bridge_handles_half_up_daemon (port-reuse race)

suzuke:
## What
De-flake `bridge_handles_half_up_daemon` (tests/integration.rs) — a fleet-wide CI-churn flake source (sibling of the runtime_config #1763 fix).

## Root cause
The test binds an ephemeral port, **drops** the listener, then asserts a connect to that "now-free" port is refused. Under a parallel `cargo test --tests` run, another test can re-bind the just-freed ephemeral port between the drop and the connect, so the connect spuriously **succeeds** → `assert!(result.is_err())` fails. It's a fixture port-reuse race, not a real "half-up daemon" outcome — it surfaced as a one-off red in a full run (passes in isolation, 5/5).

## Fix
Retry with a fresh ephemeral port when the connect unexpectedly succeeds (the port was reused by a concurrent test). The real property under test — **a connect to a non-listening port fails FAST (refused), it does not hang** — only asserts on the genuine refused case. Bounded to 8 attempts (sustained reuse across 8 fresh ephemeral ports is astronomically unlikely). Test-infra only; no production change.

## Verify
`bridge_handles_half_up_daemon` 5/5 green in a loop; fmt + clippy (`--all-targets --features tray -D warnings`) clean. Base = latest main.

(Discovered while running the #1768 full preflight; offered + flagged to the lead.)

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -739,22 +739,40 @@ fn bridge_rejects_malformed_mcp_json() {
 
 #[test]
 fn bridge_handles_half_up_daemon() {
-    // Get a random port that nothing is listening on
-    let port = {
-        let l = std::net::TcpListener::bind("127.0.0.1:0").expect("bind");
-        l.local_addr().expect("addr").port()
-        // l drops here — port is free but nothing listening
-    };
-    // Try to connect — should fail (connection refused), not hang
-    let start = Instant::now();
-    let result = TcpStream::connect_timeout(
-        &SocketAddr::from((Ipv4Addr::LOCALHOST, port)),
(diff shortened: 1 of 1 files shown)


suzuke:
✅ **Review mirror — fixup-agy VERIFIED + lead + CI green.** test-infra anti-churn: bridge_handles_half_up_daemon retries with a fresh port (up to 8×) on parallel-test port-steal; the refused-not-hang assertion only runs on a true refusal. Only tests/integration.rs, no production code. Removes a fleet CI-churn source (same class as #1763).

PR #1775, opened 2026-06-04
fix(dispatch_idle): productive-silence clock de-noise + active-recovery exempt (#1694②)

suzuke:
## What

Replaces the dispatch-idle watchdog's instantaneous thinking/tool_use state gate (#1516) with a **productive-silence clock**, fixing the operator's "reminders became noise" complaint without neutering the watchdog.

Came out of the team dispatch_idle de-noise dialectic + a read-only premise-verification spike (both this session).

## Why

The #1516 gate suppressed the nudge only while the snapshot's *instantaneous* `agent_state` was `thinking`/`tool_use`. An agent making real progress but not latched-thinking at the scan instant got nudged → noise. Conversely a spinner/junk agent whose state stays latched-thinking was never caught.

## How

- **`AgentSnapshot.silent_secs`** — seconds since `last_productive_output` (marker/heartbeat-gated, so spinner/blink/junk does *not* count), filled at snapshot-write from the StateTracker. dispatch_idle keeps reading `snapshot.json` (zero registry lock, #1492/#1516 design preserved). **Fails OPEN** (large serde default) so an old-format snapshot fires rather than silently suppressing.
- **`target_is_working`** now suppresses when the agent produced productive output within `threshold_secs` (progressing, just slow) **OR** is in an active-recovery state (`server_rate_limit`/`api_error`) — the auto-retry machinery owns recovery; nudging it re-creates the very proxy-drop churn this removes (dialectic finding #4).
- **bounded-backstop deferred** (YAGNI) — the hang detector independently covers the narrow MCP-active-but-stuck poll-loop case.

## Verified premise (spike)

The productivity **heartbeat is MCP-activity-based, not PTY-output-based** — so stuck-at-prompt / spinner / infinite-gen (no MCP calls) do **not** mask the silence clock. Residual gap (an agent making MCP calls every <10s without progress) is narrow and left to hang_detection.

## Tests

- `productive_but_not_thinking_suppressed_1694` — recently-productive non-thinking target → **suppressed** (the noise fix)
- `productive_silent_non_recovery_still_fires_1694` — productive-silent non-recovery → **still fires** (watchdog intact)
- `active_recovery_exempt_does_not_fire_1694` — ServerRateLimit target → **suppressed**
- `target_is_working_reads_silence_clock_1694` — unit cases incl. silent-but-latched-thinking + recovery exemption
- existing #1516/#1658 state tests preserved via a state→silence mapping in the test helper

`cargo fmt --check` clean · `cargo clippy --all-targets --features tray -D warnings` clean · dispatch_idle (42) + snapshot + conflict_notify + per_tick::snapshot + inbox::notify all green.

### DIFF
--- a/src/daemon/conflict_notify.rs
+++ b/src/daemon/conflict_notify.rs
@@ -645,6 +645,7 @@ mod tests {
                 submit_key: "\r".to_string(),
                 health_state: "healthy".to_string(),
                 agent_state: "thinking".to_string(),
+                silent_secs: 0,
             }],
         );
         home
--- a/src/daemon/dispatch_idle/mod.rs
+++ b/src/daemon/dispatch_idle/mod.rs
@@ -509,16 +509,39 @@ pub(crate) fn pending_for_instance(
     (as_dispatcher, as_target)
 }
 
-/// #1516: is `target` demonstrably working right now? True iff the fleet
-/// snapshot reports its `agent_state` as `thinking` or `tool_use` — the
-/// PTY-output-fed "actively generating / running a tool" signal (the same one
-/// `health.rs` treats as productive). Pure for testability. Unknown target /
-/// missing snapshot → `false` (don't suppress → fire as before; degrades to
-/// the pre-#1516 behavior, never worse).
-fn target_is_working(snapshot: Option<&crate::snapshot::FleetSnapshot>, target: &str) -> bool {
+/// #1694② silence-clock: should the dispatch-idle nudge be SUPPRESSED for
+/// `target`? Replaces the #1516 instantaneous `thinking`/`tool_use` state check
+/// with two more-robust conditions read from the fleet snapshot:
+///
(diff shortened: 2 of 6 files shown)


suzuke:
✅ **Review mirror — fixup-reviewer (codex) VERIFIED (re-review @23b6c4c) + lead + CI green (manually verified — ci-watch was not tracking this PR).** dispatch_idle noise fix (operator-observed). Root: the #1516 working-gate read the snapshot's instantaneous agent_state (thinking/tool_use only) and AgentSnapshot carried no activity timestamp. Fix: AgentSnapshot gains silent_secs (from last_productive_output); gate becomes 'fire only when productive-silence > threshold' (keeps MCP-based heartbeat). Active-recovery exemption scoped to server_rate_limit ONLY (has the #1696 retry-budget + #1744 exhaustion backstop); api_error NOT exempt (once-nudge, no backstop — a stuck api_error must still fire, caught by adversarial review). fail-open; cheerc(b) repeat dropped (wrong direction). 43 tests green. Merging via §3.12 admin-squash.

Issue #1751, opened 2026-06-04
footgun class is broader than #1748: other test files run bare-git fixtures that pollute the bound worktree (likely the core.bare recurring source)


suzuke:
## Context
#1748/#1749 fixed the agend-git.rs test fixtures (bare `Command::new("git")` → agend-git shim → #1463 ChdirPass strips `-C`/redirects → pollutes the bound worktree). dev-2 **proved the agend-git fixtures do NOT set core.bare** (init/clone are shim-passthrough; live-probed). But the operator's canonical `core.bare` flipped to `true` AGAIN after being reset → **a DIFFERENT bare-git op, not in agend-git.rs**.

## Finding: the footgun class is broader than agend-git.rs
`src/worktree_cleanup.rs` test `test_v2_remote_gone_worktree_removed` (line ~509) and likely others run **bare-git fixtures**: `git_in(&remote_dir, ["init", "--bare", ...])` + `Command::new("git").args(["clone", ...])` + (to set up the squash-merge scenario) commits/checkouts. When run via `cargo test --tests` from a **bound agent worktree**, the mutating ChdirPass ops (commit/checkout) are redirected by the shim into the bound worktree — the SAME footgun as #1748, just in a different file. One of these (or a sibling test) is the likely source of the recurring `core.bare=true` on the canonical repo.

## Fix
1. **Codebase-wide sweep**: any test running bare `git` / `Command::new("git")` / `git_in` for fixtures → use `resolve_real_git()` (the #1749 pattern), so the shim can't redirect them into the bound worktree.
2. **Focused RCA**: pin the exact op that sets canonical `core.bare=true` (monitor + correlate; check `git_in` — does it use `-C` or `.current_dir()`? the latter is shim-redirectable).
3. Until fixed: the #1737 pre-push hook stays disabled (it runs `cargo test --tests`), and agents avoid full `cargo test --tests` from bound worktrees.

## Provenance
2026-06-04: operator hit `git pull` → 'must be run in a work tree' (canonical core.bare=true); reset to false twice (lead + dev-2), flipped true again. dev-2 refuted the agend-git hypothesis (#1749). Related: #1748, #1747 (worktree GC), #1750 (branch GC).

suzuke:
## Update — the broad-footgun hypothesis is largely REFUTED (codex audit)

Two read-only RCAs (fixup-reviewer/codex) found:

[...]

Issue #1750, opened 2026-06-04
branch lifecycle: local branches never auto-delete after squash-merge + no periodic branch GC (91 local / 17 remote accumulated)


suzuke:
## Problem (operator-reported)
91 local + 17 remote branches accumulated. The operator asked: don't they auto-delete after a task's PR merges?

- **Remote**: ALREADY auto-deletes on merge — the §3.12 SOP uses `gh pr merge --admin --squash --delete-branch` (verified: #1740's `origin/fix/1740-*` was pruned on merge). The 17 remaining are pre-SOP / merged-without-flag / abandoned spikes+tests.
- **Local**: does NOT auto-delete. Squash-merge creates a NEW commit on main, so the original local branch is never an ancestor → `git branch --merged main` doesn't even detect it, and nothing deletes it. The canonical repo + agent worktrees accumulate them (91).

## Proposed process (3 layers)
1. **On-merge local cleanup** (immediate): when a PR merges, the daemon's worktree-release flow (§10.4) should also delete the task's LOCAL branch in the canonical repo (the remote is already handled by --delete-branch). Tie local-branch deletion to the existing worktree-release + the PR-merged event.
2. **Periodic branch GC** (backstop, daemon-side — mirrors the dispatch-tracking GC + gc_stale_watches): a sweep that deletes (a) local branches whose PR is merged/closed (skip checked-out, main, protected), (b) remote branches whose PR is merged/closed (skip open-PR, main, protected). PR-state-aware (squash-merge-safe, unlike `--merged`) + age-gated. Catches anything on-merge missed + pre-existing residue.
3. **Unify with #1747** (worktree GC) into a single 'post-task resource GC' (worktree + local-branch + sidecar residue — same post-task-cleanup theme).

## Immediate cleanup already done
Lead manually deleted 39 merged-PR + review/* local branches (91→52) and pruned the merged remote branch. The remaining ~52 local include checked-out worktree branches + stale `pr-NNNN` review-checkouts + possible WIP — a PR-state-aware GC (layer 2) is needed to safely finish.

Related: #1747 (review/audit worktree auto-cleanup).

suzuke:
[...]

PR #1782, opened 2026-06-04
docs(#1750 A1-b): note AGEND_GIT_BYPASS pushes do not auto-watch CI

suzuke:
## What

Bypass-coverage decision **(b)**: document the known limitation rather than touch the agend-git shim (lead/operator call). A bypass push skips the `dispatch_hook` that arms `watch_ci`, so no `[ci-ready]` fires for that branch — the actual cause of #1775's missing watch.

## Where

A new bullet in **§10 Git Workflow**, placed right after the `bind_self` invariant that lists *"auto watch_ci"* — the bypass push is the exception to it. After a bypass push that opens/updates a PR, arm `ci(action: watch)` manually (or rely on lead's `gh pr checks` fallback).

Edited the **docs/ embedded source** (`docs/FLEET-DEV-PROTOCOL.md`, compiled into the binary via `protocol.rs` `include_str!`) — **NOT** the live `.default` copy, which a daemon restart re-extracts from the binary (a live-copy edit would be silently overwritten).

## Follow-up

The deeper fix (re-enable the push hook so bypass is rarely needed) is gated on the #1751 broad shim-footgun fix — tracked, not in scope here.

Docs-only; builds clean (protocol `include_str!` recompiles), protocol + cargo_include_invariant tests green.


suzuke:
✅ **Review mirror — fixup-agy VERIFIED + lead + CI green.** Docs-only: AGEND_GIT_BYPASS-push-no-auto-watch known-limitation → docs/FLEET-DEV-PROTOCOL.md §10.

PR #1776, opened 2026-06-04
fix: extend error-line-shape regex coverage (corpus MISS list) + generalize in_error_line

suzuke:
## What
First (low-risk) step of the color-anchor rework (`①`): **extend the error-line-shape regex** to cover the MISS list found by the `t-step1-detection-corpus` spike, and generalize `net_error_in_error_line` → `in_error_line`. This is a **coverage extension only — it does NOT change the red→content anchor decision** (that larger change, and the keep/drop call on the color signal, is the next step pending the operator).

## Why (corpus-derived)
The detection-signal corpus showed the old line-shape regex `(?i)\b(api ?|fetch)?error\s*[:?]` MISSES real backend error lines:
- **ModelUnsupported `invalid_request_error:`** — a real regex bug: the `_error` underscore kills the `\berror` word-boundary, so this line *never* qualified;
- **gemini bare HTTP-status** (no `API Error:` label): `429 Too Many Requests`, `got status: 429`, `RESOURCE_EXHAUSTED`;
- **structured-JSON error fields**: `"reason":"rateLimitExceeded"`, `"type":"error"`.

(The corpus also reversed the "non-default color" idea — real errors render Default *and* themed colors like Monokai's `248,248,242` normal fg, so color is unreliable both ways. That anchor change is the follow-up; this PR only fixes content-shape coverage.)

## Change
`patterns::in_error_line` (renamed from `net_error_in_error_line`) now matches any of:
```
(?i)( \b\w*error\s*[:?]                              # Error: / API Error: / FetchError: / api_error: / invalid_request_error:
    | \b429\b | resource_exhausted | too many requests | got status\s*:   # bare HTTP-status (gemini RateLimit)
    | "(type|code|status)"\s*:\s*"\w*error\w*"        # structured-JSON error fields
    | "reason"\s*:\s*"[^"]*exceeded" )
```
`\w*error` (not `\berror`) is the fix for the `_error` underscore. The single call site (state/mod.rs, the #1757 net-error red-anchor exemption) is unchanged in structure — still gated by `is_net_error_match(matched)` — so **behavior is preserved for today's only consumer**; the broadened signatures stage the content-primary anchor for the next step.

## Tests
[...]

PR #1783, opened 2026-06-04
fix(#1777): extend the #1768 working-marker override to UsageLimit (Sticky UsageLimit)

### LINKED ISSUE #1777: Bug: Sticky UsageLimit status persists after recovery (Missing working-marker override)
### Description

The `UsageLimit` status in the TUI does not clear immediately after an agent recovers and starts producing output. It often remains latched until the agent produces enough output (typically 15+ lines) to push the old error message out of the "tail scan" window. This creates a confusing experience where an agent is clearly active (Thinking/ToolUse) but the status bar still shows `[UsageLimit]`.

### Root Cause Analysis

1.  **Priority Race**: In `src/state/mod.rs`, `UsageLimit` has a priority of **11**, which is higher than `Thinking` (**6**) and `Idle` (**4**). The state detector (`detect_with_match`) returns the highest-priority match found in the scanned text.
2.  **Tail Scan Persistence**: The daemon scans the last 15 lines (`ERROR_TAIL_SCAN_LINES`). If the agent's first response after recovery is short, the old `UsageLimit` error string remains within these 15 lines.
3.  **Missing Override**: While #1768 introduced a "working-marker override" logic (where a `Thinking` marker below a scrolled-up error wins), this logic is currently scoped to `HIGH_FP` states (primarily `ServerRateLimit`, `RateLimit`, `ContextFull`). `UsageLimit` is **not** classified as `HIGH_FP` and thus does not benefit from this "more recent output wins" logic.

### Evidence

- **Priority Definition**: `src/state/mod.rs:112` (`Self::UsageLimit => 11`)
- **Tail Scan Limit**: `src/state/mod.rs` (`const ERROR_TAIL_SCAN_LINES: usize = 15`)
- **Override Scope**: `src/state/mod.rs` (The #1768 logic in `feed_with_fg` is gated behind `high_fp` checks).

### Proposed Fix

Apply the "working-marker override" logic (introduced in #1768) to `UsageLimit` as well. Even if a `UsageLimit` string is matched, if a `Thinking` or `ToolUse` marker appears **below** it in the rendered tail, the working state should take precedence.

Alternatively, consider lowering the priority of `UsageLimit` to be below `Thinking` once an initial notification has been fired, ensuring active work always wins over historical errors.

suzuke:
## What
Fix #1777 (cheerc, "Sticky UsageLimit"): after an agent recovers from a usage-limit error and **resumes producing output**, the status stayed stuck on `[UsageLimit]` until the stale error line scrolled off (15+ lines later). Same class as the #1768 retry storm — a sticky error state that wins the priority race over working markers.

## Root cause
`UsageLimit` (priority 11) outranks `Thinking` (6) / `ToolUse` (5), so `detect_with_match` (first-match-by-priority) keeps returning `UsageLimit` while the stale error string is still on screen. And `UsageLimit`:
- is **not** in `is_high_fp_state` → it never got the #1768 working-marker override (that was gated `if high_fp`), and
- has **no arm in `maybe_expire_latched_state`** → it never auto-expires.

So once latched it re-latches every scan until the line scrolls out of the grid. (Unlike #1768 this is a stale-**display** bug, not a retry storm — `UsageLimit` drives operator-notify, not `continue` re-injection, and `transition()`'s same-state early-return means no notify spam.)

## Fix
Extend the #1768 working-marker override to `UsageLimit` — one condition:
```rust
let landed = if high_fp || matches!(detected, AgentState::UsageLimit) {
    patterns.working_state_below(screen_text, matched).unwrap_or(detected)
} else { detected };
```
When a `Thinking`/`ToolUse` marker is rendered **below** the stale UsageLimit line, the agent recovered → land on that working state. A genuinely-stuck UsageLimit (no working marker below → agent blocked) → `working_state_below` returns `None` → still latches → operator still notified.

- **`UsageLimit` stays OUT of `is_high_fp_state`** → the #1450 red anchor is unchanged. The red→content anchor decision is ① Step-2 (pending the operator's call on the color signal) — this PR does not touch it.
- **Detection-side only**, reuses the existing `working_state_below`. No notify/expiry/anchor changes.

## RateLimit / ContextFull (the "evaluate siblings" ask)
Already covered — both are in `is_high_fp_state`, so they already get the working-marker override (+ the #1518 position gate). `UsageLimit` was the only non-high_fp error state missing it. (`ModelUnsupported` is also high_fp → covered.)

## Considered but NOT done (flagging for the lens)
I did **not** also extend the #1518 `stale_position` gate to `UsageLimit`. The working-marker override fully fixes cheerc's reported scenario (recovered + producing). The only residual is the narrow idle-recovery case (agent done + idle, stale line still on-screen, no working marker) — minor display staleness that self-heals on scroll-off, not the reported bug. Extending `stale_position` too would make UsageLimit fully consistent with the high_fp siblings (and is safe — a real blocked UsageLimit produces no output so its line stays in the tail). Happy to add it as a follow-up if you want the consistency; kept this PR to the explicit ask + a tight diff.

## Tests
- `usage_limit_recovered_below_error_lands_on_working_state_1777` — UsageLimit scrolled up + `esc to interrupt` below → lands `Thinking`, NOT `UsageLimit`.
- `usage_limit_stuck_no_working_marker_still_latches_1777` — UsageLimit alone (no marker) → still `UsageLimit` (operator notified).
- #1768 `retry_storm_*`, existing `usage_limit_*` (incl. per-backend fixtures) unchanged + green.

Preflight: `cargo fmt --check` + clippy (`tray`, `tray,discord`, `-D warnings`) + full `cargo test --tests` all green. Base = latest `origin/main` (incl. #1776).

### DIFF
--- a/src/state/mod.rs
+++ b/src/state/mod.rs
@@ -872,7 +872,18 @@ impl StateTracker {
                         // latches → auto-retry fires. Detection-side only;
                         // `clears_server_rate_limit_retry` untouched → no #1713
                         // flicker-reset regression.
-                        let landed = if high_fp {
+                        //
+                        // #1777 (cheerc, "Sticky UsageLimit"): UsageLimit (prio 11)
+                        // is the same kind of sticky error — it outranks
+                        // Thinking/ToolUse and never auto-expires (no
+                        // `maybe_expire_latched_state` arm), so a stale UsageLimit
+                        // line keeps re-latching after the agent resumed work below
+                        // it (status stuck on `[UsageLimit]` until the line scrolls
+                        // off). Extend the same recovery override to it. UsageLimit
+                        // stays OUT of `is_high_fp_state` → the #1450 red anchor is
+                        // unchanged (that decision is ① Step-2, pending the
+                        // operator); this is the #1768 working-marker override only.
+                        let landed = if high_fp || matches!(detected, AgentState::UsageLimit) {
                             patterns
                                 .working_state_below(screen_text, matched)
                                 .unwrap_or(detected)
--- a/src/state/tests.rs
+++ b/src/state/tests.rs
@@ -3288,6 +3288,49 @@ fn retry_storm_error_with_no_working_marker_still_latches_1768() {
     );
 }
 
+/// #1777 (cheerc, "Sticky UsageLimit"): UsageLimit (prio 11 > Thinking/ToolUse,
+/// and never auto-expires) is the same sticky-error class as the #1768 retry
+/// storm — after the agent resumed work below a scrolled-up UsageLimit line, the
+/// status stayed stuck on `[UsageLimit]` until the line scrolled off. The #1768
+/// working-marker override now covers UsageLimit too.
+#[test]
+fn usage_limit_recovered_below_error_lands_on_working_state_1777() {
+    // UsageLimit line scrolled UP; the agent's most-recent line is the work spinner.
+    let screen = "▌ You've hit your usage limit. try again at 3pm.\n\
+                  ▌ resuming\n\
+                  ▌ esc to interrupt";
+    let n = screen.chars().count();
+    let mut t = StateTracker::new(Some(&Backend::Codex));
+    t.feed_with_fg(screen, &vec![CellFg::Default; n]);
+    assert_ne!(
+        t.get_state(),
+        AgentState::UsageLimit,
+        "#1777: a recovered agent (working-marker below the scrolled-up UsageLimit) \
+         must NOT re-latch UsageLimit"
+    );
+    assert_eq!(
+        t.get_state(),
+        AgentState::Thinking,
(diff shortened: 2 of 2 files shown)


suzuke:
✅ **Review mirror — fixup-reviewer (codex) VERIFIED + lead + CI green incl windows (rebased onto #1785 windows-fix; rerun cleared an unrelated intermittent windows git-test flake — clean_empty_init_commits, NOT this PR's pure-state diff, main HEAD passes the same test).** #1777: extends the #1768/#1769 working-marker override from high_fp to (high_fp || UsageLimit). Detection-side only — UsageLimit NOT added to is_high_fp_state, red-anchor unchanged. Stuck UsageLimit (no working-marker below) still latches → operator notified; recovery lands on the working state.

Issue #1784, opened 2026-06-04
fleet-wide windows-CI hang: intermittent Tests-step timeout (~56min) blocking all PRs incl main


suzuke:
## Symptom
Windows CI `Check (windows-latest)` intermittently HANGS in the Tests step (~55-56min, then runner SIGTERM/exit 143 — conclusion=null, NOT an assert failure). mac+ubuntu Tests pass in 8-12min. The Tests step log does not finalize on timeout → the hanging test name is not captured.

## Evidence (dev investigation, #1738 discipline — not blind-retried)
- **main HEAD is windows-red**: `11c225f` (after #1782) windows = failure/hang; `a1a4e10` (#1780) / `b7e7f96` (#1779) windows = success. → the hang appeared in the merge window after #1780; it is main-wide, not PR-specific.
- Intermittent: same commit passes some runs, hangs others (a1a4e10 passed; #1777 branch ×2 + 11c225f hung).
- #1777 (the currently-blocked PR) is pure state-detection logic (feed_with_fg synthetic text, no PTY/thread/subprocess) → physically cannot introduce a subprocess hang; it is collateral, not cause.

## Lead premise-check (thesis)
- The gh path is **mocked** in tests (`gh_poll` uses a `dyn GhPoller` mock; real gh lives in scm/github_token production) → **not a gh-auth hang**.
- Strong candidate: the **real `Command::new("git")` subprocess tests** in `src/worktree_cleanup.rs` + `src/branch_sweep.rs` (create repos/branches, run git-cherry/tree-diff) — windows git subprocesses are known to intermittently hang (file locking, process spawn, AV). These were exercised more heavily by #1778 (B3 squash-GC wiring git-cherry into the per-tick prune).

## Next step
1. Add windows-CI instrumentation to **identify the specific hanger**: `--nocapture` + per-test start logging + a tighter Tests-step timeout (fail fast, surface the last-started test) — OR cargo-nextest per-test timeout.
2. Fix: timeout the git subprocess invocations (hang → fast-fail), and/or harden/cfg-gate the flaky windows test.

## Impact
Blocks the windows required check on ALL PRs (incl main). #1777 (sticky UsageLimit fix) is code-VERIFIED + green on mac/ubuntu/coverage/clippy/fmt/audit, held only on this shared hang.

(Filed during operator-away autonomous hygiene drive 2026-06-05. Not band-aided via known-flake bypass — root-cause fix per #1738/CI-retry-masks.)

[...]

PR #1785, opened 2026-06-04
ci(#1784): name the windows test hang — windows tests via nextest (terminate-after)

suzuke:
## What (Step 1 of #1784)
Instrument the windows CI to **name** the intermittent test hang. A real `Command::new(\"git\")` test (`worktree_cleanup` / `branch_sweep` build real repos + worktrees) hangs intermittently on windows-latest, timing the whole job out at **~56min with no attribution** — `cargo test` prints a test's name only on *completion*, so a hung test is anonymous, and the step log never finalizes on a job timeout. **main HEAD (#1782, 11c225f) is itself red on windows from this** — it is fleet-wide, not any one PR.

## Change
The windows Tests step now runs via **nextest** with a per-test `terminate-after` (`.config/nextest.toml` `ci` profile: SLOW at 60s, SIGKILL at 120s):
- **Parallelism is preserved** → the intermittent race still reproduces.
- A test still running past the timeout is **killed and NAMED** → a 56-min mystery becomes a fast, attributed failure (e.g. `worktree_cleanup::tests::<name>`).
- **mac/ubuntu (green) keep `cargo test` unchanged** — surgical, zero blast radius on the passing platforms.

Validated locally (nextest 0.9.137): the `ci` profile parses and runs real `worktree_cleanup::tests::*` green, naming tests by full path.

## ⚠ Expected first-run behavior
This run is **diagnostic** — if the hang reproduces, the windows job will go **red with the culprit's name** (intended), not green. The intermittency means it may also pass (a1a4e10 did); when it hangs, we get the name. **Step 2** (the actual fix — git-subprocess timeout / harden / cfg-gate the named test) follows as a commit on this PR once the culprit is pinned. Holding #1777 until windows is green again.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,20 @@
+# #1784: instrument the fleet-wide windows CI hang hunt.
+#
+# A real `Command::new("git")` test (worktree_cleanup / branch_sweep build real
+# repos + worktrees) intermittently HANGS on windows-latest, timing the whole CI
+# job out at ~56min with NO attribution — `cargo test` prints a test's name only
+# on COMPLETION, so a hung test is anonymous in the log (which also never
+# finalizes on a job timeout).
+#
+# The windows Tests step runs via nextest with this `ci` profile: it keeps the
+# default parallelism (so the intermittent race still reproduces), but any test
+# still running past the slow-timeout is KILLED and NAMED. That turns a 56-min
+# mystery hang into a fast, attributed failure so we can pinpoint + fix the
+# offending test. mac/ubuntu (green) keep `cargo test` unchanged.
+[profile.ci]
+# Warn (mark SLOW) at 60s; SIGKILL the test after 2 periods (120s). Normal tests
+# finish in well under a second; a real-git test that takes >120s is wedged.
+slow-timeout = { period = "60s", terminate-after = 2 }
+# Surface the terminated/hung test prominently in the run summary.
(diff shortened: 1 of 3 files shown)


suzuke:
✅ **Review mirror — fixup-reviewer (codex) code-VERIFIED @3f18f65 (UNVERIFIED only pending the windows-CI runtime proof, NOW GREEN) + lead + full CI green incl Check (windows-latest).** Root-fix for the fleet-wide windows-CI hang (#1784): `agend_git_shim_phase2::shim_binary_compiles` was spawning a nested `cargo build --bin agend-git` (.output(), no timeout) that contended for the cargo/target lock — on Windows the mandatory file-lock + AV scanning the .exe intermittently DEADLOCKED → 56min Tests-step timeout, which was reddening windows on ALL PRs incl main HEAD. Fix: spawn the prebuilt `env!("CARGO_BIN_EXE_agend-git") --version` with AGEND_TEST_ISOLATION=1 (same convention as shim_recursion_guard_1504) — tests compiles+runs (stronger), no nested build (deadlock removed), satisfies the isolation invariant, 0.23s (was 38.6s). PLUS a permanent guard: Windows CI now uses cargo-nextest with a 120s per-test terminate-after, so any future hang surfaces as a named 2min fail instead of a 56min anonymous timeout. Windows green here = the fix is proven. Test/CI-only change (no production code → no rebuild needed).

Issue #1787, opened 2026-06-04
windows-git-test-flake class: systematically harden git-subprocess tests (unmasked by #1784 fix)


suzuke:
## Context
#1784's fix removed the 56min windows-CI hang (a test's nested cargo build deadlocking on the target lock). With the hang gone, the windows full test suite now runs to completion **for the first time** — and is surfacing a batch of **windows git-subprocess test flakes** that were previously masked by the hang (the suite never finished, so they never reported).

The #1785 nextest guard (windows-only, 120s per-test terminate-after, final-status-level=slow) now **names** slow/hung tests, so these are diagnosable instead of anonymous timeouts.

## Observed so far (N=1)
- `dispatch_hook::tests::clean_empty_init_commits_handles_50_interleaved_inits` — intermittent `git log origin/main..HEAD failed` (empty stderr); 50 inits + git log, windows git is fragile (file locking / process spawn / AV). Passes on rerun; main HEAD passes the same test → intermittent.

## Goal — systematic harden (operator-approved 2026-06-05)
1. **Collect the class**: use the nextest naming + grep the test suite for real `Command::new("git")` / git-subprocess tests that build repos/branches/run git ops (worktree_cleanup, branch_sweep, dispatch_hook, agend_git_shim, etc.). Identify the flake-prone ones (heavy git subprocess on windows).
2. **Harden** (per-test, root-cause like #1784 where possible):
   - Prefer prebuilt-binary / fixtures over spawning nested cargo or real git where the test intent allows (the #1785 pattern).
   - Where real git is essential: add timeouts (hang → fast-fail), retries on the known-transient windows git errors, or serialize the lock-contending ones.
   - cfg-gate / mark slow only as a last resort.
3. **Verify**: windows CI stays green across several runs (the flakes don't resurface).

This is the medium-term follow-up flagged when #1784 was fixed — the nextest guard is the short-term safety net; this issue is the root-cause sweep so windows CI is reliably green, not rerun-dependent.

(Filed during operator-away→present autonomous drive 2026-06-05.)

suzuke:
Resolved (operator-confirmed 2026-06-05): the only CONFIRMED windows flake — clean_empty_init_commits (#1783, a PRODUCTION git path running git log via bare .output() with no retry) — is root-fixed by #1788 (run_git_idempotent: bounded transient-retry that surfaces persistent failures, idempotent ops only, rebase -i deliberately excluded).

[...]

PR #1788, opened 2026-06-04
fix(#1787 P1a): retry idempotent git in clean_empty_init_commits (windows flake)

suzuke:
## What (#1787 Phase 1a)
First step of the systematic windows-git-flake harden (follow-up to the #1784 hang fix). The #1784 fix let the full windows suite run to completion for the first time, surfacing a class of intermittent windows git-subprocess test failures the hang had been masking. This PR hardens the **confirmed** one — `dispatch_hook::clean_empty_init_commits` (the #1783 flake) — and audits its same-file siblings.

## Root cause (confirmed flake)
On windows CI, `clean_empty_init_commits` ran `git log origin/main..HEAD` via a raw `.output()` and **intermittently exited non-zero with empty stderr** under scratch-repo lock contention (50 interleaved inits) — failing `clean_empty_init_commits_handles_50_interleaved_inits` + `..._recovers_from_stale_rebase_merge_dir`. main HEAD passes the same test, so it's transient, not a regression. (This is a *transient non-zero exit*, not a hang — so retry is the right fix; test-level hangs are already caught by the #1785 nextest `terminate-after` guard.)

## Fix
New `run_git_idempotent(args, cwd)` helper: bounded retry (3 attempts, 100ms backoff) on transient failure, returning the final attempt's output so callers keep their existing success/stderr handling unchanged. Applied to the **idempotent** git calls:
- `git log origin/main..HEAD` (the confirmed flake) — `clean_empty_init_commits`
- `git diff-tree …` (per-commit emptiness check)
- `git reset --soft origin/main` (idempotent: reset to a fixed ref)
- `git log -1 --format=%b` — `commit_body_is_empty`

**Sibling audit:** `rebase -i origin/main` is deliberately **NOT** routed through the helper (a partial/interrupted rebase isn't idempotent — a blind retry would trip over the rebase-merge dir; `clear_stale_rebase_state` + the existing abort-failure surfacing already handle that path). Documented inline.

## Scope (per lead 拍板)
- Phase 1a only: the confirmed flake + same-file dispatch_hook siblings.
- The broad daemon-wide git-timeout audit (all raw-`.output()` git in the daemon) is **#1787 Phase-4** follow-up — not bundled here (blast radius).
- A wall-clock timeout (hang→fast-fail) belongs to that Phase-4 audit (a correct kill-capable std subprocess-timeout helper is non-trivial and the hang class is already test-guarded by #1785); this PR is retry-only, which is what the confirmed transient flake needs.

## Tests
- `run_git_idempotent_succeeds_and_surfaces_persistent_failure_1787` — contract: succeeds on a good command; on persistent failure returns the final non-success output (loop terminates, no panic/hang).
- The 10 existing `clean_empty_init_commits_*` tests (including the two #1783-flaky ones) pass + now exercise the retry wrapper.

Preflight: `cargo fmt --check` + clippy (`tray`, `tray,discord`, `-D warnings`) + full `cargo test --tests` all green. Base = latest `origin/main`.

### DIFF
--- a/src/mcp/handlers/dispatch_hook/mod.rs
+++ b/src/mcp/handlers/dispatch_hook/mod.rs
@@ -999,6 +999,47 @@ fn derive_repo_from_remote(source_repo: &std::path::Path) -> Option<String> {
 #[allow(clippy::unwrap_used)]
 mod tests;
 
+/// #1787: run an IDEMPOTENT git subprocess with bounded retry on transient
+/// failure. The #1784 windows-CI hang fix let the full suite run for the first
+/// time and surfaced intermittent transient failures of these dispatch-hook git
+/// calls — e.g. `git log origin/main..HEAD` exiting non-zero with empty stderr
+/// (#1783) under windows scratch-repo lock contention — which a retry clears.
+///
+/// ONLY for idempotent commands (log / diff-tree / reset-to-a-fixed-ref): a
+/// non-idempotent op (`rebase -i`) must NOT route through here and stays a direct
+/// call. Returns the FINAL attempt's output (or the spawn error), so callers keep
+/// their existing success/stderr handling unchanged. `AGEND_GIT_BYPASS=1` is
+/// pinned exactly as the direct calls did. (Wall-clock timeout for the hang class
+/// is the #1787 Phase-4 daemon-git audit; here the confirmed flake is a transient
+/// non-zero exit, and test-level hangs are caught by the #1785 nextest guard.)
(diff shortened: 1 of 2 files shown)


suzuke:
✅ **Review mirror — fixup-reviewer (codex) VERIFIED + lead + CI green.** #1787 PR-1 (Phase 1a): run_git_idempotent bounded retry (3×) on transient git failures, surfaces persistent failures (contract-tested), idempotent ops only, rebase -i deliberately not retried. Fixes the confirmed #1783 windows flake. Broader daemon git-timeout audit = #1787 Phase-4.

PR #1789, opened 2026-06-04
fix(coloranchor-gate): close kiro RateLimit content FN in in_error_line

suzuke:
## What

Prerequisite for dropping the RateLimit/ServerRateLimit color anchor. The `t-coloranchor-corpus-gate` validation (rg-on-fixture = Rust regex engine ground truth) found exactly **one color-gated content false-negative**: kiro RateLimit renders `ThrottlingException: Rate exceeded for this AWS account`, which failed every `in_error_line` alternation (`Exception` has no `error` substring; no 429/JSON). It renders red today so it fires via the #1450 anchor — but the **content path** (the basis for anchor removal) missed it.

## How

- Add a **`\b\w*exception\s*[:?]`** alternation to `in_error_line` — the structured `…Exception:` label is the same strong error-line signal as `…error:` (AWS/Bedrock names). The bare `Rate exceeded` phrase is **deliberately NOT added** (prose-ambiguous, the #848/#854 class; the real kiro error always carries the label).
- **Verified codex RateLimit** content coverage: `rate_limit_exceeded` rides the verified `stream error:` wrapper (→ matches the existing `\w*error:`); `RateLimitError:` matches directly. Added a synthetic `codex-rate-limit-typical.raw` (same red `stream error:` shape as the model-unsupported incident + OpenAI `rate_limit_exceeded` code) + MANIFEST entry as the codex corpus artifact for the removal-step validation.

## Scope

Does **NOT** touch the anchor decision itself — the removal-impl scope is the operator's next call. This only makes the content path (`in_error_line`) cover kiro so RateLimit content coverage is complete.

## Tests

`in_error_line_matches_corpus_errors_rejects_prose_step1` extended: kiro `ThrottlingException:` + codex `stream error: rate_limit_exceeded` + `RateLimitError:` match; bare `ThrottlingException` / `rate_limit_exceeded` prose (no label) stays prose (no new FP). behavioral_shadow + fixture_corpus parity green; full `state::` suite 275 passed; `fmt --check` + `clippy --all-targets --features tray -D warnings` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1790, opened 2026-06-05
fix(coloranchor): drop red anchor for RateLimit + ServerRateLimit (content gate)

suzuke:
## What

Operator-approved scope after the `t-coloranchor-corpus-gate` go/no-go. The corpus proved color is an unreliable HIGH_FP discriminator (same-text color unstable; Monokai's normal foreground is non-default), and that **content + position + working-marker preserves detection** for RateLimit/ServerRateLimit (5/5 prose suppressed; FN-covered across claude/gemini/opencode/kiro incl. the kiro `ThrottlingException` fix in #1789).

## How

New `requires_red_anchor` splits the anchor gate:

| State | Anchor | Why |
|-------|--------|-----|
| ContextFull, ModelUnsupported | **keep RED** | ContextFull has no fixture corpus; ModelUnsupported never auto-clears + suppresses hang-check → verbatim-FP cost too high (#1634) |
| RateLimit, ServerRateLimit | **CONTENT** (`in_error_line`) | corpus-proven; a real grey/default fault now latches |

For RateLimit/ServerRateLimit the marker must sit on an error-line-shaped line (`in_error_line`) — no red. The old #1757 net-error red-exemption becomes the **general rule** (so `is_net_error_match` is removed, its const stays as the detection pattern). `is_high_fp_state` stays the full set — the #1518 position gate + #1769 working-marker apply to all four regardless of anchor regime.

**Accepted residual FP** (operator call): a real error line pasted verbatim into the live tail with no working-marker below now latches RateLimit/ServerRateLimit. Acceptable because both **auto-clear + are retry-driven** — a one-off mis-latch self-corrects (unlike ModelUnsupported's never-clear).

## Tests (5 new, pinning the new FP/FN profile)

- ServerRateLimit + RateLimit **plain (no red) → FIRE** via content
- plain prose **without an error-label → still suppressed**
- ContextFull + ModelUnsupported **plain → NOT fire / red → fire** (anchor kept)

The full #1450 FP/FN matrix + `discussion_text` (5 prose) + behavioral_shadow stay green. **Full bin suite 3288 passed**; `fmt --check` + `clippy --all-targets --features tray -D warnings` clean.

> Sensitive — changes the detection FP/FN profile. Requesting **codex adversarial review**: does the content+position+working-marker gate actually preserve detection, and is the accepted verbatim-FP truly self-correcting for these two states?

### DIFF
--- a/src/state/mod.rs
+++ b/src/state/mod.rs
@@ -315,7 +315,13 @@ fn oscillation_guard_window() -> Duration {
     Duration::from_secs(secs)
 }
 
-/// HIGH_FP states require the red-SGR anchor before transitioning.
+/// HIGH_FP states: FP-prone error states whose markers appear in dialectic prose
+/// / JSON dumps, so they get the extra FP defenses (the #1518 position gate + the
+/// #1769 working-marker override + an anchor gate). The ANCHOR gate splits by
+/// [`requires_red_anchor`] — RateLimit/ServerRateLimit now use a content anchor
+/// (`in_error_line`), only ContextFull/ModelUnsupported still require red. This
+/// predicate stays the full set because position + working-marker apply to all
+/// four regardless of which anchor regime they use.
 /// Per #919 spike + dev-2 cross-audit:
 /// - ServerRateLimit / RateLimit: server-side throttle alternations
 ///   include `api_error|timeout_error|overloaded_error` etc which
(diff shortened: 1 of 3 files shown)


suzuke:
✅ **Review mirror — fixup-reviewer (codex) ADVERSARIAL-VERIFIED + lead + CI green incl windows.** ① core: removes the #1450 red-anchor for RateLimit + ServerRateLimit (operator-confirmed, corpus-gate-proven). requires_red_anchor splits HIGH_FP: {ContextFull, ModelUnsupported} keep red; {RateLimit, ServerRateLimit} → content anchor (in_error_line) + #1518 position + #1769 working-marker. #1757 net-error exemption generalized. codex adversarial: FN maintained (plain/grey real errors fire), FP held (no-label prose suppressed), residual verbatim-FP self-corrects (RateLimit 300s expiry + ServerRateLimit retry-storm defenses, no never-clear), ContextFull/ModelUnsupported still red-gated, #1757/#1768 preserved. 3288 tests. This is the data-backed answer to 'can detection stay correct after removing color?' — yes for these two, kept-safe for the other two.

Issue #685, opened 2026-05-12
Hang detection: validate detection accuracy first, then add auto-recovery (F8)


suzuke:
## Background
Surfaced by health audit 2026-05 (5-backend dialectic, raised by auditor-gemini as C-1, AGREE 5/5).

Daemon main loop detects `Hung` state via `check_hang`, but the only consumer-side action is:

```rust
// src/daemon/mod.rs:585-605
tracing::warn!(agent=%name, "hang detected");
```

**No automated recovery action.** No interrupt, no restart, no escalation. Operator must notice the log and manually intervene (typically pressing ESC in the agent's TUI pane).

Direct match to **open #659** (gemini/kiro stuck-in-thinking pattern) and adjacent to **F10** (health decay) and **F39** (Thinking pattern stale screen — already demoted per audit Phase 2).

## Operator decision (2026-05-12)

**Two-phase scope**:

### Phase 1 (prerequisite): validate hang detection accuracy

**Before** building any auto-recovery, harden the detection signal itself. Currently `check_hang` is vulnerable to multiple known false-positive and false-negative vectors:

- **False negative (grey failure)** — F9: any output (1 byte / spinner / log line) resets the silence timer. Agent in infinite loop producing minimal output never triggers detection.
- **False positive (stale screen)** — F39: `AgentState::Thinking` pattern matches against vterm screen text; stale text in scrollback can latch state incorrectly. LATCHED_STATE_EXPIRY (30s) bounds it but not perfectly.
- **Mechanism uncertainty** — F10: `maybe_decay` actually decays on `last_crash.elapsed()`, not `last_output.elapsed()`. The Hung→Healthy transition path is in `check_hang` itself (silence drops below threshold = "recovered"), not in decay. State recovery has no productive-work evidence requirement.

**Phase 1 deliverables**:
1. Audit all transitions into/out of Hung state, document precondition + postcondition per transition
2. Identify productive-work signals beyond raw byte-count silence (e.g. PR push events, MCP tool calls succeeded, structured log markers)
3. Differentiate "legitimately thinking" (long but progressing) from "stuck" (silent + no progress markers)
4. Backend-specific tuning hooks (kiro/gemini stuck pattern from #659 may need different threshold than claude)
5. Cross-reference replay against captured fixtures from past stuck-thinking incidents (#659 + others)

**Acceptance criterion**: false-positive rate < 1% on fixture replay, false-negative rate < 10% on known-stuck cases.

[...]

PR #1791, opened 2026-06-05
fix(#1694a): wire #685 recovery ladder into app mode (Stage1) + Stage2 no-consumer → Stage3

suzuke:
## What

The #685 hung-agent recovery ladder (Stage1 ESC-nudge → Stage2 restart → Stage3 escalate) was allowlisted **out** of app mode (`APP_TICK_ALLOWLIST`), so it was **silently dead in the live daemon** — the live daemon runs `app::run_app`, never `run_core` (the #1720 / #1002 class). The original incident (proxy-drop leaves an agent stuck at the prompt) was never even evaluated for recovery.

Operator-approved scope: **Stage1-only, gradual opt-in**.

## How

1. **Run in app mode** — remove `recovery_dispatcher` from `APP_TICK_ALLOWLIST`. Stage1 (ESC byte → PTY) needs no `crash_rx`. The app-tick completeness invariant now requires + verifies it runs there.
2. **Shadow-gate kept** — all stages stay gated off by default; zero behavior change unless an operator opts in. Stage1-granular opt-in uses the existing `AGEND_AUTO_RECOVERY_STAGE1=1` (no new config — KISS).
3. **Stage2 no-consumer → Stage3 (not silent-drop)** — app passes a throwaway `crash_tx` (rx dropped), so a Stage2 `try_send` would silently vanish (the M-class silent-drop). Thread `stage2_dispatch_available` through `build_default_handlers` (run_core=true, app=false); a new `Stage2Decision {Shadow, EscalateNoConsumer, Fire}` makes `handle_stage2_fire` escalate to Stage3 (graceful pause + operator escalation) when no consumer exists. Stage2 stays gated-off by default, so this only fires if an operator enables Stage2 in app mode.
4. **Anti-thrash** — the existing Stage1 cooldown already dedups rapid nudges.

## Tests

- `recovery_dispatcher_runs_in_app_mode_1694a` — pins it runs in app + is not allowlisted
- `stage2_decision_escalates_on_no_consumer_else_fires_1694a` — gate-off→Shadow, consumer→Fire, **no-consumer→EscalateNoConsumer** (the silent-drop fix; mirrors the `Stage1Branch`/`classify_branch` testable-decision pattern since no AgentHandle/PTY harness exists)

Completeness invariant + full bin suite **3290 passed**; fmt + clippy clean.

## Follow-ups (out of Stage1-only scope)

- Stage2 → app mode's pane-based respawn (the "proper" Stage2-in-app fix)
- runtime-config toggle (vs env-only)

> Sensitive — autonomous-recovery behavior. Requesting **codex adversarial review**: is the shadow-default truly zero-behavior-change, is the Stage1 wiring correct, and does the no-consumer escalation actually prevent the silent-drop?

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -135,22 +135,30 @@ pub fn run(fleet_path_override: Option<&str>) -> Result<()> {
 /// NEW handler is added to `build_default_handlers` but is neither run in app nor
 /// listed here — so additions are a conscious decision, not silent drift.
 ///
-/// - `recovery_dispatcher`: coupled to run_core's `crash_rx` + `handle_crash_respawn`,
-///   which app-standalone has no equivalent of — it does pane-based respawn in the
-///   TUI loop. (#685 dispatch-recovery is shadow-default-off anyway.)
 /// - `snapshot_rotation`: app owns session persistence via `session::save_session_if_changed`.
 /// - `thread_dump`: env-gated diagnostic, not needed in the interactive TUI.
-const APP_TICK_ALLOWLIST: &[&str] = &["recovery_dispatcher", "snapshot_rotation", "thread_dump"];
+///
+/// #1694(a): `recovery_dispatcher` was REMOVED from this allowlist — the live
+/// daemon runs in app mode (`app::run_app`, never `run_core`), so allowlisting it
+/// out meant the #685 recovery ladder was silently dead in the live daemon (the
+/// #1720 class). It now runs in app mode: Stage1 (ESC-nudge to the PTY) needs no
+/// `crash_rx`; Stage2 (restart) has no app-mode consumer, so it escalates to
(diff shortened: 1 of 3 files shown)


suzuke:
✅ **Review mirror — fixup-reviewer (codex) ADVERSARIAL-VERIFIED + lead + CI green incl windows.** #1694 (a): wires #685 recovery_dispatcher into app mode (was allowlisted out — #1720-class). Adversarial-confirmed: shadow-default = ZERO behavior change (opt-in per-stage via AGEND_AUTO_RECOVERY_STAGE1/2/3); Stage1 writes exactly ESC via timeout-safe write_to_pty; no-consumer Stage2 → Stage3 escalation (not silent-drop); gates independent; anti-thrash intact.

Issue #1814, opened 2026-06-06
Root fix: self-healing restart_daemon via successor handoff (retire supervision-detection brick gate; supersedes #1812 class)


suzuke:
### Why (root cause, from #1812 dialectic)
`restart_daemon` today is "cooperative exit + external supervisor respawn": set RESTART_PENDING → graceful shutdown → `exit(42)` → an external supervisor (launchd/systemd/wrapper) respawns. Whether the fleet survives a restart hinges on `is_restart_supervised()` correctly answering **"will something respawn me after I exit?"**

That detection is a **static proxy (env vars) for a dynamic fact (is a live respawner present now)**. Both #851 (exit with no respawner → brick) and #1812 (ambient `XPC_SERVICE_NAME` false-positive → exit with no respawner → brick) are the *same failure class*: the guess is wrong → the daemon exits → nothing brings it back → control plane bricks until manual recovery. The #1812 sentinel patch reduces false-positives but does NOT leave the class (orphaned / post-install-disabled supervisors still fail-open).

### Proposed model change
Make `restart_daemon` **own its own respawn** so it never depends on external supervision being correctly detected:

1. On restart request, the running daemon **starts its successor itself** (fork+exec / re-exec of the new on-disk binary), handing off:
   - the API listening socket (design choice — see open questions),
   - re-attachment to the already-running agent processes (the existing post-restart re-attach path already does this; the successor inherits the same agents).
2. The old daemon **confirms the successor is up, listening, and healthy** (handshake) **before** it exits.
3. If the successor fails to come up / fails health → the old daemon **aborts the restart and stays alive** (no brick), surfaces an error.
4. External launchd/systemd/wrapper then only matter for **crash recovery**, not operator-initiated restart. `is_restart_supervised()` downgrades to advisory.

Net: operator/agent restart can never brick the control plane, regardless of supervision detection. Resolves the #851/#1812 class at the root.

### Open design questions (for dialectic)
- **Socket handoff**: close-and-rebind-with-retry vs `SO_REUSEADDR/SO_REUSEPORT` overlap vs fd-passing the listener (SCM_RIGHTS) vs re-exec-in-place preserving the fd via `execve`. Which avoids both "double-bound" and "window with no listener"?
- **Agent preservation**: confirm the successor re-attaches to running agent PTYs/processes exactly as the current external-respawn path does; no orphaned panes.
[...]

Issue #1812, opened 2026-06-06
Bug: is_restart_supervised() false-positives on macOS (ambient XPC_SERVICE_NAME) — #851 fail-closed guard defeated


suzuke:
### Severity
P1 — would brick the daemon on `restart_daemon` (the exact #851 incident the guard exists to prevent). Currently masked only by the operator-Away gate.

### Root cause
`is_restart_supervised()` (`src/daemon/restart.rs:54`) returns true if ANY of `AGEND_WRAPPED` / `XPC_SERVICE_NAME` / `INVOCATION_ID` is *present* (`has_env` = `var_os().is_some()`, value-agnostic).

The module doc claims `XPC_SERVICE_NAME` is "set by macOS launchd for every service it spawns. Reliable indicator that the daemon is running under launchd with a KeepAlive policy." **This premise is empirically false.** `XPC_SERVICE_NAME` is ambient in any macOS GUI/Aqua session — set on essentially every process, value `0` meaning "not itself an XPC service." It does NOT indicate a launchd KeepAlive supervisor will respawn on exit.

### Evidence (this machine, 2026-06-06)
Live daemon PID 53352 (`agend-terminal app`): parent = `/bin/zsh --login` (manual launch from kiro-cli-term), **no LaunchAgent plist, not in `launchctl list`, no wrapper** — i.e. genuinely unsupervised. Yet:
- daemon 53352 env: `XPC_SERVICE_NAME=0`
- unrelated claude agent procs (fixup-dev): `XPC_SERVICE_NAME=0`
- unrelated kiro terminal zsh: `XPC_SERVICE_NAME=0`

All three are NOT launchd-KeepAlive services, all carry the var ⇒ `is_restart_supervised()` returns true for the whole fleet on macOS ⇒ the #851 fail-closed guard is a no-op on the primary platform (only correctly fails-closed on Linux bare-start, which has no XPC).

### Impact scenario
With operator mode = active, any `restart_daemon` call → guard says "supervised" → daemon sets RESTART_PENDING, graceful shutdown, `exit(42)` → **nothing respawns it** → daemon/TUI dead, all MCP calls hang until manual relaunch (#851 incident reproduced).

### Proposed fix direction (to be confirmed by dialectic)
Stop trusting ambient OS env. Detect supervision via a **positive sentinel we control**:
1. Drop the `XPC_SERVICE_NAME` signal (false-positive source).
[...]

Issue #1339, opened 2026-05-27
Operator Mode: active/away/sleep/dnd with delegation and daemon enforcement


suzuke:
## Summary

Add an operator mode system that changes fleet behavior based on the operator's availability. Modes display on the TUI status bar, and agents adapt their autonomy level accordingly.

## Modes

| Mode | Operator Status | Decision Authority | Notifications |
|---|---|---|---|
| `active` | At TUI | Operator | All |
| `away` | Reachable via Telegram | Operator (via TG) | Important only |
| `sleep` | Unreachable | Delegated agent/lead | P0 only |
| `dnd` | At computer, don't interrupt | Operator | P0 only |

## Behavior Matrix

| Behavior | active | away | sleep | dnd |
|---|---|---|---|---|
| New task dispatch | Manual | Manual (via TG) | Delegate can dispatch | Queued for active |
| Structural ops (create/delete instance, daemon restart) | Allowed | Blocked | Blocked | Blocked |
| PR merge | Normal flow | Normal flow | Auto (CI green + verified) | Queued for active |
| Fleet changes | Allowed | Blocked | Blocked | Blocked |

## Delegation Mechanism (sleep mode core)

```yaml
mode:
  current: sleep
  delegate_to: fixup-lead
  delegate_scope:
    - task_dispatch
    - decision_approve
    - pr_merge
  blocked_scope:
    - fleet_structure
    - protocol_change
    - budget_sensitive
```

Delegate gets limited proxy authority, not full operator access. Operations outside scope are queued until operator returns.

## Daemon Enforcement

```rust
fn check_operation_allowed(op: &Operation, mode: &OperatorMode) -> Result<()> {
    match mode.current {
        Mode::Sleep | Mode::Dnd => {
            if op.is_structural() { return Err("blocked in sleep/dnd mode") }
        }
        Mode::Away => {
            if op.is_structural() { return Err("blocked in away mode") }
        }
        Mode::Active => {}
    }
    if mode.current == Mode::Sleep && op.requires_operator() {
        if !mode.delegate_scope.contains(op.kind()) {
            return Err("outside delegate scope, queued for operator")
        }
    }
    Ok(())
}
```

## Mode Switching

**Manual:**
- TUI: `Ctrl+B M` → mode selector overlay
- CLI: `agend-terminal mode sleep --delegate fixup-lead`
- Telegram: `/mode sleep` or `/away`

**Automatic (optional):**
[...]

Issue #851, opened 2026-05-16
restart_daemon MCP tool: wrapper script does not restart daemon after exit(42)


suzuke:
## Symptom

\`restart_daemon\` MCP tool returns \`{ok: true, restart: \"pending\", note: \"daemon will exit(42) after graceful shutdown; wrapper script restarts\"}\` and the daemon does exit with code 42. **But the wrapper script does not restart the daemon — operator must manually restart.**

## Repro (verified 2026-05-16 06:08 UTC)

Sequence:
1. \`mcp__agend-terminal__restart_daemon\` called → response \`{ok: true, restart: pending}\`
2. Old daemon (PID X) shuts down gracefully + exits 42
3. **No new daemon process spawns** — confirmed via \`ps aux | grep agend-terminal\` showing only the old leftover process briefly
4. All subsequent MCP calls return \`MCP error -32603: daemon proxy error: Resource temporarily unavailable (os error 35)\`
5. Operator manually starts daemon → fleet recovers

Operator quote: \"你 restart 失敗了，是我手動幫你 restart 的\"

## Root cause hypotheses

The wrapper script that should respawn daemon on exit-42 either:
1. Isn't installed in this operator's environment (assumed-default that isn't actually default)
2. Is installed but doesn't fire — maybe shebang issue, permission issue, or runs in wrong cwd
3. The exit-42-detect mechanism (parent process / launchd / supervisord pattern) isn't wired up

Reference: PR #648 \`safe daemon restart via exit(42) + wrapper script\` introduced the mechanism. Worth checking that PR's wrapper script ships in default installs.

## Suggested fix

1. **Audit wrapper script installation path** — does \`agend-terminal start\` actually wrap the daemon with the exit-42-respawn loop? Or is that only when started via specific entry point (e.g., \`launchctl\` plist generated by \`agend-terminal service install\`)?
2. **Document the requirement** — \`restart_daemon\` tool description should say \"requires wrapper script (\`agend-terminal service install\` or manual launchctl/systemd setup); falls back to no-op restart if wrapper missing\"
3. **Detect + warn** — if daemon doesn't see a wrapper indicator (env var? parent PID check?), \`restart_daemon\` should return error instead of \`{ok: true}\` so caller knows to escalate
4. **Operator-side recovery** — until the above lands, document the fallback: \"if MCP calls return \`Resource temporarily unavailable\` after \`restart_daemon\`, manually run \`agend-terminal start\` to recover\"

## Acceptance test

[...]

PR #1817, opened 2026-06-06
feat(#1814): Stage 1 — flag-gated self-respawn via successor handoff

suzuke:
## What

Stage 1 of #1814: make `restart_daemon` self-healing so it never bricks the control plane regardless of supervision detection (the #851/#1812 failure class). Behind flag `AGEND_RESTART_HANDOFF=1`; **flag OFF = byte-identical legacy `exit(42)` path** (zero regression).

Per decision `d-20260606104934346030-2` + the 4-way design dialectic. Scope is **P1** (control-plane self-respawn; agents re-spawn + backend `Resume`, exactly as a supervised restart does today). **P2** (zero-downtime live-PTY survival) is explicitly out of scope/deferred.

## How (flock = commit point)

1. **HANDOFF boot path** — a successor spawned with a legitimate `AGEND_SUCCESSOR_HANDOFF=<old_pid>:<token>` marker takes a minimal pre-lock path: create run dir + cookie, bind its own ephemeral api, write `control-ready`, then **block on the flock**. It does **not** run the destructive sole-daemon reconciles / agent-resolve / agent-spawn during the two-daemon overlap window — those would race the predecessor (e.g. `release_inprogress_orphans`, which assumes `live = ∅`, would falsely release the predecessor's in-progress tasks; `boot_sweep_zombies` could kill the predecessor). A normal/operator `start` never carries the marker, so it always runs the full `prepare`.
2. **Phase-1 health gate** — the predecessor (in its `mcp_tool` worker thread) confirms the successor: process alive (`try_wait`, which also reaps a crash-on-launch so a zombie can't masquerade as alive) + `control-ready` present + a real cookie-authenticated api round-trip (`api::call_at` against the successor's own run dir).
3. **Commit** — set `RESTART_PENDING` only; the existing `handle_session` bridge flips the shutdown flag → main loop breaks → `run_core` tail sees `RESTART_PENDING` + self-respawn → `exit(0)`, releasing the flock so the (blocked) successor promotes. We deliberately do **not** `api::call(SHUTDOWN)` — during overlap its `find_active_run_dir` is ambiguous and was observed delivering SHUTDOWN to the freshly-bound successor.
4. **Abort-stay-alive** — on Phase-1 failure: SIGKILL + reap + remove the successor's run dir; the predecessor was never signalled, so it stays fully alive with its agents intact.

## Key pieces
- `daemon/restart.rs`: `AGEND_RESTART_HANDOFF` flag + `AGEND_SUCCESSOR_HANDOFF` marker (well-formed-token guard).
- `bootstrap`: split `prepare` into reusable `boot_hygiene_sweeps` + `resolve_fleet_and_reconcile` (run post-lock by the handoff path) + `prepare_handoff_prelock` + `acquire_daemon_lock_blocking`; `OwnedFleet.lock → Option`.
- `daemon`: `FleetSource::HandoffDeferred` `run_core` branch; `exit(0)` on self-respawn.
- `bootstrap/daemon_spawn::spawn_successor_handoff`; `api::call_at` / `ipc::connect_run_dir_api`; `restart_daemon` `tool_timeout` 30s→60s.

## Known/intended Stage-1 interactions (NOT bugs)
- **launchd `KeepAlive` double-respawn**: if an external supervisor is also active, the predecessor's exit triggers both the successor and an external respawn. **Self-corrects via the flock** — whoever acquires `.daemon.lock` first wins; the loser hits the singleton attach-reject guard and exits. Documented inline.
- **Operator-mode gate**: `restart_daemon` is still gated by `operator_mode` when the operator is Away (existing #1339/#1576 authority gate) — self-respawn does not bypass it. The integration tests set mode Active to exercise the mechanism.

## Tests (§3.9 real-entry, not mocked)
`tests/self_respawn_handoff.rs` drives the **real** flow: real daemon binary → real `restart_daemon` over the api → real spawned successor → real Phase-1 gate.
- `self_respawn_succeeds_with_no_external_supervisor` — `AGEND_RESTART_HANDOFF=1`, **no** supervisor env: a NEW pid serves, the old pid is gone, agents re-spawned, exactly one active run dir.
- `self_respawn_aborts_and_old_stays_alive_when_successor_fails` — `AGEND_FORCE_SUCCESSOR_FAIL` injection: restart returns `ok:false`, the SAME predecessor keeps serving.

Unix-first (Stage 1); Windows keeps `exit(42)` + Task Scheduler.

Verification: `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, restart unit tests (11), and both integration tests all green locally (rebased onto `origin/main` incl. #1812).

### DIFF
--- a/src/api/handlers/mcp_proxy.rs
+++ b/src/api/handlers/mcp_proxy.rs
@@ -25,9 +25,12 @@ pub(crate) fn tool_timeout(tool: &str) -> Duration {
         | "list_deployments" | "set_waiting_on" | "set_display_name" | "set_description"
         | "react" | "report_health" => Duration::from_secs(5),
 
-        // Slow tools that spawn processes or do network I/O (~60s)
+        // Slow tools that spawn processes or do network I/O (~60s).
+        // #1814: restart_daemon (self-respawn) spawns a successor + runs a
+        // ≤30s Phase-1 health gate before returning — give it the 60s budget so
+        // the gate can't collide with the default 30s tool timeout.
         "create_instance" | "deploy_template" | "replace_instance" | "watch_ci"
-        | "checkout_repo" => Duration::from_secs(60),
+        | "checkout_repo" | "restart_daemon" => Duration::from_secs(60),
 
         // Default for everything else (~30s)
         _ => Duration::from_secs(30),
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -696,6 +696,29 @@ fn spawn_one(
 /// if the server does not reply `{"ok":true}`. The cookie file has mode
 /// 0600 so only the daemon's user can read it — this is the peer-UID
 /// substitute for TCP loopback (see `auth_cookie.rs`).
+/// #1814: like [`call`] but targets a SPECIFIC run dir (cookie + api.port read
+/// from `run_dir`), not the active daemon. Used by the self-respawn Phase-1
+/// gate so the predecessor can do a real cookie-authenticated round-trip
+/// against the successor's own control plane while both are briefly alive. A
+/// short fixed read timeout (the gate retries) keeps a flaky successor from
+/// hanging the handler. No self-IPC guard: this connects to a DIFFERENT
+/// process's socket, so the registry-lock deadlock class does not apply.
+pub fn call_at(run_dir: &Path, request: &Value) -> anyhow::Result<Value> {
+    let stream = crate::ipc::connect_run_dir_api(run_dir)?;
+    stream
+        .set_read_timeout(Some(std::time::Duration::from_secs(2)))
+        .context("set call_at read timeout")?;
+    let cookie = crate::auth_cookie::read_cookie(run_dir)?;
+    let mut writer = stream.try_clone()?;
+    let mut reader = BufReader::new(stream);
+    crate::auth_cookie::client_handshake_ndjson(&mut reader, &mut writer, &cookie)?;
+    writeln!(writer, "{request}")?;
+    writer.flush()?;
+    let mut line = String::new();
+    reader.read_line(&mut line)?;
+    Ok(serde_json::from_str(&line)?)
+}
+
 pub fn call(home: &Path, request: &Value) -> anyhow::Result<Value> {
     // #1492: self-IPC over the loopback socket. If the caller holds the
     // registry lock, the API handler servicing this call needs the same lock →
--- a/src/bootstrap/daemon_spawn.rs
+++ b/src/bootstrap/daemon_spawn.rs
@@ -113,3 +113,68 @@ pub struct DaemonHandle {
     pub run_dir: PathBuf,
     pub log_path: PathBuf,
 }
+
+/// #1814: handle to a spawned successor daemon (self-respawn handoff).
+pub struct SuccessorHandle {
+    /// The successor daemon's pid. `start --foreground` does NOT re-exec, so
+    /// the spawned child IS the daemon — `child.id()` is authoritative.
+    pub pid: u32,
+    /// The successor's run dir (`home/run/<pid>`) — where the Phase-1 gate
+    /// looks for `control-ready` + `api.port` + `api.cookie`.
+    pub run_dir: PathBuf,
+    /// The live child handle. Held so the Phase-1 gate can `try_wait()` —
+    /// detecting a crash-on-launch immediately AND reaping it (vs. `kill(pid,
+    /// 0)`, which reports a zombie as still alive). On the COMMIT path the
+    /// handle is dropped when the predecessor exits; `std::process::Child::drop`
+    /// neither waits nor kills, so the promoted successor keeps running.
+    pub child: std::process::Child,
+}
+
+/// #1814: spawn a successor daemon for self-respawn handoff. Like
+/// [`spawn_detached`] but (a) injects `AGEND_SUCCESSOR_HANDOFF=<value>` so the
+/// child takes the minimal pre-lock handoff boot path (bypassing the singleton
+/// attach-reject guard, deferring flock + destructive reconciles), and (b)
+/// does NOT wait for readiness — the caller (restart handler) runs the Phase-1
+/// health gate against the returned run dir, then either commits (signals the
+/// predecessor to exit) or aborts (kills this successor). The child inherits
+/// the predecessor's env (incl. `AGEND_RESTART_HANDOFF=1`), so the successor will
+/// itself self-respawn on a later restart.
+pub fn spawn_successor_handoff(home: &Path, handoff_value: &str) -> Result<SuccessorHandle> {
+    let exe = std::env::current_exe().context("resolve current_exe for successor spawn")?;
+    std::fs::create_dir_all(home).with_context(|| format!("create home {}", home.display()))?;
+
+    let mut cmd = Command::new(&exe);
+    cmd.arg("start").arg("--foreground");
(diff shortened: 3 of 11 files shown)


suzuke:
REJECTED

### Evidence
ran: `git fetch origin pull/1817/head && git rev-parse FETCH_HEAD` -> `f04f071e2d248590258cf662d5ea4113621d38e0`.
ran: `cargo test --test self_respawn_handoff --features tray` -> passed, 2 tests.
ran: `gh pr checks 1817` -> failed/non-green: Coverage failed; macOS, Ubuntu, and Windows checks pending.
cited: `src/bootstrap/mod.rs:392`-`398` writes a normal successor run dir with `.daemon` and `api.cookie` before flock ownership.
cited: `src/api/mod.rs:226`-`238` publishes successor `api.port` immediately.
cited: `src/daemon/mod.rs:223`-`245` accepts any live `.daemon` run dir as active; `src/ipc.rs:67`-`72` routes normal clients through that discovery.
cited: `src/mcp/handlers/restart.rs:140`-`147` commits Phase-1 on `control-ready` plus STATUS before lock/reconcile/agent spawn; `src/daemon/mod.rs:620`-`635` shows lock, reconciles, and agent spawn happen later.

Findings:

1. High: normal clients can discover and talk to the staged successor during the overlap window. The PR fixed the internal SHUTDOWN misroute by avoiding `api::call(home, SHUTDOWN)`, but external clients still use `find_active_run_dir` via `connect_api`. Because the successor writes `.daemon`, issues `api.cookie`, and publishes `api.port` before it owns the flock, a CLI/MCP bridge/app can attach to the half-promoted successor while the predecessor is still the real owner. That is still a split-brain discovery window; it can serve STATUS/LIST and potentially route commands against an incomplete daemon state.

2. High: the Phase-1 commit point is still too early for the stated lifecycle invariant. `phase1_gate` returns true once the successor has `control-ready` and answers STATUS, but `run_core` only then waits for predecessor flock release, runs hygiene/reconcile, resolves the fleet, and spawns agents. If the successor fails after Phase-1 but before or during promotion, the predecessor has already set `RESTART_PENDING` and will exit. The current tests cover crash-before-control-ready and common happy path, but not a half-ready successor or failure after Phase-1.

3. Blocking: PR CI is not green. `gh pr checks 1817` currently exits non-zero with Coverage failed and the three platform checks pending, while the dispatch explicitly required green checks.

Notes: flag-off behavior appears correctly gated by `AGEND_RESTART_HANDOFF == "1"`, and the RESTART_PENDING-only commit bridge is directionally correct for avoiding the old self-SHUTDOWN misdelivery. Abort-stay-alive passes for the injected crash-before-control-ready path.

suzuke:
REJECTED

### Evidence
ran: `git fetch origin pull/1817/head && git rev-parse FETCH_HEAD` -> `6a997c7de0e3778185380473cdcf127e61ea8852`.
ran: `cargo test --test self_respawn_handoff --features tray` -> passed, 3 tests.
ran: `cargo test find_active_run_dir_skips_dir_without_daemon_identity --features tray` -> passed, 1 matching test.
ran: `gh pr checks 1817` -> all listed checks pass: macOS, Ubuntu, Windows, Coverage, LOC overrun, audit.
cited: `src/bootstrap/mod.rs:392`-`407` no longer writes `.daemon` in pre-lock successor prep.
cited: `src/daemon/mod.rs:281`-`329` skips alive run dirs without valid `.daemon`; `src/daemon/mod.rs:707`-`712` writes `.daemon` only after acquiring the flock.
cited: `src/ipc.rs:75`-`85` keeps Phase-1 successor probing pinned to the successor run dir, not active discovery.
cited: `src/mcp/handlers/restart.rs:91`-`99` parks the successor handle before setting `RESTART_PENDING`; `src/daemon/mod.rs:194`-`210` clears `RESTART_PENDING` and keeps running if `try_wait` shows the successor died.
cited: `src/daemon/mod.rs:738`-`745` runs that liveness check at the main-loop shutdown decision; `src/daemon/mod.rs:800`-`831` still performs teardown, removes the predecessor run dir, sleeps, and exits after that check.

Fixed from the prior review:

1. The split-brain discovery finding is closed. The successor can still publish `api.port` for the predecessor's explicit `connect_run_dir_api` Phase-1 probe, but without `.daemon` it is invisible to normal `find_active_run_dir` clients until it owns the flock. The guard test covers the no-`.daemon` shape and preserves the run dir instead of deleting it.

2. The originally tested early-death-after-Phase-1 path is covered. The successor child handle is parked before `RESTART_PENDING`, and the main loop aborts-stay-alive if `try_wait` observes that the successor died before the shutdown decision. The new real-entry test covers this seam.

Remaining blocker:

High: the liveness recheck still has a TOCTOU window between `confirm_shutdown_or_abort_respawn` returning true and the predecessor actually exiting/releasing the flock. After the check, the predecessor proceeds through shutdown teardown, writes `daemon_stop`, removes its own run dir, sleeps for 1s, and then exits. If the successor dies during that interval, the predecessor no longer rechecks and will still exit, so the control plane can still brick in the same broad "successor died before promotion" class. The new test kills the successor early enough for the next main-loop check to catch it; it does not cover death after that check but before predecessor exit.

suzuke:
VERIFIED

### Evidence
ran: `git fetch origin pull/1817/head && git rev-parse FETCH_HEAD` -> `828935d908a9b508c30c4c87b5a1f8cc3e00e06e`.
ran: `cargo test --test self_respawn_handoff self_respawn_recovers_as_primary_when_successor_dies_during_teardown --features tray` -> passed, 1 test.
ran: `cargo test --test self_respawn_handoff --features tray` -> passed, 4 tests.
ran: `gh pr checks 1817` -> all listed checks pass: macOS, Ubuntu, Windows, Coverage, LOC overrun, audit.
cited: `src/daemon/mod.rs:828`-`840` runs `shutdown_sequence` and logs `daemon_stop` before the final gate, while the run dir/API are still intact.
cited: `src/daemon/mod.rs:842`-`877` places the final self-respawn recheck before `remove_dir_all(run_dir(home))`; dead successor clears `RESTART_PENDING`, clears shutdown, removes `restart-requested`, re-spawns agents, and `continue 'serve`; live successor removes run dir and exits immediately.
cited: `src/daemon/mod.rs:762`-`767` wraps the serve loop so recover-as-primary can resume the primary without leaving `run_core`.
cited: `src/daemon/mod.rs:1169`-`1185` drains the registry during shutdown, and `src/daemon/mod.rs:1015`-`1045` re-spawns/registers the same fleet agents and rewrites `.ready`, which is sufficient because the run dir, cookie, API server, tick infrastructure, and daemon context remain alive.
cited: `src/daemon/mod.rs:717`-`727` provides the `AGEND_FORCE_SUCCESSOR_FAIL_DURING_TEARDOWN` seam, and `tests/self_respawn_handoff.rs:419`-`467` verifies same-pid recover-as-primary plus re-served agent after that seam.

The round-2 TOCTOU finding is closed. The irreversible final gate now sits after agent teardown but before run-dir removal/flock release, so the predecessor can recover as primary while its API identity is still live. If the successor is alive at that final gate, the code removes the run dir and calls `exit(0)` immediately, with no extra sleep/teardown between check and exit; the remaining check-to-exit syscall interval is the expected irreducible residual.

I do not see a new race from the `'serve` loop refactor: recovery clears the restart/shutdown flags, drops the dead successor handle, reuses the still-live daemon context, re-spawns the drained fleet, and returns to the same select loop. The existing earlier abort path still handles death before teardown, and the final test covers death during teardown.

Issue #1809, opened 2026-06-05
Bug: Feedback-loop retry storm triggered by stale error strings in VTerm buffer


cheerc:
### Description

A critical "feedback loop" bug exists where stale error strings in the active VTerm buffer (the visible screen) trigger redundant auto-retries, even if the agent is already recovered. This is a severe side-effect of #1757 (which removed the red anchor requirement for network errors).

### Root Cause Analysis

1.  **VTerm Buffer Persistence**: When an agent recovers from a `ServerRateLimit` but produces only a small amount of output (e.g., a short task summary), the original `API Error: InvalidHTTPResponse` or `ECONNRESET` string remains sitting in the active VTerm buffer (`0..rows`).
2.  **Detection Bias**: The daemon's state detection scans this active buffer. Because #1757 removed the red color requirement for these network errors, the presence of the grey/white string is sufficient to latch the `ServerRateLimit` state again.
3.  **The "Ghost" Injection**: The daemon perceives the old string as a *new* error event and schedules a retry. 5 seconds later, it injects `[AGEND-AUTO kind=ratelimit-retry] continue` into the PTY.
4.  **Feedback Loop**: In a particularly meta scenario, if an agent (like the one writing this issue) prints a log or a discussion containing these error tokens, the daemon scans the agent's *own output*, detects the tokens as errors, and injects a "continue" back into that same agent.

### Evidence (Redacted Logs)

The following logs show `gemini-40b379` (this agent) being hit by a ghost retry simply because it was printing an analysis of a previous error:

```
// Agent is printing an analysis containing the error token "ECONNRESET"
2026-06-06T02:42:57.307099+08:00  WARN #1450: HIGH_FP pattern matched but rendered fg not red ... matched=ECONNRESET line_context=... (短暫的 TCP 連結錯誤（如 ECONNRESET）後 ...

// Daemon sees the token in the buffer, schedules a retry
2026-06-06T02:45:13.677767+08:00  INFO ServerRateLimit detected, scheduling retry (Phase A) agent=gemini-40b379 delay_secs=5

// Ghost "continue" is injected into the active prompt
2026-06-06T02:45:25.499588+08:00  INFO ServerRateLimit: injected "continue" (attempt 1) agent=gemini-40b379 retry=1
```

### Proposed Fix

State detection needs a concept of "freshness" or "positional relevance" to distinguish between active errors and historical artifacts.

[...]

Issue #1808, opened 2026-06-05
API Error/ServerRateLimit recovery logic prevents `continue` inject when agent resumes on its own


cheerc:
The user reported an API Error/ServerRateLimit around 02:30 AM where the TUI did not inject `continue`. Upon investigation, the agent (dev-team-lead) did recover and transition to Idle on its own after 42 seconds. Commit 44b2795's RECOVERY_SILENCE (45s) protection prevented the `continue` inject during this time because the agent had recent productive output. The system cleared the rate limit retry track properly when the agent transitioned to Idle at 02:31:55. This confirms that the lack of `continue` injection was not a bug but a successful mitigation of a retry storm during an auto-recovering transient network glitch.

cheerc:
UPDATE: Critical Bug Evidence Found

After re-evaluating the user's report (`my screen was frozen, over 45s agent had no action and no PTY nudge was injected`) and cross-referencing `state-transitions.jsonl`, I discovered a severe flaw in the #1713 / #1795 recovery logic for backends like ClaudeCode that abort to `Idle` after a transient error.

### What Actually Happened:
1. **02:31:13**: `ServerRateLimit` detected. Track created.
2. **02:31:19**: The backend (Claude) entered an internal wait/timeout and printed a `Thinking` marker. The daemon's `working_state_below` parsed this, transitioning the state from `ServerRateLimit` to `Thinking`.
   * *Because the state was no longer `ServerRateLimit`, Phase 1 logic bypassed the `continue` injection.*
3. **02:31:49**: Claude's internal timeout gave up. It aborted the request and printed the `Idle` prompt (`❯`).
4. **02:31:55**: The daemon tick observed the `Idle` state and hit the `clears_server_rate_limit_retry(state)` branch. It completely wiped the retry track, logging `agent recovered (Idle)`.
5. **02:32:35**: The agent sat dead at the `Idle` prompt for 46 seconds until `waiting_on_stale` eventually fired.

### The Bug:
The assumption in #1713 that **`Idle` == Genuine Recovery** is false when an agent aborts due to an API Error. 
When Claude hits a fatal network error, it drops back to the `❯` prompt. The daemon sees `Idle`, assumes the error is resolved, and clears the track. The agent never receives the `continue` nudge needed to resume the aborted work, leaving the TUI permanently frozen.

### Required Fix:
[...]

PR #1832, opened 2026-06-07
instrument: SRL-swallow probe — all backends + previously-silent Path B (refs #1809)

suzuke:
## What

Instrument-only (**zero behavior change**) — adds logging so the next live ServerRateLimit incident says exactly which gate swallowed the SRL. Does **not** resolve the underlying bug (refs #1809; intentionally not closing).

The live incident — a claude worker hit SRL but the daemon never latched it → no auto-retry → stuck — was invisible to the existing `#1808-flaw2-probe` because that probe was scoped to **Agy/Kiro** and covered only **one** of the two gates that can swallow an SRL.

## Changes (`src/state/mod.rs`, logging only — `landed` byte-identical)

- **Path A** (`working_state_below` override): drop the Agy/Kiro backend scope → log for **every backend incl. claude**. Records the winning marker, `landed_state`, and the **raw** `recovered_within` bool + `productive_silent_secs` + `dist_from_bottom` as independent fields (not just the pre-judged WARN/INFO level — the level is derived from `recovered_within`, itself a suspect). WARN/INFO split retained.
- **Path B** (`recovered_within`→Idle fallback): previously **SILENT** — the gap that hid the claude incident. Now emits an INFO probe with the same raw fields.
- **trivial**: `log_anchor_suppress` WARN wording was stale (claimed "not red"). Post-#1790 the anchor gate has two regimes — red-SGR (ContextFull/ModelUnsupported) and content/error-line (`in_error_line`, for RateLimit/ServerRateLimit). Reworded + shows `<unset>` for an empty `agent=` field.

## Tests (§3.9)

Two real-entry `feed_with_fg` tests assert the probe names each gate (`path=recovered_within_idle` / `path=working_state_below`). The probe's custom `target: "state_detection"` is dropped by `tracing_test`'s crate-scoped filter, so the tests install a manual all-targets capturing subscriber.

## Verification

- `cargo fmt --check` ✓
- `cargo test --tests --features tray` → **3682 passed, 0 failed** ✓

### DIFF
--- a/src/state/mod.rs
+++ b/src/state/mod.rs
@@ -1001,46 +1001,79 @@ impl StateTracker {
                                 detected
                             };
                             let working_below = patterns.working_state_below(screen_text, matched);
-                            // #1808-flaw2-probe (instrumentation-only, NO behavior
-                            // change): cheerc claims a static bottom status bar (Agy/
-                            // Kiro Thinking = the bare `esc to cancel` token) can sit
-                            // BELOW the error → `working_state_below` overrides a
-                            // ServerRateLimit → masks a stuck throttle. Record it
-                            // empirically: when such an override actually fires on
-                            // Agy/Kiro, log the winning marker + productive silence. NO
-                            // recent productive output ⇒ the override may be masking a
-                            // genuinely-stuck agent (the suspicious case cheerc
-                            // describes) ⇒ WARN; recent output ⇒ INFO (genuine
-                            // recovery). Scoped to Agy/Kiro + SRL-override only (low
-                            // noise); does NOT alter `landed`.
-                            if let Some((_, marker)) = working_below {
-                                if matches!(detected, AgentState::ServerRateLimit)
-                                    && (self.backend_name == Backend::Agy.name()
-                                        || self.backend_name == Backend::KiroCli.name())
-                                {
-                                    let productive_silent_secs =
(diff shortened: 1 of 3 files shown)


suzuke:
**Merge audit (§3.12 mirror)** — dual review VERIFIED + all CI green at head `250870161862e6157d13133740c3a2e66940ce4f`:

- **reviewer (codex)** VERIFIED — zero-behavior lens: byte-diff confirms `fallback`/`working_below`/`.map().unwrap_or(fallback)`/`gate_on_heartbeat`/`transition` unchanged; probe is borrow-only logging, no panic path, gated by hash-early-return + `detected==ServerRateLimit`.
- **reviewer-2** VERIFIED — content: Path A + Path B both covered (Path B = the previously-SILENT `recovered_within→Idle` gap that hid the live claude SRL-swallow bug); all-backend (Agy/Kiro scope removed); marker verbatim + raw `recovered_within`/`productive_silent_secs` as independent fields (not pre-judged by the WARN/INFO discriminator); §3.9 two real-entry `feed_with_fg` tests prove the probe names the right gate.
- **CI**: macos/ubuntu/windows/Coverage/LOC/audit all pass (windows first-run flaked on unrelated `#1744-h2` health-rehydrate underflow → re-run green).

Instrument-only, zero behavior change. Merging --admin squash.

PR #1835, opened 2026-06-08
inbox: quiet compact-clear + GC TTL/size-cap (task t-20260607080714370222-2)

suzuke:
Two complementary inbox-hygiene fixes from synthesis decision `d-20260607081209372642-1` (reviewer (a) + agy (b) + lead synthesis).

## (a) Cheap quiet-clear

`inbox::storage::clear_compact` — sibling of `drain()` (NOT raw drain, which marks everything read and violates trust):
- Same `with_inbox_lock` + tmp+fsync+rename, but sets `read_at` **selectively** and returns **compact** structs (`ClearCompactResult` — never full `InboxMessage`, so it structurally cannot reintroduce the multi-MB blowup).
- New MCP `inbox` **`action="clear"`** branch (explicit — does not overload the no-arg drain). Returns `{cleared_count, kept_unread_count, summaries[≤200]+summaries_omitted, requires_response[]}` with sanitised one-line previews.
- **Trust rule** (`comms_inbox::obligation_reason`): `read_at` = "non-obligation cleared from attention", NOT "obligation accepted". Unanswered **queries** + **tasks** the board can't prove terminal (Done/Cancelled) stay **UNREAD** and surface in `requires_response` (never capped). Uncertain → keep (failure mode = noise, not hidden work).
- Phase 2 consumes `notification_dedup` for cleared rows but **deliberately does NOT** arm the reply-ledger / touch turn-state — clearing historical channel backlog must never fabricate a "must-reply" turn (pinned by a source-scan invariant test).

## (b) GC TTL + size cap

`sweep_expired` already runs per-tick (not a missing-call bug); the root was the **7-day read TTL**.
- Read (drained) **non-blocker** TTL: **7d → 48h** (the high-volume `update`/`report`/`ci`/`poll` rows).
- **Blocker exemption**: drained `query`/`task` rows keep the original **7d** window AND are exempt from the size cap. **OPEN QUESTION resolved**: `has_drained_blocker_for_correlation` (ack-absorption / reply-routing, `messaging.rs:336`) needs these for the *full task turnaround*, which has no finite upper bound (overnight / multi-day) — so 24h was **not** provably safe; the exemption keeps ack-absorption zero-regression.
- **Per-inbox size cap** of **300** most-recent read non-blocker rows — the robust bound a TTL alone can't provide (a burst inside any window still overruns). Unread + blocker rows are never counted nor evicted.

Boundary: (a) mutates `read_at` only (never deletes); (b) deletes by TTL/size but skips unread + blocker rows; both serialise via `with_inbox_lock`.

## Tests (§3.9)

- Real `handle_tool("inbox", {action:"clear"})` entry with a multi-kind backlog (query+task+update+report+superseded) → obligations stay unread + in `requires_response`, chatter cleared.
- Real `task` create→done + clear → terminal task is cleared (board check end-to-end).
- GC: 48h non-blocker boundary, blocker exemption until 7d (+ `has_drained_blocker_for_correlation` still answers after sweep), size-cap keeps newest 300 / never evicts unread+blocker.
- Source-scan: `clear_compact` never references reply_ledger / heartbeat_pair.

## Verification

- `cargo fmt --check` ✓
- `cargo test --tests --features tray` → **3691 passed, 0 failed** ✓
- `file_size_invariant` ✓ (clear handler placed in `comms_inbox.rs`; comms.rs stays <750 LOC)

### DIFF
--- a/src/inbox/mod.rs
+++ b/src/inbox/mod.rs
@@ -25,7 +25,7 @@ pub use disk::{check_disk_space, recover_half_writes};
 
 // Storage CRUD (pub)
 pub use storage::{
-    describe_message, drain, enqueue, find_message, get_thread,
+    clear_compact, describe_message, drain, enqueue, find_message, get_thread,
     has_drained_blocker_for_correlation, mark_ci_watch_superseded, sweep_expired, unread_count,
     unread_of_kind,
 };
--- a/src/inbox/storage.rs
+++ b/src/inbox/storage.rs
@@ -3,6 +3,49 @@ use std::path::{Path, PathBuf};
 
 use super::message::{InboxMessage, MessageStatus};
 
+// ── #inbox-gc retention bounds (decision d-20260607081209372642-1, part b) ──
+//
+// Root cause of the unbounded-looking inbox files: read (drained) messages were
+// retained for 7 DAYS, so a high-throughput agent accumulates 1000s of read
+// rows within that window. Two complementary bounds replace the single 7d TTL:
+//
+// 1. A shorter read TTL for the bulk (`update`/`report`/`ci`/`poll` …), and
+// 2. A per-inbox SIZE CAP on retained read rows — the robust bound a TTL alone
+//    can't provide (a burst inside ANY window still blows past the cap).
+//
+// EXEMPTION: drained `query`/`task` rows are "blockers" — they are read by
+// `has_drained_blocker_for_correlation` (ack-absorption / reply-routing, see
+// storage.rs `has_drained_blocker_for_correlation`) for the full task
+// turnaround, which has no finite upper bound (overnight / multi-day tasks).
+// They keep the original 7d window AND are exempt from the size cap so the
+// audit path never regresses. Unread rows (obligations) keep the 30d window.
+
+/// Read (drained) NON-blocker messages expire this many hours after their
+/// timestamp. Lowered from 7 days — these are the high-volume `update`/`report`/
+/// `ci`/`poll` rows that flood the file.
+const READ_TTL_HOURS: i64 = 48;
+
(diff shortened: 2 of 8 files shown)


suzuke:
**Merge audit (§3.12 mirror)** — dual VERIFIED + CI 6/6 green at head `dd406a0463e3b869a8479596e49beb2aa5df849c`:

- **reviewer-2** (trust-invariant lens) VERIFIED — `clear` can NEVER silently mark a real obligation read: obligation arm never touches `read_at` (kept rows written back unread on disk); the ONLY clear path is an explicit terminal task (Done/Cancelled); board-lookup-fail / ambiguous / no-id / any-non-terminal / query all fail-safe KEEP; `requires_response` has no cap guard (obligations never omitted); never deletes rows; doesn't touch reply-ledger (source-scan test pinned).
- **dev-2** (GC-correctness + ack-absorption zero-regression lens) VERIFIED, tests RAN (9+138 passed incl. concurrency) — `is_blocker_row` (read_at+kind) ⊇ `has_drained_blocker_for_correlation` predicate → exemption is a superset, no needed row swept/capped; size-cap keeps exactly 300 read_non_blocker (deterministic tie-break, no off-by-one); same `with_inbox_lock`+tmp+fsync+rename as drain/sweep; clear Phase-2 only `mark_consumed` (reply_ledger/heartbeat_pair untouched); 48h read TTL only affects ack-absorption which keeps blocker rows 7d → zero regression.

Non-blocking notes (follow-up): §3.9 could add explicit board-resolved boundary fixtures (Done→clear / on-board-non-terminal→keep); `schema_version>current` rows dropped on clear write-back (forward-compat, error-logged); 48h–7d old update/report now `describe_message` NotFound (diagnostic-only, by design).

Merging --admin squash.

PR #1839, opened 2026-06-08
mcp: timeout → accepted_in_progress for side-effect tools (R3#1 candidate 2)

suzuke:
R3#1 dialectic converged on **candidate 2 (timeout-IN_PROGRESS)** after dev-2's adversarial lens vetoed the content-cache (couldn't distinguish a retry from a legitimate identical repeat; no safe TTL) and the IMPL-FEASIBILITY pass ruled candidate 1 (request_id reuse) infeasible (the bridge mints a fresh request_id per logical call; a model-initiated retry is a new logical call it can't link to the original).

## Root cause

On a per-tool timeout, `mcp_proxy` does **not** kill the execution thread — it runs to completion in the background — but returned `{ok:false, error:"timed out"}`. The agent (Claude Code) reads that as a **failure** and **re-issues** the tool call: a fresh logical call with a new bridge `request_id` that `request_dedup` can't catch → a non-idempotent side-effect (send / reply / task-create / spawn) fires **twice**. That's the retry-storm.

## Fix (`src/api/handlers/mcp_proxy.rs`)

- **Timeout arm** branches on `is_side_effect_tool`:
  - side-effect → `{ok:true, status:"accepted_in_progress", note:"…do NOT resend…"}` → the agent sees **acceptance, not failure** → no retry → the orphaned thread completes the action **exactly once**.
  - read/idempotent → keep the retryable `{ok:false, error:"timed out"}` (the agent needs the real result; a repeat is harmless).
- **`is_side_effect_tool`**: explicit read/idempotent allow-list (inbox, lists, `set_*`, health, ci, bind/release, snapshots, download_attachment); everything else — incl. **new** and multi-action-with-write tools — **defaults to side-effect** (conservative: a lost read result is recoverable; a duplicated action is the bug). No runtime dedup / hashing / TTL — exactly the problems that sank the content-cache.
- **Instrument** (`#r3-1-timeout-probe`): emits tool + side_effect + a stable `(tool,args)` content_key on every timeout. This PR's effect (agent stops retrying) is only observable after a daemon **restart** (self-dogfood anti-pattern), so the log lets a post-restart grep confirm the same content_key no longer reappears (retry rate → 0). content_key is log-only, never a dedup decision.
- **Candidate 3 (free defense-in-depth)**, documented: `task` done/update/claim are `task_id`-keyed (a resend is an illegal-transition no-op); `create_instance` is name-keyed (a resend errors, no second spawn). `send`/`reply`/`decision`-post have no natural key — they rely on this gate.

`handle_mcp_tool` refactored to `handle_mcp_tool_inner` with an injectable executor + explicit timeout — the §3.9 seam (drive the timeout path with a sleeping executor + tiny budget; no real tool, no 30s wait).

## Tests (§3.9, real entry)

- side-effect tool timeout → `accepted_in_progress` (ok:true, **no error field** → agent won't resend)
- read tool timeout → retryable `timed out` error
- completed tool → real result (accepted_in_progress is timeout-only)
- classifier covers the storm offenders + reads + unknown-defaults-to-side-effect

## Verification

- `cargo fmt --check` ✓
- `cargo test --tests --features tray` → green (a rare pre-existing global-state flake surfaced once across 5 runs in an unrelated test; the mcp_proxy suite is consistently 7/7 — CI is the arbiter)
- Updated `tests/mcp_proxy_parity.rs` source-scan: `execute_tool` is now passed as the injectable executor (fn pointer), so the invariant asserts the symbol is referenced rather than a literal `execute_tool(` call.

### DIFF
--- a/src/api/handlers/mcp_proxy.rs
+++ b/src/api/handlers/mcp_proxy.rs
@@ -37,6 +37,107 @@ pub(crate) fn tool_timeout(tool: &str) -> Duration {
     }
 }
 
+/// R3#1 candidate 2 (timeout-IN_PROGRESS): does a TIMEOUT on this tool need to
+/// be hidden from the agent as "accepted, completing in background" rather than
+/// surfaced as a retryable error?
+///
+/// **Why**: on timeout the execution thread is NOT killed — it runs to
+/// completion in the background (see `handle_mcp_tool_inner`). If the timeout
+/// surfaces as `{ok:false, error:"timed out"}`, the agent (Claude Code) treats
+/// it as a failure and RE-ISSUES the tool call (a fresh logical call with a new
+/// bridge `request_id`, which `request_dedup` cannot catch) → the action fires
+/// TWICE. For a non-idempotent side-effect tool (send / reply / task-create /
+/// decision-post / spawn …) that is a duplicate message / task / spawn — the
+/// retry-storm root cause. Returning `accepted_in_progress` instead tells the
+/// agent the action was taken (once) and NOT to resend.
+///
+/// The list below is the SAFE set — read-only or idempotent tools where a
+/// repeat is harmless AND the agent genuinely needs the real result, so they
+/// keep the retryable error. Everything else (incl. NEW tools and multi-action
+/// tools with any mutating action) defaults to side-effect — the conservative
+/// choice: a lost read result is recoverable, a duplicated action is the bug.
+///
+/// **Defense-in-depth (candidate 3, free)**: several side-effect tools ALSO have
+/// handler-side natural idempotency keys, so even a model that resends despite
+/// `accepted_in_progress` is protected: `task` done/update/claim are keyed on
+/// `task_id` (a repeat is an illegal-transition no-op), and `create_instance`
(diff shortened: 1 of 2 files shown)


suzuke:
**Merge audit (§3.12 mirror)** — reviewer-2 VERIFIED + CI 6/6 green at head `d2800301`. R3#1 mcp-timeout double-side-effect root-fix (design selected via full dialectic: dev-2 adversarially rejected content-cache → dev feasibility selected timeout-semantics). Only the Timeout arm changed (Ok/Disconnected paths byte-equivalent); is_side_effect_tool whitelist verified all-17 read/idempotent with NO non-idempotent side-effect tool misplaced (storm offenders send/reply/decision-post/task-create/spawn-lifecycle all correctly side-effect → accepted_in_progress; unknown/multi-write conservatively side-effect); accepted_in_progress response (ok:true, no error, 'Do NOT call it again') prevents model retry at the source; #r3-1-timeout-probe instruments for post-restart empirical verification (retry→0). §3.9 bidirectional + success-path real-entry tests. NOTE: core anti-retry effect is empirical, to be confirmed by the probe after the next restart (self-dogfood limit, acknowledged). Merging --admin squash.

PR #1840, opened 2026-06-08
feat(#1813): nudge agent when it answers a channel-origin message without reply

### LINKED ISSUE #1813: Improvement: nudge agent when it answers a channel-origin message without the reply tool (tracked-but-not-enforced)
### Problem
When an agent receives an input carrying a channel prefix (`[user:... via telegram]`), the contract is to respond via the `reply` MCP tool (so the answer routes back to telegram). If the agent instead emits direct text, that text only lands in the TUI pane — **the operator on telegram sees nothing**, silently.

Today the daemon already:
- appends a reminder hint to every channel-origin message (`src/inbox/message.rs:35` — "(Reply using the reply tool — do NOT respond with direct text)"), and
- tracks whether the agent used the reply tool since its last input (the `reply_tool_called_since_input` flag surfaced in instance state).

But the signal is **measured, not acted upon** — there is no code that warns/nudges/auto-corrects when an agent finishes a turn responding to a channel-origin input with `reply_tool_called_since_input == false`. (Same "instrument-only" shape as #1808.) Result: a recurring silent failure where the operator never receives the answer and has to re-prompt ("why didn't you reply on telegram?").

### Observed
2026-06-06: fixup-lead repeatedly answered telegram-origin operator messages with direct TUI text (3+ times in one session); operator received nothing each time and had to re-ask. Passive reminder alone was insufficient.

### Proposed
When a turn ends (agent → idle / produced a final assistant message) AND the most recent input was channel-origin AND `reply_tool_called_since_input == false`:
- inject a short corrective nudge to the agent, e.g. `[AGEND-AUTO kind=channel-reply-missing] Your last answer went to the TUI only — the input came via telegram. Re-send via the reply tool.`
- (mark kind so the agent treats it as a low-priority system nudge, like other AGEND-AUTO injects)

### Open questions for design (dialectic)
- Detection point: where in the daemon turn-completion path is `reply_tool_called_since_input` finalized, and is idle/turn-end reliably observable for all backends?
- Debounce / one-shot: nudge at most once per offending turn; don't loop.
- False positives: turns that legitimately don't need a reply (pure ack absorption, kind=report/update) must NOT be nudged.
- Should the nudge be agent-facing only, or also surface a hint to the operator ("agent replied to wrong channel, nudged")?

### Relation
Sibling to the discipline gap that produced repeated silent misses; complements (does not replace) the existing passive reminder.

suzuke:
Closes #1813.

## Problem

When an agent answers a `[user:… via telegram]` input with direct TUI text instead of the `reply` tool, the operator on the channel **silently receives nothing**. The daemon already *tracked* this (the #1665 reply-ledger audits whether the channel turn was closed) but only *warned* — there was no code that nudged the agent to fix it (the "instrument-only" gap, same shape as #1808). Observed 2026-06-06: a lead answered telegram-origin messages with direct TUI text 3+ times in one session; the operator got nothing each time.

## Fix — wire the existing signal to an action

Reuses the #1665 machinery end-to-end (minimal new surface):

- **`reply_ledger::sweep`** already detects the exact condition at the supervisor turn-end tick — channel-origin turn (armed only for user channel messages), agent settled (`!agent_is_busy`), past the 30s grace, classified `SilentNoReply` (reply tool never called) — and clears the turn (one-shot). It now returns a `SweepAction`; `NudgeChannelReplyMissing { channel }` carries the origin so the message names telegram/discord accurately.
- **The supervisor** injects a one-shot `[AGEND-AUTO kind=channel-reply-missing]` nudge via the existing draft-gated inject path (`force=false`, same as the continue-inject). Kept as a sibling helper so the #1680 source-guard on the continue-inject literal stays intact.

## False-positive safety (all inherited from the reply-ledger)

- grace window + `agent_is_busy` gate — a slow-but-working agent is never nudged;
- never-warn kinds (`update`/`report`/`ci`/`system`) — **ack-absorption is exempt**;
- mirror-dispatched / `Delivered` closures → no nudge;
- **Gap D (`ReplySendFailed`) is deliberately NOT nudged** — the agent *did* call reply (it just failed); "you didn't reply" would be wrong, and the #1665 warn already covers it;
- **one-shot per turn** — `sweep` clears the turn, so it never loops; a new channel input re-arms.

## §3.9 test

`sweep_nudge_channel_reply_missing_four_cases_1813` drives the **real `sweep`** (supervisor turn-end path) across the four cases: channel + no-reply → **nudge**; channel + reply → none; non-channel (no armed turn) → none; second sweep → none (**no loop**). Plus Gap D → none.

Local CI-parity preflight green (`fmt`/`clippy --all-targets --features tray -D warnings`/`cargo test --tests`; reply_ledger + supervisor + snapshot suites pass, incl. the `continue_inject_is_draft_gated_force_false_1680` source-guard unbroken).

### DIFF
--- a/src/daemon/supervisor.rs
+++ b/src/daemon/supervisor.rs
@@ -942,7 +942,16 @@ fn tick(
         // never hit a clear site (no reply, no mirror, no takeover). Lock-free
         // snapshot read; warns only past the grace window AND when the agent has
         // settled. Infallible — never blocks the supervisor loop.
-        crate::reply_ledger::sweep(home, &name);
+        // #1813: when the swept turn was a channel-origin input the agent
+        // answered WITHOUT the reply tool, additionally inject a one-shot
+        // agent-facing nudge so it re-sends via `reply` (the operator otherwise
+        // gets nothing). `sweep` cleared the turn, so this fires at most once per
+        // offending turn — no loop.
+        if let crate::reply_ledger::SweepAction::NudgeChannelReplyMissing { channel } =
+            crate::reply_ledger::sweep(home, &name)
+        {
+            inject_channel_reply_missing_gated(home, registry, &name, channel);
+        }
         // Mutate state + pull the tail under the core lock, then drop it
         // before running `format!` and the Telegram spawn. `tail_lines`
         // allocates a fresh String, so the lock window is bounded by the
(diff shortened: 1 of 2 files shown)


suzuke:
**Merge audit (§3.12 mirror)** — reviewer-2 VERIFIED + CI 6/6 green at head `7d7ee773`. Implements #1813 channel-reply-missing nudge by wiring the existing #1665 reply-ledger sweep (which already detects channel-origin turn + agent settled + 30s grace + SilentNoReply, one-shot clear_turn) to a SweepAction → draft-gated [AGEND-AUTO kind=channel-reply-missing] inject carrying the origin channel. False-positive STRUCTURALLY avoided: arm gated by m.channel.is_some() (TUI-direct/agent-to-agent never armed; ChannelKind enum is {Telegram,Discord} only); one-shot (no loop); only SilentNoReply nudged (reply-delivered/send-failed correctly excluded); force=false draft-gate doesn't interrupt typing. §3.9 four-case + Gap-D real-sweep-entry test. Merging --admin squash.

PR #1842, opened 2026-06-08
worktree-gc: real .trash recovery window (H1) + force-reclaim binding lock (M3)

suzuke:
Fixes the two compounding force-reclaim worktree-GC bugs from the daemon-core hunt (H1 + M3 of that report).

## [H1] HIGH — `.trash` retention window defeated for force-reclaimed worktrees

`try_archive` uses `fs::rename`, which **preserves the source dir's original mtime**. `purge_trash` keyed retention age on `metadata.modified()`, so a freshly-archived force-reclaimed worktree (source untouched for days) looked ancient → purged the **same sweep** it was archived. The "7-day recoverable in `.trash`" promise was false for exactly the irrecoverable (force-reclaim) case.

**Fix:** the archive dir name already embeds the true archive time (`{agent}-{secs}-{nanos}`, stamped atomically). `purge_trash` now derives age from it via `archived_at_secs`, falling back to mtime only for unrecognised names.

## [M3] MED — force-reclaim archive+unbind ran without the binding lock

Clean-release GC takes `.binding.json.lock` (the same `bind_full` holds) to make deletion and re-lease mutually exclusive; the force-reclaim path (`gc_remove_one → maybe_remove_candidate`, and the retention `sweep`) skipped it. A concurrent `bind_full` could interleave between the liveness re-check and `fs::rename`, archiving a freshly-leased **live** worktree (which H1 then purged same-sweep → unrecoverable).

**Fix:** `maybe_remove_candidate` now holds the binding lock across the force-reclaim re-validate → archive → unbind (scoped to force-reclaim; clean-release locking unchanged).

## Tests (§3.9, real purge/archive entry)

- `purge_uses_embedded_archive_time_not_inherited_mtime_h1`: a fresh-archive entry with a 30-day-old inherited mtime **survives** (would be purged under the mtime bug).
- `purge_deletes_when_embedded_archive_time_is_old_h1`: an old-embedded-time / fresh-mtime entry **is purged** (proves name-keyed).
- `force_reclaim_archive_blocked_while_binding_lock_held_m3`: force-reclaim does **not** archive while `.binding.json.lock` is held; proceeds on release.

## Verification

- `cargo fmt --check` ✓
- `cargo test --tests --features tray` → green. (One unrelated test, `dispatch_branch_skips_metadata_stub_..._1834` in messaging.rs, false-fails **only** under the local fleet `git` shim; with real `/usr/bin/git` it passes — environmental, confirmed, not introduced here. CI uses real git.)
- `daemon::retention::worktrees` suite 23/23.

### DIFF
--- a/src/daemon/retention/worktrees.rs
+++ b/src/daemon/retention/worktrees.rs
@@ -56,6 +56,33 @@ fn archive_ts() -> String {
     format!("{}-{:09}", d.as_secs(), d.subsec_nanos())
 }
 
+/// [H1] Extract the archive time (unix secs) embedded in a `.trash/worktrees/`
+/// entry's directory NAME. The name is `{agent}-{secs}-{nanos}` (see
+/// [`archive_ts`], which itself contains a `-`), so the last two `-`-segments are
+/// the timestamp stamped atomically at archive time.
+///
+/// This is the only trustworthy archive clock: [`try_archive`] uses `fs::rename`,
+/// which PRESERVES the source worktree's original mtime. A freshly-archived
+/// force-reclaimed worktree (whose source was untouched for days) therefore has a
+/// stale `metadata.modified()`, and keying the retention age on mtime would purge
+/// it the SAME sweep it was archived — defeating the recovery window for exactly
+/// the irrecoverable case. Returns `None` for names that don't match the expected
+/// shape (caller falls back to mtime).
+fn archived_at_secs(dir_name: &str) -> Option<u64> {
+    let mut parts = dir_name.rsplitn(3, '-');
+    let nanos = parts.next()?;
+    let secs = parts.next()?;
+    parts.next()?; // agent segment (may itself contain '-') must be present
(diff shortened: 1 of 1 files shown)


suzuke:
**Merge audit (§3.12 mirror)** — reviewer-2 VERIFIED + all REQUIRED checks green (macos/ubuntu/windows Check + LOC + audit). Coverage (NON-required) fails only because the H1 mtime test (purge_uses_embedded_archive_time...) is now #[cfg(unix)]-gated — windows std::fs::File::open on a directory needs FILE_FLAG_BACKUP_SEMANTICS, so that test setup is unix-only; H1 is still proven on windows by the sibling purge_deletes_when_embedded_archive_time_is_old test (no File::open) which passes on windows. All three platform Checks pass = tests pass everywhere; the Coverage delta is just the one unix-gated test. Merging --admin squash (Coverage non-required per CI policy).

PR #1843, opened 2026-06-08
fix(HIGH-1): forward dispatch directives through send(request_kind=task)

suzuke:
Fixes the HIGH-1 finding from the request/storage/comms bug-hunt.

## Bug

`handle_delegate_task` re-marshals the incoming `send` args into a fresh SEND params object from an explicit allowlist and silently **dropped nine operator-supplied dispatch directives** — the same re-marshal allowlist-drop class as #1833/#1837:

`sequencing`, `eta_minutes`, `reporting_cadence`, **`worktree_binding_required`**, `correlation_id`, `reviewed_head`, `thread_id`, `parent_id`, `terminal`

The sibling `handle_send_to_instance` (the default / `update` / `report` path) forwards all of them, and the SEND handler (`messaging.rs`) reads each into the delivered `InboxMessage`. So the same field was delivered with `request_kind=update` but **lost with `request_kind=task`** — the path where dispatch directives matter most.

Worst case: `worktree_binding_required` is a **delivery gate** (`check_worktree_enforcement`, can reject with `worktree_not_managed`); dropping it silently disabled that safety control for the very dispatch path it exists for.

**Repro:** `send(target=dev-1, request_kind=task, task_id=t-100-1, worktree_binding_required=true, sequencing="sequential-merge-only")` → dev-1's InboxMessage has all those fields null; the same call with `request_kind=update` delivers them populated.

## Fix

Forward all nine fields in `handle_delegate_task`, mirroring the sibling — both on the happy path (`api::call` SEND params) and on the inbox-fallback delivery (so an API-blip dispatch doesn't lose them either). Task-specific fields (`kind`/`task_id`/`force_meta`/`provenance`) untouched. The happy-path params are compacted (rustfmt doesn't reformat inside `json!`) to stay under the mcp-handler 750-LOC `file_size_invariant`.

## §3.9 test

`delegate_task_forwards_dispatch_directives_high1` drives the **real `send` MCP entry** with `request_kind=task` + the directives and asserts they reach the receiver's `InboxMessage` (`worktree_binding_required=Some(true)`, `sequencing`, `correlation_id`, `reviewed_head`, `eta_minutes`, `reporting_cadence`, `terminal`), with task-specific fields intact. Regression-proof.

Local CI-parity preflight green (`fmt`/`clippy --all-targets --features tray -D warnings`/`cargo test --tests` incl. the `file_size_invariant`). HIGH-2 (telegram reply dedup record-before-send) is a separate follow-up.

### DIFF
--- a/src/mcp/handlers/comms.rs
+++ b/src/mcp/handlers/comms.rs
@@ -356,19 +356,16 @@ pub(super) fn handle_delegate_task(home: &Path, args: &Value, sender: &Option<Se
         &json!({
             "request_id": uuid::Uuid::new_v4().to_string(),
             "method": crate::api::method::SEND,
+            // HIGH-1: forward the dispatch directives the SEND handler reads
+            // (messaging.rs) that this re-marshal dropped (#1833 class; the
+            // sibling forwards them all; `worktree_binding_required` is a gate).
             "params": {
-                "from": sender.as_str(),
-                "target": target,
-                "text": msg,
-                "kind": "task",
-                "task_id": task_id_str,
-                "force_meta": force_meta_json,
-                "provenance": {
-                    "from": sender.as_str(),
-                    "task": task,
-                },
-                "branch": args["branch"].as_str(),
-                "expect_reply_within_secs": args["expect_reply_within_secs"].as_i64(),
+                "from": sender.as_str(), "target": target, "text": msg, "kind": "task",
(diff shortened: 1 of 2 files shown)


suzuke:
**Merge audit (§3.12 mirror)** — reviewer-2 VERIFIED + CI green. (bug-hunt fix) Merging --admin squash.

PR #1844, opened 2026-06-08
health: 1-unit-per-window crash/recovery decay (M1 — fix decay-anchor tick-burst)

suzuke:
Fixes M1 from the daemon-core hunt.

## Bug

`maybe_decay_at` (health.rs, called every ~10s by `hang_detection.rs:70`) decayed `total_crashes` and `recovery_restart_count` but **never advanced their decay anchors** — the crash anchor was `crash_times.back()` (unchanged on decrement) and the recovery anchor was `last_stage2_fired_at` (likewise). So once one `STABILITY_WINDOW` (30 min) elapsed, **every subsequent tick decremented**, draining the entire budget in seconds instead of 1 unit per 30 min (the documented contract).

A slow-flapping agent (crash → stable 30 min → crash → …) thus wiped its whole crash budget each window and **never reached `Failed`** → unbounded respawn-loop, defeating the max-retries cap.

## Fix — advance the anchor on each decrement

- **crash:** new in-memory `crash_decay_at` (re-derives from `crash_times` on restart; not persisted). Decay clock = `max(crash_times.back(), crash_decay_at)`; set to `now` on decrement, cleared with the budget at 0. `crash_times` itself is untouched, so `record_crash` window-pruning + the escalation snapshot stay intact.
- **recovery:** re-stamp `last_stage2_fired_at = now` on decrement (its only reader is this decay; it's set on stage-2 fire elsewhere).

Both counters stay independent (1 unit per window each).

## Tests (§3.9)

- `crash_decay_one_unit_per_window_not_tick_burst_m1`: 4 crashes, then 10 `maybe_decay_at` calls within one window → decays exactly 1 (was: burst to 0); next window → exactly 1 more.
- `recovery_restart_count_decay_one_unit_per_window_m1`: same discipline for the recovery counter.

(The existing decay tests called `maybe_decay_at` once, so never exercised the per-tick repeat.)

## Verification

- `health::tests` 52/52 · `recovery_dispatcher` decay tests 24/24 (existing behaviour preserved)
- `cargo test --tests --features tray` → green except the unrelated `dispatch_branch_skips_metadata_stub_..._1834` (messaging.rs git-checkout) which false-fails **only** under the local fleet `git` shim; passes with real `/usr/bin/git` (environmental, not introduced here; CI uses real git).

### DIFF
--- a/src/health.rs
+++ b/src/health.rs
@@ -326,6 +326,12 @@ pub struct HealthTracker {
     pub state: HealthState,
     pub crash_times: VecDeque<Instant>,
     pub total_crashes: u32,
+    /// [M1] decay clock for `total_crashes`. Advanced to `now` on every
+    /// `maybe_decay` decrement so decay is genuinely 1 unit per `STABILITY_WINDOW`
+    /// — NOT keyed on `crash_times.back()` alone (which never moves on decrement,
+    /// so once one window elapsed the whole budget drained in consecutive ticks).
+    /// In-memory only (re-derives from `crash_times` on restart; not persisted).
+    crash_decay_at: Option<Instant>,
     max_retries: u32,
     /// #1744-H3: per-class notify cooldown — crash escalations stamp/read THIS,
     /// hung escalations use `last_hung_notification`. Split from the former shared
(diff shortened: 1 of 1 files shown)


suzuke:
**Merge audit (§3.12 mirror)** — reviewer-2 VERIFIED + CI green. M1 fix: maybe_decay_at advances the decay anchor on each decrement (crash via in-mem crash_decay_at=max(crash_times.back(), crash_decay_at); recovery via last_stage2_fired_at re-stamp) → genuinely 1-unit-per-30min instead of draining the whole crash/recovery budget in a tick-burst after the first STABILITY_WINDOW (which defeated the max-retries crash-loop cap for slow-flapping agents). §3.9 repeated-call test. Merging --admin squash.

PR #1845, opened 2026-06-08
fix(HIGH-2): evict telegram reply dedup claim on send failure

suzuke:
Fixes the HIGH-2 finding from the request/storage/comms bug-hunt. Defeats the #1665/#1813 missing-reply safety nets.

## Bug

`try_telegram_reply_from` recorded the dedup key **before** the send and never evicted it on failure. So an agent `reply` whose first send hit a transient failure (network / `RetryAfter` / the #945 background-init invalid-token window) recorded `SendFailed`, and a **retry of the same text within the 5s dedup TTL** — exactly what the #1813 channel-reply-missing nudge and ordinary transient retries produce — hit the cached key, returned a **synthesized `Ok((0, group_id))` without sending**, and `handle_reply` mis-recorded it `Delivered`. The operator never received the message, yet the reply-ledger believed it was delivered.

## Fix

Keep record-before-send (so the RC2 race against a concurrent duplicate is still caught atomically), but **evict the claim on any terminal `Err`** — so a retry after a failed send actually sends, and nothing is recorded `Delivered` until a send succeeds. A **successful** send keeps its key (legitimate RC2 dedup of a true duplicate is preserved).

- Added `DedupCache::evict`.
- The send + the migration/topic-recreate healing retries are wrapped so the eviction fires on whichever path returns `Err`.

## Tests

- `dedup::evict_rolls_back_claim_but_kept_claim_stays_deduped_high2` — evict un-blocks a retry while a non-evicted (success-path) claim still suppresses a duplicate (**RC2 preserved**); absent-key evict is a no-op.
- §3.9 `reply::reply_failed_send_then_retry_actually_sends_high2` (real reply path via the forced-send-error seam): first send fails → retry of the same text **reaches the send** (returns `Err`, consuming the forced error) instead of being deduped into a synthesized `Ok`. Regression-proof.

Local CI-parity preflight green (`fmt`/`clippy --all-targets --features tray -D warnings`/`cargo test --tests`). The existing migration/topic-recreate reply tests still pass — the closure refactor preserved that healing behavior.

### DIFF
--- a/src/channel/dedup.rs
+++ b/src/channel/dedup.rs
@@ -222,6 +222,20 @@ impl DedupCache {
         true
     }
 
+    /// HIGH-2: roll back a [`record_and_check`] claim — remove `key` if present
+    /// (no-op otherwise). Used when the send that the claim guarded ultimately
+    /// FAILED: leaving the key recorded would suppress a legitimate retry of the
+    /// same content within the TTL (returning a synthesized success that gets
+    /// mis-recorded as delivered). Recording stays BEFORE the send so the RC2
+    /// race against a concurrent duplicate is still caught atomically; only a
+    /// terminal failure evicts. A SUCCESSFUL send keeps its key.
+    pub fn evict(&self, key: &DedupKey) {
+        let mut entries = self.entries.lock();
+        if let Some(pos) = entries.iter().position(|e| &e.key == key) {
+            entries.remove(pos);
+        }
+    }
+
     /// Total suppressions observed for the lifetime of this cache.
     /// Surfaced via doctor / debug — operator-visible health signal.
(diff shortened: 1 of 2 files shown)


suzuke:
**Merge audit (§3.12 mirror)** — reviewer-2 VERIFIED + required checks green. HIGH-2 fix: telegram reply dedup evict-on-failure — record-before-send preserved (RC2 concurrent-dup protection), dedup key evicted ONLY on terminal Err (healing-retry success returns Ok → not evicted → no double-send) so a retry after a failed send actually sends and an unsuccessful send is not mis-recorded Delivered (restores the #1665/#1813 reply safety net). Merging --admin squash.

PR #1847, opened 2026-06-08
dispatch-idle: stop team-nudge resurrecting a resolved sidecar (M2)

suzuke:
Fixes M2 from the daemon-core hunt.

## Bug

`team_nudge::scan_and_nudge` read a pending sidecar from an unlocked `list_pending` snapshot, ran the **slow** `emit_nudge` (inbox I/O), then wrote it back with an **unconditional `atomic_write`** (no re-check that the file still existed). If a report arrived during `emit_nudge`, `mark_resolved` deleted the sidecar — and the write then **re-created** it (`status=Exceeded`, `nudge_sent_at` set). A resolved dispatch's sidecar thus leaked forever (the report already came; nothing clears it again). It also clobbered any concurrent on-disk update (lost update). The L1 sibling `scan_and_emit` already uses flock + re-read; this path didn't.

## Fix

- **`stamp_nudge_sent`** — a locked read-modify-write (`with_json_state`) that stamps `nudge_sent_at` only if the sidecar still exists AND is still `Exceeded` AND not already nudged. `with_json_state` returns `Ok(None)` for a missing file and **never recreates** it → no resurrection; the under-lock status re-read closes the lost-update.
- **`mark_resolved`** — holds the same `{dispatch_id}.lock` across the delete, making it mutually exclusive with the RMW (the unlocked delete could otherwise still race the RMW's read→write window — the lock is what makes elimination airtight, not just window-shrinking). Also fixes the stale lock-file cleanup that removed the wrong `{dispatch_id}.json.lock` and orphaned the real `{dispatch_id}.lock`.

## Tests (§3.9)

- `resolved_sidecar_not_resurrected_by_nudge_write`: delete then stamp → not recreated.
- `mark_resolved_then_nudge_write_does_not_resurrect`: end-to-end via the real `mark_resolved`.
- `live_exceeded_sidecar_stamped_once`: a live Exceeded sidecar is stamped exactly once (dedup preserved).

## Verification

- `dispatch_idle` suite 57/57
- `cargo test --tests --features tray` → green except the unrelated `dispatch_branch_skips_metadata_stub_..._1834` (messaging.rs git-checkout) which false-fails only under the local fleet `git` shim; passes with real `/usr/bin/git` (CI uses real git).

### DIFF
--- a/src/daemon/dispatch_idle/mod.rs
+++ b/src/daemon/dispatch_idle/mod.rs
@@ -137,6 +137,28 @@ fn dispatch_lock_path(home: &Path, dispatch_id: &str) -> PathBuf {
     pending_dir(home).join(format!("{dispatch_id}.lock"))
 }
 
+/// [M2] Delete a sidecar UNDER its `{dispatch_id}.lock`, so the delete is
+/// mutually exclusive with the team-nudge / L1 locked read-modify-write
+/// (`with_json_state` / `scan_and_emit`, which take the same lock). Without the
+/// lock, an unlocked `remove_file` can land inside an RMW's read→write window
+/// and the RMW then re-creates (resurrects) the just-deleted sidecar, leaking a
+/// resolved dispatch forever. Returns `true` iff the sidecar file was removed.
+///
+/// Also removes the CORRECT lock file (`{dispatch_id}.lock`); the pre-fix delete
+/// sites removed `{dispatch_id}.json.lock` (wrong name) and orphaned the real
+/// one. Single correct implementation shared by every sidecar-delete site
+/// (`mark_resolved`, `cleanup_pending_for_task_id`, `cleanup_pending_for_instance`).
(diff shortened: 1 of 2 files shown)


suzuke:
**Merge audit (§3.12 mirror)** — reviewer-2 VERIFIED (3 iterations: main paths → :339/:381 → :723 4th site; reviewer-2's own exhaustive grep confirmed no 5th site) + CI 6/6 green. M2: team-nudge sidecar resurrection class FULLY closed — extracted delete_sidecar_locked(home, dispatch_id) helper (holds {id}.lock, deletes {id}.json + the correct {id}.lock) used by ALL 4 sidecar-delete sites (mark_resolved / cleanup_pending_for_task_id / cleanup_pending_for_instance / scan_and_emit #1018-A) → no unlocked delete can race the stamp_nudge_sent RMW to resurrect a resolved sidecar; also fixed the wrong-lock-name ({id}.json.lock) orphan. §3.9 6 tests covering all 4 delete paths. Merging --admin squash.

PR #1850, opened 2026-06-08
fix(MED-1,MED-2): notify dedup evict-on-failure; deploy spawns command override

suzuke:
Two round-2 bug-hunt findings (bundled; lead self-review).

## MED-1 — notify dedup record-before-send, no evict-on-failure

The un-patched twin of the reply.rs HIGH-2 fix. `notify_telegram_inner` records the dedup key **before** the spawned send, whose result was only logged. A transient send failure left the claim live for the TTL, so a legitimate re-emit of the same text within the window was suppressed into a no-op — the operator never received it.

**Fix:** evict the dedup claim on terminal failure (initial send + the topic-recreate retry), mirroring `reply.rs:166-168`. A successful send keeps the key (RC1/RC2 dedup preserved).

## MED-2 — deploy Phase-3 SPAWN dropped the `command:` override

The re-marshal allowlist-drop class (3rd instance). `spawn_instances` passed raw `entry.backend`, but the SPAWN handler runs `params["backend"]` **as the command**, and a template's `command:` is persisted to fleet.yaml in Phase 2 — so the override was silently ignored, spawning the preset binary on first deploy.

**Fix:** extract `resolve_spawn_backend` and pass `resolve_instance().backend_command` (which honors `command:` over the preset), mirroring every sibling spawn path (start/restart/replace/cold-boot).

## Testability refactor (MED-1)

`notify_telegram_inner` now returns the spawned `JoinHandle` (public wrappers discard it) so a test can drive the current-thread runtime to completion via the Handle-guarded `block_on_value` (**#1476**: never a raw `telegram_runtime().block_on` — the source-scan invariant enforces this), and a `#[cfg(test)]` forced-send-error seam fails the send without a network call (mirrors reply.rs).

## §3.9 tests

- `notify_failed_send_evicts_dedup_med1` — drives the **real spawned send** (forced failure) and asserts the dedup claim is evicted (a retry would actually send).
- `resolve_spawn_backend_honors_command_override_med2` — a template `command:` reaches the spawn binary; resolve-miss falls back to `entry.backend`.

Local CI-parity preflight green (`fmt`/`clippy --all-targets --features tray -D warnings`/`cargo test --tests` incl. the block_on guard invariant + file_size). channel suite 179 pass, deployments/notify 76 pass; existing migration/topic-recreate reply tests unbroken.

### DIFF
--- a/src/channel/telegram/notify.rs
+++ b/src/channel/telegram/notify.rs
@@ -6,25 +6,25 @@ use crate::channel::telegram::state::*;
 
 /// Send a notification to Telegram (instance topic or general).
 pub fn notify_telegram(home: &std::path::Path, instance_name: &str, text: &str) {
-    notify_telegram_inner(home, instance_name, text, false);
+    let _ = notify_telegram_inner(home, instance_name, text, false);
 }
 
 /// Send a notification with Telegram's `disable_notification` flag set — the
 /// message still appears in the topic but does not push/vibrate the operator.
 pub fn notify_telegram_silent(home: &std::path::Path, instance_name: &str, text: &str) {
-    notify_telegram_inner(home, instance_name, text, true);
+    let _ = notify_telegram_inner(home, instance_name, text, true);
 }
 
+/// Returns the spawned send's `JoinHandle` (or `None` when nothing was sent:
+/// no telegram channel / dedup-suppressed). The public wrappers discard it;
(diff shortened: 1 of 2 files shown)


suzuke:
**Merge audit (§3.12 mirror)** — VERIFIED + required checks green. (bug-hunt round-2 fix) Merging --admin squash.

Issue #826, opened 2026-05-15
force_release_worktree: doesn't clean git-level worktree metadata when agent already disbanded


suzuke:
## Symptom

`force_release_worktree(agent=X, branch=Y)` returns `released: true` but **does not actually clean** the git-level worktree under `<canonical>/.git/worktrees/...` if the target agent's binding has already been removed (e.g. agent was disbanded). The daemon thinks success because its own pool is clean (`dir_existed: false`), but `git worktree list` (and `git branch --no-color` showing `+` prefix) still report the stale worktree.

This compounds with #817 (`repo action=cleanup_merged_branches`): the sweep correctly identifies stale branches as squash-merged candidates, but their `git branch -D` will fail because the branch is still attached to the now-orphan worktree. Operator has to manually shell out to `git worktree remove` (which itself requires `AGEND_GIT_BYPASS=1` to bypass the shim) before the sweep can apply.

## Repro (verified 2026-05-15 ~12:45 UTC, post-#809 daemon restart)

After today's daemon restart, ran `repo action=cleanup_merged_branches` (dry-run). Sweep correctly classified 4 squash-merged candidates:

```
docs/704-release-smoke-checklist           (+ worktree-attached, owner: dev-fast-1, disbanded)
refactor/694-block1-vec-collapse           (+ worktree-attached, owner: dev-arch-1, disbanded)
refactor/694-block2-fifth-cut              (+ worktree-attached, owner: dev-arch-2, disbanded)
review/pr-777                              (+ worktree-attached, owner: operator-side review pool)
```

Tried to clean the 3 daemon-owned ones via `force_release_worktree`:

```
{"agent": "dev-arch-1", "branch": "refactor/694-block1-vec-collapse"}
→ {
    "binding_outcome": {
      "binding_removed": false,
      "branch_deleted": false,
      "error": "no binding for agent 'dev-arch-1'",
      "released": false,
      "worktree_removed": false
    },
    "dir_existed": false,
    "dir_removed": false,
    "released": true       ← misleading: outer says success, inner says no-op
  }
```

After the call:
- `<canonical>/.git/worktrees/dev-arch-1-.../HEAD` still says `ref: refs/heads/refactor/694-block1-vec-collapse`
- `git branch` still shows `+ refactor/694-block1-vec-collapse`
- `cleanup_merged_branches apply=true` would still fail to `git branch -D` it

Same pattern for dev-arch-2 + dev-fast-1.

## Root cause

[...]

PR #1853, opened 2026-06-08
fix(MED-4): deploy teardown removes branch-mode worktree + orphan branch

suzuke:
Round-2 bug-hunt finding (lead self-review).

## Bug

A branch-mode deploy (`deployment deploy ... branch=<base>`) creates a git worktree per instance via `git worktree add -b {deploy}/{suffix}` in the deploy directory (which **is** the source repo in branch mode), but `cleanup_deployment_dirs` only `remove_dir_all`'d the subdir. That left:
- a prunable `.git/worktrees/<seg>` **registry entry**, and
- the orphan **`{deploy}/{suffix}` branch**

behind — so a same-name re-deploy failed (`branch already exists` / `is already checked out`). #826 gave the *bind* path a worktree GC; the *deploy* path had no equivalent.

## Fix

When the deploy directory is a git repo, teardown now tears each instance's worktree + branch down **first** — `git worktree remove --force <subdir>` + `git branch -D {deploy}/{suffix}` + `git worktree prune` — via the daemon's bypass git (mirrors `worktree_pool::release_full`), then the existing `remove_dir_all`. All best-effort: harmless no-ops for a non-branch deploy (the subdir isn't a worktree, the branch doesn't exist). The branch name is reconstructed deterministically (instances are `{deploy}-{suffix}`, branch `{deploy}/{suffix}`).

## §3.9 test

`teardown_removes_branch_mode_worktree_and_orphan_branch_med4` builds a real source repo, creates the worktree exactly as `prepare_work_dir` does, runs `cleanup_deployment_dirs`, and asserts the worktree dir + branch are gone AND a same-name `worktree add -b` re-deploy **succeeds**. Regression-proof — revert the GC and the re-add fails.

Local CI-parity preflight green (`fmt`/`clippy --all-targets --features tray -D warnings`/`cargo test --tests`; deployments teardown/cleanup suite 12 pass, no regression).

### DIFF
--- a/src/deployments.rs
+++ b/src/deployments.rs
@@ -559,9 +559,36 @@ pub fn list(home: &Path) -> Value {
 /// a single per-instance failure doesn't abort the rest of the sweep.
 fn cleanup_deployment_dirs(home: &Path, deployment: &Deployment) {
     let custom_root = std::path::Path::new(&deployment.directory);
+    // MED-4: a branch-mode deploy creates a git worktree per instance via
+    // `git worktree add -b {deploy}/{suffix}` in the deploy directory (which IS
+    // the source repo in branch mode). A bare `remove_dir_all` left a prunable
+    // `.git/worktrees/<seg>` registry entry + the orphan branch behind, so a
+    // same-name re-deploy failed ("already exists" / "already checked out").
+    // When the deploy dir is a git repo, tear the worktree + branch down FIRST,
+    // via the daemon's bypass git (mirrors `worktree_pool::release_full`). All
+    // best-effort: harmless no-ops for a non-branch deploy (subdir isn't a
+    // worktree, branch doesn't exist).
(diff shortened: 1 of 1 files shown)


suzuke:
**Merge audit (§3.12 mirror)** — lead-review VERIFIED + CI green. MED-4: cleanup_deployment_dirs now, when the deploy dir is a git repo, tears down each branch-mode instance's worktree+branch FIRST (git worktree remove --force + git branch -D {deploy}/{suffix} + git worktree prune via git_bypass, mirroring worktree_pool::release_full) before remove_dir_all — so a same-name re-deploy no longer fails on a leaked .git/worktrees registry entry / orphan branch. All best-effort (non-branch deploy = harmless no-op). §3.9 teardown_removes_branch_mode_worktree_and_orphan_branch_med4 proves worktree+branch removed + same-name re-add succeeds (regression-proof). Merging --admin squash.

PR #1856, opened 2026-06-08
fix(MED-3,MED-5): channels-plural group_id persistence + AGEND_HOME re-inject

suzuke:
Round-2 bug-hunt findings (lead self-review). Two independent fixes, one PR.

## MED-3 — channels-plural group_id never persisted

`update_channel_telegram_group_id` only handled the **singular** `channel:` form. For a channels-only fleet (plural `channels:`, which `normalize()` supports by collapsing the first entry by sorted name into the active channel), it fell through to `Ok(false)` — silently treated as success. So a telegram **supergroup migration was never persisted**, the stale `group_id` reloaded on the next restart, and the migration error recurred in a loop.

**Fix:** also handle `channels:`, mirroring normalize's first-by-sorted-name selection (= the active/migrated channel), and surface a genuine no-telegram miss via `warn` instead of a silent `Ok(false)`.

## MED-5 — AGEND_HOME lost under env-isolation

After `cmd.env_clear()` (env-isolation), AGEND_HOME is on the SENSITIVE deny-list, so the passthrough loop never re-adds it and it was never re-injected. The spawned agent's in-pane `agend-terminal` subcommands then lost AGEND_HOME and fell back to the default `~/.agend-terminal`, pointing at the **wrong daemon**.

**Fix:** re-inject AGEND_HOME unconditionally after the clear — exactly like `AGEND_INSTANCE_NAME` on the adjacent line.

## §3.9 tests

- `update_group_id_persists_into_channels_plural_med3`: channels-only fleet → update → asserts `channels.<name>.group_id` rewritten AND `load()`+`normalize()` surfaces the new id as the active channel (the on-restart path). Plus `update_group_id_no_telegram_channel_writes_nothing_med3` (no silent success).
- `build_command_reinjects_agend_home_under_env_isolation_med5`: under `AGEND_ENV_ISOLATION=1`, `build_command`'s cmd carries `AGEND_HOME=home`. Regression-proof — revert and `get_env` returns None (the allowlist excludes AGEND_HOME).

### DIFF
--- a/src/agent/mod.rs
+++ b/src/agent/mod.rs
@@ -730,6 +730,15 @@ fn build_command(config: &SpawnConfig) -> anyhow::Result<(CommandBuilder, Option
     cmd.env("COLORTERM", "truecolor");
     cmd.env("FORCE_COLOR", "1");
     cmd.env("AGEND_INSTANCE_NAME", name);
+    // MED-5: AGEND_HOME is on the env-isolation SENSITIVE deny-list (so
+    // `env_clear` drops it and the fleet passthrough loop never re-adds it), but
+    // the spawned agent's in-pane `agend-terminal` subcommands need it to resolve
+    // the daemon's home — without it they fall back to the default
+    // `~/.agend-terminal`, pointing at the wrong daemon. Re-inject unconditionally
+    // after the clear, exactly like AGEND_INSTANCE_NAME above.
+    if let Some(h) = *home {
+        cmd.env("AGEND_HOME", h);
+    }
 
     // Phase A Piece-3: GIT_EDITOR + friends = `true` (Unix no-op
     // binary that exits 0 without producing output). Prevents git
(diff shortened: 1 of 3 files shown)


suzuke:
Merge audit (section 3.12 mirror) — lead-review VERIFIED + CI 6/6 green. MED-3: update_channel_telegram_group_id now also handles the plural channels: form (set_group_id on the matching channel, mirroring normalize's first-by-sorted-name active-channel selection), and a no-telegram fleet now warn-surfaces instead of a silent Ok(false) — so a chat-to-supergroup group_id migration is persisted into a channels-only fleet.yaml (no more stale-id reload + migration-error loop on restart). MED-5: build_command unconditionally re-injects AGEND_HOME after env_clear (mirroring the adjacent AGEND_INSTANCE_NAME) so a spawned agent under AGEND_ENV_ISOLATION=1 keeps AGEND_HOME (it is on the SENSITIVE deny-list + excluded from the allowlist, so nothing else set it post-clear). Both with regression-proof section-3.9 tests. Merging --admin squash.

PR #1851, opened 2026-06-08
fix(#1808/#1809): wrap-tolerant SRL detection + upstream instrument (Phase 1)

suzuke:
## Problem
The long-standing live SRL bug: agents hit \`API Error: Server is temporarily limiting requests (not your usage limit) · Rate limited\` but the daemon never latches \`ServerRateLimit\` → no auto-retry → the agent sits idle, stuck. Mystifying lead-vs-worker asymmetry (lead recovered, narrow-pane workers hung).

## RCA (live daemon log + code, confirmed)
The detection feed `VTerm::tail_lines_with_fg` → `StateTracker::feed_with_fg` (agent/mod.rs:1436) builds `screen_text` by joining alacritty grid **physical rows** with `\n` **unconditionally** (`tail_lines_impl`). In a **narrow** (split) pane the SRL line soft-wraps across rows, so `\n` lands **mid-phrase** → the single-line SRL regex (`backend_profile.rs`) misses → `detect_with_match` returns `None` → never latches. **Wide** panes keep the phrase on one row → match → recover (the asymmetry).

The same blindness hid the existing #1562 instrument: `unclassified_throttle_tail` used `str::contains`, which a wrapped phrase also defeats → it captured nothing on the live miss.

Daemon-log evidence: the only `ServerRateLimit` events that reached the anchor gate were agents **discussing** the bug (contiguous phrase, correctly suppressed by the `in_error_line` prose anchor). The real worker errors logged **nothing** (regex returned `None` on the wrapped phrase) — matching "0 detection activity".

## Fix — Phase 1 (low-risk; Phase 2 held pending instrument confirmation)
- **`vterm::tail_lines_impl`**: `WRAPLINE`-aware **de-wrap** on the detection path (`collect_fg`) only. A soft-wrapped continuation row is joined **without** a `\n` (and its trailing inter-word space preserved so words don't fuse). One logical line the terminal wrapped now stays one line. Text-only `tail_lines` is **byte-identical** (display/dialog callers unaffected). Fully fixes the **soft-wrap** case; **no-op + zero-risk** if the wrap is a hard `\n` (Ink layout).
[...]

PR #1857, opened 2026-06-08
state: detect Ink hard-wrapped SRL + re-detect static stuck pane (SRL Phase 2)

suzuke:
Phase 1 (#1851) didn't fix the live SRL-stuck bug — dev/dev-2 went stuck again with **0 detection AND 0 instrument**. Two confirmed root causes (from the live pane):

## 1. HARD-wrap, not soft-wrap
On a narrow pane, Ink **word-wraps** the SRL/RateLimit line into rows joined by **real `\n`**, so the single-line `detect_with_match` regex never matches it. Phase 1's WRAPLINE de-wrap only merges **alacritty soft-wraps** → no-op on Ink hard-wrap.

**Fix:** `flattened_throttle_detect` — when raw `detect_with_match` misses, retry on the **flattened bottom-N tail** (whitespace runs incl. `\n` → one space). Scoped tight:
- auto-clear throttle states **only** (`ServerRateLimit`/`RateLimit`) — never `ModelUnsupported`/`ContextFull` (no auto-clear → a flatten-FP would silently wedge a healthy agent);
- **`in_error_line` on the flattened text** is the prose-FP guard (a merged prose block lacking a real error indicator is rejected; the residual verbatim-quote FP is the same accepted, self-correcting residual as the raw path).
- Wired into the detect **None arm** — the raw path (non-wrapped) is untouched → no regression. Lands via the same `recovered_within` gate.

## 2. hash-dedup blind spot
`feed_with_fg`'s screen-hash early-return sits **before** detection; a settled (static) stuck-SRL pane has an unchanged hash → returns early forever → never re-detected.

**Fix:** on a dedup hit, if the pane carries a throttle hint (cheap raw single-token pre-filter — tokens chosen to survive Ink word-wrap) **and** we're not already latched on a throttle, fall through to re-detect. Once latched, identical frames short-circuit again (perf-safe).

## Tests (§3.9)

- `hardwrap_srl_latches_via_flattened_tail_phase2` — asserts the raw single-line regex misses (Phase-1 precondition), then the hard-wrapped frame latches SRL.
- `static_srl_pane_redetected_after_dedup_blindspot_phase2` — clear-while-static → the identical next frame re-latches (would dedup-skip without the fix).
- `multiline_prose_mentioning_srl_does_not_latch_phase2` — prose with the SRL phrase but no error indicator → `in_error_line` rejects → no latch.

## Verification

- `state::` suite **310/310** (no regression — the riskiest axis)
- `cargo test --tests --features tray` → green except the unrelated `dispatch_branch_skips_metadata_stub_..._1834` (messaging.rs git-checkout) which false-fails only under the local fleet `git` shim; passes with real `/usr/bin/git` (CI uses real git).

⚠️ This is daemon-side detection — effect (stuck SRL panes recover) is visible after operator restart.

### DIFF
--- a/src/state/mod.rs
+++ b/src/state/mod.rs
@@ -520,6 +520,28 @@ const THROTTLE_DIAG_PHRASES: &[&str] = &[
     "hit a rate limit",
 ];
 
+/// #SRL-phase2: cheap raw single-token pre-filter for the hash-dedup blind-spot
+/// bypass. A settled (static) stuck-SRL pane has an unchanged screen hash, so
+/// `feed_with_fg`'s dedup would skip detection forever and it never recovers. We
+/// only override the skip when one of these distinctive tokens is on screen —
+/// and they are chosen to survive Ink WORD-wrap (each is a single token that sits
+/// intact on one wrapped row), so a hard-wrapped SRL line still trips the
+/// pre-filter even though the full phrase is split across rows. Kept tight to
+/// avoid re-detecting every static idle pane.
+const THROTTLE_HINT_TOKENS: &[&str] = &[
+    "limiting",
+    "Overloaded",
+    "overloaded_error",
+    "rate_limit_error",
+    "RESOURCE_EXHAUSTED",
+    "429",
+    "Rate limited",
+];
+
+fn screen_has_throttle_hint(screen_text: &str) -> bool {
+    THROTTLE_HINT_TOKENS.iter().any(|t| screen_text.contains(t))
+}
+
 /// #1562: rows of the live tail captured into the diagnostic record. Matches
 /// the `ERROR_TAIL_SCAN_LINES` horizon so the captured context lines up with
 /// the bottom-N window the error gates actually look at.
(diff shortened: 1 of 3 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED (two adversarial rounds) + CI 6/6 green. SRL Phase 2 fixes the two confirmed root causes behind the session-long SRL-stuck-no-recovery bug: (1) Ink HARD-wrap — narrow panes word-wrap the SRL phrase across lines with real newlines, defeating the single-line regex; flattened_throttle_detect flattens the bottom-N tail before matching (SRL/RateLimit only, never ModelUnsupported/ContextFull), runs only on the detect-None arm. (2) hash-dedup blind spot — a settled-static SRL pane was never re-fed (feed_with_fg early-returns on unchanged hash), so it was never re-detected; an override now re-evaluates when the screen carries an unlatched throttle hint (cheap word pre-filter, perf-guarded). The prose-FP guard that round 1 defeated (flattening collapsed in_error_line scope to the whole tail) is fixed via proximity-scope: an error indicator must sit within 80 chars before the throttle match (legit hard-wrap is adjacent; a stray indicator from another line is far) — reviewer-2 confirmed this converges the flatten-path FP surface to the same level as the raw path, with line_has_error_indicator extracted so in_error_line stays byte-identical. section-3.9: hard-wrap latch, static re-detect, prose-no-indicator, and prose-with-distant-indicator (regression test for the round-1 hole). Merging --admin squash. Verifies on the next live SRL event after restart.

PR #1860, opened 2026-06-08
ci-handoff: daemon-side re-nudge of the target + fix orchestrator-as-next_after_ci hole (#1859 Fix A)

suzuke:
Addresses #1859 Scenario A only (Scenario B deferred — see below). refs #1859

## Root cause (from the spike — refutes the issue's own hypothesis)

A dispatch with `next_after_ci=<lead>` auto-arms a ci-watch; on CI pass the poller enqueues `[ci-ready-for-action]` to the target via `enqueue_with_idle_hint` — **the identical path subscribers use** (so cheerc's "next_after_ci not in subscribers → no push" is wrong). The durable inbox write always lands; the **active wake** goes through `should_defer_inject`, which defers actionable wakes whenever the target is mid-token (`thinking`/`tool_use`) into the `notification_queue`. That queue's **only** flush is the TUI loop (`app/mod.rs::flush_idle_notifications`) — there is **no daemon-side redelivery** — so a busy orchestrator's wake strands: inbox has it, no active nudge.

The handoff-timeout watchdog (#1491B) was the safety net, but it escalates to `team_orchestrator_for(target)` and `continue`s when `recipient == target` — silently skipping the whole handoff exactly when `next_after_ci` IS the orchestrator (Scenario A's `dev-team-lead`). Net: no recovery.

## Fix A (analysis-approved by lead)

- **Daemon-side re-nudge of the target** in `handoff_timeout_watchdog`, before the lead escalation: handoff unread > `RENUDGE_AFTER_MINS` (2) **and** target idle (`snapshot::agent_is_busy` false) → re-fire an actionable PTY pointer (`inbox::notify::renudge_actionable_unread`), bypassing the queue/TUI-flush dependency.
- **Busy target → skipped** (the ~2min watchdog tick is the retry loop): no mid-token PTY corruption, no queueing.
- **Anti-storm**: idempotent `last_renudged` interval map (`RENUDGE_INTERVAL_MINS=2`); stops the instant the target reads the handoff (drops from the unread scan → both dedup maps reaped). Pure WAKE pointer (`id="renudge"`), **not** a new inbox row → can't self-amplify.
- **`recipient == target` hole fixed**: no longer a silent total-skip — the re-nudge already redelivers to the orchestrator itself (correctly no self-escalation).
- **`should_defer_inject` untouched** (the mid-token guard is legitimate).
- Handler cadence 30→12 ticks (~2min) so re-nudge is timely; escalation stays gated by its own 10min/30min windows.

## Tests (§3.9 — real watchdog entry via `scan_and_emit_with` + captured emitter seam, not a synthetic inject)

- `orchestrator_as_next_after_ci_is_renudged_not_silently_skipped`
- `busy_target_skipped_idle_target_renudged`
- `renudge_is_interval_gated_and_stops_when_read` (anti-storm + reap)
- existing escalation tests retained (7), all green.

## Verification

- `handoff_timeout` 10/10 · `per_tick::handoff` 3/3

⚠️ Daemon-side behaviour — effect (stranded ci-ready handoffs recover) is visible after operator restart.

Scenario B (rerun dedup, attempt-blind) is deferred to a follow-up per lead.

### DIFF
--- a/src/daemon/handoff_timeout_watchdog.rs
+++ b/src/daemon/handoff_timeout_watchdog.rs
@@ -23,39 +23,98 @@ use std::path::Path;
 const HANDOFF_TIMEOUT_MINS: i64 = 10;
 /// Don't re-escalate the same (target, handoff) more often than this.
 const REALERT_AFTER_MINS: i64 = 30;
+/// #1859 Fix A: a handoff unread for at least this long (minutes) is RE-NUDGED to
+/// the target itself (daemon-side redelivery — earlier + cheaper than the 10-min
+/// lead escalation, and the only recovery when `next_after_ci` IS the orchestrator).
+const RENUDGE_AFTER_MINS: i64 = 2;
+/// #1859 Fix A: don't re-nudge the same (target, handoff) more often than this
+/// (anti-storm). Combined with the idle gate, a busy/working target is retried at
+/// most once per interval and a target that has read the handoff stops entirely.
+const RENUDGE_INTERVAL_MINS: i64 = 2;
 /// Fallback recipient when the target isn't in any team.
 const FALLBACK_RECIPIENT: &str = "lead";
 /// Inbox kind of a CI handoff message.
 const HANDOFF_KIND: &str = "ci-ready-for-action";
 
 /// Scan every fleet instance for `ci-ready-for-action` handoffs it received but
-/// never read, and escalate timed-out ones to the target's team lead.
-/// `last_escalated` (keyed by `(target, correlation)`) is owned by the caller
-/// so dedup survives across ticks; `now` is injected for deterministic tests.
+/// never read; (1) #1859: RE-NUDGE the target itself (daemon-side redelivery) and
+/// (2) escalate timed-out ones to the target's team lead. `last_escalated` /
+/// `last_renudged` (keyed by `(target, correlation)`) are owned by the caller so
+/// dedup survives across ticks; `now` is injected for deterministic tests.
 pub(crate) fn scan_and_emit(
     home: &Path,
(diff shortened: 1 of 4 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED (two adversarial rounds) + CI 6/6 green. #1859 Fix A: makes the next_after_ci CI hand-off reliable. Root cause (disproving the issue's subscriber hypothesis — delivery is symmetric): a busy orchestrator's [ci-ready-for-action] wake is deferred by should_defer_inject, the deferred queue has no daemon-side flush (only the TUI flushes), and handoff_timeout_watchdog skipped escalation entirely when recipient==target (i.e. when next_after_ci IS the orchestrator) — so a busy lead never got nudged. Fix: the watchdog now daemon-side re-nudges the target itself (bypassing the TUI flush) once it goes idle, at a 2min interval, and the recipient==target hole is closed (orchestrator-as-target re-nudges self instead of silent-skip). Anti-storm: round 2 closed a real per-target bypass — the per-(target,corr) interval-gate sat inside the per-handoff loop, so K simultaneous unread handoffs to one target fired K redundant injects (the multi-branch orchestrator, the headline beneficiary, being the most likely to hold K). Now collapsed to at most one re-nudge per target per scan (scan-local HashSet), with stamp-all-due-keys so a collapsed key neither re-fires within the interval nor is permanently suppressed (monotonic expiry). should_defer_inject untouched (deferral is legitimate). section-3.9 through the real watchdog entry incl. the multi-handoff regression test. Scenario B (re-run dedup) deferred to backlog (Fix B). Merging --admin squash. Takes effect on next daemon restart.

Issue #1858, opened 2026-06-08
Inconsistent working_directory and Auto-worktree behavior across spawn vs. restart/replace lifecycle


cheerc:
### Observed Behavior
There is a significant behavioral inconsistency between the initial instance creation and subsequent lifecycle operations:

1. **Initial Creation (Spawn/Deploy):**
   When an instance is created via `spawn` or `deployment deploy`, it starts in the explicitly provided `working_directory` (typically under `workspace/`). No Auto-worktree logic is triggered, and no git branch is bound at this stage.

2. **Restart/Replace:**
   When the same instance is restarted via `restart_instance` or replaced via `replace_instance`, the daemon triggers the `resolve_one` / `resolve_instance` logic. If `source_repo` is present in `fleet.yaml`, this triggers the Auto-worktree mechanism, which:
   - Redirects the `working_directory` to `~/.agend/worktrees/<agent>/<branch>/`.
   - Automatically binds a branch (defaulting to `agend/<name>`) if no `git_branch` was specified.

### Impact
The runtime state of an instance diverges significantly after a restart. An instance that was originally operating in a plain workspace directory is suddenly moved into a git worktree with a newly created branch, despite `fleet.yaml` remaining unchanged.

### Requirement
The behavior and environment state (working directory, branch binding) of an instance should be consistent across its entire lifecycle. An instance's operating context should not radically change simply because it was restarted or replaced.

suzuke:
Fixed by #1862 (Fix A). The auto-worktree decision is now a single launch-idempotent shared gate (worktree::resolve_auto_worktree) that never auto-worktrees a daemon-managed default workspace dir — killing the is_git_repo side-channel that made restart drift an instance from its plain workspace into a worktree of the empty git-init'd dir. Legit opt-ins (explicit repo / branch-mode deploy / git_branch) preserved. (The create_instance sparse-entry half is tracked separately as the d-0 complete-entry backlog.)

Issue #888, opened 2026-05-17
App bootstrap incorrectly creates worktree for orchestrator/admin agents — relocates working_dir, breaks claude --continue


suzuke:
## Problem

Two intentional features interact destructively in app bootstrap:

### Feature 1 — \`ensure_project_root\` (src/instructions.rs:64)
Calls \`git init\` on every backend agent's workspace dir so Gemini/Codex stop walking up the dir tree when looking for hierarchical \`GEMINI.md\` / \`AGENTS.md\` (they'd otherwise reach \`\$HOME\` and find unrelated instruction files). Load-bearing for Gemini/Codex correctness.

### Feature 2 — \`resolve_one\` worktree auto-create (src/bootstrap/agent_resolve.rs:55-71)
When \`working_directory\` is a git repo, automatically creates a worktree via \`worktree::create\`. Designed for dev/reviewer agents bound to source_repo — gives them isolated branches.

### Interaction (the bug)
1. \`ensure_project_root\` git-inits \`workspace/<name>/\` for every backend agent
2. \`resolve_one\` sees \`is_git_repo(workspace/<name>)\` → TRUE
3. Calls \`worktree::create(home, workspace/<name>, name, None)\`
4. \`worktree::create\` synthesizes branch \`agend/<name>\` (default fallback)
5. Creates worktree at \`worktrees/<name>/agend/<name>/\`
6. \`resolved.working_directory\` redirected to that nested path
7. claude / codex / gemini launched there
8. \`claude --continue\` finds no prior session at that NEW path → resumes wrong session OR fresh
9. Result: instance names correct (daemon-side tracking), but agent context completely wrong

## Evidence (operator smoke 2026-05-17)

```bash
export AGEND_HOME=/tmp/agend-smoke
rm -rf \$AGEND_HOME && mkdir -p \$AGEND_HOME
./target/release/agend-terminal app
# (in another shell)
ls -la \$AGEND_HOME/workspace/*/
```

Output: every workspace dir has \`.git\` + \`.gitignore\` + \`.mcp-config.json.lock\` auto-created. \`shell/\` is the only one without (shell role doesn't run backend agent bootstrap).

Then:
```bash
ls -la \$AGEND_HOME/worktrees/
# → general/agend/general/, fixup-lead/agend/fixup-lead/, etc.
```

[...]

PR #1862, opened 2026-06-08
fix(#1858 Fix A): kill auto-worktree is_git_repo side-channel + share the gate

suzuke:
Spike #1858 → lead approved **Fix A** (the (A) lever). Fix B (complete-entry) queued separately.

## Bug

A `source_repo`/`git_branch` agent whose `working_directory` is the daemon-managed default `workspace/<name>` ran **plain on launch 1** but **drifted into a git worktree on launch 2** (restart_instance / replace_instance / daemon reboot). Both the boot path (`agent_resolve::resolve_one`) and the live-spawn path (`app::pane_factory`) **duplicated** the predicate and gated `worktree::create` on the **transient `is_git_repo(working_directory)`**. `instructions::ensure_project_root` git-inits that workspace dir tail-side of the decision, so the gate flipped between launches — redirecting the dir into a worktree of the **empty git-init'd workspace** (not the real `source_repo`, which is only ever read as a boolean signal). Prime victim: deploy non-branch (`deployments.rs` writes `source_repo` + a `workspace/<name>` working_directory).

## Fix A

- **One shared gate** `crate::worktree::resolve_auto_worktree(home, name, resolved)` — the only place the decision lives now. Both `resolve_one` and `pane_factory` call it → no duplicated predicate that can drift.
- **Persisted-intent only**: `worktree:false` veto + `source_repo`/`git_branch` opt-in (#888), AND the base dir must be an explicit real repo. The default `workspace_dir(home)/<name>` is **never** auto-worktree'd, regardless of `is_git_repo`. → ensure_project_root's git-init can no longer flip the decision; launch-1 and launch-2 are identical.
- **Opt-in preserved**: an explicit real-repo `working_directory`, a branch-mode deploy (dir already a repo), or `git_branch` still worktrees.
- `wants_auto_worktree` (pure #888 signal predicate) moved alongside in `worktree.rs`; truth-table test re-points to it.

## §3.9 (real entry point, representative)

- **(a)** `resolve_one_source_repo_in_default_workspace_no_drift_1858` — real boot entry, **twice**: source_repo agent in the default workspace dir stays plain on launch 1 **and** launch 2 (after the dir is git-init'd, the #1858 trigger). Regression-proof — revert the workspace-default skip and the launch-2 assert lands under `worktrees/`.
- **(b)+(c)** `resolve_auto_worktree_skips_workspace_default_allows_explicit_repo_1858` — (b) explicit real-repo working_directory + source_repo **still** worktrees (no over-kill); (c) deploy non-branch shape (source_repo + git-init'd default workspace dir) does **not**. Existing `resolve_one_creates_worktree_when_source_repo_set` stays green.

## Reviewer focus (cross-cutting spawn/restart behavior)

- Did Fix A **over-kill** a legitimate opt-in worktree? (explicit-repo / branch-mode / git_branch must still worktree — see (b) + existing test.)
- Are the **two** call sites (`agent_resolve.rs`, `pane_factory.rs`) truly routed through the **one** shared fn (no residual duplicated gate)?

### DIFF
--- a/src/app/pane_factory.rs
+++ b/src/app/pane_factory.rs
@@ -336,23 +336,14 @@ pub(super) fn create_pane_from_resolved(
 
     let mut working_dir = resolved.working_directory.clone();
 
-    // #888: Auto-create git worktree when working_directory is inside a repo
-    // AND the instance has an explicit `source_repo` or `git_branch`
-    // config (mirror logic in bootstrap/agent_resolve.rs).
-    let wants_worktree = resolved.worktree != Some(false)
-        && (resolved.source_repo.is_some() || resolved.git_branch.is_some());
-
-    if wants_worktree {
-        if let Some(ref base_dir) = working_dir {
-            if crate::worktree::is_git_repo(base_dir) {
-                let custom_branch = resolved.git_branch.as_deref();
-                if let Some(info) =
-                    crate::worktree::create(home, base_dir, fleet_name, custom_branch)
-                {
-                    working_dir = Some(info.path);
-                }
-            }
-        }
+    // #1858: auto-worktree decision is the single shared gate in
+    // `crate::worktree::resolve_auto_worktree` — same fn the boot path
+    // (`bootstrap::agent_resolve::resolve_one`) calls, so live-spawn and boot
+    // can't drift. Opts in on `source_repo` / `git_branch` (#888) but only for
+    // an explicit real-repo `working_directory`, never the daemon-managed
+    // `workspace/<name>` default (which `ensure_project_root` git-inits).
+    if let Some(wt_path) = crate::worktree::resolve_auto_worktree(home, fleet_name, resolved) {
+        working_dir = Some(wt_path);
     }
 
     let mut args = resolved.args.clone();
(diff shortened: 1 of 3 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED (adversarial, picked up via the now-working ci-ready hand-off) + CI 6/6 green. #1858 Fix A: collapses the auto-worktree decision (duplicated in bootstrap/agent_resolve.rs + app/pane_factory.rs, which drifted) into a single launch-idempotent shared gate worktree::resolve_auto_worktree. Root cause: the gate was on is_git_repo(workspace dir), which ensure_project_root's unconditional git-init flips between launch 1 (plain) and launch 2 (now a repo → drifts into a worktree of the EMPTY workspace dir, not the real source_repo). The fix adds exactly one skip — a daemon-managed default workspace_dir(home)/name is never auto-worktree'd regardless of is_git_repo — killing the stateful side-channel. reviewer-2 verified no over-kill of legit opt-ins (explicit real-repo working_directory / custom-root deploy / branch-mode deploy / git_branch all still worktree; the only skipped path is the non-branch default-root deploy = the exact bug case) and both call sites route through the single shared fn with no residual gate. section-3.9 through the real resolve_one entry run twice (regression-proof: revert → launch-2 drifts to worktrees/). Two non-blocking notes recorded (exact-PathBuf-equality vs canonicalized paths; explicit branch-spawn out of scope). Merging --admin squash. Takes effect on next restart.

PR #1864, opened 2026-06-08
supervisor: persist usage_limit notify dedup so restarts stay silent (#1861)

### LINKED ISSUE #1861: member_state_change usage_limit notification re-fires on every daemon restart (in-memory dedup lost + state re-transitions)
## Symptom
Every daemon restart re-fires the `[member_state_change] <agent>: idle → usage_limit` operator notification for an agent that is STILL usage-limited with an unchanged reset time (observed repeatedly for the codex reviewer, limited until Jun 11). Pure noise — the operator already knows.

## Root cause (traced)
Two compounding causes:
1. **State not persisted across restart.** Backends boot at `initial_state: Starting` (src/backend_profile.rs) → the daemon re-detects the usage_limit screen → a fresh `Idle/Starting → UsageLimit` transition fires.
2. **Notification dedup is in-memory, lost on restart.** `maybe_notify_member_state_change` (src/daemon/supervisor.rs:557-574) dedups via `tracks: HashMap<String, NotifyTrack>` with an `Instant`-based `NOTIFY_COOLDOWN`. `Instant` is process-local + in-memory, so a restart empties `tracks` → the `is_none_or(...)` cooldown gate passes → re-notify.

So restart → transition re-happens + dedup memory gone → re-notification.

## Fix direction (KISS)
Persist a small dedup record keyed on `(member, new_state, unlock_at)` to disk. `unlock_at` is already parsed at supervisor.rs:614 (`parse_unlock_at(pane_tail)`). On a UsageLimit member-notify, skip if a persisted record matches the same `(member, unlock_at)` and `unlock_at` hasn't passed. Re-notify only when `unlock_at` changes (a genuinely new limit) or has elapsed. For a null/unparseable unlock_at, fall back to a persisted timestamp cooldown so it still dampens across restart.

Scope: operator-notice path only; don't change recovery/continue-inject behavior.

suzuke:
Closes #1861.

## Bug

The operator is re-notified of the **same** usage limit on **every daemon restart**. Two causes stack:
1. The backend boots `Starting` and re-detects `UsageLimit` → re-transitions.
2. `maybe_notify_member_state_change`'s dedup is an in-mem `NotifyTrack` with an **`Instant`-based cooldown** that is wiped on restart → the cooldown reads as elapsed → re-notify.

## Fix (KISS — scoped to UsageLimit only)

Persist the "already notified" decision to disk (`usage_limit_notify.json`), keyed by member → `{unlock_at, notified_at}`. Once `unlock_at` is parsed in the member-notify path:
- **same member + same `unlock_at` + deadline NOT passed → SKIP** (the in-mem track is also stamped so same-session ticks short-circuit at the cooldown gate without re-reading the file);
- `unlock_at` **changed** (new limit) **or** its deadline **passed** (limit reset) → notify + re-persist;
- **unparseable `unlock_at`** → fall back to a **persisted-timestamp cooldown** (NOT the in-session Instant cooldown a restart wipes).

The deadline is the next `HH:MM` (treated UTC, per "Resets at HH:MM UTC") at-or-after `notified_at`. The read is **lock-free FAIL-OPEN** (missing/corrupt record → notify, so a real new limit is never silently swallowed); the write is a locked RMW.

**Scope:** only the operator-notice / member-notify path is touched — recovery / continue-inject and non-UsageLimit error-class notifies are unchanged.

## Tests (§3.9)

- `usage_limit_notify_not_refired_across_restart_1861` — **real production entry** `maybe_notify_member_state_change`: first detection notifies + persists; a restart (fresh tracks, record remains) with the same unlock window does **not** re-notify; a different reset time **does**.
- `usage_limit_notify_suppressed_logic_1861` / `usage_limit_null_unlock_fallback_cooldown_1861` — helper tests with injected `now` (before/after deadline, changed window, no record, null-unlock fallback cooldown).

## Verification

- `supervisor::` suite 75/75

⚠️ Daemon-side behaviour — the restart-silence takes effect after the operator restarts onto this build.

### DIFF
--- a/src/daemon/supervisor.rs
+++ b/src/daemon/supervisor.rs
@@ -552,6 +552,102 @@ fn escalate_self_orch_autherror(
     tracing::info!(agent = %name, "#1595/#1744: self-orchestrator AuthError escalated to operator (all channels)");
 }
 
+/// #1861: persisted usage_limit notify dedup record (one per member). A daemon
+/// restart wipes the in-mem `NotifyTrack` cooldown, so without persistence the
+/// operator is re-notified of the SAME usage limit on every restart (the backend
+/// boots `Starting` → re-detects UsageLimit).
+#[derive(serde::Serialize, serde::Deserialize, Default)]
+struct UsageNotifyRecord {
+    /// Parsed "HH:MM" unlock string at notify time (None if the pane had no
+    /// parseable reset time).
+    unlock_at: Option<String>,
+    /// When we notified (rfc3339 UTC) — anchors the unlock deadline + the
+    /// null-unlock fallback cooldown.
+    notified_at: String,
+}
+
+fn usage_limit_notify_path(home: &std::path::Path) -> std::path::PathBuf {
+    home.join("usage_limit_notify.json")
+}
+
+/// The UTC instant an `HH:MM` unlock window elapses, anchored to `notified_at`
+/// (the next occurrence of HH:MM at-or-after the notify, treated as UTC since the
+/// pane renders e.g. "Resets at 15:14 UTC"). `None` if unparseable.
+fn unlock_deadline(
+    hhmm: &str,
+    notified_at: chrono::DateTime<chrono::Utc>,
+) -> Option<chrono::DateTime<chrono::Utc>> {
+    use chrono::TimeZone;
+    let (h, m) = hhmm.split_once(':')?;
+    let h: u32 = h.trim().parse().ok()?;
+    let m: u32 = m.trim().parse().ok()?;
+    let naive = notified_at.date_naive().and_hms_opt(h, m, 0)?;
+    let candidate = chrono::Utc.from_utc_datetime(&naive);
+    Some(if candidate >= notified_at {
+        candidate
+    } else {
+        candidate + chrono::Duration::days(1)
+    })
+}
+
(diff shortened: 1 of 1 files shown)


suzuke:
Merge audit (section 3.12 mirror) — lead-review VERIFIED + 5 required checks green; the only red is Coverage, confirmed a profraw-corruption flake (the run shows 'invalid instrumentation profile data (file header is corrupt)' + 'failed to merge profile data', the CI's own retry hit the same corruption, and every test prints ... ok — no real test FAILED). Coverage is non-required (per #1842/#1851 precedent). #1861: persists the usage_limit member-notify dedup to usage_limit_notify.json keyed (member, unlock_at) so a daemon restart — which wipes the in-mem Instant cooldown and re-detects the same limit — no longer re-fires the operator notification. Re-notifies only when unlock_at ADVANCES (a genuinely new limit) or the unlock deadline elapses (limit reset). Read is lock-free FAIL-OPEN (missing/corrupt record → notify, never swallow a new limit); write is a locked RMW. SCOPE is supervisor.rs member-notify only — recovery / continue-inject and non-UsageLimit error-class notifications are untouched. section-3.9 covers the restart-no-refire invariant. Merging --admin squash past the Coverage flake. Takes effect next restart.

Issue #1866, opened 2026-06-08
dispatch_idle_threshold_exceeded false-fires on actively-working targets + stale dispatches not cleared


suzuke:
## Symptom
The `dispatch_idle_threshold_exceeded` watchdog repeatedly false-fires during normal multi-agent work, creating orchestrator noise:

1. **Fires on actively-working targets.** A dispatched agent that is genuinely making progress (in `tool_use`/`thinking`, editing files, running a build) trips the 600s threshold and emits a 'is it stuck?' nudge — repeatedly, every check, for the whole duration of a non-trivial task (e.g. a multi-file removal or a watchdog impl that legitimately takes 12-15min). The threshold is a flat wall-clock timer with no awareness of whether the target is actually active.

2. **Stale dispatches not cleared on task hand-off.** When an agent finishes a dispatched task and the orchestrator moves it to a NEW task, the OLD dispatch's idle timer keeps running and fires a threshold-exceeded for a dispatch that is already complete (observed: a #1861 dispatch fired idle after the dev had already pushed its PR and moved to the next task).

## Impact
Constant false 'stuck?' nudges to the orchestrator during healthy work → noise, wasted orchestrator attention, erodes trust in the signal (the real 'stuck' case gets lost among false ones).

## Fix direction (spike to confirm RCA first)
- **State-aware threshold**: don't fire if the target is actively progressing — e.g. suppress when the target's agent_state is tool_use/thinking, or reset the idle clock on observed pane activity / last_input advance, rather than a flat wall-clock since-dispatch timer.
- **Clear on hand-off / completion**: when a dispatch's task is reported done / the agent is re-dispatched / the correlated task board entry is closed, cancel the dispatch's idle tracking so it can't fire stale.

Spike should locate the dispatch_idle tracking (the threshold timer + what it keys on + when it clears), confirm both failure modes with file:line, and propose the minimal fix.

suzuke:
Fixed by #1869. (b) clear-on-handoff: a new dispatch retires the agent's older same-dispatcher different-corr armed sidecar (no more stale firing after re-dispatch). (a) state-aware: dispatch_idle now suppresses when the target has recent MCP activity (heartbeat) or was recently fed input — windowed so a genuinely wedged agent (no MCP, no input) still fires; the nudge is now informational with set_waiting_on as the explicit escape hatch. Takes effect next restart.

PR #1869, opened 2026-06-08
fix(#1866): dispatch-idle clear-on-handoff + state-aware activity suppress

### LINKED ISSUE #1861: member_state_change usage_limit notification re-fires on every daemon restart (in-memory dedup lost + state re-transitions)

suzuke:
Spike #1866 → lead approved (b)+(a) in one PR. Fixes the two recurring `dispatch_idle_nudge` false-fires.

## (b) clear-on-handoff (the pure win — closes #1861)
`record_dispatch` deduped only per `(dispatcher, target, correlation_id)`, so re-dispatching the SAME agent to a NEW task left the OLD task's sidecar Armed → it kept nudging for work already done/superseded. After writing the new sidecar, retire same-`(dispatcher, target)` sidecars whose `correlation_id` differs **and** whose `issued_at` is **strictly older** (`delete_sidecar_locked`).
**Boundary (no false-kill):** a different dispatcher's parallel dispatch and any newer sidecar are left untouched; gated on the new dispatch carrying a `correlation_id`.

## (a) state-aware suppress
`target_is_working` read only the fleet snapshot (instantaneous state + pane silence), so an agent engaged via the **daemon** but quiet on the pane (heads-down inter-agent MCP work, or freshly handed input) false-fired once past the wall-clock threshold. Add two zero-lock in-mem signals from `heartbeat_pair`:
- `now − heartbeat_at_ms < threshold` (recent MCP activity), and
- `now − last_input_at_ms < threshold` (recently handed input).

Stale fields (0 / never-set / restart-reset) → huge delta → **no** suppress, so this only ADDS suppression for provably-recent activity and **never hides a real stuck**: a wedged agent makes no MCP calls + gets no input → both stale → still fires. Existing `thinking`/`tool_use`/`silent_secs`/`waiting_on`/debounce gates unchanged.

## Tone
The nudge now reads informational (`FYI: … has been quiet … No action needed if you're mid-task …`) instead of an alarm, and names `set_waiting_on(<reason>)` as the explicit escape hatch for a legitimately-blocked agent. Still fires at most once per dispatch (`nudge_sent_at` dedup) — never re-nags within a dispatch.

## §3.9 (real entry points)
- `record_dispatch_retires_stale_handoff_sidecar_1866` — re-dispatch retires the old sidecar but NOT a different-dispatcher parallel dispatch nor a newer one.
- `scan_suppresses_on_recent_heartbeat_1866` — overdue dispatch + recent heartbeat → stays Pending through `scan_and_emit`.
- `scan_still_fires_when_target_fully_idle_1866` — fully-idle target (all signals stale) past threshold STILL flips to Exceeded.

Local CI-parity preflight green (`fmt`/`clippy --all-targets --features tray -D warnings`/`cargo test --tests`; 63 dispatch_idle tests, no regression to the existing #1516/#1694/#1775/#1018/#1047 gate tests).

## Reviewer-2 focus (adversarial)
- **False-kill?** Does (b) ever retire a *legitimately-parallel* dispatch? (boundary = same dispatcher + strictly older + different corr.)
- **Swallowed stuck?** Can (a)'s heartbeat/last_input suppress a genuinely-wedged agent? (both signals must be stale for a real stuck → fires; verify no path keeps them fresh while hung.)

### DIFF
--- a/src/daemon/dispatch_idle/mod.rs
+++ b/src/daemon/dispatch_idle/mod.rs
@@ -253,6 +253,41 @@ pub(crate) fn record_dispatch(
     if crate::store::atomic_write(&pending_path(home, &dispatch_id), body.as_bytes()).is_err() {
         return None;
     }
+    // #1866 (b) clear-on-handoff: a NEW dispatch for the same (dispatcher, target)
+    // means any EARLIER still-armed dispatch from that dispatcher to this target is
+    // a stale hand-off — the agent has been re-tasked, so its old idle timer must
+    // not keep nudging for work that's been superseded (#1861 class: dev pushed the
+    // PR + moved on, the old dispatch still fired). Retire those sidecars now.
+    // Boundary (avoid clobbering a genuinely-parallel dispatch): SAME dispatcher +
+    // strictly OLDER `issued_at` + a DIFFERENT `correlation_id`. Gated on the new
+    // dispatch carrying a `correlation_id` — without one we can't tell dispatches
+    // apart (same guard as the in-place dedup above).
+    if let (Some(new_corr), Ok(new_dt)) = (
+        correlation_id,
+        chrono::DateTime::parse_from_rfc3339(&payload.issued_at),
+    ) {
+        for stale in list_pending(home).into_iter().filter(|d| {
+            matches!(d.status, DispatchStatus::Pending | DispatchStatus::Exceeded)
+                && d.dispatcher == dispatcher
+                && d.target == target
+                && d.dispatch_id != dispatch_id
+                && d.correlation_id.as_deref() != Some(new_corr)
+                && chrono::DateTime::parse_from_rfc3339(&d.issued_at)
+                    .map(|t| t < new_dt)
+                    .unwrap_or(false)
+        }) {
+            if delete_sidecar_locked(home, &stale.dispatch_id) {
+                tracing::debug!(
+                    target: "dispatch_idle",
+                    dispatch_id = %stale.dispatch_id,
+                    target = %target,
+                    superseded_by = %dispatch_id,
+                    old_correlation_id = ?stale.correlation_id,
+                    "#1866 retired stale dispatch sidecar — target re-dispatched"
+                );
+            }
+        }
+    }
     Some(dispatch_id)
 }
 
@@ -610,7 +645,7 @@ fn target_is_working(
     target: &str,
     silence_threshold_secs: i64,
 ) -> bool {
-    snapshot
+    let snapshot_working = snapshot
         .and_then(|s| s.agents.iter().find(|a| a.name == target))
         .map(|a| {
             // Working = instantaneous-working state OR active-recovery OR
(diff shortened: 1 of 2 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED (adversarial, both focus points proven safe) + CI 6/6 green. #1866 dispatch_idle false-fires: (b) clear-on-handoff — record_dispatch now retires an armed sidecar that is same-dispatcher + same-target + different-correlation_id + strictly-older (gated on the new dispatch carrying a corr), via the locked delete_sidecar_locked helper. reviewer-2 verified this can't false-kill a parallel dispatch (different dispatcher survives, newer survives, same-corr uses in-place dedup). (a) state-aware suppress — target_is_working now also suppresses on recent heartbeat_at_ms (the agent's own MCP calls — central hook, no daemon keepalive writes it) or last_input_at_ms (inbound inject, which updates last_input NOT heartbeat so receiving a message can't fake activity), windowed at the silence threshold so the dispatch's own input expires exactly at nudge-due. reviewer-2 verified a genuinely wedged agent (no MCP + no new input in-window) has both signals stale → still fires — the fix only ADDS suppression for provable recent activity, never hides a real stuck. The nudge tone is now informational (FYI ... no action needed if mid-task ... set_waiting_on as the escape hatch). section-3.9 through real record_dispatch/scan_and_emit. Merging --admin squash. Takes effect next restart.

PR #1871, opened 2026-06-08
task-board: validate done/update transitions under the append lock (#1868)

### LINKED ISSUE #1868: task-board done/update validate transition OUTSIDE the event-log lock (claim hardened, siblings not) — illegal transitions silently land
## Summary
Bug-hunt finding (fixup-dev, empirically verified). `handle_claim` is concurrency-hardened via `append_checked` (re-replays fresh state + re-runs the precondition INSIDE the event-log lock), but its siblings `handle_done` and `handle_update` are NOT — they check `can_transition_to` OUTSIDE the lock and then blind-append. An illegal task-status transition can silently land under concurrency.

## RCA (file:line)
- `src/tasks/handler.rs` handle_done:283-365 (bare `append`), handle_update:419-506 (`append_batch`) — `can_transition_to` is evaluated at handler.rs:310/506 from an out-of-lock `read_task_record`/`list_all`, then the append happens without re-validating inside the lock.
- Contrast handle_claim:231 → `append_checked` re-replays + re-checks the precondition under the lock.
- `src/task_events.rs` apply_done/apply_simple_status:823-867 blindly set `t.status = status` with NO `can_transition` check → replay is not a second gate either.

## Trigger
Concurrent mutation of the same task across the read→append window — e.g. daemon-side `sweep_overdue_claimed` (tasks/mod.rs:240, releases a claim → Open) or `auto_close` (PR-merge → Done) racing an agent-side done/update. The out-of-lock check sees the stale status, passes, blind-appends, and replay blindly applies it.

## Impact
Illegal transitions land silently (e.g. Cancelled→Done, sweep-Open→InProgress) → the task board shows a wrong status with no rejection → misleads orchestration. MEDIUM.

## Fix direction
Make handle_done / handle_update use `append_checked` (move the `can_transition_to` re-validation inside the event-log lock, consistent with handle_claim). Optionally add the can_transition guard to apply_done/apply_simple_status so replay rejects too. Same 'fixed claim, missed the sibling' asymmetry as #1866.

suzuke:
Closes #1868 (found in the dispatch/task-board bug-hunt).

## Bug

`handle_done` (plain `append`) and `handle_update` (`append_batch`) check `can_transition_to` **outside** the event-log lock, then append — while `handle_claim` was hardened to `append_checked` (fresh replay + precondition **under** the lock). Replay's `apply_done`/`apply_status` blindly set `t.status` with **no** transition guard. So a concurrent writer (daemon `sweep_overdue_claimed` / `auto_close`, or a peer mutation) that moves the task between the out-of-lock read and the append lets an **illegal transition** (e.g. `Cancelled→Done`) commit silently → wrong board state that misleads orchestration. (The claim path was fixed for exactly this TOCTOU; done/update were the un-hardened siblings.)

## Fix — close the TOCTOU symmetrically with claim

- Add `task_events::append_batch_checked` — the batch analog of `append_checked`: a precondition evaluated under the append lock against a **fresh replay**; on failure no event is written and the reason returns as `Ok(Err)`.
- `handle_done` → `append_checked`; `handle_update` → `append_batch_checked`, each re-validating `can_transition_to` against the committed state **inside** the lock. The out-of-lock checks remain as fast-rejects; the in-lock precondition is the authoritative gate.

**Deliberately skipped** the optional replay-guard layer (`apply_*` + `can_transition`): it would change the folded state of **existing** logs that already contain a pre-fix illegal transition on restart — a risky behavior change on historical data. The `append_checked` layer prevents *new* illegal transitions, which is the fix the lead approved.

## Tests (§3.9)

- `append_checked_rejects_stale_done_after_concurrent_cancel_1868` — a `done` prepared as-if-Claimed is rejected once the committed state moved to Cancelled; the task stays Cancelled (no Done event lands).
- `append_batch_checked_rejects_stale_update_after_concurrent_cancel_1868` — same for the multi-event update arm.
- `normal_done_and_update_still_succeed_1868` — the uncontended claim→done / claim→update sequence still succeeds end-to-end through the real handlers (no regression).

## Verification

- `tasks::` 105/105 · `task_events::` 32/32

### DIFF
--- a/src/task_events.rs
+++ b/src/task_events.rs
@@ -1007,6 +1007,74 @@ pub fn append_batch(
     Ok(seqs)
 }
 
+/// Batch variant of [`append_checked`] (#1868): append ALL `events` atomically
+/// iff `precondition` — evaluated under the append lock against a FRESH on-disk
+/// replay — returns `Ok`. Closes the same TOCTOU as [`append_checked`] for the
+/// multi-event `update` arm (a status transition plus priority/desc/tags/owner
+/// events emitted as one batch). On precondition failure NO event is written and
+/// the reason is returned as `Ok(Err(reason))`; the outer `Err` is reserved for
+/// IO / replay failures.
+pub fn append_batch_checked<F>(
+    home: &Path,
+    instance: &InstanceName,
+    events: Vec<TaskEvent>,
+    precondition: F,
+) -> anyhow::Result<Result<Vec<u64>, String>>
+where
+    F: FnOnce(&TaskBoardState) -> Result<(), String>,
+{
+    if events.is_empty() {
+        return Ok(Ok(Vec::new()));
+    }
+    let instance = instance.clone();
+    let count = events.len();
+    let mut seqs: Vec<u64> = Vec::with_capacity(count);
+    let now = chrono::Utc::now().to_rfc3339();
+    let emitter_id = match crate::agent::resolve_instance(home, instance.as_str()) {
+        Ok((id, _)) => Some(id.full()),
+        Err(e) => {
+            tracing::debug!(instance = %instance, error = %e, "emitter ID resolution failed");
+            None
+        }
+    };
+
+    let mut rejection: Option<String> = None;
+    crate::event_log::append_lines_under_lock(home, LOG_NAME, |log_path| {
+        // FRESH replay under the lock — authoritative committed history.
+        let state = replay_uncached(home)?;
+        if let Err(reason) = precondition(&state) {
+            rejection = Some(reason);
+            return Ok(Vec::new()); // empty ⇒ no write
+        }
+        let start_seq = max_seq_for_instance(log_path, &instance)? + 1;
+        let mut lines = Vec::with_capacity(count);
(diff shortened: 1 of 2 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED + CI green. #1868: handle_done (now append_checked) and handle_update (now append_batch_checked) re-validate can_transition_to against fresh replayed state INSIDE the event-log lock (consistent with handle_claim), so a stale out-of-lock illegal transition is rejected (Ok(Err), zero write) rather than silently landing. append_batch_checked faithfully mirrors append_checked; the batch is atomic (precondition before the build loop → all-or-nothing). reviewer-2 confirmed the replay guard was CORRECTLY omitted — adding can_transition to apply_done/apply_status would reject pre-existing historical illegal transitions on replay and regress the board state (Done → back to Cancelled), a real user-visible regression. section-3.9 covers the handler-loses-to-concurrent-daemon direction. Merging --admin squash. (Follow-up filed: the 6+ daemon auto-→Done write paths are not yet hardened — one-way protection — tracked separately.)

Issue #1870, opened 2026-06-08
Bug-hunt: 3 watchdog/GC findings — H1 worktree-GC parse-fail reclaims within grace, H2 hung-anchor TOCTOU false-P0, H3 clock-backward class


suzuke:
Bug-hunt batch (fixup-dev-2, empirically verified). 3 findings; H1 is the actionable MED, H2/H3 queued.

## H1 [MED] worktree GC: released_at parse failure skips the grace window → immediate reclaim (fail-direction)
`src/worktree_pool.rs:906-917` (gc_candidates clean-release path). The GC_GRACE_HOURS (24h) check is nested inside `if let Ok(parse_from_rfc3339(released_at))`; a marker whose `released_at=` fails to parse skips the whole block and falls through to an unconditional `Some(GcCandidate{kind: CleanRelease})` → the worktree is reclaimed/archived immediately, bypassing the grace window meant to protect operator WIP. Trigger: a corrupt/half-written `released_at=` (partial write / crash mid-release). Impact: a just-released worktree the operator may still be inspecting (PR/CI/debug) is GC'd early. Fix: parse Err must be conservative (return None / keep), not treated as 'past grace'. ~1 line. (No sibling of this fall-through in orphan_sweep/boot_sweep/auto_release/retention — isolated.)

## H2 [MED-LOW] hung-anchor clear has a TOCTOU → stale hung_since persists → one false P0 after restart
`src/daemon/per_tick/hang_detection.rs:161-180` (#1744-H2 left_hung loop). Contract: after an agent leaves Hung, persist the cleared hung_since so a restart doesn't rehydrate a stale anchor. But if the agent unregisters (crash/DELETE on another thread) between `load_for` (present) and the registry lookup, `snapshot=None` → persist is skipped → stale hung_since stays on disk. Trigger: a self-orch agent leaves Hung then crashes in the same tick. Impact: after a daemon restart it rehydrates the stale anchor → the first post-restart tick computes still-Hung → one false P0 escalation (the exact class #1744-H2 was meant to kill). Fix: when the agent isn't in the registry, clear hung_since via escalation_persist instead of skipping.

## H3 [LOW] clock-backward is a CLASS, not just cron (N2/#1852) — also anti_stall + handoff_timeout
[...]

PR #1872, opened 2026-06-08
fix(#1870 H1): worktree GC keeps a worktree with a malformed released_at (fail conservative)

suzuke:
Bug-hunt finding **H1** (filed #1870), lead-approved as the highest-value fix (protects operator WIP).

## Bug (`src/worktree_pool.rs`, `evaluate_candidate` clean-release path)
The `GC_GRACE_HOURS` (24h) check sat **inside** `if let Ok(parse_from_rfc3339(released_at))`:
```rust
if let Ok(dt) = parse_from_rfc3339(ts) {
    if age < GC_GRACE_HOURS { return None; } // within grace
}
Some(GcCandidate { kind: CleanRelease, .. }) // ← parse-Err falls through here
```
A **malformed `released_at=`** in the `.agend-managed` marker (realistically a partial-write / crash-truncated marker) fails the parse, **skips the grace check entirely**, and falls through to an unconditional `CleanRelease` candidate → the worktree is reclaimed immediately, bypassing the grace window that exists to protect a just-released worktree's WIP (operator may still be reviewing the PR / CI / debugging).

## Fix
Parse failure now **fails conservative** — `Err(_) => return None` (keep the worktree, skip GC) instead of treating it as "past grace". Valid `released_at` is unchanged (within grace → kept; past grace → candidate).

## §3.9
`gc_candidates_skips_malformed_released_at_1870`: a worktree with `released_at=not-a-timestamp` is NOT in `gc_candidates`; a sibling with a valid past-grace `released_at` still IS. Regression-proof — revert and the malformed worktree reappears as a candidate.

Local CI-parity preflight green (`fmt`/`clippy --all-targets --features tray -D warnings`/`cargo test --tests`; existing grace-ttl / new-layout / daemon-tagged gc tests all still pass).

### DIFF
--- a/src/worktree_pool.rs
+++ b/src/worktree_pool.rs
@@ -903,11 +903,21 @@ fn evaluate_candidate(
             if binding_present {
                 return None;
             }
-            if let Ok(dt) = chrono::DateTime::parse_from_rfc3339(ts) {
-                let age = chrono::Utc::now().signed_duration_since(dt.with_timezone(&chrono::Utc));
-                if age < chrono::Duration::hours(GC_GRACE_HOURS) {
-                    return None; // still within grace
+            match chrono::DateTime::parse_from_rfc3339(ts) {
+                Ok(dt) => {
+                    let age =
+                        chrono::Utc::now().signed_duration_since(dt.with_timezone(&chrono::Utc));
+                    if age < chrono::Duration::hours(GC_GRACE_HOURS) {
+                        return None; // still within grace
+                    }
                 }
+                // #1870 (H1): a malformed `released_at=` (e.g. a partial-write /
+                // crash-truncated marker) MUST NOT be treated as "past grace". The
+                // grace window exists to protect a just-released worktree's WIP, so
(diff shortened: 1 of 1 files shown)


suzuke:
Merge audit (section 3.12 mirror) — lead-review VERIFIED + the 5 required checks green (Coverage non-required, lagging). #1870 H1: the worktree-GC clean-release path's grace-window check was nested in an if-let-Ok(parse released_at), so a malformed/half-written released_at= fell through to an unconditional Some(GcCandidate{CleanRelease}) → reclaimed within the GC_GRACE_HOURS window that protects just-released operator WIP. Now a match with Err(_) => return None fails conservative (keep the worktree, skip GC); a valid released_at is unchanged (respects grace, then reclaims). section-3.9 gc_candidates_skips_malformed_released_at_1870 (malformed → no candidate; valid past-grace → still a candidate; regression-proof). Merging --admin squash. (H2 hung-anchor TOCTOU + H3 clock-backward class remain open in #1870.) Takes effect next restart.

Issue #1873, opened 2026-06-08
task-board: 6+ daemon auto-→Done write paths bypass the in-lock can_transition gate (one-way protection after #1868)


suzuke:
Follow-up to #1868 (reviewer-2 lens-3 finding). #1871 hardened the 3 HANDLER paths (claim/done/update) to re-validate can_transition inside the event-log lock. But the protection is ONE-WAY: handle_done's own comment names sweep_overdue_claimed/auto_close as the concurrent writers it defends against, yet those daemon writers themselves write TaskEvent::Done with NO in-lock re-validation and NO replay backstop. So a daemon racing a handler (e.g. the user just cancelled a task, then a sweep/auto_close fires) still lands an illegal Cancelled→Done silently.

## Unhardened daemon →Done write sites (bare append/append_batch)
- src/daemon/task_sweep.rs:362 (append_batch, PR-merged auto-close → Done — no status precheck at all)
- src/tasks/auto_close.rs:55 (append, ReportAutoClose → Done — only an out-of-lock precheck → same TOCTOU)
- src/daemon/cron_tick.rs:849
- src/daemon/auto_release.rs:1327
- src/daemon/dispatch_idle/mod.rs:1925
- src/daemon/idle_watchdog.rs:1659

## Fix direction
Route the daemon →Done paths through append_checked (write-layer guard, consistent with the #1868 preference of guarding the write layer rather than the replay layer). Each daemon writer should re-validate can_transition_to(Done) against fresh state inside the lock; if the task already moved to a terminal/incompatible state (e.g. Cancelled), skip the Done write (and log). Verify this doesn't break the legitimate auto-close/sweep flows (a normally-Open/Claimed task still closes).

suzuke:
Fixed by #1874. The 2 production daemon runtime →Done writers (auto_close, task_sweep) now route through append_done_if_legal (in-lock can_transition re-validation), so a daemon racing a handler no longer lands an illegal Cancelled→Done. Verified (dev verify-first + reviewer-2 site-by-site self-correction) that the other listed sites are test-only or a legacy-migration rebuilder (correctly excluded) — no production runtime site missed.

PR #1874, opened 2026-06-08
task-board: guard daemon auto-close →Done writes under the lock (#1873)

suzuke:
Refs #1873 (reviewer-2 lens-3 follow-up to #1868).

## Bug

#1868 hardened the user `done`/`update` handlers with `append_checked`, but the **daemon auto-close** paths still bare-appended their →Done writes — so protection was one-directional. A daemon auto-close racing a peer/operator `cancel` (its out-of-lock status check passed, then a Cancel committed before the append) silently flipped a **Cancelled (terminal)** task to Done; replay's `apply_done` does not re-guard transitions.

## Verified scope (not assumed)

Of the candidate sites, **only TWO are production daemon →Done writers**:
- `tasks/auto_close.rs` — `auto_close_on_report` (single `Done`)
- `daemon/task_sweep.rs` — sweep (`Linked` + `Done` batch)

The other listed sites — `cron_tick.rs:849`, `auto_release.rs:1327`, `idle_watchdog.rs:1659`, `dispatch_idle/mod.rs:2113`, `tasks/mod.rs:379` — are all inside `#[cfg(test)]` blocks (test-seed appends), **not** production paths. (Classified each by comparing its line to the file's `#[cfg(test)]` start.) So the real fix is the 2 production sites.

## Fix

Add `task_events::append_done_if_legal(home, instance, task_id, events)` — routes the →Done batch through `append_batch_checked` with a precondition that re-validates `can_transition_to(Done)` against **fresh committed state under the append lock**. On a terminal/incompatible state it writes nothing and returns `Ok(false)` (logged at info — the operator already holds the canonical terminal signal). Both production sites call it:
- `auto_close_on_report` gates its `cleanup_pending_for_task_id` on the result;
- the sweep gates its "auto-closed" log (a cancelled marker drops out of `open_ids` next cycle → no re-attempt).

## Tests (§3.9)

- `append_done_if_legal_rejects_cancelled_keeps_legal_1873` — the shared guard (used by **both** sites) rejects a →Done on a Cancelled task (stays Cancelled, nothing written) and still auto-closes a legal Claimed task.
- Existing auto_close happy-path + already-cancelled-idempotent tests cover the end-to-end path (no regression).

## Verification

- `auto_close::` 8/8 · `task_sweep::` 29/29 · `task_events::` 32/32

I left this as `Refs #1873` (not `Closes`) so the lead/reviewer-2 can confirm the "4 sites are test-only" finding before closing.

### DIFF
--- a/src/daemon/task_sweep.rs
+++ b/src/daemon/task_sweep.rs
@@ -359,12 +359,18 @@ fn sweep_tick(home: &Path) -> anyhow::Result<()> {
                     },
                 },
             ];
-            task_events::append_batch(home, &emitter, events)?;
-            tracing::info!(
-                pr = pr.number,
-                marker = %marker,
-                "sweep: auto-closed (Linked + Done emitted)"
-            );
+            // #1873: re-validate →Done UNDER the lock. `open_ids` was snapshotted
+            // at sweep start; a marker task cancelled since must NOT be flipped to
+            // Done (the whole Linked+Done batch is skipped — a cancelled task drops
+            // out of `open_ids` next cycle, so no re-attempt).
+            let closed = task_events::append_done_if_legal(home, &emitter, &marker, events)?;
+            if closed {
+                tracing::info!(
+                    pr = pr.number,
+                    marker = %marker,
+                    "sweep: auto-closed (Linked + Done emitted)"
+                );
+            }
         }
     }
 
--- a/src/task_events.rs
+++ b/src/task_events.rs
@@ -1145,6 +1145,45 @@ where
     Ok(Ok(0))
 }
 
+/// #1873: append a DAEMON-originated →Done write (auto-close / sweep) iff the task
+/// can still legally transition to Done UNDER the append lock against fresh
+/// committed state. #1868 hardened the user `done`/`update` handlers, but daemon
+/// auto-close paths still bare-appended — so a daemon racing a peer/operator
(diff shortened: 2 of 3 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED (with an honest self-correction of its own #1871 enumeration) + CI 6/6 green. #1873: routes the production daemon auto-→Done writers through append_done_if_legal (append_batch_checked + in-lock can_transition_to(Done) re-validation; terminal/incompatible → Ok(false) + info-log, no write). dev's verify-first correctly found that only 2 of the 6 sites reviewer-2 originally listed are production RUNTIME →Done writers (auto_close.rs:58, task_sweep.rs:362) — the other 4 (cron_tick/auto_release/dispatch_idle/idle_watchdog) are #[cfg(test)] seeds, and tasks/mod.rs:379 is production but a legacy-migration rebuilder that is correctly NOT hardened (applying can_transition would reject legitimately-already-Done legacy tasks → migration data loss). reviewer-2 re-verified site-by-site and confirmed no production runtime site is missed. Both real sites preserve legitimate auto-close/sweep (an Open/Claimed task still closes); a concurrently-cancelled task is left Cancelled (not flipped to Done). section-3.9 covers reject + legal-pass. Closes the one-way protection gap from #1868. Merging --admin squash.

PR #1875, opened 2026-06-08
fix(#1870 H2): clear hung anchor when a left-Hung agent has left the registry (TOCTOU)

suzuke:
Bug-hunt finding **H2** (filed under #1870), lead-approved after the review queue cleared.

## Bug (`src/daemon/per_tick/hang_detection.rs`, #1744-H2 `left_hung` loop)
The loop persists a self-orchestrator's CLEARED `hung_since` on Hung-exit so a restart doesn't rehydrate a stale anchor. But it did: `load_for` (entry exists) → registry lookup → persist **only if** the lookup found a live handle. If the agent **unregistered** (crash / DELETE on another thread) between the phase-1 registry lock and the phase-3 re-lock — the TOCTOU window — the lookup missed, `snapshot == None`, and the loop **skipped**. The stale `hung_since` stayed on disk → a restart rehydrated it → a **false Hung P0** on the first post-restart tick (the exact thing #1744-H2 exists to prevent).

## Fix
When the agent has a persisted record but is **absent** from the registry, clear `hung_since` **in place** (new `escalation_persist::clear_hung_since`, mirroring `clear_failed_escalated`) instead of skipping — it left Hung AND is gone, so its anchor must not survive to rehydrate. The clear is **surgical**: touches ONLY `hung_since`, preserving the crash budget / cooldowns / `failed_escalated` latch (no over-clear of a still-relevant record).

The per-name loop body is extracted into `persist_or_clear_left_hung_anchor` so the absent-agent branch is testable through a real entry. (A synchronous test can't reproduce the concurrent-unregister TOCTOU itself, but the helper drives the exact `None` branch with an empty registry.)

## §3.9
- `clear_hung_since_clears_only_anchor_preserves_rest_1870_h2` — clears `hung_since` but preserves `failed_escalated` + crash budget + hung cooldown; no-op on missing store / unknown agent.
- `left_hung_anchor_cleared_when_agent_absent_from_registry_1870_h2` — a self-orch with a persisted hung record + empty registry → the real helper clears the anchor (simulated restart sees no stale anchor). Regression-proof — pre-fix the skip leaves `hung_since` Some.

## Reviewer-2 focus
- **Over-clear?** Does the fix ever clear a record that should keep its anchor? (Gated on `left_hung` = already-left-Hung + a persisted entry + self-orch; `clear_hung_since` is field-surgical.)
- Is extracting the loop body behavior-preserving for the in-registry (Some) path?

### DIFF
--- a/src/daemon/escalation_persist.rs
+++ b/src/daemon/escalation_persist.rs
@@ -106,6 +106,30 @@ pub(crate) fn clear_failed_escalated(home: &Path, name: &str) {
     }
 }
 
+/// #1870-H2: clear ONLY the `hung_since` anchor for one agent in place,
+/// preserving the crash budget / cooldowns / `failed_escalated` latch. Used by
+/// the `left_hung` persist loop when an agent has a persisted record but has
+/// already left the registry (crashed / deleted between the existence check and
+/// the registry read): it has left Hung AND is gone, so its `hung_since` must
+/// NOT survive to rehydrate a stale anchor on the next restart (which would fire
+/// a false Hung P0 on the first post-restart tick). No-op when no store / no
+/// entry exists. Distinct from `persist` (which needs a live snapshot) precisely
+/// because here the agent is no longer in the registry.
+pub(crate) fn clear_hung_since(home: &Path, name: &str) {
+    let path = store_file(home);
+    if !path.exists() {
+        return;
+    }
+    if let Err(e) = store::mutate_versioned::<EscalationStore, _, _>(&path, |s| {
+        if let Some(entry) = s.agents.get_mut(name) {
+            entry.hung_since_epoch_ms = None;
+        }
+        Ok(())
+    }) {
+        tracing::warn!(agent = %name, error = %e, "escalation_persist: clear_hung_since failed");
+    }
+}
+
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED (adversarial) + CI 6/6 green. #1870 H2: the #1744-H2 left_hung loop skipped persisting the cleared hung_since when the agent unregistered (crash/DELETE) between load_for and the registry-snapshot lock (TOCTOU) → a stale hung_since survived on disk → after restart it rehydrated → one false Hung P0 escalation. Now the None (absent-from-registry) arm calls a new escalation_persist::clear_hung_since to remove the stale anchor instead of skipping. reviewer-2 verified it can't over-clear (the clear only fires when the agent has already left Hung AND left the registry; field-surgical — only zeroes hung_since_epoch_ms, preserves crash budget / cooldowns / failed_escalated latch; only mutates an existing entry) and that the extracted persist_or_clear_left_hung_anchor leaves the in-registry (Some) path byte-identical. section-3.9 drives the real helper with an empty registry (the post-TOCTOU None branch) + a surgical-preservation test (regression-proof). Merging --admin squash. Takes effect next restart.

PR #1876, opened 2026-06-08
fix(#1870 H3): single clamp helper for backward-clock-skew elapsed reads

suzuke:
Bug-hunt finding **H3** (filed under #1870) — the cron N2 backward-skew class beyond cron.

## Bug
A watchdog computing `elapsed = now.signed_duration_since(ts)` vs a positive threshold goes **negative** when `ts` is in the future relative to `now` — a backward clock skew (NTP correction, VM resume / snapshot restore, DST). cron N2 (#1852) fixed one site; the same shape lived in `anti_stall` + `handoff_timeout`.

## Fix
One shared `crate::daemon::utils::elapsed_since(now, ts)` = `max(0, signed_duration_since)`, routed through the class:
- `anti_stall::check_stalled` (elapsed vs `eta*multiplier`)
- `handoff_timeout_watchdog::scan_and_emit` (handoff age vs re-nudge / escalate)
- cron N2's forward-elapsed read — the dedicated `last_check_in_future` re-stamp **still owns** the backward case (clamp there is a no-op), so cron's persisted-wedge fix is unchanged; this only folds the elapsed READ into the one helper.

## Honest scope note (for the reviewer)
The clamp is **byte-identical** for the normal `ts ≤ now` case. At these threshold comparisons, negative and zero behave **the same** (both `< threshold` → neither fired), so for anti_stall + handoff this is a **defensive class-unification** that makes the negative provably-benign and guards future use — *not* a present false-fire fix. The one genuine persisted wedge in the class was cron's `last_check` (already fixed by #1852's re-stamp), now reading elapsed through the same helper. I'm flagging this rather than overclaiming an active bug at those two sites.

## §3.9
- `elapsed_since_clamps_future_to_zero_noop_for_past_1870_h3` — **the regression-proof anchor** (the one place the clamp is observable): a past ts is byte-identical to the raw `signed_duration_since`; `ts==now` and a FUTURE ts both clamp to zero (not a negative `Duration`).
- `check_stalled_future_started_at_does_not_wedge_1870_h3` — a future `started_at` produces no false stall; existing forward tests pin unchanged stall firing.

## Reviewer-2 focus
- Is cron's `last_check_in_future` re-stamp behavior **unchanged** (clamp at line 170 is a no-op for the backward case)?
- Any of the routed sites where negative≠0 actually mattered downstream (it shouldn't — all are `> / >=` positive-threshold reads)?

### DIFF
--- a/src/daemon/anti_stall.rs
+++ b/src/daemon/anti_stall.rs
@@ -139,7 +139,9 @@ pub(crate) fn check_stalled(
     // skip stall detection (can't compute elapsed without one).
     let last_alive = crate::daemon::task_progress::read_last_progress_at(home, &task.id)
         .or_else(|| task.started_at.as_deref().and_then(parse_rfc3339))?;
-    let elapsed_secs = now.signed_duration_since(last_alive).num_seconds();
+    // #1870-H3: clamp so a backward clock skew (future `last_alive`) can't make
+    // elapsed negative → stall detection silently stop firing.
+    let elapsed_secs = crate::daemon::utils::elapsed_since(now, last_alive).num_seconds();
     let stall_threshold = (eta_secs as f64 * STALL_MULTIPLIER) as i64;
     if elapsed_secs > stall_threshold {
         Some(format!(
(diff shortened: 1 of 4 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED (confirmed dev-2's honest scope note: purely defensive, zero current behavior change) + 5 required checks green (Coverage non-required, lagging). #1870 H3: a shared elapsed_since(now, ts) = signed_duration_since(ts).max(0) helper, with anti_stall + handoff_timeout + cron's forward-elapsed read routed through it. This unifies the backward-clock (N2/#1852) class. dev-2 honestly flagged (and reviewer-2 confirmed) that for anti_stall/handoff this is DEFENSIVE only — in a  comparison a negative elapsed behaves identically to 0 (both don't fire), so there was no live false-fire there; the real persisted wedge was cron's last_check, already fixed by #1852's re-stamp (whose clause is untouched here — the clamp is a no-op for it). Normal-forward elapsed is byte-identical (max(0) no-op for ts<=now); only a future ts (backward jump) clamps to 0. Prevents future misuse of the pattern. section-3.9: helper regression-proof anchor + anti_stall future-started_at real-site test. Merging --admin squash.

Issue #1031, opened 2026-05-21
P3: ci_watch [ci-pass] payload enrichment — add PR#/head_sha/task_id for actionable reviewer wake


suzuke:
## Severity
**P3** — friction-only follow-up surfaced during #1030 dialectic. After #1030 Bundle A lands and reviewers wake on CI green correctly, payload information may still require reviewer to run \`gh pr list\` / \`git rev-parse\` to map event → actionable PR context.

## Context

Per dev #1030 primary memo H2 hypothesis + reviewer + dev-2 cross-audit consensus:
- Current \`[ci-pass]\` payload lacks PR#, full head_sha, and task_id correlation
- After #1030 H1 fix, reviewers receive wake-aware notification but must still query for context
- Information friction tolerable vs the H1 wake gap — defer to follow-up

## Fix shape (TBD)

Enrich \`[ci-pass]\` and \`[ci-ready-for-action]\` payloads in \`src/daemon/ci_watch/poller.rs\` with:
- \`pr_number\` (from gh-poll observation cache)
- \`head_sha\` (full, not truncated)
- \`task_id\` (correlation_id from dispatch chain)
- \`branch\` (already present)

Likely ~40-50 LOC payload struct refactor + test fixture updates.

## Related
- #1030 — H1 wake gap fix (Bundle A)
- Hypothesis source: \`/tmp/dialectic-1030-primary-dev.md\` §H2

Issue #931, opened 2026-05-19
[P0] ci_watch subscription dropped on worktree bind/release handoff


suzuke:
## Problem

Daemon-managed worktree binding (`bind_self` / `release_worktree`) silently invalidates the previously-installed `ci_watch` subscription for the handing-off agent. CI pass notifications never arrive at reviewer/dev — they appear idle but are genuinely uninformed.

## Empirical recurrence

Overnight #872/#911/#910/#914/#922/#927 campaign:
- PR #920 (fix/914) — reviewer never got ci-ready, lead self-merged
- PR #925 (fix/910 PR3) — 5h stall before lead caught + self-merged
- PR #928 (fix/922) — reviewer never got ci-ready, lead self-merged
- PR #929 (fix/927-pr-a) — reviewer never got ci-ready, lead self-merged

4-in-a-row over a single overnight session.

## Current workaround

Merging agent must `ci action=watch repo=... branch=...` independently after worktree handoff. Memory entry exists: \`feedback_ci_watch_survives_bind_release.md\`.

## Impact

- Every multi-PR campaign requires manual ci_watch re-subscribe per agent
- Reviewer appears idle but is uninformed → process inversion (lead absorbs merge duty)
- Active-poll discipline papers over this but doesn't fix root cause

## Repro

1. Agent A: \`ci action=watch repo=... branch=...\`
2. Agent A: \`bind_self\` (daemon worktree)
3. CI completes on watched branch
4. Agent A receives no notification — ci_watch silently lost

## Scope estimate

~50-100 LOC. Decouple ci_watch subscription lifecycle from worktree bind/release lifecycle. Subscription persists across bind handoff unless explicitly \`unwatch\`ed.

## Priority

P0 — recurring pain on every multi-PR cycle.

PR #1880, opened 2026-06-08
dispatch: thread second_reviewer→review_class=dual into the auto-armed watch (#1877)

### LINKED ISSUE #1877: second_reviewer dispatch signal is validated then dropped — dual review silently degrades to single (re-marshal class)
## Summary
Bug-hunt finding (fixup-dev, verified via git grep). send(kind=task, second_reviewer=true, second_reviewer_reason=...) validates the flag (and FORCES a reason) at the MCP boundary, but the signal is never threaded to the mechanism — so the auto-armed CI watch stays review_class=single and the requested DUAL review silently never happens. The lead thinks dual review is on (the flag was accepted and a reason was required) but it isn't. Same re-marshal/allowlist-drop class as #1833 / HIGH-1 / MED-2.

## RCA (file:line)
- comms.rs:234-238 reads + validates second_reviewer / second_reviewer_reason; tools.rs:46 schema documents it as 'Signal dual review (section 3.5)'.
- git grep second_reviewer (non-test) appears ONLY at comms.rs:234-238 + the tools.rs schema — nowhere after. It is not stored on the task, not passed to the dispatch hook, not included in the dispatch message built at comms.rs:268+.
- dispatch_auto_bind_lease_with_source_and_chain (dispatch_hook:319) has no review_class / second_reviewer parameter; the auto-arm watch_args (dispatch_hook:520) doesn't set review_class.
- The dual-review EXECUTION path (review_class=dual) is only ever set by an explicit ci action=watch review_class=dual (ci/mod.rs:674) — never by the dispatch second_reviewer flag.

## Impact
HIGH — the dual-review quality gate silently fails for every dispatch that requests it via second_reviewer=true. The CI gate runs single-review while the operator/lead believes a second reviewer is enforced.

## Fix direction
Thread second_reviewer (→ review_class=dual) from the send handler (comms.rs) through the dispatch chain (dispatch_auto_bind_lease_with_source_and_chain / the auto-arm watch_args at dispatch_hook:520) so a second_reviewer=true dispatch arms a review_class=dual CI watch. Add a test that a second_reviewer=true dispatch results in a dual-class watch (would have caught the evaporation). Consider a broader guard against this recurring re-marshal-drop class (e.g. an assertion/test that MCP-accepted dispatch directives reach the armed watch).

suzuke:
Closes #1877 (bug-hunt C1 — 4th occurrence of the re-marshal-drop class).

## Bug

`send(kind=task, second_reviewer=true)` reads + **validates** the dual-review flag in `comms.rs` (requires a non-empty reason), then **never uses it again** — `git grep second_reviewer` (non-test) finds it only at the read/validate site + the tools.rs schema. It's not stored, not in the dispatch message, and never reaches the auto-armed ci-watch's `review_class`. Since `review_class` (the actual dual-review enforcement, read by the poller's `record_ci_result`) is set **only** via an explicit `ci action=watch review_class=dual`, a `second_reviewer=true` dispatch silently gates **single** review — the operator sets the flag, it's accepted + validated, and it evaporates.

## Fix

Thread the directive through the dispatch chain exactly like #931 did for `next_after_ci` and #1031 for `task_id`:
- `comms.rs::handle_delegate_task` derives `review_class = second_reviewer.then(|| "dual")` and passes it to `dispatch_auto_bind_lease_with_chain`;
- the param flows through to `dispatch_auto_bind_lease_with_source_and_chain`, which adds `review_class` to the auto-armed `watch_args` when set.

Unset (normal dispatch) leaves the watch single — no over-upgrade. The explicit `ci action=watch review_class=dual` path is untouched.

## Tests (§3.9 + regression guard)

- `all_dispatch_directives_reach_armed_watch_1877` — **the guard the lead requested**: one dispatch carrying `next_after_ci` + `task_id` + `review_class` asserts ALL THREE land on the armed watch, pinning the recurring drop class (#931 / #1031 / #1877) so a future re-marshal can't silently drop any directive.
- `dispatch_without_review_class_stays_single_1877` — a normal dispatch must NOT over-upgrade to dual.

## Verification

- `dispatch_hook::` 50/50

### DIFF
--- a/src/mcp/handlers/comms.rs
+++ b/src/mcp/handlers/comms.rs
@@ -301,6 +301,11 @@ pub(super) fn handle_delegate_task(home: &Path, args: &Value, sender: &Option<Se
         } else {
             let repo_arg = args["repository"].as_str();
             let next_after_ci_arg = args["next_after_ci"].as_str();
+            // #1877: thread the dual-review directive to the auto-armed watch. A
+            // `second_reviewer=true` dispatch (validated above) must arm a
+            // `review_class=dual` watch; without this the flag was accepted +
+            // validated then silently dropped (CI gated single review).
+            let review_class_arg = if second_reviewer { Some("dual") } else { None };
             if let Err(e) = super::dispatch_hook::dispatch_auto_bind_lease_with_chain(
                 home,
                 target,
@@ -308,6 +313,7 @@ pub(super) fn handle_delegate_task(home: &Path, args: &Value, sender: &Option<Se
                 branch,
                 repo_arg,
                 next_after_ci_arg,
+                review_class_arg,
             ) {
                 return json!({"ok": false, "error": format!("dispatch rejected: {e}")});
             }
--- a/src/mcp/handlers/comms.rs
+++ b/src/mcp/handlers/comms.rs
@@ -299,21 +299,15 @@ pub(super) fn handle_delegate_task(home: &Path, args: &Value, sender: &Option<Se
                 "dispatch_auto_bind_lease skipped (bind: false)"
             );
         } else {
-            let repo_arg = args["repository"].as_str();
-            let next_after_ci_arg = args["next_after_ci"].as_str();
-            // #1877: thread the dual-review directive to the auto-armed watch. A
-            // `second_reviewer=true` dispatch (validated above) must arm a
-            // `review_class=dual` watch; without this the flag was accepted +
-            // validated then silently dropped (CI gated single review).
-            let review_class_arg = if second_reviewer { Some("dual") } else { None };
+            // #1877: second_reviewer → review_class="dual" on the auto-armed watch.
             if let Err(e) = super::dispatch_hook::dispatch_auto_bind_lease_with_chain(
                 home,
                 target,
                 task_id_val,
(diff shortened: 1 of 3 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED (end-to-end second_reviewer→review_class=dual threading + the systemic all-directives-reach-watch guard) + all 6 checks green on e786286. The earlier macos/ubuntu red was a REAL regression (not a flake): #1877 pushed comms.rs from 750 to 756 LOC, tripping the mcp_handler_files_under_500_loc invariant — it only showed in CI because include_str! integration tests aren't always rebuilt locally (stale local pass). Fixed net-0 by inlining two single-use lets (no behavior change). #1877: send(second_reviewer=true) now threads review_class=dual through the dispatch chain into the auto-armed watch (it was validated then dropped → dual review silently degraded to single — the 4th instance of the re-marshal-drop class). A systemic guard (all_dispatch_directives_reach_armed_watch) now reads the real watch sidecar and pins next_after_ci + task_id + review_class so a future drop is caught. Merging --admin squash. Takes effect next restart.

Issue #1879, opened 2026-06-08
Instance delete/release leaks binding.json (full_delete_instance + release_full Unmanaged early-return) → leak + teardown returns Err


suzuke:
Bug-hunt finding (fixup-dev-2, verified). Two sibling sites where deleting/releasing an instance leaves its binding behind.

## BIND-1 [MED] full_delete_instance doesn't clear binding.json → teardown returns Err
src/mcp/handlers/instance_lifecycle.rs:76-176 (full_delete_instance, used by both teardown and delete). It systematically clears ~13 stores (fleet.yaml, topic, working_dir, metadata, teams, tasks, dispatch sidecar, idle activity, escalation_persist, schedules, dispatch_tracking, ci_watch, opencode data) — but NOT runtime/<name>/binding.json (no binding::unbind / clear_binding_state call). Then the trailing name_residual_anywhere audit (:153, which checks binding.json @:228-235) finds the residual → returns Err (:157-176). Trigger: delete/teardown an agent that had bind_self / repo-checkout (a binding) without releasing first. Impact: (a) the binding leaks, (b) teardown returns Err even when everything else succeeded (looks like a partial drop), (c) the residual binding can block a same-name re-bind. Fix: add crate::binding::unbind(home, name) to the cleanup sequence.

## WT-LEAK-2 [LOW] release_full Unmanaged early-return leaks binding (same root)
src/worktree_pool.rs:398-401: Unmanaged(err) => { out.error = Some(err); return out; } returns BEFORE clear_binding_state (:426). Releasing an operator-created (no .agend-managed marker) worktree that has a binding correctly refuses to delete the worktree (protects operator data) but leaves the binding (the comment even says 'binding NOT cleared'). Same root as BIND-1: clearing the binding shouldn't be gated behind the worktree-protection early-return. Fix together with BIND-1.

suzuke:
Fixed by #1881. full_delete_instance clears binding.json before its residual audit (no more leaked binding / spurious teardown Err); release_full's Unmanaged early-return also clears the binding (worktree + error preserved, only the stale daemon binding cleared).

PR #1881, opened 2026-06-08
fix(#1879): clear the worktree binding on instance delete + unmanaged release

suzuke:
Bug-hunt finding bundle **BIND-1 + WT-LEAK-2** (lead-approved). Two sibling binding-cleanup gaps.

## BIND-1 — `full_delete_instance` omitted binding cleanup
`full_delete_instance` (`instance_lifecycle.rs`) cleans ~13 stores but never cleared the worktree binding. So deleting a **bound** agent (one that ran `bind_self` / repo checkout) without a prior release both **leaked** `runtime/<name>/binding.json` (blocking a same-name re-bind) AND tripped the `name_residual_anywhere` audit → the whole teardown returned **Err** despite otherwise succeeding.
**Fix:** `binding::unbind` + `clear_bind_in_flight` just before the residual audit.

## WT-LEAK-2 — `release_full` Unmanaged arm leaked the binding
`release_full`'s `Unmanaged` arm (an operator-created worktree with no `.agend-managed` marker) early-returned **before** `clear_binding_state`. Refusing to delete the worktree (protect operator data) must not also skip binding cleanup — the daemon's binding to an unremovable worktree is stale state that leaks + blocks re-bind.
**Fix:** the arm clears the binding before returning. (It's already the non-dry_run path; the dry_run classifier above mutates nothing, so a dry run still preserves both.)

## §3.9
- `full_delete_clears_binding_and_succeeds_1879` — a bound agent → teardown returns **Ok** (no residual Err) and the binding is gone. Regression-proof: revert the unbind and the audit fails the delete.
- `p0x_release_full_unmanaged_worktree_skipped_safely` (**existing test updated to the new contract**): the unmanaged worktree is preserved but the binding **is** cleared (was: binding kept "for operator visibility" — that's the behavior the lead approved changing).

Local CI-parity preflight green (`fmt`/`clippy --all-targets --features tray -D warnings`/`cargo test --tests`; the comprehensive `full_delete_instance_*` + all `p0x_release_full_*` tests still pass).

## Reviewer-2 focus
- Does clearing the binding on the Unmanaged refusal lose any signal the operator needed? (The worktree + the error message survive; only the daemon's stale binding goes.)
- A no-binding normal delete is unaffected (`binding::unbind` is a best-effort no-op when there's nothing to remove).

### DIFF
--- a/src/mcp/handlers/instance_lifecycle.rs
+++ b/src/mcp/handlers/instance_lifecycle.rs
@@ -146,6 +146,15 @@ pub(crate) fn full_delete_instance(home: &Path, name: &str) -> Result<(), String
     // never created). Best-effort like the steps above.
     let _ = std::fs::remove_dir_all(crate::agent::opencode_data_dir(home, name));
 
+    // #1879 (BIND-1): clear the worktree binding (`runtime/<name>/binding.json`
+    // + its HMAC sidecar + the bind-in-flight flag). Every OTHER store above is
+    // cleaned here, but binding was missed — so deleting a bound agent (one that
+    // ran bind_self / repo checkout) without a prior release both leaked the
+    // binding (blocking a same-name re-bind) AND tripped the residual audit below
+    // → the whole teardown returned Err despite otherwise succeeding.
+    crate::binding::unbind(home, name);
+    crate::mcp::handlers::dispatch_hook::clear_bind_in_flight(home, name);
+
     // Sprint 54 P1-B Bug 1 audit: enumerate every store that still holds
     // the name. If any do, surface a loud error instead of returning
     // success — `auto_start_fleet` revival of a half-deleted instance is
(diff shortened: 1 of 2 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED + all 5 required checks green (the earlier Windows red was a confirmed infra flake — bash startup failure + existing release_full worktree-setup tests, not this PR's change; rerun is green). #1879: full_delete_instance now clears binding.json (binding::unbind + clear_bind_in_flight) before the residual audit, so tearing down an agent that had a binding succeeds (no spurious residual Err) and the binding doesn't leak/block re-bind; release_full's Unmanaged early-return now clears the binding too (worktree + error preserved for operator visibility — only the stale daemon binding is cleared, non-dry_run path only). reviewer-2 verified no operator signal is lost, no-binding deletes are unaffected (best-effort), and the test-contract update (the unmanaged worktree now keeps the worktree but clears the binding) is consistent. Merging --admin squash. Takes effect next restart.

Issue #1882, opened 2026-06-08
Cross-agent double-bind: branch-uniqueness (scan→bind_full) is check-then-act without a shared lock (P0-1.5 invariant break)


suzuke:
## Summary
Bug-hunt finding (fixup-dev-2, verified). The 'one branch ≤ one agent' invariant (P0-1.5) is enforced by scan_existing_branch_binding at dispatch_hook/mod.rs:358 before bind_full, but scan→bind_full is not atomic: there is no shared flock/lease across the check and the act. BindGuard (dispatch_hook:147) only prevents the SAME agent's in-flight; bind_full (binding.rs:68) only takes the per-agent .binding.json.lock. So two DIFFERENT agents can both pass the uniqueness scan and both bind the same branch.

## RCA (file:line)
- dispatch_hook/mod.rs:358 — scan_existing_branch_binding (the P0-1.5 check) reads existing bindings, finds no conflict.
- binding.rs:68 bind_full — takes only the per-agent <name>.binding.json.lock, not a branch-level lock.
- Interleaving: agent A scans (no conflict) → agent B scans (no conflict) → A bind_full(branch X) → B bind_full(branch X) (each under its own per-agent lock) → both bound to X. P0-1.5 broken.

## Trigger
Two different agents dispatched to the SAME branch within the scan→bind window (a few file ops, ms-scale). MCP tool calls run on worker threads → genuinely parallel.

## Impact
MED — double-bind: both agents believe they hold the branch, both may push → conflict. binding_state reports cross_branch_holders after the fact (detection, not prevention). Narrow window but a real invariant break.

## Fix direction
Make scan→bind atomic: wrap the uniqueness scan + bind_full in a shared branch-level lease flock (a lock keyed on the branch, not the agent), OR bind-then-re-scan and roll back the loser if a conflict is detected. Prefer the shared-lock approach if a branch-level lock path exists; otherwise the bind+re-scan+rollback (the loser releases) is acceptable. Verify a normal single-agent bind is unaffected (no new contention) and the existing per-agent BindGuard/rollback still works.

suzuke:
Fixed by #1883. The branch-uniqueness scan→bind is now atomic via a shared branch-keyed lease flock held across scan + bind_full, applied to ALL 3 production bind paths (dispatch, bind_self, repo-checkout) — so two agents racing to the same branch result in exactly one binding (the loser gets a clean cross_agent_conflict).

PR #1883, opened 2026-06-08
fix(#1882 BIND-2): atomic branch lease via per-branch flock — close double-bind race

suzuke:
Bug-hunt finding **BIND-2** (lead-approved). Closes the cross-agent double-bind CAS race.

## Bug
P0-1.5 ("a branch is held by at most one agent") was a non-atomic check-then-act: `dispatch_auto_bind_lease_with_chain` scans `scan_existing_branch_binding` (`dispatch_hook.rs:358`) then `bind_full` (~`:450`) with only **per-agent** locks (`BindGuard` + the per-agent `.binding.json.lock`). Two **different** agents racing to lease the **same** branch both pass the scan (neither has bound yet) and both bind → double-bind, both able to push.

## Fix — option (a): branch-level lease flock
New `binding::acquire_branch_lease_lock(home, branch)` takes an exclusive flock keyed on a hash of the **branch**, acquired before the scan and held (RAII) across `bind_full`. The second racer **blocks** until the first drops, then its rescan sees the first's binding and rejects with the existing `LeaseConflict` error — exactly one wins, no rollback/tiebreak needed.

- **Covers both bind paths**: dispatch auto-bind AND `bind_self` both funnel through this fn (`bind_self` → `dispatch_auto_bind_lease_with_source` → `_with_chain`), so one lock covers both.
- **Short mutex, not a lease-lifetime lock**: the persistent lease is `binding.json` (which the scan reads), so the flock only guards the bind operation — **`release_full` needs no lock cleanup** (there's no held-across-the-lease lock to clean).
- **No deadlock**: lock order is consistent (per-agent `BindGuard` → branch lock → per-agent binding lock inside `bind_full`); different branches use different lock files → no cross-branch contention; normal single-agent binds unaffected.

## §3.9
- `branch_lease_lock_serializes_same_branch_allows_different_1882` — a **real two-thread test**: the same branch serializes (the second acquirer blocks while the first is held, proceeds once it drops); different branches acquire concurrently. The regression-proof anchor for the lock's defining behavior.
- Existing `bind_self_rejects_cross_agent_branch_conflict` (loser rejected), `bind_self_idempotent_same_agent_same_branch` (no self-deadlock), and the 50-iter `dispatch_auto_bind_lease_stress` all still pass **through** the lock (no deadlock/regression).

Local CI-parity preflight green (`fmt`/`clippy --all-targets --features tray -D warnings`/`cargo test --tests`, incl. invariants).

## Reviewer-2 focus (concurrency correctness + deadlock)
- **Deadlock**: is the lock order (`BindGuard` → branch flock → per-agent binding flock) cycle-free on every path? Any path that holds the branch flock while acquiring a lock another path takes in the opposite order?
- **Coverage**: are there bind sites that scan→bind WITHOUT going through `_with_chain` (e.g. `ci/mod.rs:220` repo-checkout bind)? (I scoped to the dispatch+bind_self funnel per BIND-2; flagging ci-checkout as a possible separate follow-up if it doesn't share the scan.)
- **Lease-lifetime vs bind-mutex**: confirm the short-mutex design is correct (the durable lease is binding.json, read by the scan).

### DIFF
--- a/src/binding.rs
+++ b/src/binding.rs
@@ -18,6 +18,33 @@ fn index_key(home: &Path, agent: &str) -> String {
     format!("{}:{agent}", home.display())
 }
 
+/// #1882: lock-file path for the per-BRANCH lease flock. Keyed by a hash of the
+/// branch so it is path-safe (branches contain `/`) and collision-free; different
+/// branches never share a lock file.
+fn branch_lease_lock_path(home: &Path, branch: &str) -> PathBuf {
+    let key = crate::daemon::utils::sha256_hex(branch.as_bytes());
+    crate::paths::runtime_dir(home)
+        .join(".lease-locks")
+        .join(format!("{key}.lock"))
+}
+
+/// #1882: acquire the exclusive per-BRANCH lease flock so the cross-agent
+/// "a branch is held by at most one agent" check-then-bind
+/// (`scan_existing_branch_binding` → `bind_full`) is ATOMIC. Two DIFFERENT agents
+/// racing to lease the SAME branch serialize here: the first binds; the second
+/// blocks until the first's guard drops, then its rescan sees the first's binding
+/// and rejects — so neither double-binds. This is a SHORT-LIVED mutex around the
+/// bind operation, NOT a lease-lifetime lock: the persistent lease state is
+/// `binding.json` (which the scan reads), so `release_full` needs no lock cleanup.
+/// Per-branch keying means different branches never contend → normal single-agent
+/// binds are unaffected. Blocking `acquire_file_lock`; released when the guard drops.
+pub fn acquire_branch_lease_lock(
+    home: &Path,
+    branch: &str,
+) -> anyhow::Result<crate::store::FileFlockGuard> {
+    crate::store::acquire_file_lock(&branch_lease_lock_path(home, branch))
+}
+
 /// Write a binding for an agent (task assigned).
 ///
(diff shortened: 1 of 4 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED the fix commit 81774254 + all 5 required checks green (incl Windows). Note on the head-integrity flag: reviewer-2's git fetch origin pull/1883/head transiently showed 89fb2c54 (the #1880 merge commit), but the authoritative GitHub head (gh pr view --json headRefOid) is 81774254 — independently confirmed: ci/mod.rs at head contains acquire_branch_lease_lock (the repo-checkout lock), mergeable=MERGEABLE, CI green on headSha 81774254. The pull/N/head ref read was stale/transient; no rebuild needed. #1882 BIND-2: the branch-uniqueness scan→bind is now atomic via a shared branch-keyed lease flock (acquire_branch_lease_lock) held across scan_existing_branch_binding + bind_full. reviewer-2's REJECT (the ci/mod.rs repo-checkout was a 3rd bind path bypassing the lock) was fixed — repo-checkout now takes the same lock + an in-lock cross-agent scan reject; all 3 production bind paths (dispatch / bind_self / repo-checkout) serialize on the one branch lock, no 4th bypass, idempotent (#1494) preserved, end-to-end test proves exactly-one-binding. Merging --admin squash. Takes effect next restart.

PR #1884, opened 2026-06-08
telegram: propagate the separate follow-up text send error (#1878)

### LINKED ISSUE #1878: telegram media+long-text: the separate text send failure is swallowed (let _) and the call returns fake success
## Summary
Bug-hunt finding (fixup-dev, verified). When a telegram message has both an attachment and text longer than the 1024-char caption limit, the text is sent as a SEPARATE message after the media — but that send's error is discarded with let _, and the function returns Ok(MsgRef) (success) anyway. The attachment arrives, the body is silently lost, and the caller sees success (fail-direction).

## RCA (file:line)
adapter.rs:165-167: after send_media succeeds and yields a msg_id, `if needs_separate_text { let _ = block_on_value(send_with_topic(... &msg.text ...)); }` then `Ok(MsgRef { id: media_id })`. The separate-text send result is dropped by `let _`.

## Trigger
attachment + text > 1024 chars (caption cap) → text goes as a separate message → that send hits a transient failure (stale topic / 429 / network).

## Impact
MEDIUM (narrow trigger — agend messages with both an attachment AND long text are uncommon): the message body is silently lost while the caller is told it succeeded.

## Fix direction
Don't swallow the separate-text send error. On failure, surface it (return Err, or a partial-delivery indicator) rather than returning Ok with only the media id — at minimum log at error and propagate a failure so the caller can retry/handle, consistent with the rest of the send path's error handling.

suzuke:
Closes #1878 (bug-hunt C2).

## Bug

In `TelegramChannel::send`, when a media message needs a **separate follow-up text** (text > the 1024 caption limit, or an explicit caption), the text send's error was swallowed with `let _ =` while the fn still returned `Ok(MsgRef { media id })`. That's a **fake success**: the attachment is delivered but the message body is silently lost, and the caller (which has send-error handling — retry / surface) sees success.

## Fix

Propagate the follow-up text send error instead of swallowing it, with context naming the partial state (media sent, follow-up text failed) so logs / retry decisions are clear — consistent with the media send's own `?` above.

## Test (§3.9)

No mock-bot harness exists to behavior-test "media ok + text fails" — every `send` test hits the bot-uninitialised path before reaching the media arm. So `send_propagates_separate_text_error_1878` is a **source invariant** (same precedent as `handle_message_body_has_no_block_on`): it asserts the media+text arm propagates the follow-up send (`?` / `map_err`) and never swallows it with `let _ =`. Combined with `-D warnings` (an ignored `Result` is `must_use`), this pins the fix.

## Verification

- `cargo test --tests --features tray` (with `touch tests/` to force-rebuild the integration tests) → green except the unrelated `dispatch_branch_skips_metadata_stub_..._1834` (false-fails only under the local fleet `git` shim; passes with real `/usr/bin/git`; CI uses real git).

### DIFF
--- a/src/channel/telegram/adapter.rs
+++ b/src/channel/telegram/adapter.rs
@@ -163,8 +163,19 @@ impl crate::channel::Channel for TelegramChannel {
                     caption.as_deref(),
                 ))?;
                 if needs_separate_text(&msg.text, att) {
-                    let _ =
-                        block_on_value(send_with_topic(&bot, group_id, topic_id, &msg.text, None));
+                    // #1878: do NOT swallow the follow-up text send. Pre-fix this
+                    // discarded its error while still returning Ok(media id) — a
+                    // fake success that delivered the attachment but silently lost
+                    // the message body. Propagate so the caller sees the failure
+                    // (its existing send-error handling: retry / surface),
+                    // consistent with the media send above.
+                    block_on_value(send_with_topic(&bot, group_id, topic_id, &msg.text, None))
+                        .map_err(|e| {
+                            anyhow::anyhow!(
+                                "telegram: media sent (id {msg_id}) but the separate \
+                                 follow-up text failed: {e}"
+                            )
+                        })?;
                 }
                 Ok(crate::channel::MsgRef {
                     binding: build_telegram_msg_binding(topic_id),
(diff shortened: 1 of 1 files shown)


suzuke:
Merge audit (section 3.12 mirror) — lead-review VERIFIED + macos/ubuntu/Coverage/audit/LOC all green. The windows-latest red is a CONFIRMED pre-existing windows-env flake, NOT a regression from this change: the failing tests are 3 unrelated worktree_pool::tests (test_release_outcome_real_failure_emits_error_key / resolve_source_repo_parses_gitdir_pointer / resolve_source_repo_returns_none_for_regular_repo), all ABORTED with 0xc0000142 'DLL initialization routine failed (os error 1114)' — a windows OS-level subprocess process-init transient (git worktree subprocesses under nextest parallelism), plus the job also hit the nextest-install bash-startup flake (partner-runner-images#169). This PR only touches src/channel/telegram/adapter.rs; its own test (send_propagates_separate_text_error_1878) PASSES, and the same code is green on macos + ubuntu + Coverage. #1877/#1880 (also touched this session) passed windows. #1878: the telegram media+long-text send no longer swallows the separate follow-up text send error (let _ → map_err + ?) — the body is no longer silently lost on a transient failure; the caller now observes it. Merging --admin squash past the documented windows env flake. (The worktree_pool windows DLL-init abort flake is pre-existing — worth a separate windows-flake hardening, noted.)

Issue #1888, opened 2026-06-08
ci-handoff re-nudge (#1860) is coupled to inbox read-state — any drain marks the [ci-ready] read and blinds the watchdog


suzuke:
Spike RCA (fixup-dev-2, code-confident). The #1860 Fix-A re-nudge auto-delivers the next_after_ci hand-off intermittently — some PRs auto-deliver to the reviewer, others need a manual dispatch.

## RCA (file:line)
- handoff_timeout_watchdog.rs scan_and_emit_with (:80) re-nudges by scanning unread_of_kind(home, target, 'ci-ready-for-action') (:97); gate (:116) age_min >= RENUDGE_AFTER_MINS(2) && !agent_is_busy(target) + last_renudged >= 2min → renudge_actionable_unread.
- unread_of_kind (inbox/storage.rs) returns ONLY messages with read_at.is_none(). Once a handoff is marked-read, the watchdog never sees it → never re-nudges.
- inbox::drain (storage.rs:304) unconditionally sets read_at on every drained unread, regardless of kind (the query/task exemption at :46 is for re-surface elsewhere; drain itself does not exempt). In production only the reviewer's own MCP  call drains its inbox.

## Mechanism / intermittency
CI green → poller enqueues an unread [ci-ready] + a wake (deferred into the notification_queue if the reviewer is busy → TUI-flush only). The reviewer's ANY drain inbox (routine pending check) marks the [ci-ready] read → the watchdog is blind from then on. re-nudge only helps while the handoff stays unread, AND it waits 2min — a reviewer checking inbox within 2min of CI-green (common) reads it before the first re-nudge fires. Auto-delivered PRs = the reviewer acted at the first wake or noticed during the drain (before read closed the window); stranded PRs = the reviewer drained (marking it read) without acting (buried in a batch) → watchdog scans empty → never re-nudges → manual dispatch needed.

## Plan
1. **Instrument-only first** (zero-behavior tracing) — deploy and confirm in production on the next stranded PR: (a) at the re-nudge decision point (watchdog:116) log per (target, correlation): unread_found / age_min / agent_is_busy / renudge_due / fired-or-gated-why; (b) at the [ci-ready] read transition (drain storage.rs:304) log agent / correlation / age_at_read.
[...]

PR #1889, opened 2026-06-09
chore(#1888): instrument-only tracing for ci-handoff re-nudge (zero behavior)

suzuke:
Instrument-only follow-up to the #1860 ci-handoff re-nudge spike (lead-approved — **no behavior change**; confirm the RCA in production before the real fix).

## RCA recap
The re-nudge scans only **unread** `[ci-ready-for-action]` (`unread_of_kind` filters `read_at.is_none()`), but the inbox `drain` marks the handoff read on the first drain regardless of kind — after which the watchdog can never find it. Intermittency = whether the reviewer acted before the drain that read it.

## What this adds (two `info!` traces, zero behavior)
- **`tag=#1888-renudge-decision`** at the `handoff_timeout_watchdog` re-nudge gate: per (target, correlation) — `unread_found` / `age_min` / `agent_is_busy` / `renudge_due` / `age_ok` / `will_fire`. Bounded: ≤ once per **unread** ci-handoff per tick.
- **`tag=#1888-ciready-read`** at the inbox `drain` read-marking: when a `ci-ready-for-action` message transitions to read — `agent` / `correlation` / `age_at_read_secs`. Rare: only that kind, only on the read transition.

Together they show, for a stranded PR, whether the `[ci-ready]` was **read before it was acted on**, with the watchdog then finding nothing to re-nudge.

## Why `info!` and no custom `target:`
The production filter is `agend_terminal=info` (logging.rs:73). `debug!` would be **invisible in production** (and the test caught that a custom `target:"ci_handoff"` wouldn't match `agend_terminal=*` either — both would silently filter the instrument out). So: `info!`, default module-path target.

## Zero behavior
Both are read-only tracing statements; the re-nudge gate and the drain read-marking are **byte-identical** (the trace locals don't feed control flow).

## §3.9
- `renudge_decision_is_instrumented_1888` + `ciready_read_on_drain_is_instrumented_1888` (`tracing_test::traced_test`): assert the traces fire AND the surrounding behavior is unchanged (the unread idle handoff still re-nudges; the drain still returns the handoff marked read). All existing `handoff_timeout_watchdog` tests still pass.

## Follow-up (after production confirms)
The real fix decouples re-nudge from the inbox read-state — a correlation-tracked sidecar + clear-on-resolution (same shape as #1866 dispatch_idle).

### DIFF
--- a/src/daemon/handoff_timeout_watchdog.rs
+++ b/src/daemon/handoff_timeout_watchdog.rs
@@ -113,6 +113,34 @@ pub(crate) fn scan_and_emit_with<F>(
             // target reads the handoff (drops from the unread scan → reaped). This
             // also covers the orchestrator-as-`next_after_ci` case below, where
             // there is no lead to escalate to.
+            // #1888 instrument-only (zero-behavior): record WHY this UNREAD handoff
+            // did / didn't get re-nudged this tick, so production can confirm the
+            // #1860 read-state-coupling RCA. A handoff that was already marked read
+            // never enters this loop at all (`unread_of_kind` filters it) — that
+            // case is captured by the `#1888-ciready-read` trace at the drain site.
+            // Read-only: these locals don't feed the byte-identical gate below.
+            {
+                let busy = crate::snapshot::agent_is_busy(home, target);
+                let renudge_due = last_renudged.get(&key).is_none_or(|prev| {
+                    now.signed_duration_since(*prev).num_minutes() >= RENUDGE_INTERVAL_MINS
+                });
+                // info!-level so it lands in the production daemon.log (default
+                // filter is `agend_terminal=info`); bounded — at most once per
+                // UNREAD ci-handoff per tick, and a handoff stops being scanned the
+                // moment it's read.
+                tracing::info!(
+                    tag = "#1888-renudge-decision",
+                    agent = %target,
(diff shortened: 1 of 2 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED (truly zero behavior: +102/-0 pure addition, no original line touched; the re-nudge gate re-evaluates independently of the trace locals, and the drain trace only reads msg fields after read_at is set) + all 5 required checks green. #1888 instrument-only: two info!-level traces — tag=#1888-renudge-decision (handoff_timeout_watchdog re-nudge gate: unread_found/age_min/agent_is_busy/renudge_due/will_fire) and tag=#1888-ciready-read (inbox drain: agent/correlation/age_at_read when a ci-ready-for-action is marked read). Bounded volume, info! so it's visible under the production agend_terminal=info filter (the §3.9 logs_contain test caught that the original debug! would be filtered out). After this deploys, the next stranded next_after_ci hand-off will be diagnosable from the daemon log (was the [ci-ready] drained-read before the reviewer acted → watchdog then scans empty), then the true fix (decouple re-nudge from read-state via a correlation sidecar with clear-on-resolution, like #1866) follows. Merging --admin squash. Takes effect next restart.

PR #1890, opened 2026-06-09
fix(robustness): leaf-lock invariant test + dismiss in-flight scopeguard (#1886 follow-up)

suzuke:
Two small defensive follow-ups to the merged #1886 locked-RMW fix.

## (a) Leaf-lock invariant test — fortifies reviewer-2's structural argument
The metadata/topics RMW (`save_metadata`, `register_topic`) run their mutate closure inside `with_json_state_or_create`'s file-flock. That flock must be a **leaf** (no nested file-flock) or it could deadlock against another lock-ordering path. reviewer-2 confirmed this structurally during #1887 review; this turns the argument into a test:
- `with_json_state_closure_is_leaf_lock_1886` — the closure observes flock depth **exactly 1** (the helper's own lock, nothing nested).
- `with_json_state_nested_flock_reaches_depth_2_1886` — proves the assertion has teeth: a deliberately nested flock reaches depth 2, so a future nested-lock regression fails the depth-1 test.

Adds a test-only `sync_audit::flock_depth()` accessor (`#[cfg(test)]`, used from the bin-side `store` test tree).

> Note: the `flock_depth_invariant.rs` I cited during #1887 is the **#1629** "sole flock chokepoint / no raw fs4 flock" guard — it confirms no raw flock was added, but does **not** prove the leaf-lock property. This PR adds the test that actually does.

## (b) Dismiss in-flight scopeguard — fixes the LOW latent fragility from the spawn/PTY bug-hunt
`agent/dismiss.rs` removed the agent from `DISMISS_IN_FLIGHT` as a **trailing statement** of the dismiss thread. A panic / early-return before it left a stale entry that silently no-op'd every future dismiss for that agent until daemon restart. Hoist `DISMISS_IN_FLIGHT` to module scope and free the slot via a RAII `InFlightGuard` armed at thread entry, so any exit (incl. panic) clears it. The spawn-failure path's manual removal stays (no thread spawned → no guard).
- `inflight_guard_clears_entry_on_panic_1886` (§3.9) — inject a panic after the guard is armed; assert the slot is still cleared.

## Verification
fmt + `clippy --features tray --tests -D warnings` clean; 7 new `1886` tests pass; `flock_depth_invariant` (#1629), `self_ipc_guard_1492`, `spawn_rationale_audit` (§10.5) all green.

### DIFF
--- a/src/agent/dismiss.rs
+++ b/src/agent/dismiss.rs
@@ -21,6 +21,26 @@ static DISMISS_REGEX_CACHE: std::sync::LazyLock<
     parking_lot::Mutex<std::collections::HashMap<String, Option<std::sync::Arc<regex::Regex>>>>,
 > = std::sync::LazyLock::new(|| parking_lot::Mutex::new(std::collections::HashMap::new()));
 
+/// H2: agents with a dismiss thread currently in flight — gates rapid dialog
+/// re-detection to one thread per agent. Hoisted to module scope (#1886
+/// follow-up) so the RAII [`InFlightGuard`] can clear it on drop.
+static DISMISS_IN_FLIGHT: std::sync::LazyLock<
+    parking_lot::Mutex<std::collections::HashSet<String>>,
+> = std::sync::LazyLock::new(|| parking_lot::Mutex::new(std::collections::HashSet::new()));
+
+/// #1886 follow-up: RAII guard that removes an agent from `DISMISS_IN_FLIGHT` on
+/// drop — including on a panic or early-return of the dismiss thread. Previously
+/// the removal was a trailing statement, so a panic before it left a stale entry
+/// that silently no-op'd every future dismiss for that agent until daemon
+/// restart. Arm it at thread entry; the in-flight slot is freed on any exit.
+struct InFlightGuard(String);
+
+impl Drop for InFlightGuard {
+    fn drop(&mut self) {
+        DISMISS_IN_FLIGHT.lock().remove(&self.0);
+    }
+}
+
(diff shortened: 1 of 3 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED + lead-review VERIFIED; ubuntu + windows Check (required) + audit + LOC all green. The macos-latest red is a CONFIRMED pre-existing recurring flake, NOT from this change: the macos job ran all tests through worktree_pool::tests::* OK then went silent for ~25min → 30min step timeout, terminating with an orphan agend_terminal subprocess (pid 24239) — a worktree_pool/spawn test that didn't reap its child subprocess on macos (the macos sibling of the #1884 windows worktree_pool DLL-init flake family). It failed on two consecutive attempts. This PR's three new tests use only flock + std::thread and spawn NO subprocess; dev ran all three on macOS (Darwin 24.5.0) locally and all pass in <0.01s, and ubuntu+windows pass the same code. #1886 follow-up: (a) an explicit leaf-lock invariant test (with_json_state's mutate closure runs at flock depth==1, with a nested-flock test proving the test has teeth) hardening the structural deadlock-safety argument; (b) the dismiss in-flight removal is now an RAII InFlightGuard armed at thread entry so a panic/early-return still frees the slot (was a trailing statement → a stale entry could silently no-op dismiss until restart). Merging --admin squash past the confirmed macos worktree_pool flake. (The cross-platform worktree_pool test subprocess flake — windows DLL-init + macos orphan-timeout — is worth a separate test-hardening, noted.)

Issue #1891, opened 2026-06-09
register_external defaults a missing pid to 0 → is_pid_alive(0) is always true → an unreclaimable zombie external agent


suzuke:
Bug-hunt finding (fixup-dev, verified). register_external silently defaults a missing/null/non-integer pid to 0, and is_pid_alive(0) is always true on Unix → the external entry is never reclaimed by the liveness sweep, becoming an unreclaimable zombie.

## RCA (file:line)
- src/api/handlers/external.rs:26: `let pid = params["pid"].as_u64().unwrap_or(0) as u32;` — pid is a required liveness-tracking field, but a missing/null/non-integer value silently becomes 0 (wrong fail-direction).
- per_tick external_liveness sweep (external_liveness.rs:23): `ext.retain(|_, h| is_pid_alive(h.pid))`.
- process.rs:10 is_pid_alive(0) = `libc::kill(0, 0) == 0` — on Unix, kill with pid 0 targets the CALLER's entire process group, so it returns 0 as long as the daemon itself is alive (always true). → a pid=0 external entry is ALWAYS seen as alive → never deregistered.

## Trigger
A register_external RPC missing / null / non-integer `pid` (an external caller, an mcp-bridge bug, or a hand-crafted RPC — the daemon cannot assume external callers always send a valid pid, so this is exactly the input that must be validated).

## Impact (MEDIUM)
The external entry becomes an unreclaimable zombie: it occupies the name (a later re-register is blocked by 'already exists (external)'), query/list show it alive forever, and the liveness sweep is permanently blind to it.

## Smoking gun
process.rs:33 and :65 (e.g. the kill-tree path) already guard `if pid == 0`; only is_pid_alive (:10) and register_external (unwrap_or(0)) don't.

## Fix
Both: (a) input validation — handle_register_external rejects a missing / 0 / non-integer pid with an error (a required field must be present and valid); (b) defense-in-depth — is_pid_alive(0) → false (aligning with the pid==0 guards elsewhere; pid 0 should never count as a live tracked agent). §3.9: a register_external with no pid (or pid 0) is rejected; is_pid_alive(0) returns false; a valid pid still tracks + sweeps normally.

suzuke:
Fixed by #1892. register_external now rejects a missing/null/non-integer/negative/zero/overflow pid before any registry insert (no silent default to 0), and is_pid_alive(0) returns false as defense-in-depth — closing the unreclaimable pid=0 zombie external agent.

PR #1892, opened 2026-06-09
fix(api): reject missing/zero pid in register_external + guard is_pid_alive(0) (#1891)

suzuke:
C1 from the API/schedule/GC bug-hunt (the one verified bug; the rest of that layer was sound).

## The bug
`handle_register_external` did `params["pid"].as_u64().unwrap_or(0)`, so a missing / null / non-integer `pid` silently defaulted to **0**. On Unix the `external_liveness` per-tick sweep keeps entries where `is_pid_alive(pid)` is true, and `is_pid_alive(0)` is `kill(0, 0)` — which targets the **caller's whole process group** and always succeeds. A 0 pid therefore reads as **permanently alive** → an unreapable zombie external registration that squats the name forever (re-register rejected as "already exists (external)", liveness sweep blind to it).

## Fix
- **(a) Input validation** — `register_external` now rejects a missing/null/non-integer or 0 `pid` with a clear error (it's a required field for liveness tracking), instead of defaulting to 0.
- **(b) Defense-in-depth** — `is_pid_alive(0)` now returns `false`, mirroring the existing `pid == 0` guards already in `process.rs` (kill_process_tree / process-group ops). pid 0 is never a real tracked process.

## Verification (§3.9)
- `register_external_missing_pid_rejected_1891` / `_zero_pid_rejected_1891` — rejected with error, **no entry inserted**.
- `register_external_valid_pid_registers_1891` — a real pid registers and is tracked.
- `is_pid_alive_zero_is_false_1891` — pid 0 is dead; self pid is alive (guard didn't over-reach).
- Existing `external_liveness` sweep tests (dead `999999999` / live self-pid) stay green.

fmt + `clippy --features tray --tests -D warnings` clean. (external.rs is `src/api/handlers/` — not covered by the `src/mcp/handlers` 750-cap invariant.)

### DIFF
--- a/src/api/handlers/external.rs
+++ b/src/api/handlers/external.rs
@@ -23,7 +23,21 @@ pub(crate) fn handle_register_external(params: &Value, ctx: &HandlerCtx) -> Valu
         return json!({"ok": false, "error": format!("agent '{name}' already exists (external)")});
     }
     let backend = params["backend"].as_str().unwrap_or("unknown");
-    let pid = params["pid"].as_u64().unwrap_or(0) as u32;
+    // #1891: `pid` is required for liveness tracking — a missing / null /
+    // non-integer / 0 pid must be REJECTED, not defaulted to 0. On Unix
+    // `is_pid_alive(0)` is `kill(0, 0)` which targets the caller's whole process
+    // group → always reports "alive", so a 0 pid would make the external_liveness
+    // sweep treat the entry as permanently live → an unreapable zombie
+    // registration that squats the name forever.
+    let pid = match params["pid"].as_u64() {
+        Some(p) if p > 0 && p <= u64::from(u32::MAX) => p as u32,
+        _ => {
+            return json!({
+                "ok": false,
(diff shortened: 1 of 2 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED + all 5 required checks green (Coverage passed too this run). #1891 C1: (a) handle_register_external now validates pid via as_u64() + (p>0 && p<=u32::MAX) — missing/null/non-integer/negative/zero/overflow are all rejected with a clear error BEFORE any registry insert (no more silent pid=0 zombie entry); (b) is_pid_alive(0) → false as defense-in-depth (aligning with the pid==0 guards elsewhere in process.rs). reviewer-2 verified the negative case returns None from as_u64 before the cast (no u32 wrap), the reject is before insert (test asserts the registry stays empty), and enumerated all is_pid_alive callers — every one checks a tracked pid for keep-vs-reclaim, so 0=dead is correct/beneficial for all (no caller passes 0 expecting alive). Closes the unreclaimable-zombie-external-agent bug. Merging --admin squash. Takes effect next restart.

Issue #1894, opened 2026-06-09
usage_limit re-notify (#1861) still fires across restarts when unlock_at is null — the fallback cooldown is too short for an ongoing limit


suzuke:
## Problem
#1861 was meant to stop the usage_limit operator notification re-firing on every restart, but it still fires for the codex reviewer (the operator has been annoyed by this repeatedly). Confirmed root cause: the persisted dedup record has `unlock_at: null`, so #1861 falls back to a SHORT timestamp cooldown instead of the unlock_at-based suppression — and restarts hours apart exceed that cooldown, so it re-notifies.

## Evidence
usage_limit_notify.json: `{"fixup-reviewer": {"unlock_at": null, "notified_at": "..."}}`. The member_state_change at the detection tick shows last_pane_excerpt = poll-reminders (NOT the 'try again at Jun 11th, 2026 9:04 AM' usage-limit screen), so parse_unlock_at had no unlock time to extract → null. So unlock_at is null not because codex lacks a reset time, but because the unlock-time text wasn't in the pane tail at that particular detection tick.

## Why it re-fires
With unlock_at = null, the suppression is a short persisted-timestamp cooldown (now - notified_at < cooldown). Restarts are typically hours apart, exceeding the cooldown → re-notify each restart. The member is CONTINUOUSLY usage_limited the whole time (the same ongoing limit), so re-notifying is pure noise.

## Fix direction
The dedup should suppress for the duration of an ONGOING usage_limit episode, not a short cooldown. Options: (a) when unlock_at is null, use a much longer fallback window (e.g. 12-24h) so restarts within a single usage-limit episode stay suppressed (re-notify at most ~once/day if the limit genuinely persists for days); (b) better — key the dedup on '(member) is still usage_limited + a record exists' → suppress until the member LEAVES usage_limit (recovers), regardless of unlock_at, with a long safety cap; (c) make parse_unlock_at more robust / parse from a wider window so codex's 'try again at <date>' is captured more often (reduces the null case). Likely (a)+(c): a long null-unlock_at fallback as the floor, plus better unlock_at extraction. Keep the existing unlock_at-based path for the parseable case. §3.9: a member re-detected as usage_limit across a simulated restart with a null-unlock_at record within the long fallback window → NOT re-notified; after the window (or on recovery) → re-notify.

suzuke:
[...]

PR #1895, opened 2026-06-09
fix(#1894): null-unlock_at usage_limit re-notify uses a 24h window, not 60s

suzuke:
Follow-up to #1864/#1861 — the residual the **operator kept hitting** (re-paged by codex usage-limit notifies on every restart).

## Root cause
When `parse_unlock_at` finds no "try again at <time>" in the pane at detection time (it was showing poll-reminders), the persisted dedup record stores `unlock_at: null`, and `usage_limit_notify_suppressed` falls back to a **timestamp cooldown**. That fallback used the **60s `NOTIFY_COOLDOWN`** — which a daemon restart hours later trivially exceeds → the operator is re-paged for the **same ongoing** usage limit on every boot. The member is continuously usage-limited the whole time = pure noise.

## Fix (member-notify path only; recovery / continue-inject untouched)
The null-`unlock_at` fallback now suppresses for a long `NULL_UNLOCK_NOTIFY_WINDOW` (**24h**) instead of 60s. A real usage-limit episode lasts hours-to-days, so restarts within the same episode stay silent; a multi-day limit re-notifies at most ~once/day instead of once/restart.

- **Parseable `unlock_at` suppression (#1864) unchanged** — the `Some(u)` deadline branches are untouched.
- **FAIL-OPEN preserved** — a missing/corrupt record still returns "not suppressed" → notify (a real new limit is never silently swallowed).

(Considered the lead's option (b) "suppress until recover": there's no clear-on-recover hook today, and adding one risks flicker false-re-notifies (UsageLimit→thinking→UsageLimit). The 24h window is option (a) — equivalent to (b)'s core without that risk — and fully kills the per-restart noise. A clear-on-recover refinement can be a later follow-up.)

## §3.9
`usage_limit_null_unlock_long_window_1894`:
- null + `+30s` → suppress; **`+5h` restart (same episode) → still suppress** (was re-notify at 60s — regression-proof: revert to `NOTIFY_COOLDOWN` and it flips);
- `+25h` (new episode, past window) → re-notify;
- missing record → FAIL-OPEN re-notify.

The #1864 tests (`usage_limit_notify_suppressed_logic_1861`, `usage_limit_notify_not_refired_across_restart_1861`) still pass.

### DIFF
--- a/src/daemon/supervisor.rs
+++ b/src/daemon/supervisor.rs
@@ -25,6 +25,16 @@ const TICK: Duration = Duration::from_secs(10);
 const TAIL_LINES: usize = 40;
 /// Debounce cooldown for member-state-change notify (Sprint 43).
 const NOTIFY_COOLDOWN: Duration = Duration::from_secs(60);
+/// #1894: null-unlock_at usage_limit re-notify window. When the pane had no
+/// parseable "try again at <time>" at detection (e.g. it was showing
+/// poll-reminders), `parse_unlock_at` records `unlock_at: null` and the dedup
+/// falls back to a TIMESTAMP cooldown. #1861/#1864 used the 60s `NOTIFY_COOLDOWN`
+/// here, which restarts hours apart trivially exceed → the operator was re-paged
+/// on every boot for the SAME ongoing limit. A real usage-limit episode lasts
+/// hours-to-days, so suppress for a long window: a multi-day limit re-notifies at
+/// most ~once/day instead of once/restart. Parseable-unlock_at suppression
+/// (#1864) is unchanged, and a missing/corrupt record still FAILS OPEN (notify).
+const NULL_UNLOCK_NOTIFY_WINDOW: Duration = Duration::from_secs(24 * 60 * 60);
 
 /// #1552: minimum continuous time in a runtime prompt state before escalating
 /// to AwaitingOperator (stability gate — a transient streaming flicker of the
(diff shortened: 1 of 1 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED + all 5 required checks green. #1894 (#1861 follow-up): the usage_limit member-notify dedup's null-unlock_at fallback was a 60s cooldown — so when parse_unlock_at returns null (the unlock-time text wasn't in the pane tail at that detection tick, e.g. poll-reminders shown instead of the 'try again at Jun 11' screen), restarts hours apart exceeded the cooldown and re-notified the operator every restart, even though the member was continuously usage_limited (the same ongoing limit = pure noise). Now the null-unlock_at fallback is a 24h NULL_UNLOCK_NOTIFY_WINDOW: restarts within one usage-limit episode stay suppressed; a limit persisting for days re-notifies at most ~once/day. The parseable unlock_at path (#1864) is byte-identical (only the None arm changed), FAIL-OPEN is preserved (missing/corrupt record → notify), and the suppress-until-recover option (b) was reasonably deferred (it needs a flicker-safe clear-on-recover hook that doesn't exist yet — naive clear would false-re-notify on the UsageLimit→thinking→UsageLimit streaming flicker). section-3.9 covers +5h-same-episode-suppress / +25h-new-episode-renotify / missing-FAIL-OPEN, regression-proof. Merging --admin squash. Takes effect next restart (stops the per-restart codex re-notify).

Issue #1893, opened 2026-06-09
CI flaky test cluster (worktree_pool subprocess orphan + concurrent-checkout race) keeps recurring — fix it systematically instead of admin-merging past it


suzuke:
## Problem
A cluster of tests flakes on windows + macos CI, recurring for a long time. We keep treating each instance as a one-off 'pre-existing flake' and admin-merging past it / rerunning, but the ROOT CAUSE is never fixed — so it keeps taxing every PR (blocks merges, forces admin-merges, erodes CI trust, wastes review time). This issue is to FIX it systematically.

## Known flaky members (from recent incidents)
- worktree_pool::tests::* — spawn real subprocesses (git worktree add + the agend_terminal binary). On macos: a child agend_terminal subprocess (e.g. pid 24239) is left ORPHANED → the test process hangs ~25min waiting for it → 30min step timeout. On windows (#1884): worktree subprocess ABORTS with 0xc0000142 'DLL initialization routine failed' under nextest parallelism.
- mcp::handlers::ci::tests::checkout_bind_true_concurrent_branch_create_race_idempotent (and likely the other *_concurrent_*_race / *_1882 binding tests) — pass 6/6 in isolation on macos but flake under full-suite PARALLEL load (name has concurrent..._race).
- (Enumerate the rest during the spike — grep for tests that spawn process::Command / current_exe / git worktree, and the concurrency-race tests.)

## Likely root causes
1. Subprocess orphaning: a test spawns a child (git worktree / agend_terminal) and doesn't reliably reap/kill it on test exit → orphan → the runner's step timeout.
2. Parallel interference: concurrency-race tests + subprocess-spawning tests run concurrently under nextest's default parallelism and interfere (shared fs/ports/process state).

## Existing infra (underused)
.config/nextest.toml exists; some tests use #[serial] (runtime_config/operator_mode/capture) — but the worktree_pool/concurrent-checkout cluster is NOT covered.

## Fix direction (spike to confirm + decide)
- Reap subprocesses deterministically: a test guard/teardown (RAII child-kill, or a process-group kill) so a spawned child is always reaped even on panic/timeout — OR mock the spawn where the test doesn't actually need a real subprocess.
- Serialize the flaky cluster: a nextest test-group (.config/nextest.toml [[profile.default.overrides]] / [test-groups]) that runs the subprocess + concurrency-race tests serially (max-threads=1 for that group), instead of letting them race the whole suite.
[...]

PR #1896, opened 2026-06-09
ci(#1893): nextest on all platforms + serialize the heavy git test cluster (FIX A+B)

suzuke:
Root-causes the recurring CI flaky cluster (spike #1893) at the **CI-config layer** — no production git code touched (that's the follow-up FIX C). The cluster's symptoms (windows `0xc0000142` DLL-init, macos 30-min anonymous orphan-timeout, `checkout_bind_true_concurrent_branch_create_race_idempotent` fail-under-load) are one problem.

## FIX A — uniform per-test timeout on ALL platforms
Previously only windows ran `cargo nextest --profile ci` (120s per-test terminate-after); mac/ubuntu ran plain `cargo test`, so a hung real-git test there ran to the **30-min GitHub STEP timeout, anonymous** (the #1890 macos hang). Now all 3 OSes install nextest and run the `ci` profile → a wedged test is **KILLED and NAMED in <2min** instead of an anonymous 30-min step timeout.

## FIX B — serialize the heavy git/worktree/subprocess test cluster
A nextest `git-subprocess` test-group (`max-threads = 1`). These tests spawn real `git` (worktree add / branch / fetch) and share filesystem + `.git/index.lock` state; at full parallelism they race (index.lock contention, fd pressure) and flake — e.g. the concurrent-checkout test passes in isolation but failed under full-suite load. Serializing the group gives each isolation-equivalent execution while the rest of the suite stays parallel. Filter: `worktree_pool` / `mcp::handlers::{ci,dispatch_hook,force_release}` / `daemon::{retention::worktrees,pr_state}`; mirrored under `default` + `ci` profiles.

`*_stress_50_iter_*` (50 real git branch-creates, ~60s) get a 3×120s slow-timeout override so the suite-wide 120s terminate-after can't false-kill a genuinely-slow run on a slow runner (mac/ubuntu had no per-test timeout before, so this preserves their pass behaviour).

## Verification (local, system git)
- `cargo nextest run --features tray --profile ci` → **3779/3779 passed in 243s** (faster than the old ~8min `cargo test`; no wall-time regression). The previously-flaky concurrent-checkout test passes deterministically under serialization.
- `cargo nextest show-config test-groups --profile ci` confirms the filter assigns the heavy modules to `git-subprocess`.
- YAML + TOML validated.
- **This PR's own CI is the first all-platform-nextest proof.**

Follow-up (separate PR, FIX C): bound `git_bypass`/`run_git_idempotent` with a timeout + process-group kill — hardens the production daemon git path, the deeper root cause.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -28,6 +28,17 @@ final-status-level = "slow"
 # runs the GROUP one-at-a-time (isolation-equivalent) while the rest of the suite
 # stays parallel. The override is mirrored under both `default` (local runs) and
 # `ci` (CI runs) profiles.
+#
+# The filter has two parts (reviewer-2 #1896: cover the FULL real-git set, but
+# don't drag pure-logic tests into max-threads=1):
+#  1. anchored MODULE prefixes — modules that are predominantly real-git/worktree
+#     /subprocess: worktree_pool, admin (worktree + cleanup_zombies process spawn),
+#     mcp::handlers::{ci,dispatch_hook,force_release,worktree,p0b_tests},
+#     daemon::{retention::worktrees,pr_state,auto_release,conflict_notify,
+#     per_tick::tmp_review_gc}, bootstrap::{agent_resolve,canonical_hygiene}.
+#  2. NAMED tests in otherwise-pure modules (messaging 2/47, tasks::tests 1/78,
+#     task_sweep, mcp::handlers::tests checkout_repo_*) — targeted by name so the
+#     ~200 pure-logic tests around them stay parallel.
 [test-groups]
 git-subprocess = { max-threads = 1 }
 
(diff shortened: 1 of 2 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED (@2e56e670, after adopting reviewer-2's filter-completeness finding) + the self-proving 3-platform nextest CI is GREEN (macos/ubuntu/windows Check all pass — macos, the platform that previously 30-min-hung, now passes cleanly with no orphan timeout). #1893 FIX A+B (CI-config only, no prod git code): (A) all 3 OS now run cargo nextest run --features tray --profile ci (was windows-only) so the existing [profile.ci] slow-timeout (terminate-after) applies on every platform — a hung test becomes a <2min NAMED failure instead of a 30min anonymous step-timeout + orphaned subprocess; (B) a git-subprocess nextest test-group (max-threads=1) serializes the FULL real-git test set (module-anchored: worktree_pool, admin, mcp::handlers::{ci,dispatch_hook,force_release,worktree,p0b_tests}, daemon::{retention::worktrees,pr_state,auto_release,conflict_notify,per_tick::tmp_review_gc}, bootstrap::{agent_resolve,canonical_hygiene}; plus named git tests in otherwise-pure modules) so they no longer race the suite over .git/index.lock — without over-serializing pure-logic tests. Wall-time is comparable to before. This stops the recurring CI flaky tax the operator called out (worktree_pool subprocess orphan on macos + concurrent-checkout race + windows DLL-init). Merging --admin squash. (FIX C — bound the prod git_bypass/run_git_idempotent timeouts + process-group kill, which also protects the production daemon from a hung git — is a separate follow-up.)

Issue #1897, opened 2026-06-09
Harden prod git subprocesses: bound git_bypass/run_git_idempotent timeouts + process-group kill (#1893 FIX C)


suzuke:
Follow-up to #1893 (the CI flake fix #1896 was FIX A+B, CI-config only). FIX C is the deeper root fix that also protects the PRODUCTION daemon, not just tests.

## Root cause (from the #1893 spike)
git_bypass (git_helpers.rs:14) and run_git_idempotent (dispatch_hook:1047) call .output() with NO timeout. 9 unbounded git_bypass sites vs only 3 git_bypass_timeout — the unbounded ones include git branch (dispatch_hook:781/815) and git worktree add (ci/mod.rs:238). If a git operation hangs on a contended .git/index.lock (or a stuck network helper), .output() never returns → the caller hangs indefinitely. In tests this manifested as the CI flake (now mitigated by #1896's per-test timeout), but in PRODUCTION a hung git would hang the daemon's git path with no bound.

## Fix
1. Route git_bypass + run_git_idempotent through a bounded timeout (like the existing git_bypass_timeout) so a contended/stuck git fails fast instead of hanging — protecting the production daemon, not just the tests. Pick a sensible default timeout per call site (worktree add / branch / fetch).
2. git_bypass_timeout's kill path (git_helpers.rs:43) currently only child.kill() — make it a process-group kill (spawn with process_group(0)/CREATE_NEW_PROCESS_GROUP, kill via the existing process::kill_process_tree) so git's sub-helpers (fetch → remote-https/pack) are also reaped (RC4: process-group leak).

## Care
Larger blast radius than the CI-config fix — it touches the production git path. Verify each site's timeout is generous enough for legitimate slow operations (large fetch) but bounded; verify the process-group kill doesn't kill anything it shouldn't. Do this as a careful standalone PR with adversarial review.

suzuke:
Fixed by #1898 (FIX C). Production git_bypass + run_git_idempotent now bound via git_bypass_timeout (60s local / 300s network) with a process-group kill that reviewer-2 verified can never hit the daemon — so a hung git fails fast instead of freezing the daemon's git path. (Bonus: bisection proved FIX C innocent of a macos CI failure and fixed the underlying pre-existing racy checkout_bind test assertion.) FIX C2 — the ~15 additional raw Command::new('git') prod sites found in the audit — is a separate follow-up.

PR #1898, opened 2026-06-09
fix(git): bound daemon git subprocesses + process-group kill (#1897 FIX C)

suzuke:
The deeper root-cause hardening from the #1893 spike: production daemon git calls could hang forever on a contended `.git/index.lock`, wedging the daemon (RC1). Scope = the central git helpers (the lead's enumeration). **No CI-config; this is the prod git path.**

## 1. Bounded timeout (clear `Err`, never hang)
- **`git_bypass`** (git_helpers.rs) was a bare unbounded `.output()` → now routes through `git_bypass_timeout` with **LOCAL_GIT_TIMEOUT (60s)**. Auto-bounds all 8 `git_bypass` call sites (rev-parse / branch / worktree add / status / remote get-url / update-ref). Fast path byte-identical.
- **`run_git_idempotent`** (dispatch_hook) per-attempt bounded with LOCAL_GIT_TIMEOUT; a timeout `?`-propagates (no retry of a wedge), the retry-on-transient-exit loop preserved.
- **`default_branch` / `primary_remote`** (git_helpers) bounded too (graceful "main"/"origin" fallback).
- The 3 **`fetch`** sites widened 60s → **NETWORK_GIT_TIMEOUT (300s)** — strictly wider, so it can only ADD tolerance for a large fetch, never newly false-kill (the "don't kill legit slow ops" CARE point).

## 2. Process-group kill (git's sub-helpers can't orphan)
- `git_bypass_timeout` spawns git in its **OWN process group** (`process_group(0)` / `CREATE_NEW_PROCESS_GROUP`) and on timeout kills the whole group via `process::kill_process_tree`, not just the immediate child. **Isolation is MANDATORY**: `kill_process_tree` resolves the pgid via `getpgid`, so without it the kill would resolve to the **daemon's** group and kill the daemon. (Group isolation only — NOT detached; pipes preserved.)
- Extracted `spawn_group_bounded` so the timeout/kill mechanism is unit-testable.

Existing `git_bypass_timeout` callers' observable contract is unchanged (`Ok(Output)` / `Err(TimedOut)`).

## Tests (§3.9, unix)
- `spawn_group_bounded_times_out_within_slack_1897` — a `sleep 30` stub with a 1s bound returns `Err(TimedOut)` within timeout+slack, NOT a 30s hang.
- `spawn_group_bounded_fast_command_unaffected_1897` — fast command returns its `Output` normally.
- `spawn_group_bounded_kills_whole_process_group_1897` — a backgrounded grandchild `sleep` (same group) is reaped by the process-group kill.

Local: fmt + `clippy --features tray --tests -D warnings` clean; the affected modules pass under `nextest --profile ci`.

## Scope note → FIX C2 follow-up
The spike's "9 git_bypass sites" undercounted. An audit found **~15 additional independent raw `Command::new("git")` prod sites** (admin / canonical_hygiene / auto_release / conflict_notify / tmp_review_gc / retention::worktrees / ci::mod / force_release::gc). They are **non-uniform** — one is a `push` (network), several don't set `AGEND_GIT_BYPASS` — so they can't be blanket-routed through `git_bypass` and need per-site care. **Deferred to FIX C2** rather than sprawl this careful change. Enumeration available on request.

### DIFF
--- a/src/git_helpers.rs
+++ b/src/git_helpers.rs
@@ -1,6 +1,21 @@
 //! Git helper functions — single source of truth for remote/branch detection.
 
 use std::path::Path;
+use std::time::Duration;
+
+/// #1897: bound for LOCAL git ops (branch / worktree add / rev-parse / status /
+/// remote get-url / log / diff / reset). These don't hit the network; a healthy
+/// one is sub-second and the only legitimate slowness is a contended
+/// `.git/index.lock` or a large worktree-add checkout. 60s is far above any real
+/// local op, so it never false-kills a legit op but still fails fast instead of
+/// hanging the daemon forever (the #1893/RC1 root cause).
+pub(crate) const LOCAL_GIT_TIMEOUT: Duration = Duration::from_secs(60);
+
+/// #1897: bound for NETWORK git ops (fetch / clone). A large fetch can legitimately
+/// take minutes, so this is generous — strictly wider than the prior hard-coded
+/// 60s at the fetch sites, so it can only ADD tolerance (never newly false-kills a
+/// fetch that passed before) while still capping a wedged network op.
+pub(crate) const NETWORK_GIT_TIMEOUT: Duration = Duration::from_secs(300);
 
 /// #781 Piece 6 — daemon-internal `git` subprocess wrapper that always
 /// sets `AGEND_GIT_BYPASS=1`. Centralizes the bypass-env contract so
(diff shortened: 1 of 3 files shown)


suzuke:
Merge audit (section 3.12 mirror) — reviewer-2 VERIFIED (FIX C core @65bb1240 + the delta @9ad18e2) + all 5 required checks green (macos now passes — the empirical confirmation that the pre-existing racy test is fixed). #1897 FIX C: production git_bypass (8 sites) + run_git_idempotent now route through git_bypass_timeout (LOCAL 60s / NETWORK 300s for the 3 fetch sites) so a git op that hangs on a contended .git/index.lock or a stuck network helper fails fast with a clear Err instead of hanging the daemon indefinitely; git is spawned into its OWN process group (process_group(0) / CREATE_NEW_PROCESS_GROUP) and a timeout kills the whole group via kill_process_tree — reviewer-2 verified WATERTIGHT that this can never hit the daemon (getpgid resolves to the child's own group, not the daemon's; pid==0 guard; no spawn-vs-kill race). Bonus: while debugging the macos CI failure, dev BISECTED (30 runs each: baseline 14/30 fail vs FIX-C 11/30 fail = same rate) and proved FIX C was innocent — the checkout_bind concurrent-race test was PRE-EXISTING flaky (~47%, its #1896 green runs were luck) due to a racy assert_eq!(winners==1) that counts auto_created_branch flags (a branch-creating caller that then loses the bind-lease race surfaces 0 winners while staying perfectly idempotent). Replaced with race-outcome-independent invariants (branch created exactly once + no double-bind) → 0/40 fail. The spawn_group_bounded poll went from fixed 200ms to exponential backoff (1ms→50ms) so the fast-path git latency stays ≈ .output(). Merging --admin squash. Takes effect next restart. (FIX C2 — the ~15 additional independent raw Command::new('git') prod sites the audit found — is a follow-up.)

Issue #828, opened 2026-05-15
teams::delete: cascade full_delete_instance to orphan tasks (smoking gun for ghost-owner class)


suzuke:
## Symptom

When a team is disbanded via `teams::delete` (e.g. `team action=delete`), the team's member instances are removed from `fleet.yaml` but their existing task ownerships are NOT cleared. Result: tasks remain owned by ghost agent names that no longer exist in the runtime registry. Subsequent attempts to update / cancel those tasks fail with the daemon ACL (\"task X owned by Y, caller Z not authorized\") — addressed superficially by #808 force flag, but root cause persists.

This is the **smoking gun** for the entire \"ghost-owned task\" class problem. Today's hygiene batch had to force-cancel 13 such tasks (issues #603/605/607/608/609/610/611/612/614/615/616 + #694 cuts + Sprint 54 PLAN reviewer assignments). Same pattern recurs every time a team is disbanded.

## Root cause (verified 2026-05-15 by fixup-dev)

`src/teams.rs:170` `teams::delete()` only calls `fleet::remove_team_from_yaml` — drops the yaml entry. It does NOT iterate the team's members and call `full_delete_instance` (which would cascade to `orphan_tasks_for_owner` per `instance_lifecycle.rs:95`).

Single missing wire-up. Daemon already has all the building blocks (`full_delete_instance`, `orphan_tasks_for_owner`); they're just not invoked on team disband.

## Suggested fix

In `teams::delete`, before removing yaml entry:

```rust
pub fn delete(home: &Path, name: &str) -> Result<...> {
    let team = load(home, name)?;
    // NEW: cascade member cleanup
    for member in &team.members {
        if let Err(e) = full_delete_instance(home, member) {
            tracing::warn!(member, error = %e, \"failed to cascade-delete team member; tasks may remain ghost-owned\");
            // continue — other members still need cleanup
        }
    }
    fleet::remove_team_from_yaml(home, name)?;
    Ok(...)
}
```

`full_delete_instance` already handles: registry removal + binding cleanup + `orphan_tasks_for_owner` (clears assignee on owned tasks → status remains, but assignee=null → anyone can claim).

## Acceptance test

- Create team with 2 members (assignee = each member, status = `claimed` on 2 tasks)
- `teams::delete(team_name)`
- Assert: both tasks have `assignee=null`, status unchanged
- Assert: `team action=list` no longer shows team
- Assert: registry no longer contains either member

## Severity

[...]

Issue #1903, opened 2026-06-09
Bug: team teardown orphans tasks instead of cancelling — stale open tasks surface on same-name redeploy


cheerc:
## Summary

When a team is torn down (via `teams::delete`), each member's tasks are processed by `orphan_tasks_for_owner` — which **only clears the assignee field** but leaves the task status as `open`/`in_progress`. After a same-name team is redeployed, these orphaned tasks reappear in the new agents' `task list`, causing confusion about which work is genuinely new vs. a ghost from a prior incarnation.

## Root Cause

`orphan_tasks_for_owner` in `src/tasks/orphan.rs` (L26) emits only an `OwnerAssigned` event with `owner: None`:

```rust
// src/tasks/orphan.rs:43–49
let events: Vec<TaskEvent> = affected
    .into_iter()
    .map(|id| TaskEvent::OwnerAssigned {
        task_id: id,
        by: emitter.clone(),
        owner: None,        // ← assignee cleared
        routed_to: None,
        // status NOT changed — task stays open/in_progress
    })
    .collect();
```

The intent of `orphan_tasks_for_owner` is correct for the **single-instance-delete** case (ACL gate issue #808 — unlocking tasks so other agents can pick them up). However, when called from a **full team teardown**, the correct behaviour is different: if an entire team is disbanded, there is no surviving agent to pick up those tasks, so they should be **cancelled** rather than left open.

The call chain is:
```
teams::delete()
  └─ full_delete_instance(member)   [for each member]
       └─ tasks::orphan_tasks_for_owner(member)   [open → owner=None, status=open ← bug]
```

## Impact

- After team teardown + same-name redeploy, the new agents see a `task list` populated with stale open/in_progress tasks from the previous incarnation.
- Operators cannot easily distinguish genuine new tasks from ghost tasks.
- The orphaned tasks have no assignee and will never self-resolve, accumulating indefinitely across multiple teardown/redeploy cycles.

## Suggested Fix

Two options:

**Option A (surgical):** Add a separate `cancel_tasks_for_owner` function called from `teams::delete` in addition to (or instead of) `orphan_tasks_for_owner` for the team-disband case. This preserves the existing single-instance-delete ACL semantics while correctly cancelling tasks on full disband.

**Option B (parameter):** Add a `cancel: bool` parameter to `orphan_tasks_for_owner` — when `true`, emit a `StatusChanged { status: Cancelled }` event instead of (or after) the `OwnerAssigned { owner: None }` event.

[...]

Issue #1902, opened 2026-06-09
Bug: instance inbox file not deleted on teardown — causes stale message inheritance on same-name redeploy


cheerc:
## Summary

When an instance is deleted via `full_delete_instance`, the inbox file (`inbox/{name}.jsonl` or `inbox/{uuid}.jsonl`) is **never removed**. If a new instance is later created with the same name, it silently inherits all unread messages from the previous incarnation. This causes `poll-reminder` to fire repeatedly on messages that are days or weeks old, misleading operators into thinking there is active unread work when there is none.

We observed this in practice after a team teardown + same-name redeploy cycle:
```
[AGEND-MSG] kind=poll-reminder  unread=34  oldest=18778m
```
The `oldest=18778m` (~13 days) messages belonged to the **previous** team incarnation.

## Root Cause

`full_delete_instance` in `src/mcp/handlers/instance_lifecycle.rs` executes 15 cleanup steps but has **no inbox deletion step**:

```
fleet.yaml → telegram topic → working_dir → metadata(id-based) →
teams::remove_member_from_all → tasks::orphan_tasks_for_owner →
dispatch_idle::cleanup_pending → idle_watchdog::remove_agent_activity →
escalation_persist::remove → schedules::orphan_schedules_for_target →
dispatch_tracking::cleanup_for_instance → ci_watch::cleanup_watches_for_instance →
opencode_data_dir → binding::unbind → dispatch_hook::clear_bind_in_flight →
[name_residual_anywhere audit]
```

The post-delete audit `name_residual_anywhere` (line 234–235) **does detect** the leftover inbox file:

```rust
// src/mcp/handlers/instance_lifecycle.rs:234
if home.join("inbox").join(format!("{name}.jsonl")).exists() {
    sources.push("inbox");
}
```

So the function correctly returns `Err("residual stores: inbox")` — but the **caller** (`teams::delete`) treats `full_delete_instance` as best-effort and ignores the error, allowing the file to persist silently.

## Impact

- Operators see misleading `poll-reminder` notifications about old messages after a redeploy.
- Any agent logic that reads inbox messages (ack-absorption, correlation, reply routing) may act on ghost messages from a prior incarnation.
- The behavior is especially confusing in recurring team teardown/redeploy workflows (e.g. nightly fleet rotation).

## Suggested Fix

Add an inbox deletion step inside `full_delete_instance`, before the residual audit. Something like:

```rust
[...]

Issue #1682, opened 2026-06-03
metadata writers bypass metadata_path_resolved: telegram inbound + bot_api hand-build name-path, breaking the id→name symlink


suzuke:
## Follow-up to #1680
While RCA'ing #1680 (draft-gate read/write keyed differently), found the **mechanism that desyncs the name-file from the UUID-file**, and it's a separate latent bug:

`metadata_path_resolved` creates a symlink `metadata/<name>.json → metadata/<uuid>.json` when only the name-file exists, so name and UUID converge. But several writers **hand-build `metadata/<name>.json` and `std::fs::rename` onto it**, which REPLACES the symlink with a regular file → from then on name-file and uuid-file diverge:

- `src/channel/telegram/inbound.rs:396,418` — writes `pending_pickup_ids` to the name-path via rename.
- `src/channel/telegram/bot_api.rs:37` — same pattern.
- (audit the full list: `grep -rn 'metadata".join' src/ | grep -v test` — supervisor.rs:1270/1285, agent/mod.rs:1622, instance_lifecycle, api/mod.rs:653 etc. all hand-build; confirm which WRITE vs read-only.)

### Observed
Live: `metadata/fixup-lead.json` (name) had `pending_pickup_ids` fresh (telegram writer) but `last_input_epoch_ms` frozen 14d, while `metadata/<uuid>.json` carried the real keystroke timestamps. Two writers, two files, broken symlink.

### Why #1680 doesn't fully fix it
#1680 aligned the **read** path (draft gate) to the resolver, fixing the force-submit. But the telegram/bot_api **writes** still land on the name-file. Post-#1680, `pending_pickup_ids` (telegram confirmations / ✅ on inbox drain) may write to the name-file while everything else uses the uuid-file → pickup confirmations could target the wrong file.

### Fix
Route ALL metadata reads AND writes through `agent_ops::metadata_path_resolved` (or save_metadata/save_metadata_batch). No code should hand-build `home.join("metadata").join(format!("{name}.json"))`. Add a grep-based invariant test (like file_size_invariant) forbidding the hand-built pattern outside agent_ops.

Priority: normal (force-submit — the operator-facing symptom — is fixed by #1680; this is correctness hygiene + a latent pickup-confirmation bug).

suzuke:
[...]

PR #1905, opened 2026-06-09
fix(teardown): delete inbox (name+uuid) + cancel tasks on team disband (#1902 #1903)

suzuke:
Two teardown-cleanup gaps in the same class (cheerc-reported, RCA verified), bundled. The lead sanity-checked the #1903 caller-distinction spike before impl.

## #1902 — inbox leak (+ audit blind spot)
`full_delete_instance` had **no inbox-deletion step** (only a name-residual audit). Worse: the audit checked only the name-based `inbox/{name}.jsonl`, but the current inbox is **UUID-based** → a `<uuid>.jsonl` inbox leaked **silently, past the audit too**.
- New `inbox::storage::inbox_path_for_id` (pure, mirrors `metadata_path_for_id`). fleet.yaml is already removed at delete time, so `inbox_path_resolved` can't resolve the UUID — feed the captured `instance_id` directly (the #1682 name↔UUID-resolution trap).
- `full_delete_instance` now deletes **both** the name and the uuid inbox path.
- `name_residual_anywhere` now also detects the uuid inbox (was a blind spot giving false confidence).

## #1903 — task orphan-vs-cancel on disband
`full_delete_instance` orphans owned tasks (`OwnerAssigned{owner:None}`) so a **surviving** team can still mutate them (ACL). But `teams::delete` cascades `full_delete_instance` for **every** member — a full disband has no survivor, so tasks were left Open+ownerless → a same-name redeploy saw a **ghost open task**. Fix (Option A', spike-decided):
- New `tasks::cancel_tasks_for_owner` — sibling of `orphan_tasks_for_owner` (same live-status filter) emitting terminal `TaskEvent::Cancelled` (`by: system:team_disband`).
- `teams::delete` calls it per member **before** the cascade. **Ordering is load-bearing**: cancel-before-cascade runs while the owner is still set; the cascade's internal orphan then **skips** the now-Cancelled task (Cancelled ∉ its `{Open/Claimed/InProgress/Blocked}` filter) → a **single** Cancelled transition, **no** double `OwnerAssigned` event. (Cancel-*after*-cascade can't work — the internal orphan would have nulled the owner first; this was the key spike correction.)
- **Single-instance delete** never routes through `teams::delete`, so it still **orphans-OPEN** → ACL-unlock semantics preserved.

## §3.9 (real uuid inbox paths, not synthetic name files)
- `name_residual_anywhere_detects_uuid_inbox_residual_1902` — audit sees the uuid inbox.
- `full_delete_instance_removes_uuid_inbox_1902` — single delete clears the **real** uuid inbox (driven via `inbox::enqueue` → resolver).
- `team_disband_cancels_member_tasks_no_double_event_1903` — Cancelled + **zero** `system:auto_orphan` double event (proves the before-ordering, not luck).
- `full_delete_instance_orphans_owned_tasks` strengthened: asserts `status==Open` (single delete orphans-open, doesn't cancel).
- `delete_team_cascades_..._per_member` updated to the #1903 contract (it previously pinned the orphan-on-disband bug; now asserts Cancelled).

Full suite **3786/3786** under `nextest --profile ci`; fmt + clippy `-D warnings` clean.

### DIFF
--- a/src/inbox/storage.rs
+++ b/src/inbox/storage.rs
@@ -50,6 +50,15 @@ pub(crate) fn inbox_path(home: &Path, name: &str) -> PathBuf {
     home.join("inbox").join(format!("{name}.jsonl"))
 }
 
+/// #1902: id-based inbox path (pure — no fleet.yaml lookup, unlike
+/// [`inbox_path_resolved`]). Mirrors `agent_ops::metadata_path_for_id`. For
+/// teardown paths where the `InstanceId` is known directly and fleet.yaml has
+/// already been removed (so the name→id resolver can't run) — e.g.
+/// `full_delete_instance`, whose UUID inbox would otherwise leak silently.
+pub(crate) fn inbox_path_for_id(home: &Path, id: &crate::types::InstanceId) -> PathBuf {
+    home.join("inbox").join(format!("{}.jsonl", id.full()))
+}
+
 /// Sprint 46 P2: resolve inbox path by InstanceId when available.
 /// Migrates legacy name-based files to id-based on first access.
 pub(crate) fn inbox_path_resolved(home: &Path, name: &str) -> PathBuf {
--- a/src/mcp/handlers/instance_lifecycle.rs
+++ b/src/mcp/handlers/instance_lifecycle.rs
@@ -99,6 +99,19 @@ pub(crate) fn full_delete_instance(home: &Path, name: &str) -> Result<(), String
             let _ = std::fs::remove_file(crate::agent_ops::metadata_path_for_id(home, &id));
         }
     }
+    // #1902: delete the inbox file — this step was MISSING entirely, so a deleted
+    // instance's inbox leaked. Cover BOTH paths: the name-based `inbox/{name}.jsonl`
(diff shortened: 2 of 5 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head 5508807 == PR head). Crux probe (the #828 test rewrite) independently git-blamed: the original owner.is_none() assertion traces to f8097c0a (#834, fix #828 ghost-owner prevention) whose commit message says previously-owned tasks become owner=None / today they remain ghost-owned — so it pinned the orphan-on-disband behavior as the ghost-OWNER fix, and #1903 refines its residual (full-disband orphan-OPEN is itself a ghost-OPEN task). The rewrite owner==None to status==Cancelled is a legitimate contract refinement (terminal Cancelled supersedes and strengthens the no-actionable-ghost goal), NOT edit-to-pass. Other probes: cancel-before-cascade ordering is load-bearing and test-protected (cancelled==1 AND auto_orphan==0; reverse-derived that cancel-after makes both go red); single-delete ACL preserved (test asserts status==Open); #1902 uuid inbox cleaned on the real prod uuid path. CI 6/6 green, 3786/3786 nextest --profile ci.

Issue #1907, opened 2026-06-09
name_residual_anywhere audit is systemically incomplete — 6 of ~15 per-instance stores, name-only (the teardown-leak class keeps recurring because the oracle can't catch it)


suzuke:
Root cause behind the recurring teardown-leak class (#1855 / #1902 / #1903 / #1906). The post-delete audit name_residual_anywhere is supposed to be the safety net that catches any per-instance store full_delete_instance forgot to clean. But the audit itself is leaky:

- It checks only ~6 stores (fleet.yaml/teams, metadata [name+uuid], inbox [name-only pre-#1905], runtime/binding.json [name], notification-queue [name]) while cleanup touches ~15.
- Every check except metadata is NAME-only — but live stores are frequently UUID-keyed, so uuid residual leaks WITHOUT tripping the audit (this is exactly the #1902 deeper finding).

Because the oracle is incomplete, every leak in this class (inbox #1902, tasks #1903, worktree + usage_limit_notify #1906) had to be found by hand / by an external contributor / by a bespoke audit — the safety net never fired.

## Proposed direction
Make name_residual_anywhere comprehensive: check EVERY per-instance persistent store, both name- AND uuid-keyed paths. To avoid false positives, pair it with an explicit ALLOWLIST of intentionally-retained stores (from the audit: schedules [orphan-disabled], tasks [owner-cleared, body retained], deployments.json [the deploy recipe]) — residual in an allowlisted store is expected, not a leak. Anything else residual = a real leak the audit must flag.

This turns the audit into a real oracle (like the #1900 e2e harness turned wiring-gaps into caught regressions): once comprehensive, it would have caught #1902/#1903/#1906 automatically, and will catch future ones.

## Note
The allowlist (what is intentionally retained vs a leak) is a judgment call worth a quick operator/lead confirm — some 'residual' may be deliberate (dead-code-may-be-roadmap lesson). Lower urgency than the concrete #1906 leaks but higher leverage (prevents the whole class). Source: the teardown-completeness audit under #1900-followup.

## References
- Audit: full coverage matrix in the #1900-followup audit report (fixup-dev-2)
- Class: #1855, #1902, #1903, #1906
- Analogy: #1900 e2e harness (turning a manual discipline into an automated catch)

suzuke:
[...]

Issue #1906, opened 2026-06-09
Teardown residual leaks (TIER 1): full_delete_instance leaves physical worktree + usage_limit_notify entry — same class as #1902/#1903


suzuke:
Systematic teardown-completeness audit (follow-up to #1900/#1902/#1903) empirically confirmed (isolated /tmp home, real daemon, create victim -> seed state -> delete_instance -> scan residual) two MORE residual leaks in the same teardown-cleanup class. inbox (#1902) + tasks (#1903) are already fixed in #1905; these are the remaining TIER-1 (look like genuine bugs, audit fully blind, no eager self-heal at teardown).

## Leak 1 — leased physical worktree never released
full_delete_instance's binding::unbind clears runtime/<name>/binding.json + .sig, but NEVER releases the physical worktree at worktrees/<name>/<branch>/ (+ the .agend-managed marker). grep confirms no release_worktree / worktree_pool release call in full_delete_instance. The name_residual_anywhere audit checks runtime/<name>/binding.json (now gone) so it FALSELY reports clean. Impact: a same-name + same-branch redeploy collides with the orphan worktree -> lease conflict / stale checkout. #1885 force_reclaim GC eventually reaps it (agent no longer live), but it leaks in the window between teardown and the next GC tick. This is the same 'rely on GC backstop' shape as #1881/#1885 — #1879 patched the binding-state leak but not the physical worktree.
SPIKE: determine whether eager worktree release at teardown is safe + correct (is the agent process guaranteed dead at full_delete_instance time? how does worktree_pool release interact?) vs. GC being the intended path. If eager release is safe, add it to full_delete_instance + teach name_residual_anywhere to detect worktrees/<name>.

## Leak 2 — usage_limit_notify.json name-keyed entry never cleaned
full_delete_instance has ZERO cleanup for the deleted instance's entry in usage_limit_notify.json (grep confirmed). Impact: a same-name redeploy inherits the deleted occupant's suppression record -> its FIRST genuine usage_limit is SUPPRESSED until the stale unlock_deadline / 24h null-window (#1894/#1895) elapses. This is the SAME stale-state-rehydrate class that escalation_persist (#1744-H2/#1680) added a remove() to fix — usage_limit_notify was simply missed. FIX: add a cleanup step removing the name-keyed entry (mirror escalation_persist::remove), and teach name_residual_anywhere to detect a leftover usage_limit_notify entry for the name.

## Root cause (shared) — see separate issue
[...]

Issue #1900, opened 2026-06-09
e2e workflow self-validation harness (MVP) — hermetic daemon + deterministic mock agents drive dispatch→work→report→review, assert seam invariants


suzuke:
## Why
A large fraction of this sprint's bugs were NOT bad arithmetic — they were broken SEAMS: a directive accepted at the MCP layer gets dropped before reaching the mechanism. The re-marshal/allowlist-drop class (#931, #1031, #1833, #1877) is exactly this; so are the metadata name-vs-UUID split, ci re-nudge coupled to read-state, and the worktree-leak series (×4). Our unit tests inject synthetic fixtures AT the mechanism, bypassing the seam, so they pass while prod breaks. Today the only thing that catches these is adversarial human review + a prod incident + a restart cycle — slow and expensive. §3.9 ('drive the real entry point with a representative fixture') is the manual version of the fix; this issue is the always-on automated version.

## MVP goal (KISS — ONE hermetic scenario, no GitHub)
An automated integration test that boots a REAL daemon against an ISOLATED AGEND_HOME (NEVER ~/.agend-terminal — /tmp/agend-e2e-<dated> only) with a LOCAL bare git repo as origin (NO network, NO GitHub), registers DETERMINISTIC mock agents (scripted bots, not real LLMs — free, fast, repeatable), and drives the canonical coordination loop:

  dispatch task to mock-dev WITH branch=B  →  mock-dev commits in its bound worktree  →  mock-dev sends report  →  mock-reviewer auto-woken  →  mock-reviewer sends VERIFIED report

asserting the SEAM invariants at each step:
  1. directive survival (THE class): mock-dev's bound worktree is actually checked out on branch B, not main
  2. worktree bound to mock-dev (not leaked / not double-bound)
  3. report actually routed to lead
  4. reviewer actually auto-woken by next_after_ci / dispatch

A single 'dispatch-with-directive → assert-side-effect' scenario catches the ENTIRE re-marshal/drop class automatically.

## The one piece with real engineering: the deterministic mock agent
The daemon already spawns agents with MCP access. A mock agent = 'an agent that runs a fixed script instead of an LLM': on launch, drain inbox → find the dispatched task → perform the canned action (write file, git commit in bound worktree, send report with a fixed verdict). Spike how to register it (new backend kind? a mock backend command in the fixture fleet.yaml? reuse existing test scaffolding?) — this is the spike's main question.

## Out of scope for MVP (Phase 2 later)
[...]

PR #1909, opened 2026-06-09
fix(#1907): complete teardown residual oracle + whole-home-scan regression

suzuke:
## #1907 — teardown-completeness oracle + whole-home-scan regression

Follow-up to the #1900 teardown audit (and #1905/#1906, which fixed the two Tier-1
leaks it found). This completes the `name_residual_anywhere` audit oracle and —
the high-leverage payoff — adds a **whole-home-scan regression test** that turns
ANY future per-instance store forgetting teardown cleanup into a RED test.

### The reframe (approved in the spike)
`name_residual_anywhere` is a *curated* list — completing it only helps the
production loud-Err-on-delete, and a future store nobody adds stays invisible. So
the regression test does a **whole-home scan** (name + uuid) after a fully-exercised
victim's delete, failing on any residual not in an explicit annotated allowlist.
That catches unanticipated stores; the oracle completion is the secondary,
production-facing half.

### Oracle completion (`name_residual_anywhere`)
Added detection (each mirroring its cleanup's exact predicate, via new `has_*`
accessors so audit and cleanup never disagree) for the stores `full_delete_instance`
cleaned but the audit didn't check: **agent-activity, escalation, dispatch_tracking,
pending-dispatch, ci-watch, opencode-data, runtime-dir, pr-state, workspace**.
Allowlisted/intentional-retention stores (schedules/tasks/deployments/topics) are
deliberately absent — adding them would make every delete a false `Err`.

### Cleanups the whole-home scan caught during impl (the payoff working)
The new test immediately surfaced four more residuals that the curated oracle had
never checked — all now fixed in `full_delete_instance`:
- **pr-state subscribers** — a deleted name lingered in PR subscriber lists (no
  per-instance cleanup existed). Added a Value-based scrub (an audit/cleanup must
  not be defeated by a partial-parse — the #1902 "leaky oracle" lesson).
- **`workspace/<name>`** (+ its name-path metadata) — `cleanup_working_dir` was
  *skipped entirely* for the common entries with no explicit `working_directory:`,
  leaking the daemon-created default workspace. Now defaults to `workspace/<name>`.
- **inbox / metadata `.lock` flock siblings** — the data files were removed but the
  `with_json_state` advisory-lock files lingered (name/uuid-bearing).
- **`worktrees/<name>/` for slash branches** — `feat/x` nests an intermediate dir,
  so #1906's single `remove_dir` left the empty `feat/` (and the agent dir) behind.
  Replaced with `remove_empty_dir_tree` (bottom-up empty removal; a refused
  unmanaged worktree stays put for the audit, exactly like the old `remove_dir`).

### Regression test
`tests/teardown_completeness_regression.rs` (`#![cfg(unix)]`): boots a REAL daemon
on an isolated `/tmp` home (never `~/.agend-terminal`), gives a victim EVERY
hermetic-reachable state (inbox name+uuid, task-owner, binding+worktree,
dispatch_tracking, pending-dispatch, ci-watch) + direct-seeds the daemon-internal
stores (usage_limit_notify, escalation, agent-activity, pr-state), runs the real
`delete_instance`, then asserts (a) the delete returns no `error` (⟺ oracle clean,
via the production path) AND (b) a whole-home scan finds zero name/uuid residual
except the annotated allowlist. The allowlist (schedules/tasks/deployments/topics +
structural logs) is explicit + per-entry rationale'd, so an operator/reviewer can
reclassify any entry.

### Not changed (out of scope, per spike)
Caller consumption of the audit `Err` is untouched — `full_delete_instance` still
returns `Err` on residual; `handle_delete_instance` still surfaces it; `teams::delete`
stays best-effort. The fix is detection-completeness + the test, not a change to the
prod fail-direction.

One existing unit test (`full_delete_instance_returns_err_when_residual_remains_post_cleanup`)
switched its residual seed from `metadata` to `notification-queue`: the workspace-default
fix now correctly cleans the metadata it relied on leaking, so it needed a store that
still genuinely survives the no-daemon test context to keep proving the Err contract.

### DIFF
--- a/src/daemon/ci_watch/mod.rs
+++ b/src/daemon/ci_watch/mod.rs
@@ -47,7 +47,10 @@ pub use provider::{
     MergeableState, PrState,
 };
 #[allow(unused_imports)]
-pub use registry::{ci_watches_dir, cleanup_watches_for_instance, remove_watch, watch_filename};
+pub use registry::{
+    ci_watches_dir, cleanup_watches_for_instance, has_instance_anywhere, remove_watch,
+    watch_filename,
+};
 #[allow(unused_imports)]
 pub use sweep::{gc_stale_watches, startup_sweep};
 pub use watcher::check_ci_watches;
--- a/src/daemon/ci_watch/registry.rs
+++ b/src/daemon/ci_watch/registry.rs
@@ -148,6 +148,36 @@ pub fn cleanup_watches_for_instance(home: &Path, instance: &str) -> usize {
     affected
 }
 
+/// #1907 teardown audit: does any ci-watch still reference `instance` — in its
+/// `subscribers`, the legacy `instance` field, or `next_after_ci`? Mirrors the
+/// three scrub targets of [`cleanup_watches_for_instance`] exactly so the
+/// residual audit and the cleanup never disagree.
+pub fn has_instance_anywhere(home: &Path, instance: &str) -> bool {
+    let dir = ci_watches_dir(home);
+    let Ok(entries) = std::fs::read_dir(&dir) else {
+        return false;
+    };
+    for entry in entries.flatten() {
+        let path = entry.path();
+        if path.extension().and_then(|e| e.to_str()) != Some("json") {
+            continue;
+        }
+        let Some(watch) = std::fs::read_to_string(&path)
+            .ok()
+            .and_then(|c| serde_json::from_str::<super::watch_state::WatchState>(&c).ok())
+        else {
+            continue;
+        };
+        if watch.subscriber_names().iter().any(|s| s == instance)
+            || watch.instance.as_deref() == Some(instance)
+            || watch.next_after_ci.as_deref() == Some(instance)
+        {
+            return true;
+        }
+    }
+    false
+}
+
 /// Deterministic, collision-resistant filename for a CI watch entry.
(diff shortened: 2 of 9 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head 9529255 == PR head). 7 probes pass. Teeth (the crux) proven by structure + the PR's OWN history: the whole-home-scan assertion filters the allowlist and seeds ~10 non-allowlisted stores so removing any cleanup makes it non-empty (cannot be a tautology), AND the test ALREADY went RED during this PR catching the stale-daemon's missing workspace cleanup residual — a real teeth demonstration, stronger than a synthetic mutation. workspace-default is gated by starts_with(workspace/) so explicit user working_directory is never remove_dir_all'd (only named agend files cleaned). remove_empty_dir_tree only removes empty dirs (user files preserved). allowlist 9 entries all structural/intentional with rationale. re-anchored unit test legitimate (metadata now cleaned by the workspace fix). binary_path CARGO_BIN_EXE fresh-build is the RED->GREEN root cause; e2e_workflow (#1900) no regression confirmed via nextest --profile ci. CI 6/6 green. Transparency: reviewer verified teeth structurally + via the PR's own RED history, not a fresh daemon-boot mutation (unsafe from a bound-agent worktree). Completes the teardown-cleanup class + adds the whole-home-scan regression oracle that catches future leaks.

PR #1910, opened 2026-06-09
fix(inject): longer pre-submit settle for typed_inject backends (codex 0.138)

yujunchao:
## Problem

`typed_inject` ratatui input boxes (codex's `›`) need the just-typed line to **commit** before the trailing `\r` is read as *submit*. The hardcoded `50ms` pre-submit settle in `inject_with_target` works for codex **0.118** (#1670), but codex **0.138** re-renders its input box more slowly: the `\r` arrives before the line commits and is appended as a *newline*, so the wake line stacks un-submitted in the box and the agent never wakes.

`doctor` already warns `codex … calibrated: 0.118.0, patterns may need update` — this is one concrete consequence on the newer CLI (observed on Windows: a `send`/wake to a codex 0.138 instance leaves `[AGEND-MSG-PENDING] …` sitting in the `›` input, never processed).

## Fix

Bump the pre-submit settle from a hardcoded `50ms` to `250ms` **for `typed_inject` backends only** (codex / opencode / gemini / agy). Bulk-inject backends (claude's `❯`, which commits instantly) keep `50ms`. The only cost is +200ms latency on an infrequent inject; it is inert for bulk-inject and harmless for codex 0.118 (which already worked at 50ms).

```rust
let settle_ms = if target.typed_inject { 250 } else { 50 };
std::thread::sleep(std::time::Duration::from_millis(settle_ms));
write_with_timeout(&target.pty_writer, submit)?;
```

## Validation

`cargo check` + the inject unit tests pass.

Empirically validated against **real codex 0.138** with a standalone `winpty` harness that mimics `inject_with_target` (paced 2ms/byte type → settle → `\r`), 2 runs each:

| settle | result |
|---|---|
| 50ms (current) | **STUCK** — line stays in the `›` input, no processing |
| 250ms (this PR) | **SUBMITTED** — line enters the transcript, codex starts the turn (`esc to interrupt` / `Booting MCP server`) |

## Scope

`calibrated_version` stays `0.118.0` — a full 0.138 pattern recalibration (ready / dismiss / state patterns) is out of scope; the existing patterns still match 0.138's banner and `›` prompt. This is the minimal, version-agnostic settle fix that unblocks codex 0.138 inject.

### DIFF
--- a/src/agent/mod.rs
+++ b/src/agent/mod.rs
@@ -2050,7 +2050,18 @@ fn inject_with_target(target: &InjectTarget, text: &[u8]) -> crate::error::Resul
     if target.deleted.load(std::sync::atomic::Ordering::Acquire) {
         return Ok(());
     }
-    std::thread::sleep(std::time::Duration::from_millis(50));
+    // Pre-submit settle: let the input box fully commit the just-written line
+    // before `\r` is read as submit. Bulk-inject TUIs (claude's `❯`) commit
+    // instantly, so 50ms is plenty. typed_inject ratatui boxes (codex's `›`,
+    // opencode/gemini/agy) re-render/debounce and need more headroom — 50ms is
+    // enough for codex 0.118 (#1670) but codex 0.138's input box re-renders
+    // slower, so the `\r` arrives before commit and is appended as a newline:
+    // the wake line stacks un-submitted and the agent never wakes. A longer
+    // settle for typed_inject backends fixes 0.138 without affecting 0.118
+    // (the only cost is +200ms latency on an infrequent inject) and is inert
+    // for bulk-inject backends.
+    let settle_ms = if target.typed_inject { 250 } else { 50 };
+    std::thread::sleep(std::time::Duration::from_millis(settle_ms));
     write_with_timeout(&target.pty_writer, submit)?;
     Ok(())
 }


suzuke:
Thanks for the well-researched fix and the empirical winpty validation, yujunchao — the RCA (typed_inject ratatui commit vs the trailing CR on codex 0.138) is solid and this directly unblocks codex 0.138 inject/wake, which matters to us. We're accepting this as a tracked STOPGAP and have opened #1912 for the root fix: replacing the fixed pre-submit settle with a terminal-state readback confirmation (poll the rendered input box until the typed line is committed, then submit) so there is no version-specific magic number that needs re-bumping each codex release. Running it through our review + CI now; once #1912 lands, the 250ms here can be removed. One question for our records: did 250ms have clear margin over the observed commit time, or was it close to the threshold?

suzuke:
Thank you, @yujunchao — this was an excellent, well-researched report, and your empirical winpty validation against real codex 0.138 was exactly the kind of evidence that makes a fix actionable. Your RCA (the typed_inject ratatui input needing the line to commit before the trailing CR is read as submit, and codex 0.138 re-rendering slower than 0.118) was spot-on. It directly drove the root fix: we opened #1912 from this PR and have just merged #1914, which replaces the fixed-sleep guess entirely with a terminal-state readback-confirm handshake (poll the rendered input box until the typed line appears, then submit, fail-open on timeout). That removes the version-specific magic number so it never needs re-bumping on a future codex release. Since #1914 supersedes the 250ms settle on the same code path, we are closing this PR as superseded — but it is the reason the root fix exists, and the credit is yours. We would genuinely welcome more contributions from you. Thank you again.

Issue #649, opened 2026-05-11
Schema-enforced dispatch templates: hard-code sequencing + ETA + cadence rules


suzuke:
## Problem
General orchestrator 用自然語言寫 dispatch message、容易帶 bug：sequential vs parallel、ETA、回報 cadence、worktree binding 等規則靠每次手寫、寫錯時下游 team follow 錯規範。

## Incident reference2026-05-11 gemini-team 試手 dispatch incident: 3 PRs (#645/#646/#647) CI passed but stayed unmerged for ~1hr due to dispatch-chain breakdown + parallel-merge anti-pattern + reviewer waiting_on stale + no CI-pass downstream trigger
Specifically：general dispatch wrote「3 個 parallel ship」追求 throughput、忽略「parallel IMPL 但 sequential merge」標準、導致 lead-gemini follow 錯規範 ship multi-PR base stale 風險。

## Suggested approach
Daemon-side enforce dispatch schema：
- `send` MCP tool 加結構化 fields：`sequencing` (parallel | sequential | sequential-merge-only)、`eta_target_minutes`、`reporting_cadence` (per-pr | wave-end | both)、`worktree_binding_required` (bool)
- daemon 收 dispatch 強制 validate、缺欄位或不合 spec 直接 reject
- 自然語言部分仍可寫 in `message` body、但 sequencing 等規範由 schema 帶
- 下游 lead/dev/reviewer 從 schema 直接讀規範、不靠 parse 自然語言

## Priority
P1 — root cause of repeated dispatch quality issues (today's gemini-team incident, plus past similar dispatch drift events)

PR #1911, opened 2026-06-09
fix(#1907): remove ghost send-task directives + schema-consumption invariant

suzuke:
Follow-up to the dispatch-directive-survival audit (#1907 line). Removes three `send` schema fields that have **zero consuming mechanism** — advertised to agents but silently dropped — and adds a meta-test that prevents the ghost-field class from recurring.

## Removed (3 confirmed ghosts)
| field | why it's a ghost |
|---|---|
| `working_directory` (send-task) | bare no-description field copied from `create_instance`; #1843 deliberately never forwarded it; `instructions.rs` wrongly advertised it. **Distinct** from create_instance/fleet `working_directory` (kept — heavily consumed). |
| `requires_reply` | TS-port remnant, dead since 2026-04-29; not even on the `InboxMessage` struct. |
| `task_summary` | TS-port remnant; only a `notify.rs` test fixture references the string. |

Footprint: 3 lines in `def_send` + one `working_directory` mention in `instructions.rs:226`. None of the 3 are on the `InboxMessage` struct or in marshalling — a clean schema-only removal.

## Preserved (explicitly NOT touched)
- **#649 trio** `sequencing` / `eta_minutes` / `reporting_cadence` — intentional Phase-1 passthrough, Phase-2 mechanism deferred.
- `worktree_binding_required` — enforced at `check_worktree_enforcement`.
- **create_instance/fleet `working_directory`** (`def_create_instance`) — real heavy consumption (worktree gate / token_cost / spawn). A distinct field; surgically left alone.

## Meta-test (the payoff)
`send_schema_fields_are_consumed_or_deferred_allowlisted` — the **complement** of #1505's Invariant B:
- #1505 catches a handler *read* with no schema declaration.
- This catches a schema *declaration* with no consumer.

Every `send` field must be in `CONSUMED` (each carries a consumption-point cite from the audit — self-documenting, hard to fake) or `DEFERRED_ALLOWLIST` (each carries a #649 roadmap ref). Enumerates via the built `tool_definitions()` (robust, not source-parse), with a reverse stale-entry check. A future bare schema field with no consumer → red, with a guiding message. `tags` carries an honest cite noting `is_some()` triggers broadcast mode but value-based tag-targeting is NOT implemented.

RED proof: injecting a `ghost_directive` field fails the test naming it.

## Verification
- `cargo fmt --check` ✓
- `cargo clippy --features tray -D warnings` ✓
- Full suite (system git, CI-parity): **3793/3793 passed**. One real-git test (`dispatch_branch_skips_metadata_stub_…`) false-fails only under the local `agend-git` shim, passes under system git (CI uses real git).

### DIFF
--- a/src/instructions.rs
+++ b/src/instructions.rs
@@ -223,7 +223,7 @@ pub(crate) fn build_instructions_body(
         "  - `instance` (single recipient) OR `instances` / `team` / `tags` (broadcast mode)\n",
     );
     content.push_str("  - `request_kind`: `task` (delegation, expects report back) / `report` (results back) / `query` (question, expects reply) / `update` (status) / omit (plain message)\n");
-    content.push_str("  - Task-mode optional fields: `success_criteria`, `task_id`, `force` + `force_reason`, `second_reviewer` + `second_reviewer_reason`, `branch`, `working_directory`\n");
+    content.push_str("  - Task-mode optional fields: `success_criteria`, `task_id`, `force` + `force_reason`, `second_reviewer` + `second_reviewer_reason`, `branch`\n");
     content.push_str(
         "  - Report-mode optional fields: `correlation_id`, `reviewed_head`, `artifacts`\n",
     );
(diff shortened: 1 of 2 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head c4d8b9c == PR head). Surgical: only def_send's 3 ghost fields removed (working_directory/requires_reply/task_summary) — def_create_instance is NOT in the diff, so the heavily-consumed create_instance/fleet working_directory (a DISTINCT field) is untouched. The 3 removed are truly zero-consumed (requires_reply has 0 non-test hits; task_summary's only non-tools hit is a test string in notify.rs:764; send working_directory is the bare no-desc copy). The meta-test asserts every send-schema field is in CONSUMED (with a real consumption-point cite — spot-checked worktree_binding_required→check_worktree_enforcement, next_after_ci→ci_watch poller, branch→dispatch_auto_bind_lease, all accurate) or DEFERRED_ALLOWLIST (the #649 trio with a Phase-2-deferred roadmap ref); it's a built-schema set-difference assert (structurally can't be a tautology; documented RED-proof: adding a ghost field fails the test naming it). tags honestly annotated (is_some triggers broadcast, value-targeting not implemented). CI 6/6 green.

Issue #1912, opened 2026-06-09
Root fix: replace the fixed pre-submit inject settle with terminal-state readback confirmation (typed_inject backends)


suzuke:
## Context
PR #1910 (yujunchao) is a STOPGAP for codex 0.138 inject failure: it bumps the hardcoded pre-submit settle from 50ms → 250ms for typed_inject backends, because codex 0.138's ratatui input box re-renders slower than 0.118, so the trailing `\r` arrives before the typed line commits and is read as a newline (the wake line stacks un-submitted in the `›` box, the agent never wakes). #1910 unblocks 0.138 now, empirically validated (winpty harness, 50ms STUCK / 250ms SUBMITTED).

## Why a root fix
A fixed sleep is a band-aid — it GUESSES the commit happened instead of CONFIRMING it. Any magic number is wrong eventually: too short races, too long wastes latency and STILL isn't guaranteed under load / on a slow machine / on a future codex version that re-renders even slower. We've already recalibrated this once (#1670: 50ms for 0.118 → #1910: 250ms for 0.138). The next codex version will need another bump. This is a recurring tax.

## Proposed root fix — readback-confirm handshake
Replace the fixed settle with a terminal-state confirmation:
- After typing the line (the existing 2ms/byte typed_inject), POLL the rendered terminal screen until the typed line is actually present in the input box (the `›` prompt line for codex), THEN send the submit `\r`.
- No magic number — adaptive (submits as soon as the line commits, however fast/slow), version-agnostic (any codex version, any typed_inject backend), robust under load.
- REUSE the existing terminal-state-detection infrastructure (#1523 epic — ready_pattern / viewport heuristics already read & match the rendered screen). The new check is just "does the input box now contain the line I typed?".
- Bounded by a timeout fallback (if the line never appears within N seconds → error, consistent with the existing inject timeout).
- SPIKE the exact shape: pre-confirm (readback the typed line is committed → then submit) vs post-confirm+retry (submit → verify the line LEFT the input box into the transcript → retry if still stuck). Post-confirm+retry directly verifies the GOAL (line submitted) and covers other stuck-causes, but must guard against double-submit. Decide in the spike.

## Scope
typed_inject backends (codex / opencode / gemini / agy). Bulk-inject (claude's `❯`, commits instantly) stays on the fast path — no readback needed (or a trivially-fast one). Once landed, the #1910 250ms magic number can be removed.

## Acceptance
[...]

PR #1914, opened 2026-06-09
fix(#1912): readback-confirm inject — replace fixed-sleep submit guess

suzuke:
## #1912 — readback-confirm inject (root fix for the codex `\r`-race stopgap)

Replaces the fixed pre-submit `sleep` (a version-specific "guess" — 50ms, bumped to
250ms in #1910) in the typed-inject path with a **terminal-state readback**: after
typing the payload, poll the RENDERED input area until the typed line lands, *then*
send the submit `\r`. Adaptive, version-agnostic, no magic number.

### Root cause
`inject_with_target` types the payload, then sends `\r` to submit. codex's
re-rendering `›` input widget doesn't reliably commit a typed line before the
trailing `\r` arrives → the `\r` is swallowed as an in-line newline → the wake sits
un-submitted → **the agent never wakes**. The fixed sleep just "guessed" how long
the widget needs; every codex version re-tuned it.

### Fix (pre-confirm handshake)
- `InjectTarget` gains `core: Arc<CoreMutex<AgentCore>>` so the inject path (which
  snapshots + releases the registry lock) can poll the rendered screen via
  `core.vterm.tail_lines()` — the same in-process read state-detection uses. Polled
  with brief lock-and-release so the `pty_read_loop` renders the echo between polls.
- `readback_confirm_typed`: poll until the typed line's **tail-sentinel** (last ≤24
  chars of the final non-blank line — robust to input-box wrapping) renders in the
  bottom rows, then return so the caller submits. **Single submit → zero
  double-submit risk** (vs a post-confirm+retry, whose retry `\r` is the footgun).
- **FAIL-OPEN on timeout** (2s): if the echo never renders, submit anyway + `warn`
  (tag `#1912-readback-timeout`). This path is the agent-wake lifeline — an
  unconfirmed readback must never become "don't submit" (= agent never wakes, a
  regression worse than today's blind submit).
- `observe_post_submit`: optional post-submit observability (log/metric-only, **never
  retries**) — the input area should change after submit; warn (`#1912-postsubmit-
  nochange`) if it stays byte-identical, so prod can see whether confirmed submits land.

### Scope
Gated on `target.typed_inject` (codex/opencode/gemini/agy — the re-rendering input
widgets). **claude `❯` bulk keeps its fast path byte-identical** (tolerates bulk +
`\r`; no readback, no added latency). The fixed pre-submit sleep is removed for
typed-inject (readback replaces it); the bulk 50ms is untouched.

### Tests — all hermetic, deterministic, flake-free (no daemon, no real PTY)
`src/agent/tests.rs`: `inject_sentinel` extraction; readback confirms when the line
is rendered; **readback polls and waits for an async echo** then confirms (the heart
of the fix — a background thread renders the echo after inject starts polling);
readback times out fail-open promptly without hanging; post-submit screen-change
detection.

**Test-strategy note (deviation from the spike's "AgendHarness-shell" idea, flagged
for review):** `typed_inject` is a backend-preset constant and `Shell = false`, so an
AgendHarness *shell* agent goes through the BULK path and would not exercise the
readback code at all. Rather than add a real-daemon test that can't cover readback
(and adds daemon-boot flake surface on this lifeline path), the async-echo test
covers the actual poll-until-rendered mechanism deterministically. The real codex
0.138 `›`-widget render timing remains out of hermetic reach — final fidelity is the
contributor's standalone winpty harness (honestly out-of-CI-scope).

### DIFF
--- a/src/agent/mod.rs
+++ b/src/agent/mod.rs
@@ -2050,9 +2050,137 @@ fn inject_with_target(target: &InjectTarget, text: &[u8]) -> crate::error::Resul
     if target.deleted.load(std::sync::atomic::Ordering::Acquire) {
         return Ok(());
     }
-    std::thread::sleep(std::time::Duration::from_millis(50));
+    // #1912: gate the pre-submit wait on the backend's input-widget style.
+    if target.typed_inject {
+        // Readback-confirm (#1912): poll the RENDERED input area until the typed
+        // line's tail-sentinel appears, THEN submit. Replaces the fixed-sleep
+        // "guess" that racing codex's re-rendering `›` widget required (every codex
+        // version re-tuned the magic number). FAIL-OPEN: on timeout we submit
+        // anyway (the helper warns) — this is the agent-wake lifeline, so an
+        // unconfirmed readback must never become "don't submit" (= agent never wakes).
+        let _confirmed = readback_confirm_typed(target, &stripped);
+    } else {
+        // claude `❯` bulk fast path — tolerates bulk bytes + `\r`; keep byte-identical.
+        std::thread::sleep(std::time::Duration::from_millis(50));
+    }
     write_with_timeout(&target.pty_writer, submit)?;
+    // #1912: post-submit observability (log/metric-only, NEVER retries — a second
+    // `\r` would risk double-submit). Typed-inject only; bulk's fast path stays lean.
+    if target.typed_inject {
+        let _submitted = observe_post_submit(target);
+    }
     Ok(())
+}
+
+/// #1912: tail-sentinel of a (possibly multi-line) injected payload — the last
+/// run of up to `MAX` chars on the final non-empty line, the line the submit `\r`
+/// commits. Short + drawn from the bottom line so it stays robust to input-box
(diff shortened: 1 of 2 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head ce5d229 == PR head). The agent-wake lifeline is fail-OPEN: the submit write is unconditional (not gated on the readback result), and the timeout arm only warns + returns false (never Err, never skips) — so an unconfirmed readback still submits and the agent always wakes. Single submit (readback/observe are poll-only, no retry — avoiding the post-confirm double-submit footgun). Claude bulk path byte-identical (only typed_inject runs readback). Brief core-lock polling (registry already released, no nesting) — no deadlock. Sentinel degrades to fail-open in every edge case. Tests hermetic/deterministic (async-echo test proves poll-until-rendered with teeth); codex-0.138 widget fidelity honestly scoped out-of-CI (contributor harness). CI 6/6 green. Replaces the version-specific fixed-sleep guess (incl. #1910's 250ms stopgap) with a terminal-state readback handshake — no magic number to re-bump per codex release.

PR #1917, opened 2026-06-09
docs(changelog): note #1576 fail-closed operator-mode one-time migration

yujunchao:
## What

Adds a CHANGELOG entry (EN + zh-TW) documenting the **one-time migration step** needed when upgrading across the #1576 boundary.

## Why

#1576 ("fail-closed + HMAC integrity for the operator-mode authority file", merged 2026-06-02) makes the daemon trust `operator-mode.json` only if it exists with a valid HMAC. **Any install upgrading from a pre-#1576 build has no signed file, so the #1339 authority gate fails closed to `Away` at startup** — and because the gate is enforced at the authority layer (not per-adapter), this silently suppresses operator interaction across **every** channel (Telegram, Discord via `--features discord`, future Slack/Matrix), until the operator runs `agend-terminal mode active` once.

Hit in practice after a real upgrade: the operator channel went silent with no operator-facing hint pointing at the fix. Previously undocumented in CHANGELOG, KNOWN_ISSUES, and the quickstart/USAGE docs.

`mode get` can still report the in-memory last-known-good mode while the on-disk file is absent, which makes the cause easy to misdiagnose — the note calls this out.

## Scope

- **Docs-only.** No behavior change — fail-closed is the intended security default and stays as-is.
- Mirrors the existing #1812 fail-closed migration-note convention already in `[Unreleased]`.
- 4 insertions, 0 deletions, across both `CHANGELOG.md` and `CHANGELOG.zh-TW.md`.


suzuke:
Merged — thank you, @yujunchao. This is an accurate, well-placed migration note: it documents a real upgrade gotcha (a pre-#1576 install starts fail-closed in Away with the operator channel silent until 'agend-terminal mode active' is run once), it correctly scopes the behavior as the intended channel-agnostic security default, and it calls out the 'mode get reports in-memory last-known-good while the disk file is absent' misdiagnosis trap. It mirrors the existing #1812 fail-closed migration-note convention in [Unreleased] exactly. Docs-only, CI green. Much appreciated — and thanks again for #1910, which drove the #1912/#1914 inject root fix.

Issue #1915, opened 2026-06-09
Complete the deleted-flag class: A2 stage2-restart resurrection + spawn_agent blind-insert chokepoint + 7 TIER-B wrong-action gaps


suzuke:
Follow-up to #1913 (which fixes only A1 = crash-respawn resurrection). A systematic audit of every act-on-instance path vs handle.deleted (the AtomicBool delete sets at lifecycle.rs:102, ~5s danger window before registry removal at :136) found #1913 is ONE site of a broader class.

## Root cause
`spawn_agent` (src/agent/mod.rs:1077) is a BLIND `reg.insert(id, handle)` — no check for an existing/deleted handle. Any overlapping spawn silently overwrites. Combined with no `deleted` check on the crash/stage2 respawn paths, this is the resurrection race.

## TIER A — Resurrection (HIGH; #1913 fixes A1 only)
- **A1** handle_crash_respawn → respawn_agent_worker → spawn_agent (crash_respawn.rs:58-95/276). **Fixed in #1913.**
- **A2** handle_stage2_restart (daemon/mod.rs:1683-1767) → spawn_agent. IDENTICAL mechanism, triggered by Stage2Restart instead of Crash. **NOT covered by #1913.** MEDIUM.
- **A3** recovery_dispatcher (per_tick/recovery_dispatcher.rs:261-276, 540) emits Stage2Restart while iterating the registry with no deleted check → feeds A2. LOW.

## TIER B — Wrong-action (act on a deleting agent without resurrecting)
- **B1** supervisor tick reactions_from_transitions (supervisor.rs:1091-1102) — fires spurious operator notifications (Recovered→gated_notify) for a deleting agent. MEDIUM (operator-visible).
- **B2** RecoveryDispatcher state mutation (recovery_dispatcher.rs:261-276) — advances recovery stage (e.g. Stage2Eligible) that survives delete and could fire on a re-spawned instance. MEDIUM (persisted state surviving delete).
- **B3** HangDetection health decay/escalation (hang_detection.rs:63-67) — false self-orch P0 escalation. LOW.
- **B4** Watchdog BlockedReason classification (watchdog.rs:45-46) — cosmetic stale-PTY classification. LOW.
- **B5** process_error_recovery retry-track mutation (supervisor.rs:1677-1681) — in-memory retry_count/exhausted mutated (inject leaf is guarded, track mutation is not). LOW.
[...]

PR #1918, opened 2026-06-09
fix(#1913): honor `deleted` on two spawn-vs-delete resurrection sites (crash-respawn + boot-stagger)

### LINKED ISSUE #1913: Bug: deleting an instance can RESURRECT it — crash-respawn ignores handle.deleted (race vs registry/config removal)
## Discovery
Found via the #1909 whole-home-scan teardown regression test, which was intermittently failing (workspace/victim residual after delete). Instrumented evidence (poll workspace/victim for 12s after delete, 4 runs): 3 runs GONE throughout, 1 run PRESENT for the full 12s and NEVER clearing → the residual is PERSISTENT, not a transient test race. The test correctly caught a real prod bug.

## Root cause — delete-vs-crash-respawn race
`delete_transaction` (src/daemon/lifecycle.rs) ordering: Step1 sets `handle.deleted = true` (~L102) → Step2 kills the victim's process → Step4 `registry.remove` → Step6 `configs.remove`.

When the process is killed, the daemon exit-loop (agent/mod.rs:~823) classifies the exit as **Crash** (not CleanExit) and calls `handle_crash_respawn` (agent/mod.rs:~836). `handle_crash_respawn` (src/.../crash_respawn.rs:~20) checks ctx.configs / resolve_uuid(fleet.yaml) / registry — but in the window BETWEEN the kill (Step2) and the removals (Step4/6) these all still exist, so every guard passes and it **respawns the victim**: re-creates workspace/<name>, re-registers, starts a new process. The deleted instance comes back.

**The hole**: neither the exit classification nor `handle_crash_respawn` consults `handle.deleted` — the very flag delete sets (at Step1, BEFORE the kill) to mean "this kill is intentional, do not resurrect." The crash path ignores it.

## Impact
- Deleting an instance can silently resurrect it (new process, re-registered, workspace re-created) when respawn wins the race against removal.
- This likely UNDERLIES the teardown-cleanup class (#1902/#1903/#1906/#1907): a resurrected instance re-leaks the very stores those fixes clean. It is plausibly the root that made several teardown residuals intermittent.

## Fix (deterministic)
`handle_crash_respawn`: when it obtains the handle (~L59-60), check `handle.deleted.load(Acquire)` → if true, treat as a clean/intentional exit and return WITHOUT respawning. Race-free: `deleted` is Stored at Step1 (happens-before the kill → before the crash-detection → before this load), so it is necessarily true during the respawn window. After Step4 removal the respawn already short-circuits (registry get = None), so the deleted check precisely covers the remaining window. Real crashes (deleted=false) still respawn — no regression to crash recovery.

## Test
The #1909 regression test becomes deterministically green once prod is fixed (no respawn → workspace stays removed). Add a focused §3.9 regression: delete → assert NO respawn (no new process, workspace stays gone) AND a real-crash (deleted=false) → DOES respawn (proves the gate is precise, not a blanket respawn-disable).

suzuke:
A deleted instance must never be resurrected by a concurrent spawn. Two sites re-created `workspace/<name>` + a registry handle **after** teardown cleanup ran — the intermittent residual that flaked the #1907/#1909 teardown oracle (and would block every PR rebased onto it).

## A1 — crash-respawn (`handle_crash_respawn`)
`delete_transaction` STOREs `handle.deleted = true` before killing the backend, but removes the registry/config entries only *after* the kill+wait. The kill is observed as a `Crash`, so without a gate the crash-respawn loop races those removals and respawns the just-deleted instance.

**Fix**: skip respawn when `handle.deleted` is set (stored before the kill → the `Acquire` load reliably sees it; once the registry entry is gone `reg.get` already returns `None`).

**Two regression tests prove the gate is precise** (not a blanket disable):
- `delete_does_not_respawn_1913` — deleted=true → skips the respawn path entirely (`record_crash` never runs, `total_crashes` stays 0).
- `real_crash_still_respawns_1913` — deleted=false → still enters the respawn path (`total_crashes` → 1). No crash-recovery regression.

## (b) — boot-stagger (`spawn_fleet_agents`) — *the actual #1909 flake*
The boot loop iterates a fleet snapshot and sleeps ~500ms (`spawn_stagger`) between spawns. A `delete_instance` in that window removes the agent from fleet.yaml, but the staggered spawn still fires and resurrects it.

**This — not the crash path — is what flaked #1909**, established empirically:
- A1 alone left #1909 **4/10 flaky**.
- `AGEND_SPAWN_STAGGER_MS=0` → **10/10** (isolates the boot-stagger).
- The daemon log shows the staggered boot-spawn re-creating `workspace/victim` 14ms after the delete's cleanup, **with no crash for the victim**.

**Fix**: re-check `fleet::instance_is_known` immediately before each staggered spawn; skip an agent torn down mid-boot. **#1909 now passes 25/25 at the default 500ms stagger.**

## Scope (honest)
These are **two concrete sites** of the spawn-vs-delete resurrection class — not the whole class. The load-bearing deep root (a `spawn_agent` chokepoint with a deleting-set guard that subsumes both, plus A2 `handle_stage2_restart` and the TIER-B sites from reviewer-3's audit) is deferred to **#1915**. The check→spawn gap in (b) is not atomic; #1915's chokepoint closes it fully. This PR closes the wide stagger window the oracle actually hit and unblocks the dependent PRs (#1910/#1911/#1914).

## Verification
- `cargo fmt --check` ✓ · `cargo clippy --features tray -D warnings` ✓
- A1 regression tests ✓ (both cases)
- #1909 oracle: **25/25** at default stagger (was 3/8 flaky)
- Full suite (system git, CI-parity): **3794/3794 passed**

Closes #1913.

### DIFF
--- a/src/daemon/crash_respawn.rs
+++ b/src/daemon/crash_respawn.rs
@@ -58,6 +58,27 @@ pub(super) fn handle_crash_respawn(home: &Path, crashed_name: &str, ctx: &Daemon
         let reg = agent::lock_registry(&ctx.registry);
         match reg.get(&instance_id) {
             Some(handle) => {
+                // #1913: an INTENTIONAL delete must not be mistaken for a crash.
+                // `delete_transaction` STORES `handle.deleted = true` (lifecycle.rs)
+                // BEFORE it kills the backend process — but it removes the registry
+                // and config entries only AFTER the kill + exit-wait. The kill is
+                // observed here as an exit classified `Crash`, so without this gate
+                // the crash-respawn loop races those removals and RESURRECTS the
+                // just-deleted instance: it re-spawns the process and re-creates
+                // `workspace/<name>`, which re-leaks the per-instance stores teardown
+                // just cleaned (the intermittent residual root of the #1902–#1909
+                // teardown class). Because `deleted` is Stored before the kill, this
+                // Acquire load reliably observes `true` by the time the exit lands,
+                // and once the registry entry IS removed `reg.get` returns `None`
+                // (the `None` arm below) — so this check covers exactly the racy
+                // window. Treat it as a clean exit: no respawn.
+                if handle.deleted.load(std::sync::atomic::Ordering::Acquire) {
+                    tracing::info!(
+                        agent = %crashed_name,
+                        "exit is an intentional delete (deleted flag set) — skipping crash-respawn"
+                    );
+                    return;
+                }
                 let mut core = handle.core.lock();
                 // #1701: decide the self-orch P0 BEFORE record_crash (which may
                 // itself stamp the crash cooldown on its recent>=2 path) — the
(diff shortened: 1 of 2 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head bfd8592 == PR head). A1 crash-respawn deleted-gate is race-free (deleted Stored SeqCst BEFORE the kill at lifecycle.rs:102/108, loaded Acquire at crash_respawn.rs:75 — causal chain Store→kill→exit→handler guarantees the load sees true; full window coverage: Some+deleted→skip / None→skip, no Some+deleted=false intentional-delete window, no TOCTOU). Real crashes (deleted=false) still respawn — the real_crash_still_respawns_1913 test asserts total_crashes==1, giving teeth against an over-blocking gate. (b) boot-stagger re-check uses fleet.yaml membership tight to the spawn, closing the wide stagger window the #1909 oracle actually hit; comment honestly scopes the deep-root spawn_agent chokepoint to #1915. +214/-0 pure additive, existing crash-recovery/spawn logic untouched. CI 6/6 green; dev reports #1909 25/25 deterministic. This is the real unblocker for #1909 (stops deleted-instance resurrection re-leaking teardown stores).

Issue #1919, opened 2026-06-09
Bug: Deployed team instances drift into Git worktrees on restart/reboot due to workspace directory mismatch


cheerc:
## Summary

When a team is deployed from a template (e.g. `dev-team`), instance working directories are created under a team subdirectory (e.g. `<home>/workspace/dev-team/dev-team-lead`). 
Upon restarting the instance (or daemon reboot), `resolve_auto_worktree` incorrectly identifies this subdirectory as an external Git repository and redirects it to a worktree under `<home>/worktrees/`, breaking `claude --continue` session resume.

---

## Root Cause Analysis

In `src/worktree.rs` [`resolve_auto_worktree`](file:///Users/cheerc/agend-terminal/src/worktree.rs#L326), the daemon tries to prevent default workspace directories from auto-worktreeing via this check:
```rust
    // #1858: the daemon-managed default workspace dir is never a legitimate
    // worktree source
    if *base_dir == crate::paths::workspace_dir(home).join(name) {
        return None;
    }
```

However, for a deployed team/template instance, the default workspace directory is created under the team's subdirectory:
`working_directory` = `<home>/workspace/<team_name>/<instance_name>`

For an instance named `dev-team-lead` in team `dev-team`:
* `base_dir` = `<home>/workspace/dev-team/dev-team-lead`
* `crate::paths::workspace_dir(home).join(name)` = `<home>/workspace/dev-team-lead`

These two paths do not match. The check fails, and the function falls through. Since the agent (or daemon bootstrap) initializes Git in the workspace folder (`git init`), `is_git_repo(base_dir)` evaluates to `true`. Because the instance has `source_repo` set in its template configuration, `wants_auto_worktree` returns `true`, and the daemon redirect is triggered, pushing the agent into a worktree.

---

## Proposed Fix

We can fix this by changing the exact-match default workspace check to a prefix check against the daemon's managed workspace directory (`workspace_dir(home)`). Since all daemon-managed workspace directories reside under `<home>/workspace/`, checking `starts_with` covers both single and team-nested default workspace directories:

```diff
diff --git a/src/worktree.rs b/src/worktree.rs
--- a/src/worktree.rs
+++ b/src/worktree.rs
@@ -336,3 +336,3 @@
     // git-init. Skip it regardless of is_git_repo so the decision can't drift
     // across launches.
-    if *base_dir == crate::paths::workspace_dir(home).join(name) {
[...]

PR #1920, opened 2026-06-09
fix(#1919): auto-worktree guard exact-match → prefix (team-nested workspace)

suzuke:
## #1919 — auto-worktree guard: exact-match → prefix (team-nested workspace)

`resolve_auto_worktree`'s #1858 guard skipped the daemon-managed default workspace
dir with an **exact** match (`*base_dir == workspace_dir(home).join(name)`). A
team-deploy nests the per-instance default one level deeper —
`<home>/workspace/<team>/<instance>` — which is `≠ <home>/workspace/<instance>`, so
the guard missed it. The git-init'd default then fell through to `is_git_repo` →
auto-worktree, which on restart pushed the deployed team member into a worktree and
**broke `claude --continue` session resume**.

**Fix (1 line):** prefix match — `if base_dir.starts_with(workspace_dir(home))`.
Everything under `<home>/workspace/` is daemon-managed default; a real user
working_directory is explicit-config'd *outside* it. This is the exact invariant
`agent_ops::cleanup_working_dir` already encodes (`starts_with(&workspaces)`, from
#1907) — the two gates now agree.

**Not over-broad:** the existing #1858 (b) test (an explicit real repo *outside*
`workspace/` as working_directory) still auto-worktrees — the prefix gate only
blocks the `workspace/` subtree, never a legit external repo.

**Test:** existing #1858 (b)/(c) stay green; added (d) team-nested
(`workspace/<team>/<instance>`, git-init'd, source_repo set → `resolve_auto_worktree`
returns None). Verified teeth: (d) FAILS on the pre-#1919 exact match, PASSES on the
fix.

### DIFF
--- a/src/worktree.rs
+++ b/src/worktree.rs
@@ -336,7 +336,15 @@ pub fn resolve_auto_worktree(
     // worktree source — it only becomes a "repo" via ensure_project_root's
     // git-init. Skip it regardless of is_git_repo so the decision can't drift
     // across launches.
-    if *base_dir == crate::paths::workspace_dir(home).join(name) {
+    // #1919: PREFIX match, not exact. A team-deploy nests the per-instance default
+    // under a team subdir (`<home>/workspace/<team>/<instance>`), which the old
+    // exact `== workspace_dir/<name>` check missed → the git-init'd default fell
+    // through to auto-worktree and broke `claude --continue` session resume on
+    // restart. Everything under `<home>/workspace/` is daemon-managed default (a
+    // real user working_directory is explicit-config'd OUTSIDE it) — the same
+    // invariant `agent_ops::cleanup_working_dir` already encodes via
+    // `starts_with(&workspaces)`.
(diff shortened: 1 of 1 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head 725be976 == PR head). The #1858 guard exact-match → starts_with(workspace_dir(home)) correctly brings team-deploy nested defaults (<home>/workspace/<team>/<instance>) under the no-auto-worktree rule, fixing the drift where exact-match missed nested dirs and auto-worktreed an empty git-init'd dir (breaking claude --continue session resume). NOT over-broad: the existing (b) test (explicit real repo OUTSIDE workspace/) still auto-worktrees (CI green). The (d) team-nested test has teeth (independently confirmed it FAILs on the pre-fix exact-match, PASSes on prefix). Aligns with the #1907 cleanup_working_dir gate which already uses starts_with(&workspaces) for the same daemon-managed-workspace invariant — the two gates now agree. CI 6/6 green. Thanks cheerc.

PR #1921, opened 2026-06-09
fix(#1915): deleting-set spawn chokepoint — deep-root resurrection guard

suzuke:
The deep-root completion of the spawn-vs-delete **resurrection class** (#1902–#1909): a spawn racing a delete re-creates `workspace/<name>` + a registry handle *after* teardown cleanup ran. The per-path `deleted`-flag checks (#1918 A1) only cover the window while the registry handle still exists; this adds a process-global **deleting-set** that survives handle removal and covers the whole delete — one mechanism subsuming the boot-stagger / crash-respawn / stage2 paths.

## Mechanism
`crate::agent::deleting`: a refcounted `(home, name)` set behind a parking_lot **leaf-lock** (never nested with the registry lock — #1492 deadlock class). Both spawn entry points consult `is_deleting` at their **TOP, before any side effect**:
- **`spawn_agent`** — before `build_command`/child-spawn. Covers crash-respawn-worker / stage2-restart / direct-spawn.
- **`spawn_and_register_agent`** — before **skills-install** (the actual boot-path workspace re-creation, which runs *before* `spawn_agent`).

The outermost-delete entries `mark_deleting` at their top and hold a RAII `DeletingGuard` until **all** teardown side-effects (incl. workspace cleanup) complete: `full_delete_instance` (MCP delete + TUI close) and app-mode `kill_agent`. `Drop` un-marks on **every** path — return, early `Err`, panic — so a name is never permanently stuck (a leaked mark = an un-re-creatable name).

## Scope — three complementary layers (kept by design)
The chokepoint is an **additional** layer; the #1918 narrow fixes stay as defense-in-depth covering different windows:
- **A1** (`handle.deleted` in `handle_crash_respawn`) — earliest gate, at the crash *decision*, before the respawn worker is even spawned.
- **(b)** (`instance_is_known` recheck in the boot loop) — covers the *post-delete-complete* edge (spawn lands after the delete fully finished; the bounded deleting-set can't cover that).
- **#1915 deleting-set** — covers the *in-flight teardown* window across the whole `full_delete_instance`, at both spawn entry points.

**A2** (`handle_stage2_restart`) is auto-covered by the `spawn_agent`-top check. **TIER-B**: B1 (supervisor pane-input notify) + B2 (recovery dispatcher) now filter out deleting handles; B3–B7 are a follow-up sweep.

## Verification (§3.9)
- deleting-set unit tests: mark/drop, refcount (clears on last guard), home-scoping, **panic-in-delete still un-marks**.
- `spawn_agent` **refuses** a mid-delete name (crash/stage2/direct) — bails before the insert.
- `spawn_and_register_agent` **skips** a mid-delete name (boot) **and re-creates the same name after the delete completes** (no-leak).
- `cargo fmt --check` ✓ · `clippy --all-targets --features tray -D warnings` ✓ · full suite (system git) **3806/3806**.

Refs #1915.

### DIFF
--- a/src/agent/deleting.rs
+++ b/src/agent/deleting.rs
@@ -1,0 +1,152 @@
+//! #1915: the delete-vs-spawn chokepoint — a process-global set of the
+//! `(home, name)` instances currently undergoing teardown.
+//!
+//! A spawn racing a delete RESURRECTS the instance: it re-creates
+//! `workspace/<name>` + a registry handle AFTER teardown cleanup already ran (the
+//! intermittent residual root of the #1902–#1909 teardown class — boot-stagger
+//! spawn, crash-respawn worker, stage2-restart). The per-path `deleted`-flag
+//! checks (#1918 A1) only cover the window while the registry handle still
+//! exists; this set survives handle removal and covers the whole delete.
+//!
+//! `spawn_agent` and `spawn_and_register_agent` consult [`is_deleting`] at their
+//! TOP — before ANY side effect (skills-install, `build_command`, child spawn) —
+//! and refuse to spawn a name that is mid-delete. The outermost delete entries
+//! (`full_delete_instance`, app-mode `kill_agent`, the daemon replace paths)
+//! call [`mark_deleting`] at their top and hold the returned [`DeletingGuard`]
+//! until ALL teardown side-effects (incl. workspace cleanup) complete.
+//!
+//! **Lock discipline (#1492 class):** the set is a LEAF lock — acquired
+//! standalone, NEVER nested inside the registry lock. The two spawn checks run at
+//! function top, before any `registry.lock()`; the delete marks run before the
+//! delete touches the registry. So the deleting-set lock and the registry lock
+//! are never held simultaneously → no lock-order deadlock.
+//!
+//! **No-leak guarantee:** [`DeletingGuard`]'s `Drop` decrements a refcount and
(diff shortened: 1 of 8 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head cedbc73 == PR head). The deleting-set chokepoint is the load-bearing deep root fix for the resurrection class. Leaf-lock proven deadlock-free: spawn_agent checks is_deleting at the top (before build_command / any registry.lock), mark_deleting is a temp-lock (increment then drop), DeletingGuard holds the refcount key not the lock — so the deleting-lock is never nested with the registry lock (no ABBA / #1492-class deadlock). RAII proven no-name-leak: DeletingGuard::drop un-marks on every path including panic unwind (panic_in_delete_still_unmarks via catch_unwind), refcount clears only on the last guard, re-create after delete succeeds. mark sites are exactly the 2 delete entries (full_delete_instance + kill_agent); spawn-rollback and replace correctly excluded (grep empty) so re-create isn't wrongly blocked. Three resurrection paths blocked (crash/stage2 via spawn_agent-top, boot via spawn_and_register-top before skills-install), legit spawn unbroken. Three-window defense-in-depth with #1918 (b)+A1 (deleting-set=in-flight teardown, (b)=after-removal, A1=crash deleted-flag — non-redundant). CI 6/6 green, +292/-0 pure additive. Completes the resurrection root fix; #1918's targeted fixes stay as defense-in-depth.

Issue #1916, opened 2026-06-09
dispatch_idle watchdog uses stale target after task OwnerAssigned reassignment


cheerc:
## Bug

When a task is reassigned via `OwnerAssigned` event (operator changes assignee without status transition), the dispatch_idle sidecar still targets the **original** agent. The watchdog then fires `dispatch_idle_threshold_exceeded` to the **old** agent about a task they no longer own.

## Root Cause

1. Task dispatched to agent A → `dispatch_idle::record_dispatch` creates sidecar with `target=agent_A`
2. Operator reassigns task to agent B → `OwnerAssigned` event fires, task owner updated
3. dispatch_idle sidecar **still has `target=agent_A`** (no hook updates it on reassignment)
4. Watchdog (`scan_and_emit`) fires threshold event to `agent_A` about a task they don't own

## Evidence

**File: `src/daemon/dispatch_idle/mod.rs`**

- `record_dispatch` (line 183) creates sidecar with `target` from the original dispatch
- `cleanup_pending_for_task_id` (line 387) clears sidecars when task is **done/cancelled** only
- `cleanup_pending_for_instance` (line 429) clears sidecars when instance is **deleted** only
- **NO hook** for `OwnerAssigned` event to update sidecar target

**File: `src/tasks/handler.rs`**

- `OwnerAssigned` event emitted at line 670 when `new_assignee` is set
- No call to dispatch_idle to update/clear sidecars for the task

**File: `src/task_events.rs`**

- `OwnerAssigned` event (line 313) carries `task_id`, `owner` (new), `routed_to`
- This event is NOT handled by dispatch_idle module

## Impact

- Old agent receives spurious `dispatch_idle_threshold_exceeded` nudge
- Task owner (new agent) doesn't get the nudge they should
- Watchdog threshold accuracy degraded

## Reproduction

1. Dispatch task to agent A (kind=task, target=agent_A, expect_reply_within_secs=60)
2. Before agent A replies, operator reassigns task to agent B (`update` action with assignee=agent_B)
3. Wait for threshold to exceed
4. Observe: agent_A gets nudge, agent_B gets nothing

## Suggested Fix

Add a dispatch_idle hook on `OwnerAssigned` event:

```rust
// In dispatch_idle module
pub(crate) fn update_target_for_task_id(home: &Path, task_id: &str, new_target: &str) -> usize {
    // Update target field of all sidecars with matching correlation_id
}
```

[...]

PR #1922, opened 2026-06-09
fix(#1916): dispatch_idle OwnerAssigned hook — retarget sidecar on task reassign

suzuke:
## #1916 — dispatch_idle OwnerAssigned hook (retarget sidecar on task reassign)

dispatch_idle cleaned up its sidecars on task done/cancelled and on instance-delete,
but had **no hook for `OwnerAssigned`** (task reassignment). So after a task was
reassigned A→B, the pending-dispatch sidecar still targeted A → the idle watchdog
kept nudging the **former owner** about a task they no longer held.

### Fix
One helper `dispatch_idle::reassign_pending_for_task(home, task_id, new_owner)`,
called at the two `OwnerAssigned` emit sites (verified exhaustive — only these two):
- **`handle_update`** (assignee change, `owner: Some`) → `Some(new_owner)`
- **`orphan_tasks_for_owner`** (#1903 disband/delete orphan, `owner: None`) → `None`

Behavior:
- **`Some(new_owner)` → RETARGET in place**: same `dispatch_id` file, `target =
  new_owner`, and **RESET the idle clock** (`issued_at = now`, `not_working_streak
  = 0`, revive `Exceeded → Pending`, clear `nudge_sent_at`). The new owner just took
  over — inheriting A's near-threshold clock would nudge B immediately (the #1866
  principle: nudge reflects the *current* owner's idle time, not an inherited age).
- **`None` (orphan) → CLEAR** the sidecar (delegates to `cleanup_pending_for_task_id`)
  — there is no owner to nudge; never leaves `target=None`.

### Why call-site (not a central reducer / not inside append)
There is no central side-effect reducer (`handler.rs`'s `OwnerAssigned` match is a
display formatter). The codebase convention is call-site side effects — the existing
`cleanup_pending_for_task_id` is at `handle_update`/`handle_done`, keeping
`task_events::append` free of a dispatch_idle dependency. This hook follows that.

### Dedup-key safety
`record_dispatch`'s dedup key is `(dispatcher, target, correlation_id)` — target-based.
In-place RMW (same `dispatch_id` file) is safe: no duplicate (same file), no orphan
(file preserved + retargeted); the key recomputes correctly, so a later re-dispatch
to the new owner refreshes THIS sidecar rather than duplicating it.

### Tests (§3.9)
- helper: reassign A→B retargets + resets the clock (target=B, Pending, streak=0,
  issued_at recent); orphan(None) clears the sidecar.
- **wiring (real entry point):** `task update(assignee=B)` through `handle()` retargets
  the sidecar — verified teeth: FAILS without the `handle_update` hook, PASSES with it
  (an injected-input helper test alone wouldn't prove the wiring reaches the hook).
- done/cancelled cleanup unchanged (existing tests stay green).

### DIFF
--- a/src/daemon/dispatch_idle/mod.rs
+++ b/src/daemon/dispatch_idle/mod.rs
@@ -419,6 +419,67 @@ pub(crate) fn cleanup_pending_for_task_id(home: &Path, task_id: &str) -> usize {
     count
 }
 
+/// #1916: a task REASSIGN (`OwnerAssigned`) must move the dispatch-idle sidecar to
+/// the new owner — otherwise the watchdog keeps nudging the FORMER owner about a
+/// task they no longer hold. Call-site hook (mirrors `cleanup_pending_for_task_id`;
+/// keeps `task_events` free of a dispatch_idle dependency).
+///
+/// - `Some(new_owner)`: RETARGET every sidecar for `task_id` IN PLACE (same
+///   `dispatch_id` file) → `target = new_owner`, and RESET the idle clock
+///   (`issued_at = now`, `not_working_streak = 0`, revive `Exceeded` → `Pending`,
+///   clear `nudge_sent_at`). The new owner just took over; inheriting the prior
+///   owner's near-threshold clock would nudge them immediately (#1866 principle:
+///   the nudge must reflect the CURRENT owner's idle time, not an inherited age).
+/// - `None` (orphan / #1903 disband-orphan): CLEAR the sidecars — there is no owner
+///   to nudge. Delegates to [`cleanup_pending_for_task_id`] (never sets `target=None`).
+///
+/// In-place RMW preserves the `dispatch_id` file identity, so `record_dispatch`'s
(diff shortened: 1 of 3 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head 4caa0c62 == PR head). reassign_pending_for_task hooked at both OwnerAssigned emit sites (handler.rs:670 Some / orphan.rs:52 None — git grep confirms exactly 2). Some(new_owner) → in-place retarget (same dispatch_id file A→B, dedup-key recomputes correctly so a future re-dispatch to B refreshes THIS file not a duplicate, no leftover target=A orphan, already-B skip) + timer reset (fresh issued_at/streak=0/Exceeded→Pending/nudge_sent_at=None so B doesn't inherit A's near-threshold clock, per #1866). None → clear (cleanup_pending, never target=None). The wiring test drives the real crate::tasks::handle(update, assignee) entry (removing the handle_update hook makes it red — proves wiring, not an injected helper). done/cancelled cleanup gated on assignee-change, no regression. CI 6/6 green. Thanks cheerc.

Issue #1923, opened 2026-06-09
Correlation-tracking accuracy class (audit): unified owner-assign refresh (next_after_ci/dispatch_tracking/dispatch_idle) + in-memory cleanup-on-delete sweep


suzuke:
Systematic audit (fixup-reviewer-3) of all 10+ watchdog/sidecar mechanisms for 3 correlation-accuracy properties: tracks-right-state, clears-on-right-signal, targets-right-agent. #1916 (dispatch_idle stale target) + #1888 (re-nudge read-state coupling) were the seeds; the audit found the rest of the class.

## ⭐ Cross-cutting root (the headline)
THREE mechanisms freeze the target at dispatch time and never update it on task reassignment (OwnerAssigned): next_after_ci (G1), dispatch_tracking.to (G3), dispatch_idle.target (#1916, fixed in #1922). A SINGLE owner-assign refresh hook (extending #1922's reassign_pending_for_task) that also refreshes next_after_ci + dispatch_tracking would close this sub-class.

## HIGH
- **G1 — ci_watch next_after_ci write-once** (poller.rs:1719-1720, watch_state.rs:100): next_after_ci is set at dispatch, never updated. Task reassigned to a new reviewer → CI-pass [ci-ready-for-action] routes to the OLD reviewer (new one never notified), and handoff_timeout_watchdog re-nudges the same stale target. Cross-module, compounds with re-nudge. Hermetic-testable.

## MEDIUM
- **G2 — dispatch_idle ghost dispatcher** (dispatch_idle/mod.rs:876-889): stale_sidecar_reason checks target_in_fleet but never dispatcher_in_fleet → a deleted dispatcher gets dispatch_idle_threshold_exceeded to a ghost inbox.
- **G3 — dispatch_tracking sweep_stuck stale entry.to** (dispatch_tracking.rs:1401-1408 + :45-53): reassign inserts a NEW DispatchEntry alongside the stale one (track_dispatch pushes without dedup → multiple entries per task_id); sweep_stuck fires stuck-check at the old assignee for up to 24h.
- **G4 — supervisor notify_tracks never retained on delete** (supervisor.rs:257; cf retry_tracks .retain at :1824-1827): name-keyed cooldown map, same-name redeploy inherits old NOTIFY_COOLDOWN → first notification suppressed. (Same class as #1906 usage_limit_notify.)
[...]

PR #1924, opened 2026-06-09
fix(#1923): correlation-tracking PR-A — keying (G8) + in-memory-stale cleanup (G4/G5/G12)

suzuke:
**PR-A** of the correlation-tracking class (reviewer-3 audit #1923) — the **zero-dependency** findings. (PR-B = G1/G3/G2, the owner-assign reassign hooks, builds on the now-merged #1922 — coming next.)

## G8 — correlation-key mismatch (`messaging.rs`)
The dispatch-idle sidecar is **recorded** under `correlation_id.or(task_id)` but a `kind=update/query` reply **refreshed** it under `correlation_id` only. So a reply carrying just `task_id` (no explicit correlation_id) never refreshed its sidecar → the dispatch-idle watchdog fired a **false idle nudge** despite the reply arriving. Fix: align the refresh key with the write side (`correlation_id.or(task_id)`) — a **superset**, behaviour unchanged when correlation_id is set. (Likely explains the spurious dispatch_idle nudges seen this session.)

## G4 (notify_tracks) + G5 (conflict_notify) — in-memory-stale-on-redeploy
These per-agent maps live in the supervisor `run_loop`, **not reachable from the cross-thread `full_delete_instance`** delete path — so the cleanup is a **per-tick `.retain(live agents)` sweep**, mirroring the existing #1470 `retry_tracks` sweep. A deleted agent leaves the registry → drops out of `live_agents` → its entries prune next tick. Without it a same-name redeploy inherits the dead instance's crash-notify dedup (false-**suppress**) or conflict-observation/escalation timestamps (false-**fire**/suppress). Adds `ConflictNotifyTracker::retain_active`.

## G12 — escalation rehydrate (verify-only)
`full_delete_instance` **already** removes the escalation-persist store, so a same-name redeploy can't rehydrate the dead instance's escalation state. Adds a regression test pinning that; **no prod change**.

## Verification
- G5 `retain_active` prune unit test + G12 delete-removes-escalation-store regression test.
- `cargo fmt --check` ✓ · `clippy --all-targets --features tray -D warnings` ✓ · full suite (system git) **3811/3811**.
- Based on post-#1922 main (84f3f1f).

Refs #1923. (LOW G6–G11 details now in hand from reviewer-3; a final-sweep decision to follow.)

### DIFF
--- a/src/api/handlers/messaging.rs
+++ b/src/api/handlers/messaging.rs
@@ -535,7 +535,14 @@ fn track_dispatch(
             }
         }
     } else if matches!(kind_str, "update" | "query") {
-        if let Some(corr) = msg.correlation_id.as_deref() {
+        // #1923 G8: key the dispatch-idle refresh by the SAME correlation as the
+        // WRITE side above (~:496 records the pending-dispatch sidecar under
+        // `correlation_id.or(task_id)`). A reply carrying only `task_id` (no
+        // explicit `correlation_id`) was refreshed under `correlation_id` (None)
+        // → its sidecar never got refreshed → the dispatch-idle watchdog fired a
+        // FALSE idle nudge despite the reply arriving. Aligning the key is a
+        // superset — behaviour is unchanged when `correlation_id` is set.
+        if let Some(corr) = msg.correlation_id.as_deref().or(msg.task_id.as_deref()) {
             let _ = crate::daemon::dispatch_idle::refresh_issued_at(home, corr);
         }
     }
--- a/src/daemon/conflict_notify.rs
+++ b/src/daemon/conflict_notify.rs
@@ -68,6 +68,20 @@ pub(crate) struct ConflictNotifyTracker {
 }
 
 impl ConflictNotifyTracker {
+    /// #1923 G5: prune per-agent state for agents no longer in the registry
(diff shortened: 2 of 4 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head e1eb1190 == PR head). Diff hygiene confirmed clean: exactly 4 intended files (messaging/conflict_notify/supervisor/instance_lifecycle tests), no #1921/#1922 files touched, reverse-delete grep empty — the stale-worktree (pre-#1922) restore succeeded, no accidental revert of the resurrection chokepoint or reassign hook (+105/-1, the -1 being the G8 key change). G8 aligns the dispatch_idle refresh key to the write-side correlation_id.or(task_id) (strict superset: correlation_id-present unchanged, only adds refresh for task_id-only replies that previously never refreshed → false idle-nudge despite a reply — likely the root of this session's spurious nudges). G4/G5 per-tick .retain(live_agents) prunes deleted-agent tracker state (source correct, registry lock released before retain, no nesting; run_loop-local maps so per-tick IS their cleanup-on-delete, mirroring #1470 retry_tracks). G12 verify-only proves the existing escalation cleanup covers stale-rehydrate. CI 6/6 green. (#1923 PR-A of 2; PR-B G1/G3/G2 + LOW sweep remain.)

PR #1931, opened 2026-06-09
ci(teardown): daemon-boot flake gate (20x, path-gated) + drift-guard invariant

suzuke:
## teardown-determinism: daemon-boot flake gate + drift-guard

The resurrection-root fixes (#1918 + #1921) made the **entire daemon-boot test class
deterministic** — verified locally under `nextest --profile ci`:
- teardown_completeness_regression + e2e_workflow: **50/50** clean
- boot-heavy set (harness_smoke, restart_smoke, self_respawn_handoff,
  ready_marker_invariants, migrated_scripts, vte_gotchas): **30/30** clean

This PR adds the **gate that keeps it that way** — the daemon-boot path's race (#1909)
bit #1914 as an intermittent main red, and a reviewer can't catch a low-probability
boot flake by running a worktree once.

### Three pieces (lead-approved design)
1. **`tests/daemon_boot_gate_filter.txt`** — single source of truth: the daemon-boot
   test binaries (AgendHarness users + the direct-boot ones).
2. **`.github/workflows/daemon-boot-flake-gate.yml`** — a **path-gated** workflow
   (only runs when `src/agent|daemon|bootstrap`, the harness, or the gated test files
   change — docs/unrelated PRs pay nothing) that builds the nextest filter from the
   file and repeats the set **20×** under the `ci` profile, failing on ANY failure.
   nextest 0.9.137 has no native repeat (only `--retries`, which *masks* flakes), so
   the repeat is an explicit exit-code loop.
3. **`tests/daemon_boot_gate_invariant.rs`** — drift-guard: every `AgendHarness`-using
   test binary MUST be in the filter file, else the invariant goes RED (the
   #1907/#1911 curated-list-drift class — a new daemon-boot test can't silently
   escape the gate). Teeth-verified: removing a binary from the file fails the
   invariant, naming it.

### Caveat for the operator (branch-protection)
A `paths:`-filtered workflow does **not** trigger on PRs that touch none of those
paths, so this gate is **advisory by default** — do NOT add it to required status
checks as-is, or path-irrelevant PRs would block on a never-reported check (the known
GH Actions "path-filtered required check stays pending" gotcha). If you want it
required, it needs the skipped→success shim. Left non-required intentionally.

### DIFF
--- a/.github/workflows/daemon-boot-flake-gate.yml
+++ b/.github/workflows/daemon-boot-flake-gate.yml
@@ -1,0 +1,79 @@
+# Daemon-boot flake gate (#t-teardown-determinism).
+#
+# The daemon-booting test class (AgendHarness + `start --foreground`) shares the
+# boot path whose race (#1909) bit #1914 as an intermittent CI red. A reviewer
+# can't catch a low-probability boot flake by running a worktree once, so this
+# workflow repeats the daemon-boot subset 20x under the race-exposing `ci` nextest
+# profile and fails on ANY failure — catching a boot-race regression PRE-MERGE.
+#
+# Cost control: PATH-GATED — only runs when daemon-boot-relevant code/tests change
+# (the daemon boot path + the harness + the gated test files), so docs/unrelated
+# PRs pay nothing. The test SET is the single-source list in
+# tests/daemon_boot_gate_filter.txt (drift-guarded by
+# tests/daemon_boot_gate_invariant.rs). nextest 0.9.137 has no native repeat (only
+# `--retries`, which MASKS flakes), so the repeat is an explicit exit-code loop.
+name: Daemon-boot flake gate
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'src/agent/**'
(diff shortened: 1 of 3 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head c5441662 == PR head). The Daemon-boot flake gate (20x) check passed (20m33s) ON #1931 itself (it touches gated paths) — e2e validation that the gate mechanism actually works in CI (filter-from-file → 20x nextest --profile ci exit-code loop → red on any failure; not --retries which would mask flake), and the daemon-boot class is deterministic in CI too (20/20). filter.txt is the single source of truth read by both the workflow and the drift-guard invariant. The invariant has teeth (a new AgendHarness user not listed → red naming it; self-skip + stale-entry guard correct). path-gating means docs/unrelated PRs pay zero cost — which is exactly why it must stay non-required (a path-filtered check set as required would hang unrelated PRs, the GH Actions gotcha dev-2 flagged). +191/-0 pure additive, no prod code. NOTE for operator: do NOT add this gate to branch-protection required checks.

PR #1936, opened 2026-06-10
fix(#1809): suppress cross-cycle phantom ServerRateLimit re-latch

suzuke:
cheerc's #1809 retry storm: an agent recovers from a `ServerRateLimit` but outputs little, so the **stale error line stays in the bottom-N error-scan window**. `#badge-recovery` (`recovered_within(45s)`) suppresses it for the first 45s; once that expires, the next screen change (a CLI status-bar **clock tick** flips the screen hash) re-scans, re-matches the SAME old error, and re-latches SRL → the supervisor schedules another auto-retry → a blind `continue` inject. The window re-arms on that nudge's output → the loop repeats every ~45s (cheerc's phantom-retry-every-~4min trace).

## The fix — telemetry → behavior, 2 lines
The existing `#1808-probe0-phantom` instrumentation already computed everything: it signatures the matched error line `(line_hash, dist_from_bottom)`, persists `last_srl_match_sig` across Idle, and flags a `cross_cycle` re-fire (same signature with an intervening non-SRL state). This turns that detection into behavior: when the SRL would latch (`landed == SRL`) with no recent productive output AND it's a `cross_cycle` phantom → **land Idle instead of re-latching**. (`let mut landed` + the override.)

## Conservatism (this is the SRL-recovery lifeline path)
- **Only `cross_cycle`** is suppressed (agent left SRL → same stale error re-grabbed = the unambiguous phantom). The in-place `consecutive_rematch` case (still in SRL, never recovered) is **left to latch** — a genuine long throttle must keep retrying.
- **"New output" gate is free**: the signature is the error LINE's hash (excludes the clock elsewhere on screen), and `recovered_within` reads `last_productive_output`, already tuned (#685 PR-2 RC1) to ignore cursor-blink/clock-tick noise.
- **#badge-recovery semantics untouched** — this is an additional stale-ignore layer only.
- **Accepted narrow FP**: a genuinely-NEW SRL identical in text AND screen position to the just-processed one, with no intervening productive output, is ignored. A real new error almost always has new text/position.

## Tests (§3.9)
- `stale_srl_cross_cycle_rematch_does_not_relatch_1809` — the storm case → no re-latch.
- `new_srl_error_after_recovery_still_latches_1809` — different signature → still latches (no false-kill).
- `in_place_srl_rematch_still_latches_1809` — still-SRL genuine throttle → keeps latching (no over-suppression).

`fmt --check` ✓ · `clippy --all-targets -D warnings` ✓ · full suite (system git) **3820/3820** (incl. all 292 state-detection tests, zero regression).

**#1808** (abort-to-Idle freeze) stays open — to be re-assessed after this root fix lands (per the evaluation, it likely shrinks to a narrow case). Refs #1809.

### DIFF
--- a/src/state/mod.rs
+++ b/src/state/mod.rs
@@ -1114,7 +1114,7 @@ impl StateTracker {
                         // stays OUT of `is_high_fp_state` → the #1450 red anchor is
                         // unchanged (that decision is ① Step-2, pending the
                         // operator); this is the #1768 working-marker override only.
-                        let landed = if high_fp || matches!(detected, AgentState::UsageLimit) {
+                        let mut landed = if high_fp || matches!(detected, AgentState::UsageLimit) {
                             // #badge-recovery (state-level mirror of #1795): a
                             // ServerRateLimit whose agent produced PRODUCTIVE output
                             // within the recovery window has recovered even though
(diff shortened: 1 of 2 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head 30325ca9 == PR head). The crux — only cross_cycle phantoms are suppressed, genuine throttles are not — was line-traced conclusively: cross_cycle requires non_srl_since_last_srl (set on every non-SRL landed state at :1345, reset on SRL detection at :1248), so an in-place rematch (still SRL, no intervening non-SRL) has cross_cycle=false and keeps retrying — and the in_place_srl_rematch_still_latches_1809 test pins exactly that (goes red if the fix ever over-blocks a genuine long throttle). New errors latch (different line-hash sig); #badge-recovery 45s semantics untouched (the fix is an additional stale-ignore layer in the !recovered path only); clock/spinner ticks can't fake either side (sig hashes the error line, productive-output uses the #685 matched-substring dedup). Accepted narrow FP (identical text+position new SRL across cycles) documented. CI 6/6 green incl. 292 state-detection tests.

Issue #922, opened 2026-05-18
Windows CLI smoke race: 'shell agent missing from list' iter 1 — api.port vs registry timing


suzuke:
## Pain
Windows CI smoke test (\`Check (windows-latest) / CLI smoke (Phase C.5)\`) intermittently fails with \"shell agent missing from list\" after \"daemon ready (iter 1)\". Race window: \`api.port\` is published in ~5ms (per #896 fix #908), but smoke immediately polls \`list\` within ~30ms of \`api.port\` appearance. The agent registry may not yet contain the shell agent because \`spawn_and_register_agent\` happens AFTER \`api.port\` write in the daemon main loop (per PR #906 reorder).

## Concrete observation
- PR #920 (#914 daemon.log rotation) windows CI fail 2026-05-18 ~17:09
- Same fail on origin/main \`fd2ea188\` ~16:53 (unrelated PR)
- Pattern: smoke races daemon's spawn_and_register_agent for the FIRST agent

## Affected paths
\`tests/smoke/\` or wherever the Windows CLI smoke harness lives. Smoke polls \`list\` after detecting \`api.port\`; daemon needs additional time for first agent registration.

## Proposed fix candidates (open for spike)
1. Smoke harness: poll \`list\` with retry budget (e.g., 3× retries with 100ms backoff) instead of one-shot fail.
2. Daemon: emit a \"ready\" signal AFTER first agent registers, not just api.port written. Smoke polls for the signal.
3. CLI \`list\`: wait briefly (~500ms) for registry to populate when called within X ms of api.port-published.

## Suggested
(1) cheapest — smoke harness retry budget. Daemon-side signal is more invasive but cleaner.

## Out of scope
- Underlying race between \`api.port\` write and first \`spawn_and_register_agent\` completion — addressed in #908. The race here is between api.port and external prober, not internal sync.

## Origin
- Surfaced by fixup-dev during #914 (PR #920) CI iteration, evidenced by parallel main-branch fail (msg \`m-20260518172915152268-241\`)
- Triage: pre-existing flake, NOT #914 regression

PR #1938, opened 2026-06-10
fix(#1935): close run/<pid>/.port teardown residual race (3 layers)

### LINKED ISSUE #1935: Teardown residual: run/<pid>/<name>.port not cleaned by full_delete (caught by the #1931 flake-gate, escaped the #1907 oracle)
The #1931 daemon-boot flake-gate caught this on #1865's run (16/20 — textbook intermittent): teardown_completeness_regression failed with a residual run/<pid>/victim-agent.port. The per-agent .port file written by the #908 TUI thread into the daemon-lifecycle run-dir (run/<pid>/) is NOT cleaned by full_delete_instance — a different dir from runtime/<name>/, and the #1907 whole-home-scan oracle didn't cover it (or races it).

This is the teardown-residual class (#1902/#1903/#1906/#1907) — one more store. Intermittency suggests a write-vs-delete race (the TUI thread may write the .port after cleanup scans), so verify-first: RCA whether it's (a) simply missing from full_delete's cleanup steps (add it + oracle coverage), or (b) a race where the port file is re-written after cleanup (then the writer needs a deleted/deleting guard — the #1913/#1921 class).

Also explain why the #1907 oracle misses it today (scan exclusion? allowlist? timing?) and make the oracle catch it.

Found by: flake-gate run on #1865 (proof the #1931 investment pays — pre-existing main flake caught pre-merge on an unrelated PR). Assignee context: fixup-dev-2 (built both the oracle and the gate).

suzuke:
The #1931 daemon-boot flake-gate's **first catch**: `teardown_completeness_regression` flaked ~1-in-20 on a residual `run/<pid>/<name>.port`. Verify-first RCA (4 code-traced facts + the 1-in-20 rate).

## Root cause
- The per-agent TUI port file is written **once at boot-spawn** (`tui_bridge.rs::write_port` → `run/<pid>/<name>.port`, the daemon-**lifecycle** dir). full_delete *does* remove it (`api::call(DELETE)` → agent stop → `ipc::remove_port`), so it's **not a missed-cleanup** — it's a **(b) writer-vs-teardown race**.
- The test triggered it because `AgendHarness::spawn_with` waited for **`api.port`** (published BEFORE the spawn loop, #922) not **`.ready`** (after it) → it seeded + deleted a victim while the victim's boot-spawn was still publishing → `write_port` landed *after* `remove_port` → residual.
- **Why #1907's oracle missed it**: `name_residual_anywhere` scans `runtime/<name>/` (binding dir), never `run/<pid>/` (lifecycle dir — different family). The whole-home scan test *should* catch it but was flaky (race only sometimes loses), so it passed when #1907 shipped.

## Fix (three layers, defense-in-depth — per lead scope)
1. **`AgendHarness::spawn_with` waits for `.ready`** (boot-spawn published) — closes the test-side overlap. `.ready` is written **unconditionally** after the spawn loop (`daemon/mod.rs:1095` — empty fleet still reaches it, 0 iterations); failed boots early-exit via the existing `try_wait`.
2. **`name_residual_anywhere` scans `run/<pid>/<name>.port`** + full_delete explicitly `remove_port`s it before the audit → the production delete is now **loud** if it lingers (closes the oracle blind spot).
3. **`prepare_tui_listener_and_publish_port` refuses `write_port` while the instance is mid-delete** (`is_deleting` leaf-lock read; #1921 writer-respects-teardown) — closes the publish-after-remove window for a real operator deleting a **still-booting** instance (#1913 class), not just the test.

## Verification
- **`teardown_completeness_regression` 20× → 0 flake** (was ~1-in-20).
- **All 8 daemon-boot binaries 10× → 0 flake** (the harness `.ready` change touches all of them; `ready_marker_invariants`/`restart_smoke`/`self_respawn_handoff` included).
- **`publish_port_respects_deleting_guard`** §3.9: mid-delete → `.port` NOT written + publish refused; normal → `.port` written.
- Empty-fleet `harness_smoke` + early-exit (bad-yaml) cases unaffected (still fast).
- fmt + clippy `-D warnings` clean. This PR path-triggers the flake-gate = a natural CI 20× re-verification.

Closes #1935.

### DIFF
--- a/src/daemon/tui_bridge.rs
+++ b/src/daemon/tui_bridge.rs
@@ -42,6 +42,22 @@ pub(crate) fn prepare_tui_listener_and_publish_port(
     })?;
     let listener = crate::ipc::bind_loopback()?;
     let port = crate::ipc::local_port(&listener);
+    // #1935: refuse the port publish if the instance is mid-delete. `full_delete`
+    // → `remove_port` deletes `run/<pid>/<name>.port`, but a boot-spawn / respawn
+    // publish still in flight would re-create it AFTER the removal (the #1913
+    // writer-vs-teardown race that left a residual the #1907 oracle didn't catch).
+    // The #1915 DeletingGuard already refuses the spawn/register chokepoint; this
+    // closes the narrower window where write_port runs past it. Cheap leaf-lock
+    // read. `home` = run_dir's grandparent (run_dir is always `home/run/<pid>` via
+    // run_dir_for_pid), so the key matches full_delete's `mark_deleting(home, …)`.
+    if let Some(home) = run_dir.parent().and_then(|p| p.parent()) {
+        if crate::agent::deleting::is_deleting(home, name) {
+            return Err(std::io::Error::new(
+                std::io::ErrorKind::Interrupted,
+                format!("instance '{name}' is being deleted; TUI port publish skipped"),
+            ));
+        }
+    }
     crate::ipc::write_port(run_dir, name, port)?;
     tracing::info!(agent = name, port, "TUI socket ready");
     Ok(TuiListenerMeta { listener, cookie })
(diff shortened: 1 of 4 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head ee70b161 == PR head). Crux: the .ready wait can't hang — the marker is written unconditionally after the spawn loop (daemon/mod.rs:1095, reached even with an empty fleet), the harness early-exits on child death (try_wait) and has a 15s kill backstop; and the flake-gate ran 20x green ON THIS PR (all 8 daemon-boot binaries — the gate that caught this flake on day 2 now validates its fix). The publish deleting-guard derives home correctly (run/<pid> grandparent, same key as mark_deleting) and the boot caller rolls back gracefully on refusal. write_port / remove_port / oracle all share pub(crate) port_path (no drift). full_delete explicitly removes the port under the #1921 guard — both ends of the writer-vs-teardown race closed, and the oracle now makes any future run-dir residual loud. CI 6/6 + flake-gate green. Closes the loop on the teardown-residual class.

PR #1941, opened 2026-06-10
fix(#1939): restore restart/replace respawn pane to its original position

### LINKED ISSUE #1939: restart_instance respawns at default placement instead of preserving the original pane position (layout disruption)
## Symptom
Restarting an instance via the restart_instance MCP tool (either mode) respawns its pane at the DEFAULT placement instead of where the pane was. Observed 2026-06-10: two restarts of fixup-dev emitted InstanceCreated with layout=SameTab, spawner=None, target_pane=None — the original pane position was not captured, so the agent landed at the default spot AND the remaining panes in the tab reflowed (a sibling instance, fixup-dev-2, visually moved as a side effect). The operator noticed the layout disruption immediately.

```
09:32:42 INFO SPAWN emitting InstanceCreated agent="fixup-dev" layout=SameTab spawner=None target_pane=None
09:54:32 INFO SPAWN emitting InstanceCreated agent="fixup-dev" layout=SameTab spawner=None target_pane=None
```

## Expected
restart_instance is kill + respawn of the SAME instance — the operator's mental model is 'same agent, same place'. The respawned pane should reappear in (or as close as possible to) its original tab/split position, not at default placement.

## Direction (for the spike)
Before killing, capture the instance's current pane placement (tab index, split geometry / neighbor relation — whatever the TUI layout model exposes); pass it through the respawn path (the InstanceCreated event already carries layout/target_pane — currently hardcoded None on the restart path). Likely small: the restart path just never plumbed placement, while spawn-from-create supports layout/target_pane. Edge cases: original tab/neighbor gone by respawn time → fall back to current default; replace_instance probably has the same gap (verify + fix together if cheap).

## Context
Restarting agents to free a full context (97% after a long session) is a normal operation — it shouldn't cost the operator their layout. Found during the 2026-06-10 fixup-dev context-full restarts.

suzuke:
## Problem (#1939)
`restart_instance` / `replace_instance` respawn with `layout=SameTab`, but the #1431 memory records only the **tab name**. The respawned pane therefore lands best-effort:
- **Multi-pane tab**: `place_in_remembered_tab` = `split_focused(SplitDir::Horizontal)` — splits whatever pane happens to be focused, direction hardcoded, ratio default → within-tab slot/direction/ratio all lost. This is the operator-visible reflow from the 2026-06-10 fixup-dev restarts.
- **Single-pane tab**: the tab is closed on delete; respawn recreates it **appended at the end** → tab order lost.

The event chain itself is sound: DELETE emits `InstanceDeleted` synchronously before returning (`api/handlers/instance.rs:124-129`), so capture-before-remove has no race.

## Fix — TUI-side memory upgrade, daemon/API untouched
- `layout/tree.rs`: `parent_split_of()` captures the removed leaf's parent split (dir, ratio, side, sibling-subtree agents); `wrap_subtree_with_split()` restores by wrapping the **minimal subtree containing the surviving siblings** in the remembered split — nested sibling subtrees come back with their original shape, not a split against one leaf inside them.
- `layout/mod.rs`: `removed_tab_memory` (agent → tab name) → `removed_pane_memory` (agent → `RemovedPanePlacement{tab_name, tab_idx, split}`). `restore_removed_pane()` fallback chain: remembered split next to surviving siblings → focused split in the same tab (pre-#1939 behavior) → recreate the tab at its remembered index (`insert_tab`, clamped) instead of appending.
- `app/tui_events.rs`: `handle_instance_deleted` records the placement; the SameTab create branch restores it.

**replace_instance fixed for free** — it emits the same `layout=same-tab` with no anchor (`mcp/handlers/instance.rs:338`) and shares this TUI path.

**Headless safe by construction** — capture/restore live entirely in the TUI event loop; with no TUI attached `notifier=None` and `InstanceCreated`/`InstanceDeleted` are never emitted. No daemon-side behavior change.

## Tests (§3.9)
- `restore_2pane_exact_side_dir_ratio_1939` — exact slot: side + direction + ratio restored
- `restore_nested_sibling_subtree_shape_1939` — sibling subtree wrapped whole; original tree shape back
- `restore_recreates_closed_tab_at_original_index_1939` — single-pane tab recreated in place, order restored
- `restore_falls_back_when_siblings_gone_1939` — graceful fallback, no panic
- `restore_clamps_out_of_range_tab_idx_1939` — clamp, no panic
- #1431/#1625 replace/restart tests updated to the new API, contracts unchanged

`cargo fmt --check` ✓ · `clippy --all-targets --features tray -D warnings` ✓ · full suite **3526 passed / 0 failed** (bins) + integration tests green.

Closes #1939. Refs #1431, #1625.

### DIFF
--- a/src/app/tui_events.rs
+++ b/src/app/tui_events.rs
@@ -5,7 +5,7 @@
 //! the layout accordingly — auto-creating or removing tabs/panes.
 
 use crate::agent::{self, AgentRegistry};
-use crate::layout::{Layout, MovePlacement, Pane, SplitDir, Tab};
+use crate::layout::{Layout, MovePlacement, SplitDir, Tab};
 
 /// Events sent from the API server to the TUI event loop when agents or teams
 /// are created/deleted via MCP tools. The TUI reacts by auto-creating or
@@ -344,10 +344,11 @@ fn handle_instance_created(
             }
         }
         None => match hint {
-            // #1431: a replace_instance re-spawn returns to the tab the old
-            // same-named pane occupied (recorded on its removal).
-            LayoutHint::SameTab => match layout.take_removed_tab(name) {
-                Some(tab_name) => place_in_remembered_tab(layout, &tab_name, pane),
+            // #1431/#1939: a replace_instance / restart_instance re-spawn
+            // returns to the position the old same-named pane occupied
+            // (recorded on its removal).
+            LayoutHint::SameTab => match layout.take_removed_pane(name) {
+                Some(placement) => layout.restore_removed_pane(&placement, pane),
                 None => layout.add_tab(Tab::new(name.to_string(), pane)),
             },
             _ => layout.add_tab(Tab::new(name.to_string(), pane)),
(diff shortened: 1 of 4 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head df6bb45b == PR head). The fallback chain is panic-safe at every layer (position→Option, idx from position so no OOB, restore_split returns the pane when all sibling anchors are gone, insert_tab clamps idx.min(len), switch_active is store-only; the lone expect is the tab-root invariant, not reachable from extreme input) — with the extreme cases tested (siblings gone, tab gone, single-pane, out-of-range idx). wrap_subtree_with_split restores real tree shape (tests assert dir/ratio/side and that a sibling subtree stays a split, not flattened). #1431/#1625 tests are API-adapted with identical assertions (no weakening). Headless-safe (TUI event path only). Placement memory is take-and-remove — no accumulation, no stale placement for a same-name re-create. daemon/API untouched. CI 6/6 green. Live TUI behavior validates on next TUI restart (PR-own-dogfood).

Issue #842, opened 2026-05-16
Bridge ↔ daemon JSON-RPC: idempotent retry via request_id + server-side dedup cache


suzuke:
## Symptom

`agend-mcp-bridge` retries failed RPCs once on transport-level errors (`is_retriable_io` matches `BrokenPipe / ConnectionReset / ConnectionAborted / NotConnected / UnexpectedEof / WouldBlock / TimedOut` per `src/bin/agend-mcp-bridge.rs:208-220`). The retry transparently re-sends the same JSON envelope after reconnecting.

For idempotent ops this is fine. For ops with side-effects (`send`, `delegate_task`, channel injection, etc.), the retry can cause **double execution** when the daemon successfully processed the original request but the response delivery failed (slow channel IO + 30s read_timeout fires → bridge sees TimedOut → retry → daemon executes again).

PR #804 (cheerc fix/double-injection) surfaced this for telegram/discord channel inject path. Operator chose architectural fix (idempotent retry) over per-handler workaround (cheerc's spawn-and-forget approach).

## Root cause

JSON-RPC envelope at the bridge↔daemon boundary has no request identity. Each retry is a fresh request from daemon's perspective — no way to detect \"this is a re-send of an in-flight or completed request\".

### Wire-format change

Bridge generates a `request_id` (UUID v4) at envelope creation time, includes in JSON envelope:

```json
{\"method\": \"send\", \"params\": {...}, \"request_id\": \"550e8400-e29b-41d4-a716-446655440000\"}
```

Bridge retry uses the SAME request_id (the bridge's retry IS the same logical request, just re-transported).

### Daemon-side dedup cache

New module `src/api/request_dedup.rs`:

```rust
pub struct DedupCache {
    entries: Mutex<HashMap<String, (Instant, Value)>>,  // request_id → (completed_at, cached_response)
    ttl: Duration,  // default 10 min
}

impl DedupCache {
    pub fn check_or_register(&self, request_id: &str) -> CheckResult {
        // Returns one of:
        //   Cached(response) — request_id seen + completed; return cached response
        //   InProgress       — request_id currently processing; block or return retry-later
        //   Fresh            — first sight; register as in-progress
    }
    
    pub fn store_completion(&self, request_id: &str, response: Value) { ... }
    pub fn gc_expired(&self) { ... }  // periodic cleanup
}
```

Daemon API dispatch path (`src/api/mod.rs::serve`):

```rust
[...]

PR #1943, opened 2026-06-10
fix(#1940): cap inbox drain batch so a lost MCP response is recoverable

### LINKED ISSUE #1940: inbox drain marks messages read even when the MCP response times out — client loses the message content (mark-read-before-confirmed-delivery)
## Observed (2026-06-10, live fleet)
ai-scout received a kind=task dispatch. Its inbox drain call TIMED OUT on the MCP transport — the daemon had already marked the message read, but the client never received the drained content. The agent had to ask the sender to re-send ('my inbox tool timed out while draining and marked it as read before I could see its content'). Without the agent proactively asking, the task would have been silently stranded: read_at set → re-nudge/unread mechanisms blind (same shape as #1888), content never delivered.

## Mechanism (to verify)
inbox drain presumably: load unread → set read_at → write → return content in the MCP response. If the response leg times out (slow transport, large drain payload), the read-marking has already persisted → at-most-once delivery with loss. The #1888 class taught us read-state ≠ consumed; this is the delivery-side instance of the same lesson.

## Direction (spike)
Options to evaluate: (a) two-phase drain — return content first, mark read on a follow-up ack (heavier protocol change); (b) peek-then-ack variants; (c) pragmatic: make drain idempotent-recoverable — keep a short-lived 'last drain snapshot' per agent so a retry after timeout can re-fetch the same batch even though read_at is set; (d) mark-read-after-serialize — ensure the marking happens as late as possible and the payload is small enough not to time out (cap drain batch size, paginate). Lean KISS: (c)/(d) avoid protocol changes. Also worth checking what actually timed out (payload size? daemon load?) — if drains are paginated/capped the window may shrink to negligible.

## Severity
MED: silent task-content loss on a coordination-critical path; rare (first observed today) but the failure mode is silent-stranding. Found live; the agent's proactive re-ask was the only recovery.

suzuke:
ai-scout lost a `kind=task` dispatch: its inbox drain timed out on the MCP transport **after** the daemon marked the message read, and a re-drain returned empty (`read_at` set) — silent task-content loss (the #1888 "read-state ≠ consumed" class, delivery side).

## RCA (spike-verified, code-traced)
The recovery is **already in place**: the bridge (#842) retries a lost transport with the **same `request_id`**, and `request_dedup` returns the cached response. The only hole — an uncapped drain whose response exceeded the dedup per-entry cap (`PER_ENTRY_CAP_BYTES` = 64 KiB) was cached as **`Oversized`**, so the retry got a deterministic error and the already-`read_at`-set content was lost.

## Fix — (d) only: cap the drain batch
`drain()` now accumulates a batch up to `DRAIN_BATCH_BUDGET_BYTES` (48 KiB, comfortably under the 64 KiB dedup cap) so the MCP response is **always dedup-cacheable** → the bridge same-`request_id` retry always recovers it. The remainder stays **UNREAD** for the next drain. A message is never split; at least one is always returned (progress on a single >budget message).

## The snapshot route (spike option c) was implemented, then REJECTED
A per-agent `.draining` re-serve snapshot cannot distinguish a timeout-retry from a normal next poll without a client cursor, so it either **starves** new messages (re-serve within a TTL — a rapid drain loop never advances) or **double-delivers** (re-serve once / concurrent drains) — both break the inbox's **exactly-once** contract. Proven by the existing `test_enqueue_vs_drain_no_lost_msg` (rapid loop, exactly 10) + `test_concurrent_drain_no_duplicates` (concurrent, exactly 3). The pre-existing **zero-creator** `.draining` read/recovery dead code is removed with it (a tombstone comment records why).

## Accepted residual (validate-before-building)
Bridge-retry **exhaustion** (budget: **2 attempts × 120s read-timeout**, `agend-mcp-bridge.rs:329/430`) on a **non-oversized** response — i.e. a daemon wedged > 240 s — still loses content. A snapshot can't save a wedged daemon either. If scout-style loss recurs post-deploy, raise the bridge retry budget. No change to it in this PR.

## §3.9
- `drain_caps_batch_under_budget_leaving_remainder_unread`: backlog > budget → returned batch ≤ budget (dedup-cacheable) + remainder unread + next drain delivers the rest → total exactly N (no loss, no duplicate, no split).
- `drain_returns_single_oversized_message_alone`: a single >budget message is returned alone (never stranded).
- 9 pre-existing drain tests reconciled: 6 pass unchanged (consume-once preserved — no snapshot), 2 `.draining`-recovery tests deleted (tested removed dead code), 1 concurrent test repurposed to the live inbox.
- Full `cargo nextest run --features tray --profile ci`: **3822 passed, 0 failed**. fmt + clippy `-D warnings` clean.

Closes #1940.

### DIFF
--- a/src/inbox/storage.rs
+++ b/src/inbox/storage.rs
@@ -260,24 +260,39 @@ pub fn mark_ci_watch_superseded(
 ///
 /// Soft-delete semantics: messages stay in the JSONL file with `read_at`
 /// set; [`sweep_expired`] removes them later based on TTL rules.
+/// #1940: byte budget for one drain's returned batch — kept under
+/// `request_dedup::PER_ENTRY_CAP_BYTES` (64 KiB) so the response is always
+/// dedup-cacheable (never `Oversized`). That is the whole fix: the bridge (#842)
+/// retries a lost transport with the SAME `request_id`, and `request_dedup`
+/// returns the cached response — but only if it was cacheable. An uncapped drain
+/// that exceeded 64 KiB was cached as `Oversized`, so the retry got a
+/// deterministic error and the (already `read_at`-set) content was lost.
+pub(crate) const DRAIN_BATCH_BUDGET_BYTES: usize = 48 * 1024;
+
+/// Drain unread messages: mark `read_at` and write back.
 /// Uses atomic tmp+fsync+rename for crash safety.
+///
+/// #1940 (mark-read ≠ delivered — the #1888 class on the DELIVERY side): the MCP
+/// response can be lost AFTER drain() has persisted `read_at`. The recovery is
+/// the bridge's same-`request_id` retry + `request_dedup` cache, which is
+/// already in place; the ONLY hole was that an oversized (>64 KiB) response was
+/// cached as `Oversized` and could not be re-served. So (d): cap the returned
+/// batch under the dedup per-entry cap, leaving the remainder UNREAD for the next
+/// drain (a message is never split — at least one is always returned). A
+/// per-agent `.draining` re-serve snapshot was evaluated and REJECTED: it cannot
+/// distinguish a timeout-retry from a normal next poll without a client cursor,
+/// so it either starves new messages (re-serve within a TTL) or double-delivers
+/// (re-serve once / concurrent) — both break the inbox's exactly-once contract.
 pub fn drain(home: &Path, name: &str) -> Vec<InboxMessage> {
     let path = inbox_path_resolved(home, name);
 
-    if !path.exists() && !path.with_extension("draining").exists() {
+    if !path.exists() {
         return Vec::new();
     }
 
-    // Phase 1 (locked): read file, mark read_at, write back.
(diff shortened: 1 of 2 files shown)


suzuke:
reviewer-2 VERIFIED twice (adversarial; final reviewed_head f32ca44e == PR head). Production fix line-traced: the 48KB drain byte-cap keeps every response under the request-dedup 64KB cache cap so the bridge's same-request-id retry always recovers the content (closing the Oversized loss hole), with >=1-message progress, no message ever split, and the remainder left unread for the next drain (total preserved, no loss, no dup). The half-built .draining snapshot path was removed with a tombstone documenting why the snapshot design was rejected (no client cursor — cannot distinguish a timeout retry from the next poll, forcing starvation or double-delivery, both breaking exactly-once). The reviewer-flagged stale vacuous test was deleted in the follow-up commit; remaining .draining references are documentation tombstones only. The residual hole (bridge retries exhausted + daemon wedged >240s, non-Oversized) is documented and accepted — a snapshot wouldn't survive a wedged daemon either. CI 6/6 green.

PR #1945, opened 2026-06-10
feat: multi-backend context% detection + ≥80% orchestrator alert

suzuke:
Operator-directed (lead task t-20260610030712174426-9, spike approved): the daemon now knows each agent's context usage early and alerts the team orchestrator at ≥80%. **Detection + notification only — no automated handling.**

## Three honest layers
1. **`context_pattern` per backend** (`BackendProfile`): capture-group regex over the **bottom 6 status rows**, parsed in `StateTracker::feed_with_fg` right after the hash-dedup — screen-change only, zero new tick cost, no IO under the core lock. Patterns are **live-verified against real fleet panes** (2026-06-10 `pane_snapshot`):
   - Claude: `Ctx Used: 61.0%` statusline → `(?i)\bctx\s+used:\s*(\d+(?:\.\d+)?)\s*%` (fractional; the default-install left-form "Context left until auto-compact" has REMAINING semantics — deliberately v1-skipped, TODO'd)
   - kiro: `◔ 10%` footer → `◔\s*(\d+(?:\.\d+)?)\s*%`
   - codex / opencode / agy / shell: `None` (no usable display; codex would require editing the user's global `~/.codex/config.toml`)
   - Scope is deliberately NOT the 15-row error tail: agents routinely **discuss** context% in conversation (prose-FP). Truncated statusline (narrow pane) keeps the previous reading; readings expire after 10min (stale ≠ safe).
2. **Transcript-tail estimate** (Claude fallback): newest attributable transcript's **last** assistant message usage ≈ the context the next turn resumes from. Session-cumulative was evaluated and **rejected** — it grows across compacts and would latch a permanent false alert; the last-message estimate self-corrects after a compact. `[1m]` model ids → 1M window, others 200k. File IO runs in the per-tick handler with **no locks held** — never in the PTY feed path (lead's care item).
3. **Unknown stays unknown**: LIST gains `context_pct` + `context_source` (`pattern` | `transcript`), absent when nothing fresh is known → `list_instances`/TUI/health surfaces it automatically.

## Alerting
New `per_tick/context_alert.rs`, registered in the **shared `build_default_handlers`** (run_core + app mode both get it — #1002 wiring-drift class). Every ~6 ticks: resolve usage → on crossing ≥ threshold (80, `AGEND_CONTEXT_ALERT_PCT` override) notify `orchestrator_for(agent)` (never the agent itself; `FALLBACK_RECIPIENT` otherwise) via `inbox::notify_system`. Hysteresis: fire disarms; re-arm below threshold−5 (no 79↔81 flapping); continuously-high re-alerts every 30min. In-memory latch: a daemon restart re-fires once — accepted per spike review (single, self-limiting).

## Tests (§3.9)
- Live-capture pattern fixtures: claude statusline (incl. duplicate-frame last-wins + fractional), kiro footer
- **Prose-FP guard**: conversation mentions of context% must not read
- Truncation keeps previous reading; per-backend pattern presence pins; resolution order (pattern > transcript > unknown)
- Alert `decide()` state machine: cross/dedup/re-alert window/hysteresis wobble/compact re-arm
- Transcript estimate: **last-message-not-cumulative (compact 回落)**, 1M window, trailing non-usage lines skipped, foreign-cwd → None

Honest limits: the LIST json mapping itself has no dedicated integration test (needs a PTY-backed registry harness that doesn't exist) — it is a two-line map over the unit-tested `resolved_context()`. Codex rollout files only carry session-cumulative totals → codex stays `unknown` until a per-turn source is found.

`cargo fmt --check` ✓ · `clippy --all-targets --features tray -D warnings` ✓ · bins **3550 passed / 1 failed** — the 1 failure is `dispatch_branch_skips_metadata_stub_but_checks_out_real_source_1834`, a pre-existing **fleet git-shim environmental false-fail** (passes with system git: `PATH=/usr/bin cargo test ...` → ok; CI uses real git).

### DIFF
--- a/src/api/handlers/query.rs
+++ b/src/api/handlers/query.rs
@@ -13,7 +13,7 @@ pub(crate) fn handle_list(_params: &Value, ctx: &HandlerCtx) -> Value {
         .values()
         .map(|handle| {
             let name = handle.name.as_str();
-            let (agent_state, health_state, blocked_reason, blocked_note) = {
+            let (agent_state, health_state, blocked_reason, blocked_note, context) = {
                 let c = handle.core.lock();
                 (
                     c.state.get_state().display_name().to_string(),
@@ -23,6 +23,10 @@ pub(crate) fn handle_list(_params: &Value, ctx: &HandlerCtx) -> Value {
                     // blocked and its free-text annotation (previously internal-only).
                     c.health.current_reason.as_ref().map(|r| r.to_string()),
                     c.health.current_note.clone(),
+                    // Context% telemetry: resolved usage + producing source
+                    // ("pattern" = the agent's own statusline, "transcript" =
+                    // token-usage estimate). Absent = honestly unknown.
+                    c.state.resolved_context(),
                 )
             };
             let (dispatched_waiting_for, pending_response_to) =
@@ -36,6 +40,8 @@ pub(crate) fn handle_list(_params: &Value, ctx: &HandlerCtx) -> Value {
                 "health_state": health_state,
                 "blocked_reason": blocked_reason,
                 "blocked_note": blocked_note,
+                "context_pct": context.map(|(pct, _)| pct),
+                "context_source": context.map(|(_, source)| source),
                 "kind": "managed",
                 "dispatched_waiting_for": dispatched_waiting_for,
                 "pending_response_to": pending_response_to,
--- a/src/backend_profile.rs
+++ b/src/backend_profile.rs
@@ -38,7 +38,26 @@ pub struct BackendProfile {
     pub behavioral: BehavioralConfig,
     pub productivity: ProductivityConfig,
     pub initial_state: AgentState,
+    /// Context% telemetry: regex whose capture group 1 extracts the backend's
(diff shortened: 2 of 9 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head 6ee844a3 == PR head). Crux line-traced: the feed path (scan_context_pct) is regex-only over the in-memory screen tail — zero fs calls under the core lock; transcript IO is isolated to the per_tick handler's 3-phase flow (short-lock snapshot → file IO with NO locks held → evaluate), so a slow transcript read can never stall the PTY reader. The estimate uses the LAST assistant message's usage (falls back after compacts), not the rejected cumulative sum. Bottom-rows scan + statusline-specific patterns (Ctx Used: / ◔) are prose-FP-safe; hysteresis is a pure decide() fn (fire >=80, re-arm <75, 30min re-alert — no spam); alerts go to the orchestrator with the #1859 self-skip. Honest limits: codex/non-Claude stay unknown (no fake estimates). CI 6/6 green. Operator-directed multi-backend design.

Issue #1946, opened 2026-06-10
SRL recovery chain breaks when AUTO-retry budget exhausts during a sustained rate-limit — agent stuck at prompt until manual rescue (production-evidenced; completes #1808)


suzuke:
## Production evidence (2026-06-10, twice)
1. **08:59:20** — the #1808-flaw1-probe WARN fired in production: 'abort-to-Idle cleared an in-flight ServerRateLimit retry with no recent productive output — cheerc's claimed freeze path'. The deferred-pending-evidence condition on #1808 is met.
2. **~12:00** — operator-witnessed live: fixup-reviewer-2 hit a sustained server rate-limit while processing a ci-handoff. The AGEND-AUTO ratelimit-retry injected 'continue' TWICE; each continue triggered a retry that hit the still-active SRL and aborted again; after the AUTO budget exhausted the agent sat at an idle prompt indefinitely until the lead manually sent a task (which succeeded — the SRL had passed by then). The daemon log has near-zero trace of the whole episode (one drain mark-read line; no inject/retry/abort records) — observability gap.

## The full picture (why this class keeps recurring)
Each prior fix addressed one face: #1809/#1936 killed the phantom-retry storm; #1894/#1895 fixed notify dedup; the #1889 probe instrumented the clear-path. The REMAINING face: during a SUSTAINED rate limit, (a) AGEND-AUTO retries a fixed small number of times with no backoff — each immediate continue just re-hits the wall; (b) when its budget exhausts, the handoff SHOULD be the supervisor's ServerRateLimit retry track (the ~10s continue mechanism) — but the abort lands the agent in Idle and supervisor clears that track (#1808 Flaw 1, probe-confirmed) — so NOBODY owns recovery anymore; agent is stuck until a human notices.

## Fix direction (spike)
1. **Flaw 1 fix (cheerc's, now evidence-justified + storm-safe post-#1936)**: on abort-to-Idle with retry_count>0 && !recovered, do not just clear the track — queue a recovery nudge (or retain the track for a delayed retry). Narrow gate as designed in the #1808 re-evaluation.
2. **AUTO-retry backoff**: instead of N immediate continues during an active SRL, back off (e.g. exponential up to a few minutes) so retries land after the limit window; coordinate with (1) so exactly one mechanism owns the retry at any time (no double-continue).
[...]

PR #1950, opened 2026-06-10
fix(#1947): exclude input/user-message lines from the error content anchor

### LINKED ISSUE #1947: Error-detection scans raw screen text — operator-typed / quoted error strings trigger false SRL/ApiError latches (prose-FP)
## Observed (2026-06-10, operator-reproduced live)
The operator manually typed the SRL error string ('API Error: Server is temporarily limiting requests ... Rate limited') into an input line — and the daemon's error detection TRIGGERED on it (false ServerRateLimit latch / AUTO-retry behavior). 

## Root cause
Error detection scans the bottom ERROR_TAIL_SCAN_LINES (15) of the rendered vterm as raw text. It cannot distinguish:
- a REAL error block emitted by the agent CLI,
- the OPERATOR typing/quoting that string in the input line (bottom of screen — always inside the scan window),
- a LEAD/agent dispatch message QUOTING the error text (injected into the pane, e.g. incident discussions — today's #1946 dispatch quotes the SRL string verbatim),
- conversation prose discussing the error.

Same family as #1809 (stale string re-scan) and the same prose-FP concern the context% design handled by narrowing scan scope — but for error patterns the input line and message bodies are still in scope.

## Impact
False SRL/ApiError state latches → spurious AUTO-retry continues injected → state churn; incident discussions about errors can themselves trigger error handling (meta-loop). Likely explains some historical oddities.

## Fix directions (spike with #1946 — same SRL line)
1. **Anchor on real-error output shape**: real CLI error blocks have stable structure (e.g. claude's '⏺ API Error:' bullet prefix at line start + indent), vs quoted text inside an input box (after the prompt marker '❯'/'›') or inside an injected [AGEND-MSG] body. Exclude matches that sit on the input line (reuse the per-backend prompt-location infra from #1914/#1944) and matches inside pointer/message-body regions.\n2. **Format tightening**: require the error pattern to match at line start / with the CLI's error prefix, not as a substring anywhere.\n3. Evaluate which is robust across backends (multi-backend lens — operator directive).\n\nNote for the team meanwhile: avoid quoting raw error strings verbatim in dispatch messages (use paraphrase) until this lands.

suzuke:
Second PR of the #1946/#1947 pair (lead task t-20260610041002260921-1, spike approved). Operator-reproduced live FP: typing the SRL error string into the input line latched a false ServerRateLimit — the quoted line self-satisfies `line_has_error_indicator` ("…Error:"), and the position/working-marker gates pass at an idle prompt. The corpus-gate era "verbatim-quote FP self-corrects" assumption no longer holds post-#1946: an FP now starts a persistent recovery-ownership track.

## Fix (v1, structural)
A HIGH_FP content-anchor match whose line is the backend's **input line** (or an echoed/submitted **user-message line**) is excluded. Markers fixture/live-verified per backend:
- **claude `❯`** — both the live input line AND submitted user messages render `❯`-prefixed (live pane capture), so one rule kills the operator-typed FP **and** the dispatch-quote FP; `>` additionally covers markdown blockquotes (quoted text either way)
- **codex `›`** — echoed submitted input; real errors render bare at line start (`codex-rate-limit-typical.raw`)
- **kiro `>`** — its input prompt; real `ThrottlingException` lines bare (`kiro-rate-limit-typical.raw`)
- **opencode / agy** — no stable input prefix → **honestly uncovered in v1** (empty markers = exactly the pre-#1947 anchor)

The hard-wrap flatten fallback (#1851/#1857) drops marker lines **before** flattening — flattening destroys the line structure the exclusion needs. Known accepted edge: a multi-line typed input's continuation rows lack the marker.

**v2 (line-start anchoring) stays corpus-gated as approved** — `claude-server-throttle.raw` proves the real SRL error renders WITHOUT a `⏺` bullet prefix, so a naive line-start anchor would false-negative.

## Mechanics
`BackendProfile.input_line_markers` (same shape as #1945's `context_pattern`), cached on `StateTracker` like `anchor_on_red`; `in_error_line` → `in_error_line_excluding_input(…, markers)` (empty-markers = identical semantics; unused wrapper removed, single caller migrated).

## Tests (§3.9)
- typed SRL quote on claude `❯` → no latch · submitted dispatch quote → no latch
- **FN guards**: real bare SRL block with the prompt below → still latches (fixture shape); codex real `stream error:` → latches while the `›` echo quote doesn't; kiro real `ThrottlingException:` → latches while the `>` quote doesn't
- flatten fallback drops marker lines, still detects a genuine bare hard-wrap
- full corpus / replay suite green — zero FN regression

`cargo fmt --check` ✓ · `clippy --all-targets --features tray -D warnings` ✓ · bins **3560 passed / 1 failed** (the known pre-existing git-shim environmental false-fail `dispatch_branch_..._1834`; passes with system git).

Team note (until deployed): keep paraphrasing error strings in dispatch messages.

Closes #1947. Refs #1946, #1768, #1857.

### DIFF
--- a/src/backend_profile.rs
+++ b/src/backend_profile.rs
@@ -44,6 +44,12 @@ pub struct BackendProfile {
     /// only over the bottom status rows — NOT the error tail window — because
     /// agents routinely DISCUSS context% in conversation text (prose-FP).
     pub context_pattern: Option<&'static str>,
+    /// #1947: prompt markers identifying the backend's INPUT line (and echoed /
+    /// submitted user-message lines). An error pattern matched on a line whose
+    /// trimmed start is one of these is operator-typed / quoted text, not CLI
+    /// error output — the content anchor excludes it. Empty = no stable prompt
+    /// prefix (opencode / agy): input-line exclusion honestly unavailable.
+    pub input_line_markers: &'static [&'static str],
 }
 
 /// ClaudeCode context% — matches the fleet statusline's used-form as rendered
@@ -109,6 +115,7 @@ fn agy_profile() -> BackendProfile {
             cache_id: Some(MarkerCacheId::Gemini),
         },
         context_pattern: None,
+        input_line_markers: &[],
         initial_state: AgentState::Starting,
     }
 }
@@ -168,6 +175,7 @@ fn kirocli_profile() -> BackendProfile {
             cache_id: Some(MarkerCacheId::Kiro),
         },
         context_pattern: Some(KIRO_CONTEXT_PATTERN),
+        input_line_markers: &[">"],
         initial_state: AgentState::Starting,
     }
 }
@@ -227,6 +235,7 @@ fn opencode_profile() -> BackendProfile {
             cache_id: Some(MarkerCacheId::OpenCode),
         },
         context_pattern: None,
+        input_line_markers: &[],
         initial_state: AgentState::Starting,
     }
 }
@@ -279,6 +288,7 @@ fn codex_profile() -> BackendProfile {
             cache_id: Some(MarkerCacheId::Codex),
         },
         context_pattern: None,
+        input_line_markers: &["›"],
         initial_state: AgentState::Starting,
     }
 }
@@ -349,6 +359,7 @@ fn claudecode_profile() -> BackendProfile {
             cache_id: Some(MarkerCacheId::Claude),
         },
         context_pattern: Some(CLAUDE_CONTEXT_PATTERN),
+        input_line_markers: &["❯", ">"],
         initial_state: AgentState::Starting,
     }
 }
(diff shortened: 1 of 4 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head dc65bb86 == PR head). FN-safety line-traced: the exclusion is per-occurrence — any occurrence of the error string on a non-input error-indicator line still latches, only occurrences sitting on an input/user-message line (marker-prefixed ❯/›/>) are excluded; the real_srl_error_with_prompt_below FN-guard proves a genuine bare SRL block (no ⏺ prefix) with a prompt below still latches. opencode/agy have empty markers = byte-identical pre-#1947 behavior (honestly uncovered). The marker filter runs BEFORE the hard-wrap flatten (which would destroy the line structure the exclusion needs). Single caller migrated, no stale wrapper. No sneaky v2 (line-start anchoring stays corpus-gated — the fixture proves naive anchoring would FN on real SRL). Kills the operator-reproduced false latch (typed/quoted error strings) without losing a single real error. CI 6/6 green.

PR #1951, opened 2026-06-10
fix(#1945): disable the transcript context-estimate — pattern-only resolution

suzuke:
Operator decision (lead task t-20260610054104852923-0, scope updated mid-flight): #1945's first live minute fired a **triple false 100% alert** (dev/dev-2/reviewer-2 simultaneously, `source: transcript`, while the lead statusline ground truth read 71%). The estimate fallback is **disabled until validated for accuracy** — an unreadable statusline is now honestly `unknown` and never alerts. This stops the 30min re-alert nag immediately on deploy.

## Root cause (lead's hypothesis confirmed against live transcripts, recorded for re-enable)
The transcript `message.model` field is plain **`claude-fable-5`** — the CLI's `[1m]` display suffix NEVER reaches the transcript (verified: `tail` of this fleet's own jsonl; live usage sample ~456k = cache_creation 439,779 + cache_read 16,367 + io). The window matcher (`contains("[1m]")`) therefore resolved every 1M session against 200k → ~350% → clamped to a credible-looking 100% on every narrow pane at once.

## Changes
- **`StateTracker`**: `resolved_context` is **pattern only**; the estimate field/setter/`"transcript"` source are removed. Unreadable pane → `None` → no alert.
- **`context_alert` tick**: estimate phase removed; the handler only reads pattern-resolved values.
- **`token_cost::estimate_context_pct`**: retained, tested, deliberately uncalled — annotated **DEFERRED (not ghost; #649-trio pattern)** with the re-enable condition: *validate estimates against statusline ground truth on wide panes first*. While parked, the estimator itself is corrected and test-pinned:
  - fable → 1M window (transcript model ids carry no `[1m]`)
  - **>110% reading = window-misjudge signal → WARN + honest `None`** (never laundered into a trusted 100)
  - 100–110% rounding edge still clamps to 100
  - `<synthetic>` / zero-usage placeholder rows skipped

## Tests (§3.9)
- live-shape fable transcript (plain model id, cache-heavy 456k usage) reads **~46%, not 100** — the exact false-alert regression
- over-window (700k vs 200k id) → **unknown, not 100** · rounding edge clamps · synthetic rows skipped
- resolution pin rewritten to pattern-only: no pattern → unknown (no alert source exists); pattern → `"pattern"` — pins the disabled path stays disabled (the estimate plumbing no longer compiles into the tracker)
- pattern-path tests (statusline extraction, prose-FP, alert state machine) unchanged and green

`cargo fmt --check` ✓ · `clippy --all-targets --features tray -D warnings` ✓ · bins **3564 passed / 1 failed** (the known pre-existing git-shim environmental `dispatch_branch_..._1834`; passes with system git).

### DIFF
--- a/src/daemon/per_tick/context_alert.rs
+++ b/src/daemon/per_tick/context_alert.rs
@@ -3,15 +3,15 @@
 //! alert goes to the agent's team orchestrator (and the usage is visible via
 //! LIST `context_pct`/`context_source`); nothing is auto-restarted.
 //!
-//! Source resolution per agent (see `StateTracker::resolved_context`):
-//! 1. `pattern` — the agent's own statusline percent, parsed by the PTY
-//!    feed (cheap, in-band). Preferred when fresh.
-//! 2. `transcript` — for Claude agents whose pattern can't be read (narrow
-//!    pane truncation / no statusline), this handler computes the estimate
-//!    from the newest transcript's LAST message usage. That file IO runs
-//!    HERE — on the tick, with no registry/core lock held — never in the
-//!    PTY reader's feed path.
-//! 3. otherwise unknown — no alert, honest `null` in LIST.
+//! Source per agent (see `StateTracker::resolved_context`): the statusline
+//! `pattern` ONLY — a pane whose statusline can't be read is honestly
+//! `unknown` (no alert, `null` in LIST).
+//!
+//! #1945-disable (operator decision, 2026-06-10): the transcript-estimate
+//! fallback is DISABLED — its first live minute fired a triple false 100%
+//! alert (window misjudge). The corrected estimator survives, tested but
+//! uncalled, in `token_cost::estimate_context_pct`; re-enable ONLY after
+//! validating its readings against statusline ground truth.
 //!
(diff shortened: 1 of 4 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head b27a4ffb == PR head). The disable is structural, not a flag: the estimate phase is removed from the context_alert tick, the context_estimate field/setter/'transcript' source are removed from state, and estimate_context_pct has zero prod callers (kept #[allow(dead_code)] + test-pinned for a future validated re-enable) — re-enabling requires a deliberate code change. Unknown now never alerts (deploying stops the 30min false-alert nag); the statusline path is unchanged. Root cause recorded and the estimator itself fixed: the CLI's [1m] suffix never appears in transcripts (message.model is plain claude-fable-5), so the window resolved to 200k → ~456k usage = 228% → laundered into a credible 100% on three agents at once. Now fable→1M window, and >110% is treated as a window-misjudge signal (warn + honestly-unknown, never clamped into a trusted number). The regression test pins today's exact false alert (456k cache-heavy fable transcript → ~46%, not 100). CI 6/6 green. Per operator ruling: estimates stay off until validated against statusline ground truth.

PR #1948, opened 2026-06-10
fix(#1944): buffer-aware TUI draft gate + draft-gate/badge instrumentation

suzuke:
The TUI notification draft-gate held peer messages until the operator's *next submit*, even ones that arrived while the input line was empty (operator-reported). RCA (#1944): `notification_queue::draft_state` deferred pane `[AGEND-MSG-PENDING]` injection on `typed_ms`-vs-`submit_ms` **ORDER alone, never the actual buffer content**. A type-then-clear (typed then backspaced to empty, or typed-but-not-submitted) leaves `typed_ms > submit_ms` for up to 5 min with a **visibly empty** input line → blanket-deferred until the next real submit.

**Not a regression** (archaeology in the issue): the false-positive is inherent to #1457's order-based design; it only became *observable* when #1680 fixed the metadata-read path that had left `draft_state` permanently `None` (gate disabled = always inject). Operator's two messages were `kind=report` (ambient → correctly draft-gated by #1473); the actionable-bypass is intact.

## ① Buffer-aware gate (fixes #1457's design flaw; real-draft protection unchanged)
- **`Backend::input_prompt_marker()`** — claude `❯` / codex `›` / agy `>`, each verified against a real `tests/fixtures/state-replay` capture (e.g. `claude-discussion-text.raw` ends in exactly `❯ ` — no placeholder). kiro/opencode/gemini unverified → `None`; shell/raw have no prompt widget.
- **`notification_queue::input_box_is_empty(tail, marker)`** → `Some(true)` if the **bottom-most** prompt line (first non-blank char == marker — the #1944 prose-FP guard) has no non-whitespace after the marker; `Some(false)` if it does; `None` if no prompt line (agent mid-output / markerless).
- **`flush_notifications_for_pane`** reads **`pane.vterm`** (owned, **no lock** — cheap, addresses the tick-cost concern) only when `draft_state == Drafting`. An empty box overrides to `None` (deliver); a real draft, a markerless backend, or an undeterminable read all keep deferring (**fail toward draft-protection** — never risk clobbering a real draft).

## ② Instrumentation (the RCA had zero logs on this path)
- `#1944-draftgate-decision` (info) on every defer / buffer-override decision (agent, raw + effective state, `buffer_empty`, typed/submit ms, pending count).
- `#1944-badge-state` (info) on every pane pending-count change — to locate the operator's separate "queue badge disappeared" report at runtime (the badge code is **intact** per the RCA; this catches whether the count reaches the render with N>0).

## §3.9
[...]

PR #1952, opened 2026-06-10
fix(#1948 v2): kiro placeholder coverage + drop the non-functional codex marker

suzuke:
#1948 v1 covered backends with a stable prompt-line **marker** (claude `❯` / codex `›` / agy `>`). kiro/opencode had none → fell back to the timestamp-only gate, so the type-then-clear false-positive (messages held while the input box is empty) still bit them. The operator's live test confirmed it: kiro and opencode both held messages after the box was cleared.

## v2 — a second emptiness signal (placeholder)
A **placeholder** is a hint the TUI shows ONLY while the box is empty, and replaces the instant the operator types.

- **`Backend::input_empty_placeholder()`** — kiro renders ` ask a question or describe a task ↵` when empty (verified against a **live `pane_snapshot`** of a just-cleared kiro pane, 2026-06-10). **opencode is intentionally left `None`**: its `┃`-bordered box has no placeholder, and the `┃`-prefixed model/path footer is always non-empty → input-vs-footer geometry can't be told apart robustly (version-coupled), so fail-toward-protection fallback is safer. (Re-evaluate if a future opencode build adds a stable empty-box placeholder — noted in the code.)
- **`notification_queue::input_box_empty_probe(tail, marker, placeholder)`** — marker probe first (claude/codex/agy), then placeholder probe (kiro: visible ⇒ empty), else `None`. **Placeholder ABSENCE returns `None`** (not `Some(false)`): it could mean "typed" OR "placeholder string changed across kiro versions" → fail toward protection. A version change therefore merely *disables* the kiro path, never risks clobbering a real draft (the key fail-protect argument the lead approved).
- The gate calls the unified probe via `pane.vterm` (owned, no lock).

## Test-harness fix
`pane_with_screen` used `VTerm::new(3, 40)` — but the signature is `(cols, rows)`, so that was **3 columns wide**. The v1 short-content tests (`❯ `) passed by luck; the 32-char kiro placeholder needs a realistic width → now `VTerm::new(80, 6)`.

## §3.9
- `input_box_empty_probe`: marker path (empty/typed/mid-output→None), kiro placeholder (visible→empty / absent→hold), neither→None.
- kiro gate integration: cleared (placeholder visible) → **delivered**; typed → **deferred**.
- v1 claude/codex/agy/Shell tests unchanged (no regression).
- Full `cargo nextest run --features tray --profile ci`: **3870 passed, 0 failed**; fmt + clippy clean.

Extends #1948 (kiro coverage; opencode stays fallback by design).

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -1311,15 +1311,15 @@ where
     // (fail toward draft-protection — never risk clobbering a real draft).
     let raw_state = notification_queue::draft_state(home, &pane.agent_name);
     let buffer_empty = if raw_state == notification_queue::DraftState::Drafting {
-        pane.backend
-            .as_ref()
-            .and_then(|b| b.input_prompt_marker())
-            .and_then(|marker| {
-                notification_queue::input_box_is_empty(
-                    &pane.vterm.tail_lines(DRAFT_INPUT_TAIL_ROWS),
-                    marker,
-                )
-            })
+        pane.backend.as_ref().and_then(|b| {
+            // #1948 v2: marker probe (claude/codex/agy) → placeholder probe (kiro)
+            // → None fallback. pane.vterm is owned (no lock).
+            notification_queue::input_box_empty_probe(
+                &pane.vterm.tail_lines(DRAFT_INPUT_TAIL_ROWS),
+                b.input_prompt_marker(),
+                b.input_empty_placeholder(),
+            )
+        })
     } else {
         None
     };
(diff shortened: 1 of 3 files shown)


suzuke:
**Added the codex honesty fix (lead-approved, folded in since #1952 was not yet reviewed):**

Operator live test surfaced that codex's empty box renders a rotating ghost phrase after `›` (`› Explain this codebase`, …), so v1's `Some("›")` marker reads the ghost as typed content → always defers. The `Some("›")` claim was non-functional (fail-protect-safe but never delivered on an empty codex box — a #1493 fidelity miss: the v1 fixture only covered the typed case).

This commit drops codex's `input_prompt_marker` → `None`, so codex falls back to the timestamp behavior (held while typed>submit) like opencode, **until** a colour/dim-based empty-box signal lands ((b), gated on an operator fg capture of codex's ghost-vs-typed colours). agy (`>` clean) and claude (`❯ ` clean) were re-checked via live snapshot — only codex was affected.

suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head 44c84841 == PR head). The crux holds: a placeholder MISS returns None (never Some(true)/Some(false)) — if kiro ever changes the placeholder string the probe silently degrades to fail-protect deferring, never to wrongly delivering into a real draft. Probe chain: marker (claude/agy) → placeholder (kiro, positive-empty signal, live-verified string) → None. opencode intentionally None with the assessment recorded in-code. Also includes the honest codex correction: the v1 '›' marker was non-functional (codex renders rotating ghost phrases after the prompt, so the marker probe always read 'typed content' and never delivered on empty — operator's live test exposed it); dropping it to None is behavior-equivalent and honest, pending the dim-attribute empty signal from the raw-capture spike. The test-harness VTerm::new(3,40) was a latent 3-column-wide bug, fixed to (80,6) with all v1 tests still green (real fix, not fixture-bending). CI 6/6 green.

PR #1953, opened 2026-06-10
fix(#1948 b): codex empty-box coverage via DIM-attribute ghost detection

suzuke:
Follow-up to #1952. #1948 dropped codex's `›` marker because its empty box renders a rotating **ghost** phrase after the prompt that a plain marker probe mis-reads as typed content (so codex always deferred — fail-protect-safe but never delivered on an empty box). This restores codex coverage using the **DIM attribute** as the signal.

## RCA (raw PTY capture, isolated sandbox, 2026-06-10)
I spawned codex in a throwaway `/tmp` sandbox (never touching `~/.agend-terminal`) and captured its raw bytes. Around the prompt:

```
ESC[1m›        ← bold prompt
ESC[22m ESC[2m ESC[2m Use /skills to list available skills   ← DIM ghost
ESC[22m        ← reset
```

So the ghost is `ESC[2m` (**DIM** attribute) on the **default** foreground colour. The colour-only `CellFg` reads it as `Default` and can't tell it from real input. The operator confirmed the ghost is **visually dim** and real input is **normal/bright**, ruling out a false-deliver on a real draft.

## Fix — detect the DIM attribute
- **`VTerm::tail_lines_with_dim(n)`** → `(text, per-char DIM mask)`, collected alongside the existing fg mask in `tail_lines_impl` (same gate, 1:1 aligned; `tail_lines`/`tail_lines_with_fg` unchanged behaviour).
- **`notification_queue::input_box_dim_aware_empty(text, dim, marker)`** — the bottom-most marker line is empty iff every non-whitespace char after the marker is DIM (ghost) or there is none; a single **normal-intensity** glyph → a real draft (protect); no marker line → `None` (fail toward protection).
- **`Backend::input_dim_ghost_marker()`** = codex `›`; the gate routes codex through the dim-aware path, everyone else through the v1/v2 text-only probe (claude/agy marker, kiro placeholder). `pane.vterm` owned, no lock.

## §3.9
- `input_box_dim_aware_empty` units: dim-ghost→empty, normal→real, only-whitespace→empty, no-marker→None, bottom-most-marker-wins.
- **codex gate integration** drives the vterm with codex's **real SGR** (`ESC[1m›` + `ESC[2m`): empty (dim ghost) → **delivered**; normal-intensity input → **deferred** (the #1493-faithful test — fed the producer's actual encoding).
- `tail_lines_with_dim` alignment unit + the backend marker table.
- claude/agy/kiro paths unchanged.
- Full `cargo nextest run --features tray --profile ci`: **3878 passed, 0 failed**; fmt + clippy clean.

Closes the codex blind spot from #1948; opencode still fallback (no signal).

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -1312,13 +1312,21 @@ where
     let raw_state = notification_queue::draft_state(home, &pane.agent_name);
     let buffer_empty = if raw_state == notification_queue::DraftState::Drafting {
         pane.backend.as_ref().and_then(|b| {
-            // #1948 v2: marker probe (claude/codex/agy) → placeholder probe (kiro)
-            // → None fallback. pane.vterm is owned (no lock).
-            notification_queue::input_box_empty_probe(
-                &pane.vterm.tail_lines(DRAFT_INPUT_TAIL_ROWS),
-                b.input_prompt_marker(),
-                b.input_empty_placeholder(),
-            )
+            // #1948(b): codex's empty box shows DIM ghost text after `›`, which a
+            // plain marker probe mis-reads as typed content — route it through the
+            // DIM-aware check (needs the per-char dim mask). Everyone else uses the
+            // text-only probe: marker (claude/agy) → placeholder (kiro) → fallback.
+            // pane.vterm is owned (no lock).
+            if let Some(marker) = b.input_dim_ghost_marker() {
+                let (text, dim) = pane.vterm.tail_lines_with_dim(DRAFT_INPUT_TAIL_ROWS);
+                notification_queue::input_box_dim_aware_empty(&text, &dim, marker)
+            } else {
+                notification_queue::input_box_empty_probe(
+                    &pane.vterm.tail_lines(DRAFT_INPUT_TAIL_ROWS),
+                    b.input_prompt_marker(),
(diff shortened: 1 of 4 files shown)


suzuke:
reviewer-2 VERIFIED (adversarial, reviewed_head d8038aaf == PR head). The crux direction holds: the dim-aware probe returns Some(false) (defer) on the FIRST normal-intensity non-whitespace char after the marker — mixed dim+normal defers, only an all-dim (ghost) or whitespace-only tail delivers; the operator-confirmed fact that real input renders at normal intensity makes this a strict clobber-guard (a real draft is never delivered into). Bottom-most marker line wins (a dim quoted › above a real input line doesn't fool it — tested). The vterm extension is purely additive (tail_lines/tail_lines_with_fg byte-identical, dim is a third element dropped by existing callers) so state-detection has zero regression; the dim mask is validated against real SGR sequences (ESC[2m / ESC[22m) through vterm processing, faithful to the raw codex capture. Gate routes only codex through dim-aware (input_dim_ghost_marker), claude/agy/kiro keep the #1952 text probe. Completes the codex empty-box trilogy: non-functional v1 marker → honest #1952 disable → #1953 dim-aware re-enable. CI 6/6 green.

PR #1978, opened 2026-06-11
fix(notification-queue): daemon-side per-tick flush — deferred Telegram messages no longer strand under headless run_core

yujunchao:
## Problem

Telegram messages deferred by the #1513 busy-gate **strand forever** under headless `run_core` (`agend-terminal start --foreground`), because the notification queue's only flusher is the TUI event loop — which never runs in headless mode. Operator-reported 2026-06-10: 7 user messages sat undelivered for hours.

## Fix

Daemon-side per-tick flush so headless deployments drain the deferred queue without the TUI:

1. **`per_tick::NotificationFlushHandler`** (new, registered in `build_default_handlers`, ~10s tick): scans each fleet instance's deferred queue and drains it through the shared gating core via the submit-aware PTY path. In app-mode it also covers agents not present in the TUI layout.
2. **`inbox::notify::flush_agent_queue`** extracted verbatim from `app::flush_notifications_for_pane` — the draft-state gate, busy/typing holds, and `MAX_DEFER` anti-starvation cap now live in one shared core used by both the TUI and the new daemon handler.
3. **Claim-atomic drain** so the daemon flush and TUI flush can run concurrently without double-delivery.

## Tests (TDD: RED committed first, then GREEN)

- `per_tick::notification_flush::flushes_stranded_ambient_once_agent_settles`
- `per_tick::notification_flush::busy_agent_holds_fresh_then_cap_releases`
- `notification_queue::drain_does_not_steal_fresh_foreign_draining_file` (claim-atomicity contract)

## Soak

Deployed on the downstream fork since 2026-06-10; smoke confirmed a previously-stranded queue auto-flushed within seconds of the new binary starting. Single commit, applies directly on `main`.


suzuke:
This is a real bug and a well-built fix — thank you. I confirmed the mechanism independently: `flush_notifications_for_pane` / `flush_idle_notifications` live **only** in `src/app/mod.rs` (the TUI), and `inbox/notify.rs` even has a comment acknowledging the gap ("when the operator TUI isn't [running]…"). So deferred notifications genuinely strand under headless `run_core`. The TDD (RED-first), the shared-core extraction, the claim-atomic drain, and the downstream soak are all the right instincts. 👍 We'd like to take this.

**One thing first: it needs a rebase onto current `main` — it's currently conflicting (`mergeable: CONFLICTING`).** `main` has moved since you branched. Most relevant to *this* PR: the notification-queue / draft-gate area you're extracting from was recently hardened in a few merged PRs:

- **#1944** — buffer-aware empty-input detection in the draft gate (the `flush_notifications_for_pane` path now probes the actual input box, not just timestamps).
- **#1952 / #1953** — per-backend input-box emptiness (kiro placeholder; codex DIM-attribute ghost-text detection).

These changed exactly the gating logic inside `flush_notifications_for_pane` that your PR extracts into the shared `flush_agent_queue`. So the rebase isn't just a mechanical conflict-resolve — please make sure the extracted shared core ends up **behaviorally identical** to the *current* `flush_notifications_for_pane` (draft-state gate + buffer-aware empty probe + busy/typing holds + `MAX_DEFER` cap), so the daemon-side flush and the TUI flush apply the same gating. That input-detection behavior was recently validated against live backends and we don't want to regress it.

Once rebased, we'll do a thorough review with these focuses:
1. Extracted shared core is byte-for-byte equivalent in gating behavior to the post-#1944/#1952/#1953 TUI path (no draft-gate regression).
2. The claim-atomic drain is genuinely race-free under concurrent TUI + daemon flush (no double-delivery, no lost message).
3. The headless-strand path actually drains (the core fix).

Could you rebase onto current `main` and push? Happy to review promptly once it's green. Thanks again — this is a valuable fix for headless / server deployments.

suzuke:
@yujunchao thanks again for this fix — the mechanism is confirmed real and we want it landed. Since the rebase request has been pending a while and main has moved significantly today (18 merges), our team will carry the rebase forward to keep things moving — **your commits and authorship stay intact**. If you're back and prefer to do it yourself, just say so on this PR before we push.

suzuke:
Carried forward per the takeover note above (authorship preserved — @yujunchao remains the commit author):

- **Rebased onto current main** (`2654c9f`, 18 merges ahead of the original base). One real conflict in `src/app/mod.rs`: after this PR's base, main gained the #1944/#1948 draft-gate refinement (probing the ACTUAL rendered input box before honoring a raw `Drafting`). Resolution preserves both intents:
  - the shared core (`inbox::notify`) is split into `flush_agent_queue` (raw draft state — what the headless daemon uses; the probe needs a TUI-owned `pane.vterm`, so headless conservatively fails toward draft-protection) and `flush_agent_queue_with_state` (caller-supplied state);
  - the TUI wrapper keeps main's probe + `#1944-draftgate-decision` instrument and passes the refined state into the same core. The drain/hold/MAX_DEFER logic itself was extracted faithfully (byte-identical to main's) and is unchanged.
- One mechanical fixup: the daemon-flush test fixture gained `output_silent_secs: 0` (field added on main by #1975 after this PR's base).
- Validation: the PR's own RED tests + the #1944/#1948/#1513 draft-gate and flush_release suites all green (48/48 in the affected area); full suite 3961/3962 (the 1 = known #1834 environmental false-fail, passes solo with `AGEND_GIT_BYPASS=1`).

Review next: the team's reviewer will do the strict pass (shared-core byte-equivalence vs #1944/#1952/#1953 behavior + drain claim-atomicity under concurrent TUI/daemon flush).

suzuke:
## Review mirror (§3.12) — dual adversarial review, both VERIFIED @ b796c76a

**First pass (structure/concurrency/noise — codex)**: shared-core extraction is faithful (TUI wrapper keeps the #1944/#1948 probe + instrumentation byte-unchanged, refined state flows into the same core); headless worst-case divergence is over-holding, never force-submitting over a live draft; drain takes a per-agent OS file lock — no double-delivery between daemon and TUI flushes, failed injects requeue; backlog drain on restart is bounded exactly like the old TUI clean-buffer drain.

**Second pass (draft-gate domain history)**: `Drafting` is bounded at 5 minutes by `draft_state`'s typed-ms window — no infinite hold; in headless-only deployments `typed_ms` never sets (no TUI keystrokes) so Drafting can't even occur; the #1952/#1953 placeholder/dim-ghost edge cases live in the untouched TUI refinement layer; mixed-mode raw-vs-refined divergence is benign under the claim-atomic drain.

CI fully green. Carried forward by the team with conflict resolution against today's main (see earlier comment) — **authorship preserved**. Thanks @yujunchao for finding and fixing a real headless-mode gap; merging (squash).

PR #1981, opened 2026-06-11
feat: support customizing low disk headroom floor via AGEND_LOW_DISK_THRESHOLD

### LINKED ISSUE #1980: feat: Make low disk space threshold configurable via environment variable
Currently, `agend-terminal` has a hardcoded disk threshold of 5% (`const LOW_DISK_THRESHOLD: f64 = 0.05;` in `src/inbox/disk.rs`) before the inbox is set to read-only.

However, different machines have different disk capacities. For a large disk (e.g. 1TB+), 5% means 50GB, which is a lot of free space that is wasted just to avoid a low disk space warning. For smaller disks, a lower threshold might also be preferred.

I have implemented a change that allows configuring this threshold via the `AGEND_LOW_DISK_THRESHOLD` environment variable (e.g., setting it to `0.01` for 1%). It falls back to `0.05` (5%) if the environment variable is not set or invalid, and clamps the value between `0.0` and `1.0`.

This makes it easy to run `agend-terminal` on machines with limited disk capacity without running into persistent `inbox readonly: disk space critically low` errors, as long as a few gigabytes remain.

cheerc:
Resolves #1980

This PR updates the low disk space check to use an absolute byte headroom floor (default 1 GiB) instead of a percentage ratio. It also supports customizing this headroom threshold via the `AGEND_LOW_DISK_THRESHOLD` environment variable (in bytes).

### Changes:
- Replaced the hardcoded `LOW_DISK_THRESHOLD` 5% ratio in `src/inbox/disk.rs` with `get_low_disk_threshold_bytes()`, which defaults to 1 GiB.
- Read customizable threshold bytes from the `AGEND_LOW_DISK_THRESHOLD` environment variable.
- Updated the warning log to format the low disk space headroom dynamically in GB or MB.
- Updated doc comments in `src/inbox/mod.rs` to document the 1 GiB default and `AGEND_LOW_DISK_THRESHOLD` env variable.
- Added a unit test `test_custom_disk_threshold_env` in `src/inbox/tests.rs` to verify env parsing, invalid fallback, and defaults.

### DIFF
--- a/src/inbox/disk.rs
+++ b/src/inbox/disk.rs
@@ -4,22 +4,27 @@ use std::sync::atomic::{AtomicBool, Ordering};
 /// Global readonly flag — set when available disk space drops below threshold.
 pub(super) static DISK_READONLY: AtomicBool = AtomicBool::new(false);
 
-/// Minimum free-space ratio before entering readonly mode.
-const LOW_DISK_THRESHOLD: f64 = 0.05;
+/// Default minimum free disk space floor (1 GiB) before entering readonly mode.
+const DEFAULT_LOW_DISK_FLOOR_BYTES: u64 = 1024 * 1024 * 1024;
+
+/// Get the low disk space threshold in bytes from `AGEND_LOW_DISK_THRESHOLD` environment variable,
+/// falling back to 1 GiB.
+pub(super) fn get_low_disk_threshold_bytes() -> u64 {
+    std::env::var("AGEND_LOW_DISK_THRESHOLD")
+        .ok()
+        .and_then(|v| v.parse::<u64>().ok())
+        .unwrap_or(DEFAULT_LOW_DISK_FLOOR_BYTES)
+}
 
 /// Check available disk space at `path`. Returns true if below threshold.
 fn is_disk_low(path: &Path) -> bool {
     use fs4::available_space;
-    use fs4::total_space;
     let avail = match available_space(path) {
         Ok(s) => s,
         Err(_) => return false, // can't check → assume OK
     };
-    let total = match total_space(path) {
-        Ok(s) if s > 0 => s,
-        _ => return false,
-    };
-    (avail as f64 / total as f64) < LOW_DISK_THRESHOLD
+    let limit = get_low_disk_threshold_bytes();
+    avail < limit
 }
 
 /// Update the global readonly flag based on disk space at `home`.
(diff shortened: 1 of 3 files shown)


suzuke:
Thanks for this — the underlying observation in #1980 is a **real problem**, and it's well-articulated. A 5% *ratio* on a 1 TB disk means we flip the inbox to read-only with ~50 GB still free, which is clearly wrong. Good catch. 🙏

Before merging, I'd like to steer the **fix shape** a bit, because I think we can solve the root cause more cleanly than a configurable ratio:

### The root issue is the *ratio*, not the lack of a knob

`is_disk_low` uses `avail / total < 5%`. The problem is that a percentage is the wrong unit here — what actually matters is **absolute headroom** (will the disk fill up soon?), and that doesn't scale with total size:

| disk | 5% ratio triggers at | actually low? |
|---|---|---|
| 1 TB | 50 GB free | ❌ no (50 GB is plenty) |
| 256 GB | 12.8 GB free | ❌ not really |
| 10 GB | 0.5 GB free | ✅ yes |

### Preferred fix: an **absolute floor** (self-tuning, no env needed)

Change the predicate to something like:

```rust
const LOW_DISK_FLOOR_BYTES: u64 = 1 * 1024 * 1024 * 1024; // 1 GiB headroom
fn is_disk_low(path: &Path) -> bool {
    // ... avail = available bytes ...
    avail < LOW_DISK_FLOOR_BYTES
}
```

Why this over the configurable-ratio approach:
- **It fixes the default for everyone** — large-disk users no longer hit premature read-only *without having to set anything*. With the env-ratio approach, the default is still 5%, so anyone who doesn't know about the env var still gets bitten (and most users won't).
- **No new env flag** — this repo has been actively *reducing* its env-config surface (we've removed several magic-number env knobs as YAGNI), so we'd rather not add one if the root cause can just be fixed.
- **Self-tuning across disk sizes** — 1 GB free is genuinely low on any disk; 50 GB free is never low.

If you still want an override on top, that's fine — but please repurpose `AGEND_LOW_DISK_THRESHOLD` as an **absolute bytes/GB floor** (default 1 GiB), not a ratio, so the default is correct out of the box.

Would you be up for revising the PR along these lines? Happy to review again once updated — and thanks again for spotting this, it's a genuine improvement. If you'd rather we take it from here, just say so and we'll implement it with credit to you for the catch.

cheerc:
Thanks for the feedback! I've updated the PR to use an absolute byte headroom floor (default 1 GiB) instead of a percentage ratio. It also supports custom headroom via the `AGEND_LOW_DISK_THRESHOLD` environment variable (in bytes). Local tests have been verified and passed.

suzuke:
Perfect — this is exactly the right shape now. 👍 The default is self-tuning (1 GiB absolute floor works correctly on any disk size, no more premature read-only at 50 GB free on a 1 TB disk), and `AGEND_LOW_DISK_THRESHOLD` as an absolute byte override is a reasonable escape hatch for anyone who wants to tune it. Thanks for being responsive to the feedback and for catching the underlying issue in the first place.

We'll run it through review + CI and merge. Appreciated!

PR #1984, opened 2026-06-11
fix(service): propagate PATH environment variable to background services

### LINKED ISSUE #1983: fix: propagate PATH environment variable to background services during service install
### Description
When installing the `agend-terminal` daemon as a user-level background service via `agend-terminal service install` (on macOS via `launchd` or on Linux via `systemd`), the service runs with a minimal environment. 

Specifically, the `PATH` environment variable defaults to the system minimal value (e.g., `"/usr/bin:/bin:/usr/sbin:/sbin"`), which does not inherit the operator's custom directories (such as `/opt/homebrew/bin/` or `~/.local/bin/`).

Because of this, the background daemon fails to spawn backend instances like `claude` (installed in `~/.local/bin/`) or `opencode` (installed in `/opt/homebrew/bin/`), resulting in errors in `daemon.log` such as:
```text
ERROR spawn_and_register_agent rolled back; agent NOT in fleet agent=agy-65f671 error=Failed to spawn 'agy': Unable to spawn agy because:
No viable candidates found in PATH "/Users/cheerc/.agend/bin:/Users/cheerc/.cargo/bin:/usr/bin:/bin:/usr/sbin:/sbin"
```
Consequently, when running `agend-terminal app`, the TUI starts completely empty without any active instances.

### Proposed Solution
We propose propagating the current shell's `PATH` environment variable of the installer into the service configuration templates during `agend-terminal service install`:
- For macOS (`launchd`), we substitute a `__PATH__` placeholder inside the `<key>EnvironmentVariables</key>` dictionary of the plist file.
- For Linux (`systemd`), we substitute `__PATH__` under `Environment=PATH=__PATH__` in the service unit.

We have implemented this fix and pushed it to a branch. Is this the intended way to resolve this environment inheritance issue, or is there a preferred approach to managing paths for background daemon services?

cheerc:
### Description
This pull request addresses the environment inheritance issue when running the daemon as an OS background service (macOS `launchd` or Linux `systemd`). 

Currently, the service runs in a minimal environment with a default system `PATH` (typically `/usr/bin:/bin:/usr/sbin:/sbin`), causing the daemon to fail when spawning backend agents installed in user-level directories (e.g., `claude` in `~/.local/bin/` or `opencode` in `/opt/homebrew/bin/`).

Fixes #1983

### Changes
1. **macOS (`launchd`)**:
   - Added a `__PATH__` placeholder inside the `EnvironmentVariables` dictionary of [launchd.plist.template](file:///Users/cheerc/agend-terminal/assets/service/launchd.plist.template).
   - In [macos.rs](file:///Users/cheerc/agend-terminal/src/service/macos.rs), we fetch the installer's current shell `PATH` at runtime, XML-escape it, and substitute it.
2. **Linux (`systemd`)**:
   - Added `Environment=PATH=__PATH__` to [systemd.service.template](file:///Users/cheerc/agend-terminal/assets/service/systemd.service.template).
   - In [linux.rs](file:///Users/cheerc/agend-terminal/src/service/linux.rs), we fetch the current `PATH`, systemd-quote it, and substitute it.
3. **Tests**:
   - Verified that service-related unit tests run and pass successfully.

### Discussion
Is propagating the installer's `PATH` into the service configuration files the correct direction for this, or is there a preferred configuration pattern (e.g. asking the user to manually configure the PATH or setting it via some configuration file)?

### DIFF
--- a/assets/service/launchd.plist.template
+++ b/assets/service/launchd.plist.template
@@ -37,6 +37,8 @@ Placeholders (substituted at install time by service::install):
              ambient XPC_SERVICE_NAME false-positive). -->
         <key>AGEND_SUPERVISED</key>
         <string>1</string>
+        <key>PATH</key>
+        <string>__PATH__</string>
     </dict>
     <key>RunAtLoad</key>
     <true/>
--- a/assets/service/systemd.service.template
+++ b/assets/service/systemd.service.template
@@ -28,6 +28,7 @@ Environment=AGEND_HOME=__HOME__
 # systemd Restart=on-failure unit. (INVOCATION_ID, set by systemd at
 # runtime, remains a belt-and-suspenders signal.)
 Environment=AGEND_SUPERVISED=1
+Environment=PATH=__PATH__
 # Sprint 57 Wave 3 PR-2 (#548 Q3): the daemon does NOT supervise itself.
 # systemd is the supervisor of last resort. Restart=on-failure ensures
 # operator-visible crashes get respawned by systemd, but Q6 graceful
--- a/src/service/linux.rs
+++ b/src/service/linux.rs
@@ -35,11 +35,13 @@ pub(super) fn install(home: &Path, exe: &Path) -> Result<PathBuf, String> {
     // Paths without special chars round-trip unchanged.
     let exe_quoted = systemd_quote(&exe.display().to_string());
     let home_quoted = systemd_quote(&home.display().to_string());
+    let path_quoted = systemd_quote(&std::env::var("PATH").unwrap_or_default());
     let resolved = apply_substitutions(
         SYSTEMD_TEMPLATE,
         &[
             ("__EXECUTABLE__", exe_quoted.as_str()),
             ("__HOME__", home_quoted.as_str()),
+            ("__PATH__", path_quoted.as_str()),
         ],
     );
     if let Some(parent) = unit.parent() {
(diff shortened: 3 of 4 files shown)


suzuke:
Thanks — the problem is real (minimal service PATH can't spawn user-level backends) and **capture-at-install is the right direction**: it's the standard pattern for launchd/systemd service generators, and a re-install refreshes the snapshot. We don't want to require manual PATH editing.

Review is underway (escaping/quoting edge cases: XML entities in plist, `%` specifier expansion in systemd units, unreplaced-placeholder fallback). Will follow up with the verdict shortly.

suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ 924a033 (+2 non-blocking findings)

All five probes concluded:
- **XML escaping** ✓ — reuses the existing `xml_escape` (all five entities, `&` first); well-formed for any PATH.
- **systemd quoting** ⚠ non-blocking — `systemd_quote` doesn't escape `%` → `%%` (systemd expands specifiers inside `Environment=`). This is a **pre-existing limitation** (exe/home substitutions share it), input is the installer's own PATH (self-controlled), and a literal `%` in PATH is near-nonexistent — so not blocking this PR. Follow-up issue incoming.
- **Missing PATH** ✓ — `unwrap_or_default()` → empty string, no literal placeholder leakage.
- **Capture-at-install semantics** ✓ — install is idempotent, re-install refreshes the snapshot, matches the PR description.
- **Test coverage** ⚠ non-blocking — no dedicated `__PATH__` test / no install-level "no leftover placeholder" guard; wiring itself verified correct by inspection + CI green on 3 platforms.

Merging (squash). Thanks for another clean, well-scoped fix!

Issue #1990, opened 2026-06-11
Versioning hardening follow-ups: store migration hook, corrupt≠wipe, version unversioned stores


suzuke:
Tracking issue for the medium-priority gaps from the on-disk format inventory spike (2026-06-11). Each is independent; pick off as capacity allows.

1. **Migration hook in the store backbone**: `store::load_versioned` currently only rejects future versions and leans on serde defaults for old ones — there is no v(N-1)→v(N) upgrade step, so only additive changes are safe even for versioned stores. Add a migrate hook invoked when persisted version < CURRENT. (Deliberately deferred until the first non-additive change actually needs it — additive-only is the documented rule meanwhile.)
2. **Corrupt versioned store wipes all state**: `load_versioned` on corrupt JSON backs up to `<path>.corrupt.<ts>` then returns **default** — one bad byte loses the whole store (schedules, deployments, health escalation…). Consider per-record rescue or at least operator-visible surfacing beyond an error log line.
3. **Unversioned (b)-tier stores**: `metadata/*.json`, `decisions/*.json`, `topics.json`, `runtime-config.json` (operator-facing via config MCP — do this one first) lack schema_version; `binding.json` has a bare `version:1` field but doesn't go through the SchemaVersioned trait (no future-version guard — also reconcile the reported reader-behavior discrepancy while there).

Inbox's pattern (version field + per-line skip + boot-time half-write recovery) is the house standard to converge on.

suzuke:
Adding item 4 from the #1992 adversarial review (pre-existing, out of that PR's scope):

4. **Forward-compat ABORT is operator-invisible at the consumption sites**: `replay()` is lazy+cached and not called at boot; the tick-level callers silently swallow the error (`cron_tick.rs:380` `Err(_) => TaskGate::ReadError`, idle_watchdog `.ok()?`). So when the task board goes fail-closed (future-version / unknown-variant line), the daemon runs fine but the board silently stalls — no boot-time signal, no operator-visible event explaining why. Surface it: once-per-boot ERROR event (same pattern as #1972 crash-budget surfacing) when replay fails with the fail-closed classes.

suzuke:
## Item 3 — partial: 3 versioned, 2 deferred (free-form bags)

PR for the three **typed-struct** (b)-tier stores. `schema_version` added additive-only (old files without the field still load via serde default; written files carry it; a newer-than-supported file is fail-closed, never downgraded):

[...]

PR #1992, opened 2026-06-11
fix(#1988): task_events corrupt-line resilience — skip+quarantine, keep fwd-compat fail-closed

### LINKED ISSUE #1988: task_events.jsonl replay is fail-closed ABORT — one corrupt line bricks the entire task board
From the on-disk format inventory spike (2026-06-11). **Highest-risk robustness gap found.**

`read_envelopes_strict` (`src/task_events.rs:977/1431`) aborts the whole replay on any malformed line or future-version envelope → the entire task board becomes unreadable from a single partial write (e.g. power loss mid-append).

Contrast with **inbox** (`src/inbox/storage.rs:118/286`), which gets this right and should be the template:
- skip malformed lines (warn, keep going)
- drop future-version records per-record, not per-file
- boot-time `recover_half_writes` quarantines the bad tail

**Fix**: port the inbox tolerance pattern to task_events replay. Forward-version fail-closed semantics can stay (that part is deliberate); corrupt-line must not abort the world. Quarantine skipped lines (e.g. `.recovery/`) so nothing is silently lost.

Part of the compatibility/robustness hardening track (multi-user product: external users' boards must survive a bad byte).

suzuke:
Closes #1988. Implements the highest-risk robustness gap from the on-disk format inventory spike: `task_events.jsonl` replay used to **abort the entire task board** on a single malformed line.

## The decision the dispatch asked me to reason out
The issue floated "drop future-version records per-record." After tracing the semantics I landed on a **three-way split by cause** — only true garbage is skippable; both forward-incompat shapes stay fail-closed:

| Line shape | Cause | Behavior | Why |
|---|---|---|---|
| **Corrupt (non-JSON)** | crash-torn tail / disk glitch | **SKIP + warn** (+ boot quarantine) | carries no recoverable event; one bad byte must not brick the board |
| **Future-version** (valid JSON, `schema_version > CURRENT`) | a newer daemon wrote a real event | **whole-file fail-closed ABORT** | **correctness**: an undecodable event may have changed task state in ways this binary can't see, so serving the rest = acting on a **partial/misread board** (e.g. re-dispatching a task a newer event already closed). Fail-closed until the operator runs a new-enough binary. |
| **Well-formed but undeserializable** (valid JSON, supported version, unknown `kind`/missing field) | newer event variant w/o a version bump, or a hand-edit | **ABORT** | same partial/misread-state hazard as future-version — a real event we can't apply |

Per-record drop is **rejected** for the forward-incompat cases on correctness grounds (don't operate on a board missing events you can't interpret). Corrupt-line skip is lossless because garbage has no semantic content.

> **Rationale correction (reviewer-2, PR commit 1c537d1c):** an earlier draft justified the ABORT by claiming a later `compact()` would rewrite the log and lose the events. That is wrong — `task_events::compact()` is `#[allow(dead_code)]` (no production caller) and is line-based + non-destructive (moves to archive, replay reads archives first). The ABORT conclusion stands; the justification is **correctness**, not compaction-loss. Comments fixed in-tree.

## Changes
- **`read_envelopes_strict`**: Value-first parse → the three-way split above. Corrupt → skip+warn; future-version → keep `bail!` (forward-compat fail-closed); undeserializable-at-supported-version → keep `bail!` ("replay aborts").
- **`task_events::recover_half_writes`** (new, mirrors `inbox::recover_half_writes`): under the log's own flock, quarantines **non-JSON** lines to `task_events.recovery/<ts>/` (fsync'd forensic copy) and rewrites the hot log (tmp+fsync+rename) with the good lines only — nothing silently destroyed. Keeps valid-JSON future/unknown lines so the read-path gate still fires. Wired into daemon boot right after the inbox recovery. Only the live log is scanned (archives are atomic-written; the read-path skip is the backstop for any archive corruption).

## Tests (§3.9 — all through the **real** `replay` / `recover_half_writes` entries, real board shapes create/claim/update/done)
- `replay_skips_corrupt_midfile_line_keeps_full_lifecycle` — corrupt line mid-lifecycle → board folds intact, no abort.
- `recover_half_writes_quarantines_torn_tail` — torn tail → quarantined + rewritten out, good events survive.
- `recover_keeps_future_version_line_replay_still_fail_closed` — recovery does **not** auto-drop a newer daemon's events; read-path still fail-closes.
- Existing `invariant_4_forward_compat_fail_closed` + `replay_rejects_unknown_event_variant` still pass (fail-closed semantics unchanged).

## Verification
`cargo fmt --check`, `cargo clippy --all-targets --features tray -D warnings` clean. Full `cargo nextest --features tray`: **3900 passed**. Two unrelated reds, both confirmed NOT from this change (team_dedup_and_rejection — see #1993; `spawn_group...no_bypass_1899` — local `AGEND_GIT_BYPASS=1` env artifact).

Part of the compatibility/robustness hardening track (multi-user product: an external user's board must survive a bad byte).

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -1138,6 +1138,10 @@ fn build_tick_infrastructure(
     crate::instance_monitor::spawn_monitor_tick(home.to_path_buf(), Arc::clone(&ctx.registry));
 
     crate::inbox::recover_half_writes(home);
+    // #1988: same half-write recovery for the task-event log — quarantine a
+    // crash-torn tail line and rewrite the hot log with the good events only,
+    // so a single bad byte cannot brick the whole task board on replay.
+    crate::task_events::recover_half_writes(home);
     replay_missed_at_startup(home, &ctx.registry);
     crate::daemon::ci_watch::startup_sweep(home);
     // #1488: GC bindings (schedules/dispatch_tracking/ci_watch) left orphaned
--- a/src/task_events.rs
+++ b/src/task_events.rs
@@ -1461,11 +1461,13 @@ fn read_envelopes_strict(path: &Path, out: &mut Vec<TaskEventEnvelope>) -> anyho
         // (2) FUTURE-VERSION line — valid JSON, but `schema_version` exceeds what
         //     this binary understands: a VALID event a NEWER daemon wrote. Keep
         //     the WHOLE-FILE fail-closed ABORT (deliberately NOT a per-record
-        //     skip): if this older daemon operated on the partial board it had
-        //     skipped, a later `compact()` would rewrite the hot log WITHOUT those
-        //     newer events → permanent loss of the newer daemon's task state.
-        //     Aborting refuses to operate at all, forcing the operator back to the
-        //     newer binary. This is the intended forward-compat protection.
+        //     skip) on CORRECTNESS grounds: an event we cannot decode may change
+        //     task state in ways we cannot see (a newer done/reassign/close
+        //     variant), so skipping it and serving the rest would have this
+        //     daemon act on a PARTIAL, MISREAD board — e.g. re-dispatching a task
+        //     a newer event already closed. Aborting refuses to operate until the
+        //     operator runs a binary new enough to read the whole log. This is
+        //     the intended forward-compat protection.
         let version = value
             .get("schema_version")
             .and_then(|v| v.as_u64())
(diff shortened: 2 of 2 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ 1c537d1c (adversarial, 6 probes, iterative)

- **Trichotomy** correct: non-JSON garbage → skip+quarantine (no semantic loss); future-version & undeserializable → fail-closed ABORT. No skip-holes (`unwrap_or(0)` routes to ABORT; `deny_unknown_fields`, no `serde(other)`).
- **Crash-safety** traced: good events fsync→atomic rename, original log intact until rename, quarantine copy made durable before rewrite (review finding, fixed). Single lock shared with append/compact — no interleaving window.
- **Rationale corrected during review**: the compaction-loss argument was refuted (compact is line-based, non-destructive, no prod caller); ABORT stands on correctness grounds (an old daemon must not act on a partial, misread board). Comment now states the true reason.
- Remaining note (pre-existing, out of scope): tick-level callers swallow the fail-closed error silently — tracked as #1990 item 4.

CI fully green incl. Daemon-boot flake gate (20x). Merging (squash).

PR #1996, opened 2026-06-11
test(#1993): rewrite team_dedup_and_rejection → team_recreate_extends_roster (strict #1964 contract)

### LINKED ISSUE #1993: team_dedup_and_rejection fails deterministically on macOS (green on CI/ubuntu) — re-create of same team not rejected
Surfaced while verifying #1988 (PR #1992) — confirmed **unrelated to that change** and **pre-existing**.

## Symptom
`tests/migrated_scripts.rs::team_dedup_and_rejection` asserts that re-creating a team with an existing name is rejected. On macOS the second `create_instance {team:"test-team", count:2}` is **not** rejected — it succeeds and spawns two more members:

```
panicked at tests/migrated_scripts.rs:57:5:
re-creating same team must be rejected:
{"ok":true,"result":{"backends":["claude","claude"],"spawned":["test-team-3","test-team-4"],"team":"test-team"}}
```

(Note also `backends:["claude","claude"]` on the re-create despite the fixture's `backend: cat` — possibly a related dedup/default-resolution oddity.)

## Evidence — env confound excluded
Run on a **fresh worktree at clean `origin/main` (2e77d9de)** — no #1988 changes (`grep -c recover_half_writes src/task_events.rs` = 0):

| Condition | Result |
|---|---|
| `AGEND_GIT_BYPASS=1` exported | FAIL |
| `env -u AGEND_GIT_BYPASS` (unset) | FAIL |
| 3 consecutive runs (bypass unset) | FAIL × 3 (deterministic, not a flake) |
| CI mainline (ubuntu-latest) | green |

So it is **not** the `AGEND_GIT_BYPASS` shell-env confound (that one only false-reds `git_helpers::…no_bypass_1899`), and **not** a timing flake — it is a **deterministic darwin-vs-ubuntu divergence**.

## Hypothesis (needs root-cause)
The team-name dedup check passes on Linux but not on macOS. Candidates:
- The dedup reads team existence from a store/registry whose path or case-sensitivity differs on macOS (APFS is case-insensitive by default), so the `test-team` lookup misses.
- A persistence-ordering difference between the first create's team registration and the second call's existence check that only resolves correctly on the CI runner.

Either a platform-fragile test or a real macOS team-dedup gap. CI green means it does not block mainline, but local `cargo nextest` is red on every macOS dev box, which erodes the "local-green = CI-green" preflight contract.

## Repro
```
git worktree add --detach /tmp/wt origin/main
cd /tmp/wt
env -u AGEND_GIT_BYPASS cargo nextest run --features tray -E 'test(team_dedup_and_rejection)'
```

suzuke:
Closes #1993.

`tests/migrated_scripts.rs::team_dedup_and_rejection` failed deterministically on local macOS but was "green on CI". Root-caused (see #1993): it asserted the **pre-#1964 reject-on-duplicate-team contract**, which #1964 (PR #1966) **deliberately removed** — re-creating an existing team now **extends the roster**, numbering new members from `max(existing <team>-N)+1` (`src/api/handlers/team.rs:111-128, 290-310`). The test was never updated.

## Why it was green on CI / red locally (the loose-assertion bug)
The final fallback assertion was:
```rust
inner2.get("error").is_some() || resp_str.contains("already exists") || resp_str.contains("error")
```
`resp_str.contains("error")` passes on **any** "error" substring anywhere in the response. A clean re-create on a fast local box returns `{spawned:[test-team-3,test-team-4]}` with no "error" text → **red** (correctly reflecting #1964). On CI the response apparently carries some spawn-noise text containing "error" → **accidentally green**. So post-#1964 the test asserted nothing real.

## The rewrite — strict #1964 extend-roster contract
- first create → `ok`, spawned `[test-team-1, test-team-2]`
- re-create same team → `ok` (**not rejected**), spawned `[test-team-3, test-team-4]` (numbered from max+1)
- **no `contains("error")` substring fallback** — exact spawned-id assertions + `result.error` must be absent
- explicit `backends:[cat,cat]` instead of `count:2` (which falls back to the hardcoded `"claude"` default and would need the claude binary) → spawn is **deterministic on every host incl. CI**
- renamed `team_dedup_and_rejection` → `team_recreate_extends_roster` (the old name described a removed contract)

## Verification
`cargo clippy --all-targets --features tray -D warnings` clean. The rewritten test passes locally (the old one failed deterministically on the same machine). If CI goes red on this, that is a genuine CI-environment finding to investigate, not something to paper over.

### DIFF
--- a/tests/migrated_scripts.rs
+++ b/tests/migrated_scripts.rs
@@ -16,49 +16,87 @@ fn tmp_home(tag: &str) -> std::path::PathBuf {
     dir
 }
 
-/// Migrated from: scripts/test-team-dedup.sh
-/// Tests: team creation spawns members + re-create with same name is rejected.
+/// Originally migrated from: scripts/test-team-dedup.sh — rewritten for #1993.
+///
+/// #1964 (PR #1966) deliberately changed `create_instance(team=X)`: re-creating
+/// an existing team is NOT rejected — it EXTENDS the roster, numbering new
+/// members from `max(existing <team>-N) + 1`. The old `team_dedup_and_rejection`
+/// asserted the pre-#1964 reject-on-dup contract via a `resp_str.contains("error")`
+/// fallback so loose it passed on ANY "error" substring in the response — green
+/// on CI by accident (spawn-noise text), red locally where a clean re-create has
+/// no "error" substring (#1993). This asserts the CURRENT extend-roster contract
+/// STRICTLY: exact spawned ids, no substring fallback.
 #[test]
-fn team_dedup_and_rejection() {
-    let home = tmp_home("team-dedup");
+fn team_recreate_extends_roster() {
+    let home = tmp_home("team-extend");
     let harness = AgendHarness::spawn(home.clone(), "defaults:\n  backend: cat\ninstances: {}\n")
         .expect("spawn");
 
     let client = TuiClient::new(&harness, 80, 24);
 
-    // Create a team — should succeed
-    let result = client.call(
-        "mcp_tool",
-        &json!({
-            "tool": "create_instance",
-            "arguments": {"team": "test-team", "count": 2},
-            "instance": "test-caller"
-        }),
-    );
-    assert!(
-        result.is_ok(),
-        "team creation API call must not fail: {:?}",
-        result.err()
-    );
-
-    // Re-create same team — should be rejected with "already exists"
-    let resp2 = client
-        .call(
+    // Explicit `backends: [cat, cat]` (NOT `count`, which falls back to the
+    // hardcoded "claude" default in instance.rs and would need the claude binary)
+    // so member spawn is deterministic on every host incl. CI — this test asserts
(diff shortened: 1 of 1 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ c43251cf

- New assertions faithfully encode the current contract (`plan_member_names`: fresh→1, existing→max+1, `team.rs:67-68`) — exact spawned-id equality, `ok==true`, `error.is_none()`; the loose `contains("error")` fallback that made CI accidentally green since the #1964 contract change is gone entirely.
- No platform branches; strict equality green on all three CI platforms.
- Non-blocking note: explicit `backends:[cat,cat]` means the `count`→default-backend selection line (`team.rs:99-101`) is no longer exercised by this test — deliberate (determinism on hosts without claude); numbering contract coverage is unaffected.

Merging (squash) — closes #1993.

PR #1999, opened 2026-06-11
fix(supervisor): require srl_active_on_screen for after-abort rate-limit retry nudge

### LINKED ISSUE #1985: Bug: Repeated continue nudges during long tasks/waiting states (related to #1946, #1809, #1518)
Hi there,

We recently noticed an issue where an agent (specifically `dev-team-impl2`) got stuck in a loop of repeated `continue` nudge injections from the daemon while waiting on a long-running process (an internal PTY/MCP call taking ~9 minutes). 

After digging through the codebase and searching the issue history, we noticed this is closely related to the ongoing hardening of the ServerRateLimit/ApiError recovery path:
- **#1518**: Stale error text lingering in the viewport causing retry storm (solved by `ERROR_TAIL_SCAN_LINES = 15`).
- **#1808** / **#1809** / **#1936**: Re-latching of stale error text on screen ticks (solved by line signature matching and overriding cross-cycle matches to `Idle`).
- **#1946**: Retaining retry track across abort-to-Idle (in Idle state with `!recovered` / no productive markers within 45s, a retry nudge is scheduled and `continue` injected).

### The Root Cause
When the agent is running a long command or waiting for a long process, the screen is static and lacks new `CLAUDE_PRODUCTIVE_MARKERS`. 
1. Once `RECOVERY_SILENCE` (45s) is exceeded, `recovered` evaluates to `false`.
2. When the agent lands on `Idle` (e.g. naturally returning to the prompt or via the #1809 cross-cycle override), the `#1946` logic kicks in: since `retry_count > 0` and `!recovered`, it assumes the agent aborted back to its prompt, sets `abort_pending = true`, and repeatedly schedules and injects `continue` nudges.
3. This creates a loop of spurious nudges because the supervisor has no way of knowing whether the rate-limit error is actually still on screen or has scrolled off.

### Proposed Fix
We can distinguish between a genuine abort (where the error is still on screen) and a scrolling recovery/working state by tracking whether the rate-limit error is physically visible on the screen (`srl_active_on_screen`).

1. Add `pub srl_active_on_screen: bool` to `StateTracker` in `src/state/mod.rs`.
2. Reset it to `false` at the start of `feed_with_fg()`, and set it to `true` when a valid live `ServerRateLimit` is detected (both in `detect_with_match` after anchor/position gates and in the hard-wrap fallback `flattened_throttle_detect`).
3. Update `process_error_recovery` in `src/daemon/supervisor.rs` to extract `srl_active_on_screen` and restrict the `#1946` abort-to-Idle check to require `srl_active_on_screen == true`:
   ```rust
   Some(track) if track.retry_count > 0 && !track.exhausted && !recovered && srl_active_on_screen => {
       // retain track & schedule nudge
   }
   ```
   If the error is no longer visible on screen, the supervisor clears the track immediately.

We have drafted and tested this fix locally, and all 3600+ unit tests (including the updated supervisor tests) pass.

Since this recovery logic has been iterated on several times (and we definitely don't want to cause any disruption or regress other edge cases), we would love to get your thoughts on whether this diagnosis is correct and if this is a safe/desirable direction for a PR.

Thanks so much for your time!

cheerc:
### Description
This pull request implements the supervisor-side evidence gate to prevent repeated spurious `continue` nudges when an agent is running a long, silent command.

Specifically, it addresses the issue where `StateTracker::tick()` transitions a silent agent from `Thinking`/`ToolUse` to `Idle` after 30 seconds of inactivity. Under the previous `#1946` abort-to-Idle logic, if the agent was previously rate-limited and had a non-zero retry track, the supervisor would assume the agent aborted back to its prompt and repeatedly inject `continue` nudges—even if the rate limit error was no longer physically visible on screen (scrolled off or never occurred in the current command).

Resolves #1985

### Changes
1. **State layer**:
   - Exposed `screen_has_throttle_hint` as `pub(crate)` in `src/state/mod.rs` to allow the supervisor to check screen text directly.
2. **Supervisor layer**:
   - In `process_error_recovery` (`src/daemon/supervisor.rs`), we now read the `vterm` tail lines (using `core.vterm.tail_lines(15)`) under the core lock to check if a rate limit hint is still on the screen.
   - Restricted the `#1946` abort-to-Idle check to require `has_throttle_hint == true`.
   - If the rate limit error is no longer on screen, we immediately clear the retry track and log a warning with tag `#1946-abort-clear-no-evidence` as a production data guardrail.
3. **Tests**:
   - Updated unit tests `abort_to_idle_retains_track_and_resumes_retry_1946` and `abort_pending_budget_exhaustion_reachable_1946` to write a throttle error to the mock `vterm` so they continue to pass under the new constraint.
   - Added `abort_pending_scrolled_off_clears_track_1946` to verify that the track is cleared immediately when the error is no longer on screen.

### DIFF
--- a/src/daemon/supervisor.rs
+++ b/src/daemon/supervisor.rs
@@ -1717,11 +1717,14 @@ pub(crate) fn process_error_recovery(
             // produced (`last_productive_output == None`) is NOT recovery, so it
             // latches + injects normally (the fresh-agent edge the creation stamp
             // used to mis-suppress). `productive_silence` is for the log only.
-            let (state, recovered, productive_silence) = {
+            let (state, recovered, has_throttle_hint, productive_silence) = {
                 let core = handle.core.lock();
+                let has_hint =
+                    crate::state::screen_has_throttle_hint(&core.vterm.tail_lines(TAIL_LINES));
                 (
                     core.state.current,
                     core.state.recovered_within(RECOVERY_SILENCE),
+                    has_hint,
                     core.state.productive_silence(),
                 )
             };
@@ -1802,7 +1805,12 @@ pub(crate) fn process_error_recovery(
                     // cleanly separates this from genuine recovery — failed-retry
                     // spinners do not count as productive output (probe evidence:
                     // productive_silent_secs=600 across 4 retries).
-                    Some(track) if track.retry_count > 0 && !track.exhausted && !recovered => {
+                    Some(track)
+                        if track.retry_count > 0
+                            && !track.exhausted
+                            && !recovered
+                            && has_throttle_hint =>
+                    {
                         if !track.abort_pending {
                             track.abort_pending = true;
                             let idx = (track.retry_count as usize)
@@ -1840,15 +1848,27 @@ pub(crate) fn process_error_recovery(
                         // Idle — cross-episode reset so a later episode starts
                         // fresh at Phase A.
                         if let Some(cleared) = retry_tracks.remove(name) {
-                            tracing::info!(
-                                agent = %name,
-                                ?state,
-                                retry_count = cleared.retry_count,
-                                after_abort = cleared.abort_pending,
-                                recovered,
-                                productive_silent_secs = productive_silence.as_secs(),
-                                "ServerRateLimit retry cleared — agent recovered (Idle)"
-                            );
+                            if cleared.retry_count > 0 && !recovered {
+                                tracing::warn!(
+                                    agent = %name,
+                                    tag = "#1946-abort-clear-no-evidence",
+                                    retry_count = cleared.retry_count,
+                                    next_retry_at = ?cleared.next_retry_at,
+                                    secs_until_retry = cleared.next_retry_at.saturating_duration_since(now).as_secs(),
+                                    productive_silent_secs = productive_silence.as_secs(),
+                                    "abort-to-Idle cleared — no throttle error visible on screen (likely scrolled off or recovered)"
+                                );
+                            } else {
+                                tracing::info!(
+                                    agent = %name,
+                                    ?state,
+                                    retry_count = cleared.retry_count,
+                                    after_abort = cleared.abort_pending,
+                                    recovered,
+                                    productive_silent_secs = productive_silence.as_secs(),
+                                    "ServerRateLimit retry cleared — agent recovered (Idle)"
+                                );
+                            }
                         }
                     }
                     None => {}
@@ -3423,11 +3443,20 @@ instances:
         name: &str,
         state: crate::state::AgentState,
     ) -> (crate::agent::AgentHandle, Box<dyn std::io::Read + Send>) {
+        mock_agent_handle_with_size(name, state, 10, 80)
+    }
+
+    fn mock_agent_handle_with_size(
+        name: &str,
+        state: crate::state::AgentState,
+        rows: u16,
+        cols: u16,
+    ) -> (crate::agent::AgentHandle, Box<dyn std::io::Read + Send>) {
         let pty_system = portable_pty::native_pty_system();
         let pair = pty_system
             .openpty(portable_pty::PtySize {
-                rows: 10,
-                cols: 80,
+                rows,
+                cols,
                 pixel_width: 0,
                 pixel_height: 0,
             })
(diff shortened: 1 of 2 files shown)


suzuke:
## Review (§3.12) — verdict: changes requested (direction confirmed correct, fast path to green)

The approach is exactly what #1985 called for — lock ordering verified safe (`tail_lines` is a pure grid read under the same core guard the branch already holds), the existing-test updates are faithful to the real scenario (a genuine nudge screen does contain the throttle error), and the new scrolled-off test goes through the real `process_error_recovery` entry. Thanks for the quick, well-shaped rework.

**Blocking (mechanical):**
1. CI is red on `cargo fmt --check` only — the new `handle.core.lock().vterm.process(b"...")` fixture lines need wrapping. `cargo fmt` fixes it.
2. Branch is behind main — please rebase onto `3d5d84b7`.

**Blocking (one real detection gap):**

3. `tail_lines(15)` — the `15` is a magic number, and it's **smaller than the supervisor's existing error-context window `TAIL_LINES = 40`** (`supervisor.rs:25`). A throttle error sitting in rows 16–40 — still physically on screen — would be missed by the 15-row window → track cleared → exactly the freeze this gate is supposed to avoid. Please hoist it to a named constant and consider just reusing `TAIL_LINES` (40); the cost of scanning 25 more rows is nil, the false-negative it removes is the expensive kind.

**Non-blocking, but cheap to address while you're in there:**

4. `tail_lines` joins grid rows with `\n`, so a hint token split across a soft-wrap boundary (`limit\ning`) won't match. The hint tokens are short, so the odds are low — but a test pinning the behavior either way would document the known edge (a "hint in rows 16–40" test for item 3 would be great too).
5. The WARN guardrail fields are good (`agent`, `retry_count`, `productive_silent_secs`); adding `next_retry_at` would help production triage.

On the clear-vs-retain semantics: we traced the fallback chain — after a no-evidence clear, the idle watchdog still pings the lead at 60min, so the worst case of a true false-negative is a bounded freeze with manual rescue, which is the tradeoff #1985 accepted. That's why item 3 matters: it shrinks the false-negative surface to nearly zero.

suzuke:
Saw the fmt push (5e02780) — item 1 resolved, thanks. Two notes so you don't wait on CI unnecessarily:

- **Item 3 (the other blocking one) is still pending**: the gate still reads `core.vterm.tail_lines(15)` (`supervisor.rs:1722`). Hoisting to `TAIL_LINES` (40) — or a named constant — is what closes the rows-16-40 false-negative.
- Main has moved again since the review (now `0c2b0197`, a few merges landed this afternoon — none touch `process_error_recovery`, so the rebase should stay trivial). Fine to leave the rebase as the last step.

Items 4/5 remain optional. Once item 3 + rebase are in, we'll re-review promptly.

cheerc:
Hi @suzuke,

Thanks for the detailed feedback! We have resolved the items:
1. **Item 3 (evidence window size)**: Hoisted the scan window to use `TAIL_LINES` (40 rows) instead of the hardcoded 15 rows.
2. **Rebase**: Rebased the branch on the latest main (`0c2b0197`).
3. **Item 4 (soft-wrap split behavior)**: Added `abort_pending_split_wrap_clears_track_1946` unit test to document and pin this edge case.
4. **Item 5 (warn log improvements)**: Added `next_retry_at` and `secs_until_retry` to the warn log output for the no-evidence clear case.
5. Added another test `abort_pending_retains_track_when_error_in_rows_16_to_40` that uses a custom 50-row terminal and verifies that a throttle error sitting at row 20 is successfully matched and retains the track (confirming the fix for Item 3).
6. Cleaned up formatting with `cargo fmt`.

All tests pass locally and the PR has been updated!

suzuke:
All five items confirmed addressed — the rows-16-40 regression test (custom 50-row terminal, error at row 20) is exactly the right pin for item 3, thanks.

One last mechanical blocker: CI is red on **clippy, not tests** — this repo denies `unwrap()` outside of `expect` (`-D clippy::unwrap-used` via `[lints.clippy]`):

```
error: used `unwrap()` on an `Option` value
  --> src/daemon/supervisor.rs:3657:26
   |  let handle = reg.values().next().unwrap();
```

Swap it for `.expect("registry seeded with one agent")` (the surrounding tests use that style) and CI should go green. Local `cargo clippy --all-targets --features tray -- -D warnings` reproduces it. We'll re-review promptly on the next push — this is the last item.

cheerc:
{
  "body": "Hi @suzuke,\n\nThanks for the detailed feedback! We have resolved the items:\n1. **Item 3 (evidence window size)**: Hoisted the scan window to use `TAIL_LINES` (40 rows) instead of the hardcoded 15 rows.\n2. **Rebase**: Rebased the branch on the latest main (`0c2b0197`).\n3. **Item 4 (soft-wrap split behavior)**: Added `abort_pending_split_wrap_clears_track_1946` unit test to document and pin this edge case.\n4. **Item 5 (warn log improvements)**: Added `next_retry_at` and `secs_until_retry` to the warn log output for the no-evidence clear case.\n5. Added another test `abort_pending_retains_track_when_error_in_rows_16_to_40` that uses a custom 50-row terminal and verifies that a throttle error sitting at row 20 is successfully matched and retains the track (confirming the fix for Item 3).\n6. Cleaned up formatting with `cargo fmt`.\n\nAll tests pass locally and the PR has been updated!"
}

suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: REJECTED → VERIFIED @ 93bbab51

Every review item resolved by the author across three pushes:
- Evidence window hoisted from a bare 15 to `TAIL_LINES` (40) — and the dedicated rows-16-40 regression (50-row terminal, error at row 20, through the real `process_error_recovery` path) proves an on-screen error in the previously-missed band now retains the track instead of falsely clearing.
- Soft-wrap split edge documented and pinned (`abort_pending_split_wrap_clears_track_1946`).
- No-evidence-clear WARN enriched with `next_retry_at`/`secs_until_retry` for production triage.
- fmt, rebase, and the final clippy `unwrap → expect` all clean; CI fully green incl. the daemon-boot flake gate.

A model external-contribution arc: real bug report (#1985) → reworked to the maintainers' suggested architecture → iterated through adversarial review to green, all same-day. Thank you @cheerc — merging (squash), closes #1985.

PR #2000, opened 2026-06-11
feat(#1990 item 3): schema_version for runtime-config / decisions / binding (b-tier versioning)

suzuke:
#1990 item 3 — wire the unversioned (b)-tier stores into the future-version guard. **Additive-only**: a file written by an older daemon (no `schema_version`) still loads via serde default; written files carry the version; a newer-than-supported file is **fail-closed, never silently downgraded**.

## The three typed-struct stores
| Store | Versioning | Future-version behavior |
|---|---|---|
| **runtime-config.json** | `SchemaVersioned` (CURRENT=1), stamped on write | inline check in `reload()` routed through the existing **#1576 `fail_closed`** — keep-last-good at runtime / safe default at startup. **Not** `load_versioned` (which resets to default → would regress #1576). |
| **decisions/*.json** | per-record `task_progress` pattern (module const + read guard), stamped on every write (post / supersede / update) | skipped on read (`load_all`), refused for update/archive — no downgrade |
| **binding.json** | reconcile — it carried a bare `version:1` field but did **not** go through `SchemaVersioned` (no guard) | `read()` rejects `version > 1` → `None` = treated **absent**, which the git-shim push gate / lease checks already fail-close on. Missing-signature path unchanged. |

`runtime-config.rs::keys()` filters `schema_version` out (it's on-disk metadata, not an operator-settable `config` MCP key — caught by `keys_cover_all_settable_keys` in the full suite and fixed).

## Deferred (lead-approved): topics.json + metadata/*.json
Both are **free-form key-value bags** (a bare `HashMap`, an open `Value` KV bag), not typed structs. Adding a version field needs a non-additive shape change (envelope + tolerant dual-reader) — a breaking rewrite to add a version, on the two lowest-value stores, inside an additive-only track, is backwards. Documented as a tier (b) exception in `docs/COMPATIBILITY.md` ("only add new keys") + rationale in the [#1990 comment](https://github.com/suzuke/agend-terminal/issues/1990#issuecomment-4678119796). Both are low-risk (topics self-heals via boot orphan-sweep; metadata is operator-cosmetic).

## Tests (each store: additive old-file pin + future-version pin)
- runtime-config: `old_config_without_schema_version_loads`, `future_schema_version_config_kept_last_good`
- decisions: `old_decision_without_schema_version_is_listed`, `future_schema_version_decision_skipped_and_update_refused`
- binding: `future_version_binding_reads_as_absent`

## Verification
`cargo fmt`, `cargo clippy --all-targets --features tray -D warnings` clean. Full `cargo nextest --features tray` (no bypass, rebased on current main incl. #1996/#1997): **3912 passed**, 0 failed. `docs/COMPATIBILITY.md` updated (tier (b): the 3 newly-versioned stores + the 2 deferred free-form exceptions).

### DIFF
--- a/docs/COMPATIBILITY.md
+++ b/docs/COMPATIBILITY.md
@@ -44,9 +44,11 @@ than the tier floor). Either way: never a crash, never a silent drop.
 `runtime-config.json`, the decision log (`decisions/*.json`), and
 `binding.json` carry an explicit `schema_version` (#1990): an older file
 without it reads normally; a newer-than-supported file is fail-closed
-(runtime-config keeps the last-known-good per #1576; a decision is skipped on
-read and refused for update; a binding reads as absent, so the git-shim push
-gate denies).
+(runtime-config keeps the last-known-good per #1576 and refuses an overwriting
+write; a decision is skipped on read and refused for update; a binding reads as
+absent to **daemon-side** readers — the git shim has its own reader that
+HMAC-verifies and treats a parseable future binding as bound, so the agent stays
+restricted to its own worktree).
 
 **Two tier (b) stores are unversioned free-form key-value bags** that cannot
 carry a `schema_version` without a breaking shape change: `topics.json` (the
--- a/docs/COMPATIBILITY.md
+++ b/docs/COMPATIBILITY.md
@@ -41,6 +41,24 @@ keeps serving the rest (degrade); the task-events store fail-closes on the
 whole file (deliberate — board integrity outranks availability, stricter
 than the tier floor). Either way: never a crash, never a silent drop.
 
+`runtime-config.json`, the decision log (`decisions/*.json`), and
+`binding.json` carry an explicit `schema_version` (#1990): an older file
+without it reads normally; a newer-than-supported file is fail-closed
+(runtime-config keeps the last-known-good per #1576; a decision is skipped on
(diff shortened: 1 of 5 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ 4b107a79, delta-CONFIRMED @ 25364c16

Core additive versioning verified across all three stores: reload() keep-last-good preserved on both branches (#1576 intact, test-pinned); decisions write-sites compiler-enforced (no Default + required field); binding future-version guard.

Three review findings, all fixed in the delta:
1. `set()` now refuses to overwrite a future-version file on disk (visible error, mirrors decisions-update; test pins the on-disk file staying intact).
2. COMPATIBILITY.md / comments corrected to reality: the guard applies to daemon-side readers; the git shim has its own HMAC-verified reader and treats a parseable future binding as bound (net-safe, but not "denied").
3. Retention's destructive site now uses `present_including_future` — a future-version binding no longer reads as absent, so a newer daemon's legally bound worktree isn't reclaimed (test-pinned).

topics.json / metadata deferred by design (wrapping a bare bag in a versioned envelope is itself a breaking change) — documented as tier-b exceptions.

CI fully green on the delta head. Merging (squash) — #1990 item 3 done, tracker stays open for items 1/2/4.

PR #2002, opened 2026-06-11
feat(#1990 item 4): surface task-board replay fail-closed to the operator

suzuke:
#1990 item 4 — reviewer-2's P4 on #1992. Observability-only.

## Problem
After #1992, `replay()` fail-closes on a forward-incompatible record (a future-version or unknown-variant envelope — deliberate board-integrity protection). But the per-tick task-board readers swallow that `Err`:
- `cron_tick.rs` → `Err(_) => TaskGate::ReadError`
- idle watchdog → `.ok()?`

So the daemon keeps running, the task board silently stops advancing, and the operator has **no cause to look at**.

## Fix (follows the #1972 crash-budget surface pattern, observation-only)
On a fail-closed replay error, `surface_failclosed_replay_once`:
- `tracing::error!` **every** occurrence (greppable during an episode), and
- emits **one** operator-visible `event_log` entry **per boot** (process-global `AtomicBool` latch so the per-tick callers can't spam it).

The entry names the failure and the remedy (*upgrade the daemon, or quarantine the offending record*). It lives in `replay()`'s error path, so **both** tick sites are covered from one place. Boot-scoped — a restart re-alerts (the cause either healed or still blocks).

**Zero recovery-semantics change**: the fail-closed `Err` still propagates unchanged; transient IO errors (not the `"fail-closed"` class) are left alone — same discipline as #1972.

## Test
`replay_fail_closed_surfaces_operator_event_once` — two fail-closed replays in one boot → **exactly one** `event_log` entry (latched). The existing fail-closed replay tests (`invariant_4_forward_compat_fail_closed`, `replay_rejects_unknown_event_variant`, `recover_keeps_future_version_line_...`) share the process-global latch, so all four are in one `#[serial(task_replay_latch)]` group to keep the latch reset deterministic under nextest's parallelism.

## Verification
`cargo fmt`, `cargo clippy --all-targets --features tray -D warnings` clean. Full `cargo nextest --features tray` (no bypass): **3915 passed**, 0 failed.

Part of #1990 (item 4); the tracker stays open for items 1/2 + the deferred free-form bags.

### DIFF
--- a/src/task_events.rs
+++ b/src/task_events.rs
@@ -1305,6 +1305,14 @@ static REPLAY_FAILCLOSED_EVENT_EMITTED: std::sync::atomic::AtomicBool =
 /// per-tick callers can't spam). Observation only: the fail-closed `Err` still
 /// propagates unchanged, so no recovery semantics change (same discipline as
 /// #1972). A transient IO error (not the "fail-closed" class) is left alone.
+///
+/// Classification is by the `"fail-closed"` substring of the error — a contract
+/// pinned at `read_envelopes_strict`'s two `bail!` sites; keep them in lockstep.
+/// Known limitation (#1990 item 4, reviewer-2 minor 2): only failures that flow
+/// through [`replay`] are surfaced; the timeline queries `envelopes_for_task` /
+/// `stream_envelopes` fail-close without surfacing. The board-freeze alert here
+/// is the primary operator signal, so that narrower timeline-query gap is
+/// accepted rather than expanding scope.
 fn surface_failclosed_replay_once(home: &Path, err: &anyhow::Error) {
     let msg = err.to_string();
     if !msg.contains("fail-closed") {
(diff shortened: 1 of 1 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ 1618a366; amend delta (6ce1b268) self-reviewed by lead

- Observation-only invariant traced: the fail-closed `Err` propagates unchanged (both tick sites byte-identical), transient IO errors correctly excluded from surfacing, once-per-boot latch per #1972 pattern with sound test-reset mechanics, event content sufficient for operator self-rescue (path:line, version, fix guidance).
- Amend delta = bidirectional CONTRACT comments pinning the "fail-closed" substring coupling (bail sites ↔ classifier) + a negative test guarding the false-alarm boundary (bare IO error → no event) + documented known limitation for the timeline-query paths.

CI green on the amended head. Merging (squash) — #1990 item 4 done; tracker stays open for items 1/2.

PR #2003, opened 2026-06-11
feat(quickstart): --unattended mode — non-interactive setup for CI / scripted installs

suzuke:
## Summary

Productization stop 2 — quickstart was interactive-only (7 stdin prompt sites + a 3-minute network wait in `detect_group`), blocking CI/CD and scripted installs. This adds `quickstart --unattended` (alias `--yes`) as a KISS bypass: the interactive flow is untouched.

### Structural guarantees
`run_unattended` is a separate code path — nothing in its call graph reads stdin or touches the network:
- **Never hangs on stdin**: missing required input = clear error + non-zero exit (the CI killer the flag exists to avoid).
- **Never waits on the network**: no `detect_group` (3-min wait), no `verify_bot` — an env-provided token is stored **unverified** (noted in output; the daemon validates at startup).

### Behavior
| input | resolution |
|---|---|
| backend | first detected on PATH; **zero detected = the one hard error** (install hints on stderr, exit non-zero; no hardcoded claude assumption) |
| Telegram token | `AGEND_TELEGRAM_BOT_TOKEN` env (explicit, wins) → existing `.env` → **skipped** (commented channel block) |
| group_id | `AGEND_TELEGRAM_GROUP_ID` env → existing fleet.yaml → none (invalid env value: ignored with a note) |
| existing fleet.yaml | **never overwritten** (idempotent re-runs; interactive default `N` made unconditional) |
| differing .env token | updated only in the explicit-env case (CI token rotation) — the `.env` fallback arm is by definition equal to the stored token |

Resolution is a pure function (`resolve_unattended_telegram`) so the precedence matrix is unit-tested; every decision prints a `·` note for readable CI logs.

### Docs
README "Unattended setup (CI / scripts)" with a GitHub Actions example; CHANGELOG entry.

## Validation
- 5 unit tests (precedence + unverified note, fallback + skip, gid env-wins + invalid-ignored, never-overwrite idempotency, no-prompt token update — runs with no stdin, so a prompting regression reads EOF and fails).
- **Empirical smoke, real binary with stdin closed (`< /dev/null`)**: minimal → telegram skipped, fleet generated, exit 0; re-run → fleet kept including an operator edit; env token+gid → `.env` written `0600` + channel block carries the gid.
- fmt / clippy(tray) clean; nextest 3918/3919 (the 1 = known #1834 environmental, passes solo with bypass; team_dedup excluded per #1993).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Format based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); projec
 
 ### Added
 
+- **`quickstart --unattended` (alias `--yes`)** — non-interactive setup for CI / scripted installs: never reads stdin (missing input = clear error + non-zero exit, not a hang), never waits on the network. Backend = first detected on PATH; Telegram optionally from `AGEND_TELEGRAM_BOT_TOKEN` / `AGEND_TELEGRAM_GROUP_ID` env (token stored unverified; the daemon validates at startup), otherwise skipped. Idempotent: an existing fleet.yaml is never overwritten; an existing `.env` token is only replaced by an explicit env var.
+
 - **fleet.yaml `schema_version` + compatibility policy (#1989)** — `fleet.yaml` accepts an optional `schema_version:` field (omitted = `1`; existing files unchanged, the daemon never injects it). A file declaring a version newer than the daemon supports loads with a warning instead of being silently misread without trace. `docs/COMPATIBILITY.md` declares the on-disk interface tiers — (a) stable public (fleet.yaml, service templates, instruction blocks, MCP config), (b) internal persisted state, (c) regenerable/ephemeral — and the additive-only change rule for (a)/(b).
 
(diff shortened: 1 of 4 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ f577bf3e

- Core promise structurally guaranteed: `run_unattended`'s entire call graph is stdin/network-free (all prompts and network calls live only in the interactive `telegram_setup` path, unreachable from unattended) — and the tests run with no stdin, so any prompting regression reads EOF and fails.
- Zero-backend → hard error with install guidance on stderr, no hardcoded claude assumption; existing fleet.yaml never overwritten (guard sits before the interactive overwrite prompt, test-pinned); telegram precedence is a pure function with 5 pinned cases; .env written 0600 with CI-rotation semantics.
- Interactive flow unchanged (unattended=false callers byte-identical).

Known related issue NOT in scope here: the shared fleet template's legacy env-name bug (#2005) — fix already in flight as #2012, which also covers the unattended-generated template.

CI fully green. Merging (squash).

PR #2009, opened 2026-06-11
fix(#2004): surface four swallowed best-effort failures — warn-log batch, zero behavior change

### LINKED ISSUE #2004: Surface swallowed best-effort failures — warn-log batch (INJECT / dispatch_idle sidecar / ci-watch TTL refresh / fetch-prune)
Outcome of the silent-failure audit (reviewer-3) + adversarial verify-first pass (reviewer-2). Of the top-10 claims, 3 were FALSE, 3 PARTIAL/misdirected — this issue captures the verified-REAL remainder. All four fixes are **pure warn-log additions at swallowed failure points, zero behavior change**; one small PR.

1. **Team-spawn INJECT swallow** — `src/mcp/handlers/instance_spawn.rs:227`: `let _ = api::call(INJECT)` inside the fire-and-forget thread, no log. A slow-starting member (>3s) can miss its task text and idle with zero context, invisibly. Add warn + event_log entry.
2. **dispatch_idle sidecar writes** — `src/api/handlers/messaging.rs:506/536`: `record_dispatch` failure → task never gets its idle-timeout nudge (MED); `mark_resolved` failure → stale sidecar → spurious nudge. Add warn on both.
3. **ci-watch TTL refresh** — `src/daemon/ci_watch/poller.rs:1919` (post-#2001 head): `let _ = atomic_write` in refresh_expires_at → watch silently expires early; PR branches self-heal via PR-3 auto-arm, non-PR branches genuinely lose coverage. Add warn.
4. **fetch --prune in cleanup_merged_branch** — `src/worktree_pool.rs:162`: swallow is real but fail-direction is safe (stale refs → branch kept, self-heals on next fetch) — **downgraded from the audit's (d) to warn-log** so accumulation is at least visible.

Explicitly NOT in scope (verify-first found the original claims false or misdirected):
- replace/restart DELETE un-swallow (#1/#2): `handle_delete` hardcodes `{"ok":true}` — un-swallowing detects nothing; spawn failure already surfaces via `spawned:false`.
- telegram notify JoinHandle (#5): documented fire-and-forget, inner send already warn-drops; awaiting the handle would block the daemon hot path.
- Deeper kill-failure reporting (#3, needs delete_transaction-level status) and bind_self readback via `_outcome` path (#7, #781 follow-up) noted as optional backlog.
- store corrupt-backup copy swallow (#8) folded into #1990 item 2.

suzuke:
## Summary

Fix #2004 — the verified-REAL remainder of the silent-failure audit: pure warn/event_log additions at four swallowed failure points, **zero behavior change** (every code path's outcome is identical; only visibility changes).

| # | site | swallowed consequence | added |
|---|---|---|---|
| 1 | team-spawn INJECT (`mcp/handlers/instance_spawn.rs`, fire-and-forget thread) | member starts with zero task context, invisibly | `warn` + `event_log("team_spawn_inject_failed")` on `Err` or `ok != true` |
| 2a | `record_dispatch` (`api/handlers/messaging.rs`) | dispatch never gets its idle-timeout nudge | `warn` on `None` — at this call site every legit-skip arm is already excluded (kind=task branch, non-empty names, resolver-positive threshold), so `None` = disk failure |
| 2b | `mark_resolved` | stale sidecar later fires a spurious idle nudge | **deviation from the issue's call-site suggestion, same intent**: the fn returns `Option` where `None` is *normal* (most reports have no pending sidecar) — a call-site warn would false-alarm on every ordinary report. The warn lives at the actual failure branch inside `mark_resolved` (matched sidecar whose `delete_sidecar_locked` fails), with the `dispatch_id`. Call site documents this. |
| 3 | `refresh_expires_at` (`daemon/ci_watch/poller.rs`) | watch silently expires early (PR branches self-heal via PR-3; non-PR branches lose coverage) | `warn` on write failure |
| 4 | `fetch --prune` in `cleanup_merged_branch` (`worktree_pool.rs`) | safe fail-direction (branch kept) but persistent failure accumulates undeletable branches invisibly | `warn` on `Err` **and** on non-zero exit (with stderr); cleanup proceeds on local refs unchanged |

Out of scope (the issue's verify-first NOT list, untouched): DELETE un-swallow, telegram JoinHandle, kill-failure depth, bind_self readback, store backup-copy.

## Validation
- Zero behavior change by construction — no new tests (nothing observable changed except log output); the full suite is the regression net: nextest 3913/3914 (the 1 = known #1834 environmental, passes solo with bypass; `team_dedup_and_rejection` excluded per #1993).
- fmt / clippy(tray) `-D warnings` clean.

Closes #2004

### DIFF
--- a/src/api/handlers/messaging.rs
+++ b/src/api/handlers/messaging.rs
@@ -503,14 +503,27 @@ fn track_dispatch(
             )
         {
             let outbound_corr = outbound_corr.map(String::from);
-            let _ = crate::daemon::dispatch_idle::record_dispatch(
+            // #2004: a swallowed record failure means this dispatch never gets
+            // its idle-timeout nudge — surface it (non-fatal: the dispatch
+            // itself succeeded). At THIS call site every legit-skip arm of
+            // `record_dispatch` is already excluded (kind=task branch, non-empty
+            // names, threshold>0 from the resolver), so `None` = disk failure.
+            if crate::daemon::dispatch_idle::record_dispatch(
                 home,
                 from,
                 target,
                 outbound_corr.as_deref(),
                 kind_str,
                 threshold,
-            );
+            )
+            .is_none()
+            {
+                tracing::warn!(
+                    from = %from,
+                    target = %target,
+                    "dispatch_idle record_dispatch failed (sidecar not written) — this dispatch will get NO idle-timeout nudge"
+                );
+            }
         }
         // #1942: link the dispatched branch to the correlated task. The lead
         // dispatches `kind=task` with `branch=`, but the task is often created
@@ -533,6 +546,11 @@ fn track_dispatch(
         // fires a spurious `dispatch_idle_threshold_exceeded` nudge once the
         // target goes Idle (#1516's working-state gate does not cover that).
         if let Some(corr) = msg.correlation_id.as_deref().or(msg.task_id.as_deref()) {
+            // #2004: `None` here is NORMAL (a report whose correlation has no
+            // pending sidecar — most reports). The real swallowed failure —
+            // a matching sidecar whose DELETE fails, leaving it to fire a
+            // spurious idle nudge later — warns inside `mark_resolved` at the
+            // actual failure point, where the dispatch_id is known.
             let _ = crate::daemon::dispatch_idle::mark_resolved(home, corr);
             // #1888 phase-2: a report carrying the handoff's `repo@branch`
             // correlation (reviewer verdicts do) RESOLVES the ci-handoff track —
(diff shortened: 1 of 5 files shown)


suzuke:
## Review mirror (§3.12) — dual review: codex P1 → fixed → r2 verified; claude reviewer verified the other four sites

- **Initial REJECTED (codex)**: the new record_dispatch warn fired for legitimate `kind=query` dispatches (`record_dispatch` deliberately returns None for non-task kinds, but the call-site branch is `task | query`) — a noise-PR introducing noise. Lead adjudicated the conflicting claude-reviewer VERIFIED (which had trusted the PR's own incorrect comment) in codex's favor.
- **Fix verified non-vacuously**: warn narrowed to `kind_str == "task"`; the new query-no-warn negative test was proven real via RED mutation (reverting the condition makes it fail with the exact false WARN).
- Other four sites (INJECT, record_dispatch disk-failure, mark_resolved at the actual failure point, refresh_expires_at, fetch-prune) verified observation-only — zero behavior change, warns only on true failures.

Required checks green. Merging (squash) — closes #2004.

PR #2012, opened 2026-06-11
fix(#2005): canonical bot_token_env in quickstart template + symmetric creds fallback

### LINKED ISSUE #2005: quickstart channel template pins legacy bot_token_env (AGEND_BOT_TOKEN) while .env is written canonical — fresh Telegram setup fails to resolve at startup
Found during #2003 (quickstart unattended) — filed as a docs-consistency nit, but the code trace shows it is a **functional bug**: a fresh quickstart Telegram setup produces a channel config that fails to resolve at daemon startup.

## The mismatch

- `quickstart` writes the token to `.env` under the **canonical** key (`src/quickstart.rs` `TELEGRAM_TOKEN_KEY`):
  `AGEND_TELEGRAM_BOT_TOKEN=<tok>`
- but the generated `fleet.yaml` channel template pins the **legacy** name (`src/quickstart.rs:693`, and the commented variant at `:700`):
  `bot_token_env: AGEND_BOT_TOKEN`

## Why this is broken, not just ugly

1. `load_dotenv` (`src/main.rs:131`) exports `.env` keys **verbatim** — no aliasing. Process env gets `AGEND_TELEGRAM_BOT_TOKEN` only.
2. `creds::resolve_channel_from` (`src/channel/telegram/creds.rs:23-33`) reads `std::env::var(bot_token_env)` → `var("AGEND_BOT_TOKEN")` → **not set**.
3. The legacy fallback then tries… `var("AGEND_BOT_TOKEN")` **again** (same name, because the template already pinned the legacy name) → fails → `bot token env 'AGEND_BOT_TOKEN' not set`.

The fallback only rescues the OPPOSITE drift (fleet says canonical, .env has legacy). It never tries canonical-ward, so the combination quickstart itself generates is exactly the unrescued one.

Why nobody noticed: installs whose `.env` was written by a pre-#525-H1 quickstart carry the legacy `AGEND_BOT_TOKEN=` line (the old write key), which satisfies the template. Only **fresh** setups since the canonical `.env` migration hit it — i.e. exactly new external users.

## Fix direction (small)

1. Template: `bot_token_env: AGEND_TELEGRAM_BOT_TOKEN` (both the active and the commented variant in `generate_fleet_yaml`) — matches what the same flow writes to `.env`, and matches the serde default (`fleet/mod.rs` `default_telegram_bot_token_env`).
2. Optionally: make the creds fallback symmetric (try the canonical name when `bot_token_env` is the legacy one) so historic fleets keep working either way; keep the deprecation warn.
3. Regression test: generate a fleet via quickstart's template + a canonical-key `.env` → `resolve_channel_from` must succeed.

## Evidence

- cited: src/quickstart.rs:693 + :700 (`bot_token_env: AGEND_BOT_TOKEN` in both template arms) vs `TELEGRAM_TOKEN_KEY = "AGEND_TELEGRAM_BOT_TOKEN"` (the `.env` write key)
- cited: src/channel/telegram/creds.rs:23-33 — fallback env name is hard-coded `AGEND_BOT_TOKEN`, identical to the template's `bot_token_env`, so both lookups read the same missing variable
- cited: src/main.rs:131 `load_dotenv` — verbatim key export, no aliasing
- ran (under #2003): unattended smoke writes `.env` `AGEND_TELEGRAM_BOT_TOKEN=…` + fleet `bot_token_env: AGEND_BOT_TOKEN` — the broken pair, straight from the generator

suzuke:
## Summary

Fix #2005 — a **fresh quickstart Telegram setup failed to resolve at daemon startup**. The chain: the fleet.yaml template pinned the legacy `bot_token_env: AGEND_BOT_TOKEN` while `save_env_token` writes `.env` under the canonical `AGEND_TELEGRAM_BOT_TOKEN`; `load_dotenv` exports keys verbatim (no aliasing); and the creds fallback retried the **same** legacy name — dead code for exactly the pair quickstart itself generated. Old installs were masked by leftover legacy `.env` keys; the break hit only fresh installs, i.e. new external users.

### Changes
1. **quickstart template → canonical** (`src/quickstart.rs`, both the active and the commented channel arms) — matches what the same flow writes to `.env` and the serde default (`fleet::default_telegram_bot_token_env`). The #2003 `--unattended` path shares this generator, so it is fixed by the same edit (lead point 4 confirmed: unattended wrote the legacy name too, pre-fix).
2. **Symmetric creds fallback** (`src/channel/telegram/creds.rs`): configured name first, then whichever of canonical/legacy it isn't — covers **both** drift directions (old fleet template + migrated `.env`; new template + legacy `.env`). Legacy use keeps the deprecation warn; canonical-rescue warns naming the stale fleet field.
3. **4 regression tests through the REAL resolve entry** (`resolve_channel_from`, bumped `pub(super)` → `pub(crate)` for the cross-module test):
   - fresh-install shape (canonical fleet + canonical env) — failed pre-fix at the template level
   - old-install shape (canonical fleet + legacy env) — legacy compat preserved
   - **THE bug pin**: old-template fleet (legacy `bot_token_env`) + canonical-only env — failed pre-fix at the fallback level, now resolves
   - no-env → clear error naming the configured var
4. CHANGELOG `Fixed` entry.

### Drift-matrix coverage (post-fix)
| fleet `bot_token_env` | env var present | resolves |
|---|---|---|
| canonical | canonical | ✓ direct |
| canonical | legacy | ✓ fallback (deprecation warn) |
| legacy (old template) | canonical | ✓ fallback (stale-field warn) — **was broken** |
| legacy (old template) | legacy | ✓ direct |

## Validation
- fmt / clippy(tray) `-D warnings` clean.
- `cargo nextest -E 'test(/2005/)'` → 4/4.
- Full suite 3917/3918 (the 1 = known #1834 environmental, passes solo with bypass; `team_dedup_and_rejection` excluded per #1993).
- Note: test fixtures call the current 4-arg `generate_fleet_yaml`; if #2003 (5-arg) merges first, the auto-rebase will surface a one-line fixup.

Closes #2005

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Format based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); projec
 
 - **MSRV: declared `rust-version` corrected to 1.88 (#1994)** — the previous declaration (1.87) was false: the locked dependency set required up to rustc 1.95 (`sysinfo 0.39.3`), so `cargo install agend-terminal --locked` broke for anyone on 1.87–1.94 who trusted it. `sysinfo` is pinned to `0.38` (MSRV 1.88, no code changes — caught by the #1987 release gate on its first run, which now enforces the floor). Builders need rustc ≥ 1.88.
 
+### Fixed
+
+- **Fresh Telegram setup resolves at startup (#2005)** — the quickstart fleet.yaml template pinned the legacy `bot_token_env: AGEND_BOT_TOKEN` while `.env` was written under the canonical `AGEND_TELEGRAM_BOT_TOKEN`, and the credentials fallback retried the same legacy name — so a fresh install's Telegram channel failed to resolve at daemon startup (old installs were masked by leftover legacy `.env` keys). The template now pins the canonical name and the fallback is symmetric (configured name → the other of canonical/legacy), so all four fleet/.env drift combinations resolve; legacy use still warns.
+
 ### Added
 
 - **fleet.yaml `schema_version` + compatibility policy (#1989)** — `fleet.yaml` accepts an optional `schema_version:` field (omitted = `1`; existing files unchanged, the daemon never injects it). A file declaring a version newer than the daemon supports loads with a warning instead of being silently misread without trace. `docs/COMPATIBILITY.md` declares the on-disk interface tiers — (a) stable public (fleet.yaml, service templates, instruction blocks, MCP config), (b) internal persisted state, (c) regenerable/ephemeral — and the additive-only change rule for (a)/(b).
--- a/src/channel/telegram/creds.rs
+++ b/src/channel/telegram/creds.rs
@@ -1,6 +1,7 @@
 //! Telegram credential resolution — loads bot token + group id from fleet.yaml.
 
 /// Resolved Telegram channel credentials — avoids repeated fleet.yaml loads.
+#[derive(Debug)]
 pub(crate) struct TelegramCreds {
     pub(crate) token: String,
     pub(crate) group_id: i64,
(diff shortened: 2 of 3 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ 92006b29

- Symmetric fallback covers all four drift combinations (the previously-broken legacy-fleet + canonical-env shape included), warns only on the fallback path — no spurious warns on direct hits.
- Constant aligned across the whole chain: creds CANONICAL == fleet default == save_env_token == both template arms; the shared generator means #2003's unattended template got the same fix in one edit.
- 4 regression tests through the real `resolve_channel_from` entry, including THE bug pin (old template + canonical-only env — daemon-start death for fresh installs, pre-fix).
- #2003 merge-order resolved via merge commit (5-arg generate_fleet_yaml + fixtures), CI green on the merged head.

Merging (squash) — closes #2005. Fresh-install telegram setup no longer dies at daemon start.

Issue #2008, opened 2026-06-11
[epic] System noise reduction — inventory and fix non-actionable automated messages (renudges, false stuck-alerts, repeat notifications)


suzuke:
Operator: the system currently generates a lot of unnecessary noise; this needs to be treated seriously. Alarm fatigue buries real signals.

**Concrete evidence from one afternoon (2026-06-11)**:
1. **Stale ci-ready renudge**: after a PR head moved, the ci-ready-for-action for the *superseded* head kept renudging the lead every ~2 minutes (~10 repeats over 40 min) until merge. The notification never invalidates when its subject becomes stale.
2. **dispatch_idle false alarms**: 600s threshold fires on agents running full nextest suites (5–8 min is normal) — 3 false stuck-alerts in one day, each costing a pane-inspection round trip. Threshold doesn't account for visible activity (the pane showed active test runs every time).
3. **ci-watch cancelled storm** (#1991, fix in PR #2001 pending deploy): up to 5 duplicate bogus notifications at 60s cadence.
4. **Spurious continue nudges** (#1985, fix in PR #1999 in review): daemon repeatedly nudging an agent that was working fine.

**Design principles to enforce across all automated messaging** (each item gets evaluated against these):
- (a) **Actionable or silent**: if the receiver can't act on it, don't send it.
- (b) **Dedup + latch**: same condition → one message; re-fire only on state change, never on a timer.
- (c) **Escalate, don't repeat**: unacknowledged condition escalates once (different channel/severity), not N identical repeats.
- (d) **Auto-invalidate**: when the subject becomes stale (head moved, task completed, condition cleared), pending repeats must cancel themselves.

**Phase 1 (inventory, read-only)**: enumerate every automated injection/notification/nudge source in the daemon (renudge loop, dispatch_idle watchdog, ci-watch notify, health/escalation, schedule fire, ack-absorption gaps, telegram notify, etc.) with: trigger condition, repeat behavior, dedup/latch presence, staleness-invalidation presence. Classify each against principles (a)–(d).
**Phase 2**: fix the worst offenders (the renudge staleness bug from item 1 is a likely first target), tune thresholds (dispatch_idle vs test-suite duration), and add the missing latches.

Related: #1991 (shipped class-fix), #1985/#1999, the noise-budget clause in the context-full safety-net issue.

suzuke:
[...]

PR #2013, opened 2026-06-11
fix(#2008 Batch A): head-aware ci_handoff_track invalidation (stop renudging a dead head)

suzuke:
#2008 Batch A — the one verify-first-REAL bug from the noise epic: the operator-observed **40-min renudge loop**.

## Mechanism (codex-traced)
`ci_handoff_track` sidecars are keyed `owner/repo@branch` with **no head_sha** (`ci_handoff_track.rs`), recorded when the poller enqueues a `next_after_ci` handoff (`poller.rs:1742`). They resolve only via **report arrival / branch claim / PR-terminal / 24h sweep** — a **PR head move invalidates none of these**, so the old head's ci-ready obligation re-nudges every ~2 min until merge or 24h.

## Fix — make the track head-aware
- **record()** stores the branch head (`pr.current_sha`) at enqueue time: additive `head_sha: Option<String>` with `#[serde(default)]`, so a pre-#2008 track reads back as `None` (COMPATIBILITY tier-b additive — no `schema_version` bump).
- **`resolve_head_advanced(correlation, current_head)`** resolves a track whose recorded `head_sha` no longer matches the branch's current head. Wired into the **ci-watch poll**, which already holds `pr.current_sha` every poll — so within one poll of a head move, the stale track is resolved and the renudge stops.
- A **pre-#2008 track (`head_sha` None) is LEFT ALONE** — we can't tell if it's stale, so the old behavior (other resolve exits + 24h backstop) is preserved (no mis-kill on upgrade).
- The **24h backstop is unchanged** (symptom cap retained as the catch-all).

**Noise discipline:** this only **adds a resolve condition** — it introduces no new re-send path.

## §3.9 tests
`head_advanced_resolves_track`, `head_unchanged_keeps_track`, `pre_2008_track_without_head_sha_not_invalidated` (hand-written old-format JSON with no `head_sha` field), `head_advanced_only_affects_matching_correlation`.

## Batch B (verification, no code)
Checked whether an instance-rename path exists that could bypass `poll_reminder`'s delete cleanup: **it doesn't.** `replace_instance` is in-place same-name (kill + respawn), `set_display_name` is cosmetic metadata only, and there is **no rename action** that changes the instance name key. Nothing to fix.

## Verification
`cargo fmt`, `cargo clippy --all-targets --features tray -D warnings` clean. Full `cargo nextest --features tray` (no bypass): **3918 passed**, 0 failed.

### DIFF
--- a/src/daemon/ci_handoff_track.rs
+++ b/src/daemon/ci_handoff_track.rs
@@ -67,6 +67,13 @@ pub(crate) struct CiHandoffTrack {
     pub correlation: String,
     /// RFC3339 — when the handoff was enqueued (the re-nudge age anchor).
     pub sent_at: String,
+    /// #2008: the branch head (`pr.current_sha`) at record time. `#[serde(default)]`
+    /// → a pre-#2008 track reads back as `None`, which [`resolve_head_advanced`]
+    /// treats as "head-unknown, do not invalidate" (backward-compatible — the
+    /// other resolve exits + 24h backstop still apply). Additive (COMPATIBILITY
+    /// tier-b) — no `schema_version` bump.
+    #[serde(default)]
+    pub head_sha: Option<String>,
 }
 
 fn dir(home: &Path) -> PathBuf {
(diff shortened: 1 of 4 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer (codex): REJECTED → VERIFIED @ c465cfdf

- r1 blocker (§3.9 coverage gap) fixed: two poller-level regressions drive the **real** `run_ci_check`/`ci_check_repo` path with MockCiProvider branch-head data — stale recorded head + advanced provider head resolves the track; unchanged head keeps it. The wiring (correlation key construction, provider PR data, watch gating) is now proven, not assumed.\n- Implementation re-checked: records anchor `Some(&pr.current_sha)`; legacy `None` tracks protected; non-polled/tombstone limitation documented on `resolve_head_advanced` (existing exits + 24h backstop cover those).\n- Reviewer independently re-ran both tests and verified clean merge-tree against current main.\n\nThe operator-observed 40-minute stale-renudge loop (and the two I manually resolved today) ends with this. Merging (squash) — #2008 Batch A done, epic stays open for Phase 2 threshold tuning.

PR #2014, opened 2026-06-11
feat(#1523): promote hook-resolution to authoritative for STRONG backends

suzuke:
The day's capstone: flip the #1982 freshness-resolution layer (`resolved_state_for`) from deferred dead-code to the **authoritative** agent-state source for STRONG backends — gated, with the screen heuristic as the complete fallback.

Clean shadow data (12:56–17:00, 2636 samples) met the operator's promotion condition: noise classes zeroed, main signal 7× stable, **zero hook false-positives all day**.

## Design
- **`hook_shadow::authoritative_state(backend, name, heuristic)`** — when `AGEND_HOOK_STATE_POC=1` **and** the backend is STRONG (claude/agy), a `Fresh` hook resolution **wins**; `Stale` / `Unknown` / flag-off / non-hook backend fall back to the heuristic — **byte-identical** to pre-promotion. Wired at the **single chokepoint** (`per_tick::snapshot`), so every downstream consumer (`agent_state_of`, watchdogs, pane badge) inherits the authoritative state from one place.
- **`Backend::has_state_hooks()`** — STRONG = `ClaudeCode | Agy` (the only backends whose lifecycle hooks `mcp_config` injects; others never fire hooks → `Unknown` → heuristic).
- **ToolUse is EVENT-PAIR-closed, not freshness-bounded** (decision `d-20260611051442661249-0`): `PreToolUse` opens it; only `PostToolUse`/`Stop` closes it (by overwriting the snapshot). A long (>`HOOK_FRESHNESS`) tool is therefore **not** demoted to the screen heuristic mid-run — exactly the #1985 nudge class the promotion fixes. A crashed-mid-tool agent is reaped by the liveness watchdog independently, so leaving ToolUse open is safe — the bound is the event, not the clock.
- **`#hook-shadow` log unchanged** — keeps collecting agreement data post-promotion to verify real authoritative behavior.

## Multi-backend / fallback floor
The fallback chain is the floor — hooks **enhance**, never gate. A non-hook backend or a stale window is never worse off than the heuristic-only path. Flag off ⇒ literally the old code path.

## §3.9 tests
`flag_off_is_byte_identical_to_heuristic`, `claude_fresh_hook_wins_over_heuristic`, `stale_hook_falls_back_to_heuristic`, `long_tool_stays_tooluse_past_freshness` (PreToolUse +10min past window stays ToolUse, then PostToolUse closes it), `backend_strength_gates_promotion` (codex heuristic-only, agy strong). Tests use a split `authoritative_state_inner(enabled, …)` so the flag/backend/freshness logic is exercised without mutating the process-global env var.

## Verification
`cargo fmt`, `cargo clippy --all-targets --features tray -D warnings` clean. Full `cargo nextest --features tray` (no bypass): **3926 passed**, 0 failed (flag off in the suite ⇒ byte-identical, confirming no snapshot regression).

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -56,10 +56,15 @@ impl Backend {
     /// and that therefore emit authoritative `hook_shadow` state events. Only
     /// these have hook data to PROMOTE over the screen heuristic; every other
     /// backend always uses the heuristic (no hooks fire, so `resolved_state_for`
-    /// stays `Unknown` → heuristic fallback). claude + agy (claude-family
-    /// settings consumers).
+    /// stays `Unknown` → heuristic fallback).
+    ///
+    /// **claude-only in v1.** `configure_agy` does NOT inject the hooks, and
+    /// production data confirmed it (2026-06-11: ai-scout/agy emitted **0** hook
+    /// events all day, including during a real 16:14 task run; all 3,490 events
+    /// came from claude). Re-adding Agy requires the injection implementation AND
+    /// shadow-data evidence that its hooks fire — not just the enum membership.
     pub fn has_state_hooks(&self) -> bool {
-        matches!(self, Backend::ClaudeCode | Backend::Agy)
+        matches!(self, Backend::ClaudeCode)
     }
 
     /// #1440: credential env-var names this backend legitimately needs to
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -52,6 +52,16 @@ impl Backend {
         }
     }
 
+    /// #1523: STRONG backends — those whose lifecycle hooks `mcp_config` injects
+    /// and that therefore emit authoritative `hook_shadow` state events. Only
+    /// these have hook data to PROMOTE over the screen heuristic; every other
+    /// backend always uses the heuristic (no hooks fire, so `resolved_state_for`
+    /// stays `Unknown` → heuristic fallback). claude + agy (claude-family
+    /// settings consumers).
+    pub fn has_state_hooks(&self) -> bool {
+        matches!(self, Backend::ClaudeCode | Backend::Agy)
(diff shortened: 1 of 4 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: REJECTED → VERIFIED @ cd283ebe (+ serial-group test fix @ b97487c4 per reviewer's own prescription)

Adversarial review confirmed the promotion mechanics and forced three honesty fixes before landing:
- **Scope truth**: the "single chokepoint" claim was false — promotion is **snapshot-scoped phased v1** (covers dispatch_idle/pane/agent_state_of — the #1985 surface); per-tick deciders (supervisor #1946, hang detection, recovery dispatcher, watchdog, conflict_notify, query API) still read raw heuristic, now documented with the full list. Worst-case divergence is independently bounded by #1999's evidence gate + health-gating. Supervisor-side promotion is #1523 phase-2, post-soak.
- **Backend gate truth**: Agy dropped — production data showed 0 agy hook events all day (3,490 all claude) while configure_agy injects nothing; claude-only v1, re-add criteria documented (injection implemented AND shadow-proven).
- **ToolUse no-clock-backstop** documented: dropped-close vs long-tool is indistinguishable by state; a clock bound would re-admit the #1985 failure mode; mitigation chain = Stop-per-turn-end + SessionStart clear + ongoing #hook-shadow visibility.
- New env-based integration test covers the real flag gate; cross-module env race fixed via named `#[serial(hook_state_poc)]` group.

Backed by two rounds of shadow data (375 + 2,636 events, zero hook misfires). Flag-gated: off = byte-identical. Merging (squash) — #1523 epic stays open for phase-2.

PR #2017, opened 2026-06-11
fix(#1990 item 2 + #2008 #8): corrupt store — robust backup + operator-visible surface

suzuke:
#1990 item 2 (corrupt versioned store ≠ silent whole-file wipe) + the #2008 #8 addendum (verified REAL). A corrupt store still resets to DEFAULT — per-record rescue is deferred — but it is no longer silent or data-losing-on-backup-failure.

## #2008 #8 — the backup was a lie on failure
`store.rs` backed up via `let _ = std::fs::copy(path, &backup)`, swallowing the failure. When the copy failed, the "backing up corrupt file" warn was **false**: the corrupt file stayed on the live path and the **next save overwrote it → permanent loss**.

New `backup_corrupt_file(path, backup) -> bool`:
- prefer an **atomic `rename`** — moves the corrupt bytes OFF the live path, so the next save can't race-overwrite them;
- fall back to `copy` across filesystems (EXDEV);
- on **total failure**: log at ERROR + **leave the original in place** (the load path never writes/overwrites it — no silent loss, the warn is honest).

## #1990 item 2 — operator visibility
Corrupt → whole-store reset (schedules/deployments/health emptying) was **silent**. New `handle_corrupt_store`: ERROR-log every occurrence (greppable) + **one `event_log` entry per (boot, path)** (the #1972/#2002 latch) — *"store X corrupt, backed up to Y, running empty"*. Loss semantics unchanged; the **save path is untouched** (no added complexity there).

Both corrupt sites — `load` (full `T` decode) and `load_versioned` (the `Value` peek) — route through the one helper.

## §3.9 tests
- `corrupt_versioned_store_renames_backup_and_surfaces` — corrupt → default, bytes RENAMED to a backup (original gone from the live path), one `store_corrupt` event_log entry.
- `backup_corrupt_file_total_failure_keeps_original` — backup target is a directory → rename + copy both fail → returns false + the original stays in place.
- Existing `test_load_corrupt_creates_backup` / `test_corrupt_json_returns_default` / `test_load_versioned_*` still pass.

## Verification
`cargo fmt`, `cargo clippy --all-targets --features tray -D warnings` clean. Full `cargo nextest --features tray` (no bypass): **3954 passed**, 0 failed.

Part of #1990 (item 2; items 1 + the 2 deferred free-form bags remain).

### DIFF
--- a/src/store.rs
+++ b/src/store.rs
@@ -13,8 +13,11 @@ static CORRUPT_SURFACED: std::sync::OnceLock<std::sync::Mutex<std::collections::
 
 /// #2008 #8: move a corrupt store file to `backup` before the caller falls back
 /// to DEFAULT. Prefer an atomic `rename` — it takes the corrupt bytes OFF the
-/// live path, so the next `save` cannot race-overwrite them; fall back to `copy`
-/// across filesystems (EXDEV). Returns whether the bytes were preserved. The
+/// live path, so the next `save` cannot race-overwrite them. `copy` is a
+/// best-effort second shot for a generic rename failure: `backup` is the original
+/// path with a swapped extension (same directory → same filesystem), so EXDEV
+/// can't happen — a rename failure here is a permission / I/O error, which the
+/// copy may or may not survive. Returns whether the bytes were preserved. The
 /// prior `let _ = std::fs::copy(...)` swallowed the failure, making the "backing
 /// up" warn a LIE and letting the next save destroy the only copy — so a total
 /// failure now logs at ERROR.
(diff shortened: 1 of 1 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ 04a59baf; comment-only delta (56d749d9) self-reviewed

- **Backup lie fixed** (closes the verified #2008 #8 finding): `backup_corrupt_file` prefers atomic rename (corrupt bytes leave the live path — the next save can't clobber them), generic-failure copy as best-effort second shot, both-fail → ERROR + event + original left in place. The "no *silent* loss" boundary is stated precisely — a both-fail file is still lost on the next mutate, but loudly.
- **Visibility** (#1990 item 2): ERROR on every corrupt load + once-per-(boot,path) operator event — finer-grained than #2002's latch.
- Rename semantics traced safe: file locks live on `<path>.lock` (untouched), all `.exists()` callers are pre-load or different paths, save_atomic creates fresh files identically.
- Review findings (false EXDEV comment, test-trigger stand-in note) fixed in the comment-only delta.

CI green. Merging (squash) — #1990 down to item 1 (migration hook, deliberately deferred) + two documented bag exceptions.

PR #2021, opened 2026-06-11
fix(#2020): false AwaitingOperator after restart — opencode resumed-idle pattern + productive-output veto

### LINKED ISSUE #2020: False AwaitingOperator latch after restart: startup-stall fallback + hook-staleness fallback re-exposes it (first live #1523 phase-2 sample)
Observed live 2026-06-11 20:09 (operator asked why fixup-lead showed AwaitingOperator while working normally).

**Chain**:
1. Daemon restart respawns agents; a busy agent (injected work immediately) never renders the clean ready-prompt the heuristic's ready-pattern expects → heuristic stays `Starting`.
2. The startup-stall fallback (`set_awaiting_operator`, `src/state/mod.rs:1965`) fires → forced `AwaitingOperator` (transition logged 12:09:52Z, `starting → awaiting_operator`).
3. Hook promotion (#2014) can't correct it between turns: hooks only fire during activity; >600s after the last Stop event the hook state goes Stale → `authoritative_state` falls back to the heuristic → the forced AwaitingOperator resurfaces whenever the agent is idle.
4. Result: badge (and `is_blocked` semantics) oscillate — correct during turns, false AwaitingOperator when idle.

**Fix directions** (pick during #1523 phase-2):
- (a) A fresh hook `Stop → Idle` should qualify as the "ready" evidence that lifts AwaitingOperator (today only the screen ready-pattern does, via `transition()` priority). Cheap and targeted.
- (b) For hook-capable backends, the startup-stall fallback could require *no recent hook activity* before forcing AwaitingOperator (a SessionStart + UserPromptSubmit stream is proof the agent isn't stalled at a login prompt).
- (c) Larger: promote the supervisor-side deciders (the phase-2 umbrella).

(a)+(b) are both narrow and noise-reducing; the false-blocked state also feeds G.3 AwaitingOperator notifications (#2008 inventory) so this is a noise source too.

suzuke:
## Summary

Fix #2020 — both operator-caught restart shapes traced to one root: **after a daemon restart the ready-pattern fails to match for respawned agents** → `Starting` lingers past the 30s silence gate → the startup-stall fallback (whose `Starting` arm has NO position/stability/engagement gates — only silence + role) forces a false `AwaitingOperator`.

### Shape 1 — idle opencode (fixup-reviewer-3, 3/3 restarts today)
**Spike answer (Q1)**: a respawned pane *resuming a session* renders **no `Ask anything` placeholder** — the input box is bare `┃` lines (captured live from the stuck agent). Both existing Idle patterns key on that placeholder → Idle never matches.
**Fix**: lowest-priority Idle pattern on the persistent statusline hint `ctrl\+p commands`. The statusline persists during work/perm states, so **first-match order protects it** (`esc interrupt` / tool markers / perm chrome precede it in the vec) — pinned by a unit test on the live capture + a working-pane contrast test, and **grid-validated by the full opencode replay-fixture suite** (#1559 discipline: `ctrl+p` IS raw-present in the thinking/tooluse/perm fixtures; every expected transition sequence still passes).

### Shape 2 — busy claude (fixup-lead, 20:09)
**Spike answer (Q2)**: the `Starting` arm's only gates are silence>30s + role — a busy respawned agent (injected work immediately, never renders the pristine prompt) sails through during any 30s output lull.
**Fix**: veto the `Starting` arm when the agent has rendered **backend productive markers since this spawn**. Deviation from the issue's "received inject + any output" suggestion, deliberately: a *real* login prompt can **echo** injected text (= output!) but never renders tool chrome — markers keep the fallback's actual job (true login-stall detection) intact, the under-force-never-falsely-lift direction the dispatch required. `last_productive_output` is in-memory per spawn (StateTracker resets on respawn) → no cross-life leakage.

### Deferred (documented in the commit)
Hook-side (a)/(b) from the issue: with the veto, hook-backend agents no longer false-latch **at the source**, and the heuristic fix covers non-hook backends — (a)/(b) belong to the #1523 phase-2 umbrella.

## Tests
- `resumed_idle_pane_detects_idle` — the live-captured stuck pane tail → `Idle`.
- `working_pane_with_statusline_still_detects_thinking` — order pin.
- `starting_stall_vetoed_by_productive_output_2020` / `starting_stall_still_fires_without_productive_output_2020` — veto + legacy-fire guard on the guard.
- `replay_manifest_regression` green across all opencode fixtures (grid validation).

## Preflight
fmt / clippy(tray) `-D warnings` clean; nextest **3975/3976** (the 1 = known #1834 environmental, passes solo with bypass; `team_dedup` excluded per #1993).

Closes #2020

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Format based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/); projec
 
 ### Fixed
 
+- **False AwaitingOperator after daemon restart (#2020)** — two restart shapes forced a healthy agent into `AwaitingOperator`: an idle respawned opencode pane resuming a session renders no `Ask anything` placeholder (the Idle pattern never matched → `Starting` lingered → the startup-stall fallback fired, 3× live), and a busy respawned agent injected work immediately never renders the clean ready-prompt (same fallback during a silence window). Fixes: the opencode profile gains a lowest-priority Idle pattern on the persistent statusline hint (`ctrl+p commands`; working/error patterns still win first-match — verified against the full opencode replay-fixture suite), and the startup-stall arm is vetoed when the agent has rendered backend productive markers since this spawn (a real login prompt never renders tool chrome, so the fallback's actual job is preserved; echoed injected text does not count).
+
 - **Fresh Telegram setup resolves at startup (#2005)** — the quickstart fleet.yaml template pinned the legacy `bot_token_env: AGEND_BOT_TOKEN` while `.env` was written under the canonical `AGEND_TELEGRAM_BOT_TOKEN`, and the credentials fallback retried the same legacy name — so a fresh install's Telegram channel failed to resolve at daemon startup (old installs were masked by leftover legacy `.env` keys). The template now pins the canonical name and the fallback is symmetric (configured name → the other of canonical/legacy), so all four fleet/.env drift combinations resolve; legacy use still warns.
 
 ### Added
--- a/src/backend_profile.rs
+++ b/src/backend_profile.rs
@@ -227,6 +227,17 @@ fn opencode_profile() -> BackendProfile {
             ),
             (AgentState::Idle, r"Ask anything"),
             (AgentState::Idle, r"Ask anything|tab agents"),
+            // #2020: a RESPAWNED opencode pane resuming a session renders NO
+            // "Ask anything" placeholder (the input box is bare `┃` lines) —
+            // the only stable idle chrome is the bottom statusline hint. Listed
+            // LAST so every working/error pattern above wins first-match (the
+            // statusline persists during Thinking/ToolUse/PermissionPrompt —
+            // `esc interrupt` / tool markers / perm chrome match first; grid
+            // validation = the opencode replay fixtures, #1559 discipline).
+            // Without this, a restarted idle opencode agent never leaves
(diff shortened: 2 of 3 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ 0faeaaa9 (state-detection domain review)

- `ctrl+p commands` Idle pattern is provably lowest-priority (vec position verified; ToolUse/Thinking patterns intercept first, order-pinned + grid-validated against all opencode fixtures per #1559).
- The veto's safety direction holds by string-level proof: every backend's productive markers are file-op outputs + tool chrome — things a real login/auth screen never renders, so a login prompt echoing injected text cannot falsely lift the stall detection. Under-force-never-falsely-lift, as required.
- Both observed shapes covered: idle-opencode-resume (pattern) and busy-claude (veto); live-captured stuck pane is the test fixture.

Two non-blocking phase-2 notes (tracked in #1523/#2020): the always-present fallback shifts uncovered working states from false-AwaitingOperator to the more recoverable false-Idle — making opencode working-pattern completeness worth an audit; and the Starting arm could gain the position/stability gates the prompt arms already have, for the theoretical working-without-markers shape.

CI fully green. Merging (squash) — closes #2020. **Acceptance check after next restart: reviewer-3 should show Idle, not AwaitingOperator.**

PR #2026, opened 2026-06-11
fix(#1986): systemd_quote escapes % → %% + per-template no-residual-placeholder round-trips

### LINKED ISSUE #1986: Harden service template substitution: systemd_quote %→%% + no-leftover-placeholder install guard
Follow-up from #1984 review (both pre-existing / non-blocking findings, surfaced while reviewing the PATH propagation fix):

1. **`systemd_quote` does not escape `%` → `%%`** (`src/service/mod.rs:133` — match only handles `\\` and `"`). systemd performs specifier expansion inside `Environment=`/`ExecStart=`, and quoting does **not** suppress it — a literal `%` in a substituted value (PATH, exe path, home) would be mis-expanded (`%h` → home, etc.). Input is self-controlled (installer's own env), so no injection risk — pure correctness hardening. Fix covers the pre-existing `__EXECUTABLE__`/`__HOME__` exposure too.

2. **No install-level "no leftover placeholder" round-trip guard**: `apply_substitutions` is tested generically, but nothing asserts that a rendered template contains no remaining `__*__` placeholders. If a future template adds a placeholder the substitution array misses, nothing catches it. One test per template (render with dummy values → assert no `__` substring) closes the class.

Small, well-scoped, good first issue. @cheerc feel free to take it if interested — your #1984 touched exactly this code.

suzuke:
## Summary

Fix #1986 (both #1984-review findings, taken over per the 8h-silence policy — lead noting on the issue):

1. **`systemd_quote` doubles `%` → `%%`**: systemd expands specifiers (`%h`, `%i`, …) inside `Environment=`/`ExecStart=` and quoting does **not** suppress it — a literal `%` in PATH/exe/home was silently mis-expanded. `%` is already outside the quote-trigger safe set, so percent-only values route through quoting correctly.
2. **Per-template no-residual-placeholder round-trips**: one test per template (systemd / launchd / windows task XML), each rendering with the **exact key set its production installer passes** → asserts no `__` substring remains. Closes the "future placeholder misses the substitution array → literal `__NEW_THING__` ships" class.

## Tests
`systemd_quote_doubles_percent_1986` (+ quote-trigger pin) and the three template round-trips — 35/35 in the service area.

## Preflight
fmt / clippy(tray) clean; nextest 3985/3986 (the 1 = known #1834 environmental, bypass-solo passed; team_dedup excluded per #1993).

Closes #1986

### DIFF
--- a/src/service/mod.rs
+++ b/src/service/mod.rs
@@ -143,6 +143,12 @@ pub fn systemd_quote(value: &str) -> String {
         match ch {
             '\\' => out.push_str(r"\\"),
             '"' => out.push_str(r#"\""#),
+            // #1986: systemd performs SPECIFIER expansion (%h, %i, …) inside
+            // Environment=/ExecStart= values and quoting does NOT suppress it —
+            // a literal `%` must be doubled or e.g. `%h` silently becomes the
+            // home dir. Input is installer-controlled (no injection), pure
+            // correctness for exe/home/PATH values that contain `%`.
+            '%' => out.push_str("%%"),
             other => out.push(other),
         }
     }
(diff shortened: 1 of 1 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ ee4c9da5

Closes out the reviewer's own two #1984 findings: `systemd_quote` now doubles every `%` (specifier expansion isn't suppressed by quoting; percent-only values verified through the quoting path), and each of the three service templates gets a round-trip test rendered with its installer's **exact production key set** (key sets cross-checked against linux/macos/windows install code — the future-placeholder-ships-literally class is closed). CI fully green. Merging (squash) — closes #1986.

PR #2029, opened 2026-06-11
fix(#2028): drain contention is Unavailable, not empty — single-shot callers retry

### LINKED ISSUE #2028: flaky: drain_one_pops_oldest_leaves_rest — second pop returns None under llvm-cov parallelism (non-blocking drain lock turns contention into false-empty)
First observed 2026-06-11 ~14:00Z on PR #2026's Coverage job (run 27352063313) — a branch that does NOT touch notification_queue (service-template hardening). The #2006 signature-gated retry correctly labeled it REAL (no profraw signature) and failed fast:

```
test notification_queue::tests::drain_one_pops_oldest_leaves_rest ... FAILED
thread '…drain_one_pops_oldest_leaves_rest' panicked at src/notification_queue.rs:846:42:
second pop
test result: FAILED. 3682 passed; 1 failed
```

The test (landed today via #1978's carry, part of the drain claim-atomicity work): enqueue 3 → `drain_one` ×3 expects first/second/third. The failure is the SECOND `drain_one` returning `None`.

## Evidence

- CI: failed under `cargo llvm-cov` (Coverage job = plain `cargo test`, full thread-parallelism, instrumented, slow runner).
- Local: 5× consecutive `cargo test --bin agend-terminal notification_queue::tests` → 28/28 green each time (light load). Timing-sensitive, not deterministic.

## Mechanism pointers (for the verify-first owner)

`drain_one` → `drain_with_stale_threshold` (notification_queue.rs:405):
1. `try_acquire_file_lock(drain_lock_path)` is **non-blocking by design** — `None` ("held by a peer") returns EMPTY and the comment says "caller retries next tick". A `drain_one` caller has no next tick: empty = `None` = the test's "second pop" panic. Under llvm-cov full-parallelism, anything that briefly holds (or slows the release of) that per-agent lock between pop #1's drop and pop #2's acquire turns into a false-empty. Candidate: pop #1's lock-file release latency on a loaded runner (flock drop is in-process, but the lock FILE handling / FLOCK_DEPTH guard interplay under instrumented threads is unverified).
2. Alternative: pop #1's `requeue_all` (append to a fresh queue file) racing pop #2's `path.exists()` — fs visibility should be synchronous in-process, so (1) is the stronger suspect.

Note the PROD impact of the same semantics: the TUI/daemon flush callers tolerate false-empty (next tick retries), so this is primarily a TEST-determinism issue — but `drain_one`'s Abandoned-trickle caller would silently skip a release cycle under the same race. Low severity, real.

## Fix directions

- Test-side: drive `drain_one` through a small retry (or use the `stale_ms`-injected deterministic variant) — but don't paper over (1) without confirming the lock is the mechanism.
- Or make `drain_one` distinguish "queue genuinely empty" from "lock unavailable" (`Option<Option<…>>` / retry-once) so single-shot callers aren't lied to.

suzuke:
## Summary

Fix #2028 — the drain layer collapsed **three** distinct "could not claim" shapes into an empty `Vec`: drain lock held by a peer, lock file unopenable (`Err` — e.g. fd pressure under llvm-cov-grade load), and a failed claim-rename over a still-existing queue file. Flushers tolerate that (next-tick retry); single-shot callers (`drain_one` = the Abandoned-trickle release + tests) believed the lie — the #2026 Coverage run caught the second pop returning `None`.

### Design (per dispatch)
1. **"True empty" vs "contended" distinguished**: new core `try_drain_with_stale_threshold` → `DrainAttempt::{Drained, Unavailable}`. `Unavailable` = lock held/unopenable OR claim-rename failed with the queue still present. Consumed stale leftovers are always **delivered** (a rename failure after leftover recovery returns `Drained` with them — claimed lines are never dropped).
2. **Flusher behavior untouched (byte-identical)**: `drain`/`drain_with_stale_threshold` collapse `Unavailable` → empty, pinned by `flusher_drain_still_collapses_contention_to_empty_2028`.
3. **`drain_one` bounded retry**: 5 × 10ms on `Unavailable`, then loud warn + `None` (defers to the next trickle cycle — never claims empty). The original non-blocking rationale (no double-drain dead-wait) is preserved: the wait is bounded at 50ms against a microseconds-scale healthy hold. True empty stays an immediate `None` (pinned with a latency assert).
4. **Bonus loss-visibility**: `requeue_all`'s swallowed append is MESSAGE LOSS (items already claimed out of the queue) — now error-logged with the lost text head (#2004 discipline, zero behavior change).

### Determinism (the #1909 routine)
The previously-flaky test's whole module ran **10× consecutive green** under the same harness shape that failed in CI (single-process threaded `cargo test`, the llvm-cov form). Plus 4 new §3.9 tests: Unavailable-while-held → drains-after-release; flusher-collapse pin; `drain_one` outlasting a 15ms contention window (wide margin vs the 50ms budget); true-empty immediate-`None`.

## Preflight
fmt / clippy(tray) `-D warnings` clean; nextest **3990/3991** (the 1 = known #1834 environmental, bypass-solo passed; team_dedup excluded per #1993).

Closes #2028

### DIFF
--- a/src/notification_queue.rs
+++ b/src/notification_queue.rs
@@ -256,15 +256,39 @@ pub fn input_box_dim_aware_empty(text: &str, dim: &[bool], marker: &str) -> Opti
 /// Routes through the claim-atomic `drain` (then requeues the tail) so a
 /// concurrent flusher can never read the same lines mid-rewrite.
 pub fn drain_one(home: &Path, agent_name: &str) -> Option<QueuedNotification> {
-    let mut all = drain(home, agent_name);
-    if all.is_empty() {
-        return None;
+    // #2028: drain_one is a SINGLE-SHOT caller (the Abandoned-trickle path and
+    // tests) — unlike the flushers it has no "next tick" to absorb a transient
+    // false-empty, so a lock/claim hiccup must be retried, not reported as
+    // "queue empty". Bounded (no double-drain dead-wait regression): a live
+    // peer holds the drain lock for microseconds (rename + read), so a few
+    // short retries comfortably outlast any healthy contention window.
+    const RETRIES: u32 = 5;
+    const RETRY_SLEEP: std::time::Duration = std::time::Duration::from_millis(10);
+    for attempt in 0..=RETRIES {
+        match try_drain_with_stale_threshold(home, agent_name, STALE_DRAINING_MS) {
+            DrainAttempt::Drained(mut all) => {
+                if all.is_empty() {
+                    return None; // TRUE empty — claim succeeded, nothing queued.
+                }
+                let oldest = all.remove(0);
+                if !all.is_empty() {
+                    requeue_all(home, agent_name, &all);
+                }
+                return Some(oldest);
+            }
+            DrainAttempt::Unavailable => {
+                if attempt < RETRIES {
+                    std::thread::sleep(RETRY_SLEEP);
+                }
+            }
+        }
     }
-    let oldest = all.remove(0);
-    if !all.is_empty() {
-        requeue_all(home, agent_name, &all);
-    }
-    Some(oldest)
+    tracing::warn!(
+        agent = agent_name,
+        "#2028: drain_one gave up after {RETRIES} contended attempts — \
+         deferring to the next trickle cycle (not claiming empty)"
+    );
+    None
 }
 
 pub fn enqueue(home: &Path, agent_name: &str, text: &str) -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ 2858db37 (notification_queue domain review)

- Three fake-empty shapes (lock held / lock-open error / rename-failed-queue-present) correctly unified as `Unavailable`, distinct from true-empty. Claimed-rows-never-dropped verified at lock-scope level: leftover consumption and the rename failure live under the same held lock, the rename-fail branch delivers consumed leftovers and leaves queue rows in the file for the next round — no reverse lie either direction.
- Flusher behavior byte-identical (both TUI and daemon callers go through the collapsing `drain`, pin-tested); only the single-shot `drain_one` gains the bounded retry (worst-case 50ms, negligible vs the ~10s trickle cadence and #1784 hang thresholds).
- Bonus: `requeue_all` swallowed-append (= outright message loss) now error-logs the lost text head.

One non-blocking note for #2008 phase-2: the drain_one give-up warn has no latch, so a pathological *persistent* Unavailable (permanent fd issue / stuck holder) would warn once per trickle cycle — rare, bounded, latch later.

CI fully green incl. determinism evidence (10× same-harness runs + contention-window test). Merging (squash) — closes #2028.

Issue #821, opened 2026-05-15
Test isolation invariant: fixture subprocess git ops can pollute host worktree .git


suzuke:
## Symptom

Tests using subprocess `git` commands (e.g. `git checkout -b feat-b` in fixture setup) under specific conditions inherit the parent process's cwd and **modify the host worktree's `.git` directory** instead of an isolated test repo. This was caught during PR #820 (#817 fix) when a test fixture's `git checkout -b feat-b` ended up creating `feat-b` branch on the *real* worktree, causing a subsequent C2 commit to land on `[feat-b ad6b413]` instead of `[fix/817-... ...]`.

Recovery required:
- Detect the contamination (commit landed on wrong branch)
- `git cherry-pick <stray-commit-on-feat-b>` onto the intended branch
- `git branch -D feat-b` to delete the polluting branch

This is a **test correctness + host-environment safety bug**. Tests should never be able to modify state outside their fixture's temp directory.

## Repro (observed 2026-05-15 ~09:00 UTC)

During fixup-team's PR #820 (issue #817 fix) implementation, dev's test fixture for the new `branch_sweep` module included subprocess `git` calls (likely something like `Command::new(\"git\").args([\"checkout\", \"-b\", \"feat-b\"]).output()`). The `Command::new` inherits the **current process's cwd** by default. If the test was run from inside an active worktree (i.e. `cargo test` invoked from `/Users/suzuke/.agend-terminal/worktrees/.../`), the subprocess `git` operated on that worktree's `.git`, not the test fixture's temp `.git`.

Lead's report (verbatim, m-20260515090157384386-218):

> Note 2 — TEST FIXTURE ISOLATION INVARIANT:
> - Dev's test fixture's `git checkout -b feat-b` somehow polluted the WORKTREE'S branch (not temp test repo)
> - Discovered when C2 commit landed as `[feat-b ad6b413]` instead of `[fix/817 ...]`
> - Recovered via cherry-pick + branch -D
> - Root cause: subprocess cwd inheritance quirk under specific conditions
> - Mitigation: tests should use `GIT_DIR` env var to fully isolate from cwd
> - Worth backlog issue: \"tests in src/<...>/tests MUST NOT modify the worktree's .git\"

### Layer 1 — fixture authoring guidance
Add to test-writing docs: any test fixture spawning `git` subprocess MUST do one of:
- `Command::new(\"git\").current_dir(&temp_dir)...` — explicit cwd, never inherit
[...]

PR #2030, opened 2026-06-11
fix(#2027): non-fleet repo `git branch <name>` passes through, never lies success

### LINKED ISSUE #2027: git shim: no-bypass 'git branch' in non-fleet repos silently no-ops with exit 0 (success lie) or fakes 'already exists' from global state — loud-deny violation
Found while debugging a #2011 test (and initially mis-attributed — see "correction" below). The agend git shim's **no-bypass handling of `git branch <name>` violates loud-deny**: in a repo the fleet doesn't manage, it either silently no-ops with **exit 0** (a success lie) or fakes a git error from the wrong scope. The bypass contract itself is fine (`AGEND_GIT_BYPASS=1` passes everything through correctly — an earlier "bypass doesn't help" claim was wrong).

## Repro (deterministic, macOS, shim = `~/.agend-terminal/bin/git` first on PATH)

```sh
# fresh NON-fleet repo
T=$(mktemp -d); cd $T
AGEND_GIT_BYPASS=1 git init -q -b main repo && cd repo
AGEND_GIT_BYPASS=1 git commit -q --allow-empty -m init

# Shape 1 — never-seen branch name, NO bypass:
git branch zz-never-used-$(date +%s)
echo $?                       # → 0  (!!)
AGEND_GIT_BYPASS=1 git branch -a   # → only `main` — branch was NOT created
# SILENT NO-OP WITH SUCCESS EXIT — any `&&` chain proceeds on a lie.

# Shape 2 — name the shim has seen before (e.g. created earlier via bypass elsewhere):
git branch feat-x
# → fatal: a branch named 'feat-x' already exists   (exit 128)
# …in a repo that has NO feat-x. Cross-repo/global state masquerading as git.
```

Contrast the CORRECT loud-deny the same shim produces for `worktree add` without bypass:
```
agend-git: ERROR git worktree denied
           agent=fixup-dev, reason: fleet-managed — use agend-terminal worktree tools
           HINT: … AGEND_GIT_BYPASS=1 …
```
(exit 1, honest, actionable). `git branch` should behave the same way — or pass through entirely for repos the fleet doesn't manage.

## Bypass-state matrix (verified today)

| command | no bypass | AGEND_GIT_BYPASS=1 |
|---|---|---|
| `git branch <new>` (non-fleet repo) | **silent no-op, exit 0** | ✓ creates |
| `git branch <seen-before>` (non-fleet repo) | **fake `fatal: already exists`, exit 128** | ✓ creates |
| `git worktree add` (non-fleet repo) | loud agend-git denial, exit 1 ✓ | ✓ works |

## Impact assessment

- **Production daemon paths: unaffected** — `git_helpers::git_bypass` sets the env on every call.
- **Agent/operator shells + scripts + tests**: any plain `git branch` in a shim-on-PATH shell silently does nothing while reporting success; `&&` chains continue and fail later with a CONFUSING downstream error (today's bite: a test's `git branch feat-x && git worktree add … feat-x` → "fatal: invalid reference: feat-x", which sent the debugging toward worktree-add for three rounds).
- **Non-fleet repos**: the shim applies fleet semantics (and apparently fleet-global branch-name state) to repos it doesn't manage at all — scope leak.

## Correction (for the record)

The earlier in-flight report "shim mangles `worktree add`, bypass doesn't help" was a mis-attribution: the worktree-add failure was downstream of Shape 1's silent no-op (branch never existed), and a separate doubled-`#[test]` artifact muddied the repro. Verified today: worktree add + bypass behave correctly.

## Fix directions (verify-first triage tomorrow, per lead)

1. `git branch` no-bypass arm: same loud agend-git denial as worktree (exit 1, honest message). Never exit 0 without doing the work.
2. Scope check: commands in repos without fleet markers (no `.agend-managed`, not the canonical repo, not under the worktrees root) should pass through untouched.
3. The "already exists" fake: whatever registry produced it must not answer for foreign repos.

suzuke:
Closes #2027.

## The bug (deterministic repro in the issue)

The shim groups `branch`/`tag` with the **read-only** commands (they default to
*listing*), so a **bound** agent's ref-naming `git branch <name>` is
`ChdirPass`'d into its worktree. In a **foreign** repo (a separate object store —
e.g. a test scratch repo) that is a **success-lie**:

| shape | command | before | after |
|---|---|---|---|
| 1 | `git branch <new>` (foreign, no bypass) | **silent no-op, exit 0** (create ran against the worktree) | creates in the foreign repo ✓ |
| 2 | `git branch <name-the-worktree-has>` | **`fatal: already exists`, exit 128** from the wrong repo | creates in the foreign repo ✓ |

Shape 1's exit-0 poisons `&&` chains; shape 2 leaks the worktree's ref state.

## Root cause (verify-first correction to the issue)

Both shapes are the **worktree-redirect**, *not* a separate "global registry" as
shape 2's framing suspected — the `already exists` is real `git` answering from
the redirected worktree. `apply_foreign_repo_passthrough` already flips a foreign
`ChdirPass`→`Passthrough` for mutating-local commands (#1463), but `branch`/`tag`
were never in `is_mutating_local`, so the conversion never reached them.

## Fix — lead's "non-fleet repo → real git" direction

Extend the foreign-passthrough to a **ref-naming** `branch`/`tag`
(`branch_tag_names_ref`: a positional ref name, or a `-d`/`-D`/`-m`/`-M`/`-c`/`-C`
delete/move/copy flag — the bare LIST form is excluded). Gated on
`cwd_is_foreign_repo` (commondir-identity, **fail-closed**: canonical + every
sibling worktree share canonical's commondir, so `true` means a *genuinely*
separate store — the lead's "how reliable is the detection?" concern is answered
by this existing detector). **Deny is never needed** because passthrough is
honest — the iron rule (deny is OK, lie is NOT) holds with the cleaner option.

Conservative by design: over-classifying (`git branch --list <pattern>`) only
adds the *correct* foreign-passthrough; under-classifying is the bug.

## Fleet behavior is byte-identical

`cwd_foreign=false` (worktree / canonical / siblings) → not converted → unchanged
`ChdirPass`. The bare LIST form stays `ChdirPass` even when foreign (read-only, no
lie). Only genuinely-foreign repos change. The shim is the #821/#1463 security
mechanism — the interception path is untouched.

## §3.9 tests (3, pinning the full classify→apply chain)

- `branch_tag_names_ref_classifier_2027` — create/delete/move vs the LIST form
- `foreign_passthrough_branch_tag_matrix_2027` — issue shapes 1+2 + tag + fleet-byte-identical + bare-list
- `bound_branch_create_classifies_to_chdirpass_2027` — chain precondition: bound `branch <name>` → `ChdirPass` (the exact input the conversion flips)

## Verification

- `cargo nextest --bin agend-git`: **53/53 pass** (50 prior + 3 new)
- `scripts/preflight.sh --quick`: fmt --check / clippy `--all-targets --features tray -D warnings` / `--tests --features tray` all **clean** (Windows skipped per `--quick` — pure-logic change, no platform-specific code)

### DIFF
--- a/src/bin/agend-git.rs
+++ b/src/bin/agend-git.rs
@@ -120,6 +120,7 @@ fn main() {
             is_agent_caller,
         ),
         subcommand,
+        &args,
         cwd_is_foreign_repo(&binding),
     );
 
@@ -494,13 +495,51 @@ fn is_mutating_local(subcmd: &str) -> bool {
     )
 }
 
-/// #1463 (A): convert a bound-agent `ChdirPass` into `Passthrough` when the
-/// command is a local mutating one AND cwd is a foreign repo. Pure so the
-/// matrix is unit-testable; everything else (push/checkout ChdirPass, Deny,
-/// already-Passthrough) is returned unchanged.
-fn apply_foreign_repo_passthrough(action: Action, subcmd: &str, cwd_foreign: bool) -> Action {
+/// #2027: a `git branch` / `git tag` invocation that NAMES a ref — it creates,
+/// deletes, moves, or inspects a SPECIFIC ref (a positional ref name, or a
+/// `-d`/`-D`/`-m`/`-M`/`-c`/`-C`/`--delete`/`--move`/`--copy` flag) — as opposed to
+/// the bare `git branch` LIST form. `classify` groups `branch`/`tag` with the
+/// read-only commands (they DEFAULT to listing), so a bound agent's
+/// `git branch <name>` is `ChdirPass`'d into the worktree. In a FOREIGN repo that
+/// is the #2027 success-lie: `git branch <new>` runs against the worktree, so the
+/// foreign repo silently gets nothing yet exits 0; a name the worktree already
+/// holds reports `fatal: already exists` from the wrong repo. A ref-naming
+/// branch/tag in a foreign repo must run against THAT repo (passthrough).
+///
+/// Conservative by design: over-classifying (e.g. `git branch --list <pattern>`)
+/// only adds the CORRECT foreign-passthrough (the pattern list should run on the
+/// foreign repo too); under-classifying is the bug. Only `args[0]` (the
+/// subcommand) is skipped — `args[1..]` are scanned for a ref-mutating flag or a
+/// positional ref name.
+fn branch_tag_names_ref(subcmd: &str, args: &[String]) -> bool {
+    if !matches!(subcmd, "branch" | "tag") {
+        return false;
+    }
+    args.iter().skip(1).any(|a| {
+        matches!(
+            a.as_str(),
+            "-d" | "-D" | "--delete" | "-m" | "-M" | "--move" | "-c" | "-C" | "--copy"
+        ) || !a.starts_with('-') // a positional ref name / pattern
+    })
+}
+
+/// #1463 (A) + #2027: convert a bound-agent `ChdirPass` into `Passthrough` when
+/// cwd is a foreign repo AND the command is a local mutating one (#1463) or a
+/// ref-naming `branch`/`tag` (#2027). Pure so the matrix is unit-testable;
+/// everything else (push/checkout ChdirPass, Deny, already-Passthrough) is
+/// returned unchanged.
+fn apply_foreign_repo_passthrough(
+    action: Action,
+    subcmd: &str,
+    args: &[String],
+    cwd_foreign: bool,
+) -> Action {
     match action {
-        Action::ChdirPass(_) if cwd_foreign && is_mutating_local(subcmd) => Action::Passthrough,
+        Action::ChdirPass(_)
+            if cwd_foreign && (is_mutating_local(subcmd) || branch_tag_names_ref(subcmd, args)) =>
+        {
+            Action::Passthrough
+        }
         other => other,
     }
 }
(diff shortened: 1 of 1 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer (codex): REJECTED ×2 → r3 VERIFIED @ fd215e9a

A security-surface review done right — three adversarial rounds, each catch proven by running real Git:
- **r1**: no-positional current-branch write forms (`--set-upstream-to`/`-u`/`--unset-upstream`/`--edit-description`) escaped the classifier — still lying in foreign repos. Proven via mktemp repos writing `branch.main.merge`/`description` with zero positional tokens.
- **r2**: the attached short form `-u<upstream>` (`-uorigin/main`) escaped the exact-match — proven the same way.
- **r3**: completeness claim re-scanned against `git branch -h`: no remaining no-positional write forms; `--create-reflog` confirmed list-only, `--set-upstream` over-include harmless (rejected by Git ≥2.39).

The fix itself: foreign repos now get honest pass-through to real Git for ref-naming `branch`/`tag` (the #1463 foreign-passthrough extended), gated on the existing fail-closed `cwd_is_foreign_repo` detector — fleet interception byte-identical (canonical/sibling/crafted-gitfile/symlink probes all green). Can deny, can pass through, can never lie.

CI fully green; BEHIND-but-clean merge-tree per standard. Merging (squash) — closes #2027.

PR #2034, opened 2026-06-11
fix(#2033): gate the recovery telegram notice to acted-upon blockages

### LINKED ISSUE #2033: Recovery-notice telegram (G.5) fires for self-resolved/false blocks — gate it to acted-upon blockages
Operator flagged (2026-06-12 morning): the `recovered from blocked state — notifying telegram` message is noise.

**The instance**: 2026-06-11 20:12, `agent=fixup-lead` — the "blockage" was the #2020 false AwaitingOperator (agent was never blocked), so the daemon notified the operator about recovering from a state that never truly existed. Double noise: false source + non-actionable content.

**Beyond the false-source case** (#2021 already killed that): a recovery notice is only useful when (a) the operator was actually told about the blockage on the same channel, and (b) the blockage lasted long enough that the operator might be reacting to it. A transient self-resolving block currently can produce notification traffic the operator can do nothing with.

**Fix (principle (a) actionable-or-silent, #2008)**: gate the telegram recovery notice on `blocked notice was sent to telegram for this episode AND block duration > threshold` — otherwise log-only. Pair the block/recovery notices as an episode (the block notice presumably has the same latch state to key on).

Release-blocking for 0.8.0 (#2023 A-queue): operator-flagged noise.

suzuke:
Closes #2033. 0.8.0 A-queue, operator-flagged noise.

## Problem

The `recovered from blocked state — notifying telegram` notice (G.5, `NoticeAction::Recovered`) fired whenever an agent left `InteractivePrompt`/`AwaitingOperator` — **regardless of whether the operator was ever told about the block**. The 2026-06-11 20:12 instance was the #2020 false `AwaitingOperator` (agent never blocked), so the operator got a "recovered" for a state that never existed. #2021 killed that false source; this gates the notice itself.

## Fix (#2008 actionable-or-silent)

A recovery notice is useful only when **(a)** the operator was actually told about the block on telegram **AND (b)** it lasted past a threshold so they might be mid-reaction. Otherwise → log-only.

- **Pair block/recovery as an episode.** New per-episode state on `StateTracker` (`blocked_since`, `blocked_notice_sent`), tracked in `record_set` — the single transition funnel, so it catches both the `set_awaiting_operator` bypass and `transition()`, and spans intra-block transitions (`InteractivePrompt → AwaitingOperator`) for the **full** operator-facing duration. The supervisor calls `mark_blocked_notice_sent()` when it forwards either `Stall`.
- `take_recovery_notice()` now returns `Option<RecoveryEpisode>` carrying the gate inputs. The supervisor emits `Recovered` only when `recovery_notice_is_actionable` = `notice_sent && block_duration >= RECOVERY_NOTICE_MIN_BLOCK` (30s — the AwaitingOperator silence floor). The common AwaitingOperator path always clears this: the ≥30s silence accrues while already in (raw) `InteractivePrompt`, which the episode clock counts.
- Health-reason cleanup still runs in **both** branches (internal state, not operator-facing). A missing episode (e.g. daemon restart mid-block) defaults to non-actionable → errs silent.

## Sweep (issue point 3)

`NoticeAction` has only `Stall`/`Recovered`. The `ServerRateLimit`/`RateLimit` recoveries auto-clear with **log-only `tracing`** (no telegram). **G.5 is the only operator-facing all-clear path** — no other path needs the gate.

## §3.9 tests

- `recovery_episode_captures_notified_long_block_2033` — notified + long → actionable, full duration captured
- `recovery_episode_unnotified_block_2033` — never-notified block → `notice_sent=false` (→ silent)
- `blocked_episode_resets_notice_sent_on_reentry_2033` — the latch resets per episode (no stale carry-over)
- `recovery_notice_gate_actionable_or_silent_2033` — the predicate's 5-way matrix incl. the `>=` boundary
- Existing recovery-arming tests carried to the `Option` API

## Verification

`scripts/preflight.sh --quick` — fmt / clippy `--all-targets --features tray -D warnings` / `--tests --features tray` all green (390/390 state+supervisor). Windows skipped per `--quick`.

### DIFF
--- a/src/daemon/supervisor.rs
+++ b/src/daemon/supervisor.rs
@@ -40,6 +40,14 @@ const NULL_UNLOCK_NOTIFY_WINDOW: Duration = Duration::from_secs(24 * 60 * 60);
 /// to AwaitingOperator (stability gate — a transient streaming flicker of the
 /// permission chrome never holds this long).
 const AWAITING_STABILITY: Duration = Duration::from_secs(8);
+/// #2033: minimum blocked-episode duration for the "recovered from blocked state"
+/// Telegram notice to be actionable. Set to the AwaitingOperator silence floor
+/// (30s) — an episode that reached genuine operator-attention territory. Below it
+/// is an InteractivePrompt that self-resolved before the operator could plausibly
+/// engage, so the recovery notice would be non-actionable noise. The common
+/// AwaitingOperator path always clears this bar: the ≥30s silence accrues while
+/// already in (raw) InteractivePrompt, which the episode clock counts.
+const RECOVERY_NOTICE_MIN_BLOCK: Duration = Duration::from_secs(30);
 /// #1552: how many live bottom rows the permission chrome must render in for
 /// the escalation position gate. A scrollback footer (the meta-FP: a working
 /// agent whose pane merely echoes the chrome) scrolls above this and fails.
@@ -1325,6 +1333,9 @@ fn tick(
                         prev_state = agent_state.display_name(),
                         "awaiting operator (stalled on prompt) — escalating"
                     );
+                    // #2033: pair this blocked notice with the recovery notice —
+                    // record that the operator was actually told about the block.
+                    core.state.mark_blocked_notice_sent();
                     Some(NoticeAction::Stall {
                         tail: core.vterm.tail_lines(TAIL_LINES),
                         silent_secs: Some(silent.as_secs()),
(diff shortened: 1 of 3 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ f2d3bb22 (+ documented restart-window note @ 36a17b24, lead-reviewed)

Operator-flagged noise closed at the notification layer (the false-state source was already killed by #2021):
- Recovery telegram now requires **the blocked notice actually reached the operator this episode AND the block lasted ≥30s** — transient self-resolving blocks are silent on both ends. Actionable-or-silent (#2008 principle a).
- Episode tracking lives at the single `record_set` funnel (grep-proven: only `self.current =` site) — covers both forced and normal transitions, spans intra-block state changes for full duration; both Stall-forward arms mark the notice (no reverse under-notification); re-entry resets cleanly; the Option-API migration keeps health cleanup byte-identical.
- The 30s floor is meaningful (filters short self-resolving InteractivePrompts that did notify); the narrow restart-mid-block missed-all-clear window is documented as a known limitation (safe direction).

CI green. Merging (squash) — closes #2033.

Issue #1080, opened 2026-05-22
bug: skills install missing for non-Claude-Code backends on daemon spawn


suzuke:
## Symptom

After `agend-terminal skills add` + daemon restart, skill symlinks are only created for **Claude Code backend instances** (+ 1 kiro-cli instance). Codex, OpenCode, Gemini, and some Kiro instances have **no symlinks** — daemon `install_for_agent` at spawn skips them.

## Evidence (empirical, 2026-05-22)

```
HAS symlinks:  claude-585a78, claude-ba5740, claude-d27301, fixup-dev, fixup-dev-2, fixup-lead, general, kiro-cli-ea377a
NO  symlinks:  fixup-reviewer (codex), kiro-research (kiro), gemini-4315be (gemini), opencode-f068d5 (opencode), mas-eval
```

All symlink-present instances are Claude Code backend, except kiro-cli-ea377a (anomaly — same backend as kiro-research but different result).

Newly spawned instances of ANY backend also do not get skills installed.

## Expected

`install_for_agent` should run for ALL instances at spawn, creating symlinks (or copies on Windows) for all 5 backend skill paths regardless of the instance's actual backend.

## Scope

Bug in `daemon/mod.rs` `spawn_and_register_agent` → `skills::install_for_agent` call path. Either:
1. The call is conditional on backend type (shouldn't be)
2. The call fails silently for some instances
3. The call is never reached for certain spawn paths

## Canary skill used for testing

```
~/.agend-terminal/skills/agend-skill-canary/SKILL.md
```

PR #2039, opened 2026-06-11
fix(#2038): fleet.yaml model applies on runtime respawns, not just daemon boot

### LINKED ISSUE #900: bug: spawn_one hardcodes env: None — template env vars never reach backend process
## Summary

`api::spawn_one` (`src/api/mod.rs:518`) passes `env: None` to `SpawnConfig`, which means environment variables set in fleet.yaml (both in `instances.<name>.env` and template `instances.<name>.env`) are silently dropped at spawn time for all API/daemon-spawned instances.

## Root cause

There are two spawn code paths:

| Path | File | Line | env |
|---|---|---|---|
| TUI (`create_pane_from_resolved`) | `app/pane_factory.rs` | 332 | ✅ `&resolved.env` |
| API (`spawn_one`) | `api/mod.rs` | 518 | ❌ `None` |

`handle_spawn` (instance handler) and template `deploy()` both call `spawn_one`, which hardcodes `env: None`. The TUI path (`create_pane_from_resolved`) correctly reads `env` from `resolve_instance()` and passes it through.

`fleet.rs::resolve_instance()` already merges `defaults.env` + `inst.env` into a `HashMap<String, String>` on `ResolvedInstance`, so the data is available — it's just never consumed by `spawn_one`.

## Affected scenarios

- Template-deployed instances (via `deployment deploy`) — `env:` in template stanzas is ignored
- MCP `SPAWN` / `create_instance` calls — fleet.yaml `env:` is ignored
- Team spawn (CREATE_TEAM) — same

## Fix direction

In `spawn_one` (around line 518), load fleet.yaml and resolve the instance to get `env`, then pass it to `SpawnConfig`:

```rust
// Before (line 518):
env: None,

// After:
env: {
    let fleet = crate::fleet::FleetConfig::load(&crate::fleet::fleet_yaml_path(home)).ok();
    fleet.as_ref()
        .and_then(|f| f.resolve_instance(name))
        .map(|r| r.env)
        .as_ref()
},
```

Or alternatively, add an `env: Option<&HashMap<String, String>>` parameter to `spawn_one` and let each caller supply it from their own resolution context.

### LINKED ISSUE #2038: fleet.yaml  is only applied by daemon-boot spawn — restart_instance / replace_instance respawn without --model (same class as #900 env fix)
Found live (2026-06-12): set `model: claude-opus-4-8` on three instances in fleet.yaml, then `restart_instance` (fresh) and `replace_instance` — **both respawned without the `--model` flag** (verified via `ps`: no `--model` in the spawned claude command line; pane shows the CLI default model).

The boot path (`src/bootstrap/agent_resolve.rs:128`) does push `--model` from the resolved fleet config, so a full daemon restart applies it — but the runtime respawn flows (restart/replace, and presumably crash-respawn) resolve from a config snapshot that drops/ignores `model`.

This is the exact class #900 fixed for `env` ("persisted to the fleet.yaml entry so replace/restart flows re-apply them") — `model` needs the same treatment: respawn flows must re-resolve `model` (and audit the other per-instance fields — `args`? — for the same drift).

§3.9 test: fleet entry with `model` set → restart_instance → assert spawned argv contains `--model <formatted>`; same for replace and crash-respawn paths.

Workaround meanwhile: model changes take effect at the next full daemon restart. 0.8.0 A-queue candidate (config-honesty class: a written config that silently doesn't apply).

suzuke:
Closes #2038.

## Problem

`fleet.yaml`'s per-instance `model:` was only applied by the daemon-boot spawn path (`bootstrap/agent_resolve.rs` pushes `--model`). Every runtime respawn flow issues a SPAWN RPC whose `args` never carry the flag, so the written config silently didn't apply until the next full daemon restart (verified live via `ps` in the issue):

| Path | pre-fix |
|---|---|
| `restart_instance` | sends fleet-resolved `args` **without** model |
| `replace_instance` | sends **no `args` at all** — fleet `args` AND `model` both dropped |
| `start_instance` | same as restart |
| deploy Phase-3 spawn | sends `params.model`, which `handle_spawn` ignored (dead wire field) |
| CREATE_TEAM Phase-2 | spawns with `&[]` args, ignores `defaults.model` / hand-edited entries |
| crash-respawn | ✅ unaffected — replays the boot-resolved `AgentConfig.args`, which already carry `--model` |

## Fix — same seam as #900's env fix

#900 fixed the exact same class for `env` with a fleet.yaml fallback at the `handle_spawn` handler boundary (`spawn_one` is documented as a pure data consumer, not a re-resolver — kept that contract). This PR applies the identical pattern for `model` + `args`:

- **`handle_spawn`**: one fleet resolve (shared with the existing #900 env fallback). Model precedence (r1: now fully specified): **explicit `--model` in args** (create_instance pre-formats it) **> `params.model`** (deploy Phase 3's wire field — honoured instead of silently ignored, consistent with #900's params.env-over-fleet rule) **> fleet entry's resolved model** (instance > defaults via `resolve_instance`). `Backend::push_model_arg` enforces tier 1 by never duplicating an existing flag and formats through `format_model_arg` (OpenCode provider prefix). An args-less SPAWN (the `replace_instance` wire shape) additionally falls back to the fleet entry's resolved `args`.
- **`team.rs` Phase-2**: same boot-parity rule (resolves the full instance once instead of env-only).

### Per-instance field audit (boot vs respawn)

- `args`: drift in replace only → fixed here. `env`: fixed by #900. `working_directory`: resolved from fleet in both flows ✅. `submit_key`: both sides use the backend preset (`resolve.rs:114`) — no drift. `skills`: `spawn_one` handles (#1080) ✅. `instructions`/`role`: `prepare_instructions` ✅.

### Backend `--model` flag support evidence

- claude: `--model` documented; the boot path has always pushed it.
- codex (local CLI, `codex-cli 0.139.0`): `codex --help` → `-m, --model <MODEL>  Model the agent should use` ✅ (long form `--model <value>` accepted; same flag shape the boot path already emits).
- opencode: `--model provider/model`, prefix handled by `format_model_arg`.

## §3.9 tests (all real-spawn argv-capture through the actual entry points, mirroring the #900 test pattern)

- `handle_spawn_falls_back_to_fleet_yaml_model_2038` — restart shape: args without `--model` → fleet model lands in the spawned argv.
- `handle_spawn_falls_back_to_fleet_yaml_args_and_model_2038` — replace shape: no `args` key → fleet args + model both respawn.
- `handle_spawn_caller_model_flag_wins_over_fleet_2038` — tier-1 precedence + no-duplicate.
- `handle_spawn_params_model_wins_over_fleet_2038` (r1) — deploy Phase-3 shape: `params.model` honoured, wins over a diverging fleet entry.
- `handle_spawn_instance_model_overrides_defaults_2038` (r1) — `resolve.rs:124` instance-over-defaults chain proven at the SPAWN entry.
- `create_team_spawns_members_with_defaults_model_2038` (r1) — real CREATE_TEAM Phase-2 entry: member argv carries `--model` from `defaults.model` (script via `defaults.args`; pre-fix members spawned with empty argv).
- `push_model_arg_appends_and_dedupes_2038` — unit: formatting, glued `--model=` form, empty no-op.
- Crash-respawn has no SPAWN-RPC seam to assert without a live daemon → doc-noted in the test header (it preserves model by construction).
- RED-verified per ingress: fleet-model + fleet-args tests fail with the handle_spawn resolve removed; create_team test fails with the `member_args` revert (`&[]`); params.model test fails with the params tier removed.

## Verification

- `cargo fmt --check` + `cargo clippy --all-targets --features tray -- -D warnings` ✅
- `cargo test --tests --features tray`: 3702+ passed, 1 failed — `dispatch_branch_skips_metadata_stub_but_checks_out_real_source_1834`, **pre-existing on origin/main in my agent environment only** (agend-git shim intercepts its real-git checkout; passes with `AGEND_GIT_BYPASS=1`, unaffected in CI).
- Live guinea-pig check (fleet entry with `model: claude-opus-4-8` → restart → `ps` argv) queued after merge/deploy of this build, per the task contract.


suzuke:
## Review mirror (§3.12) — fixup-reviewer (codex): REJECTED → r2 VERIFIED @ 036f9f12

- Root fix: respawn flows (restart/replace/start, team Phase-2) now resolve fleet `model`+`args` at the handle_spawn boundary — the exact class #900 fixed for env. Crash-respawn verified already-correct (replays boot-resolved args).
- r1 gap closed with entry-level tests: real `handle_create_team` with `defaults.model` → member argv; instance-overrides-defaults proven at the SPAWN entry; deploy's dead `params.model` wire upgraded to a meaningful three-tier precedence (explicit `--model` in args > `params.model` > fleet resolved), all RED-mutation-verified (reviewer independently reproduced the RED).
- `push_model_arg` dedupes both `--model x` and `--model=x` forms; per-backend flag shapes evidenced (incl. codex `-m, --model`).

CI fully green. Merging (squash) — closes #2038. Operator's Opus 4.8 tiering now applies on individual restarts, no daemon restart required (after next deploy).

PR #2045, opened 2026-06-12
feat(telegram): resolve sender display name from user_allowlist {id,name} entries

yujunchao:
## Problem
Inbound Telegram messages from accounts **without a public `@username`** are surfaced to agents as `[user:unknown via telegram]` — even though the numeric `sender_id` is known and on the allowlist. Operators whose Telegram account has no @username always show up as `unknown`.

## Fix
Let `user_allowlist` entries optionally carry a display name, and resolve the inbound sender's name from it.

- **fleet**: `user_allowlist` entries accept either a bare id (legacy `- 12345`) **or** `{ id, name }` — new `AllowlistEntry` enum, `#[serde(untagged)]`, **backward compatible** (old bare-id lists still parse). The type is **channel-agnostic** by design (documented for Discord/Slack reuse).
- **bootstrap**: split entries into the bare-id list (authz path **unchanged**) and an `id -> name` map stored on `TelegramState`.
- **inbound**: `username = @username ?? allowlist_name(sender_id) ?? "unknown"` (main + status-summary paths). `NotifySource::Channel(name, kind)` already renders `user:NAME via {channel}` for any `ChannelKind`.
- **tests**: untagged deserialization (bare id + `{id,name}`), `username_for` resolution; existing `is_user_allowed` / doctor / yaml round-trip unchanged.

### fleet.yaml (new optional form)
```yaml
user_allowlist:
  - { id: 5315487661, name: "Yu" }
  - 8700664262            # bare id still works
```

## Verification
- `cargo check` + `cargo check --tests`: clean on upstream/main
- `cargo test --bin agend-terminal`: allowlist (13), user_allowlist (7), username_for (1), is_user_allowed (4) — all pass, no regressions
- Deployed + verified in production: operator messages now arrive as `[user:Yu via telegram]`

### DIFF
--- a/src/bootstrap/doctor.rs
+++ b/src/bootstrap/doctor.rs
@@ -218,7 +218,8 @@ mod tests {
             bot_token_env: "AGEND_TELEGRAM_BOT_TOKEN".into(),
             group_id: -100123,
             mode: "topic".into(),
-            user_allowlist,
+            user_allowlist: user_allowlist
+                .map(|v| v.into_iter().map(crate::fleet::AllowlistEntry::Id).collect()),
             fleet_binding: None,
         }
     }
--- a/src/bootstrap/doctor.rs
+++ b/src/bootstrap/doctor.rs
@@ -218,8 +218,11 @@ mod tests {
             bot_token_env: "AGEND_TELEGRAM_BOT_TOKEN".into(),
             group_id: -100123,
             mode: "topic".into(),
-            user_allowlist: user_allowlist
-                .map(|v| v.into_iter().map(crate::fleet::AllowlistEntry::Id).collect()),
+            user_allowlist: user_allowlist.map(|v| {
+                v.into_iter()
+                    .map(crate::fleet::AllowlistEntry::Id)
+                    .collect()
+            }),
             fleet_binding: None,
         }
     }
--- a/src/channel/telegram/bootstrap.rs
+++ b/src/channel/telegram/bootstrap.rs
@@ -63,7 +63,20 @@ pub fn init_from_config(
         }
         Some(list) => tracing::info!(count = list.len(), "telegram user_allowlist active"),
     }
-    let allowlist = user_allowlist.clone();
+    // Split the allowlist into the bare-id list (authz, unchanged downstream)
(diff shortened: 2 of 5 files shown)


suzuke:
Review complete — **the design passed all six adversarial probes; the only blocker is `cargo fmt`**. This is a run-one-command-and-merge PR. Details:

**What we verified and liked**:
- **Fail-closed direction confirmed empirically** (we independently reproduced the serde behavior): a malformed entry (`{id: "x"}`, missing field, quoted id) fails the whole list → fleet load errors → downstream auth denies everything (`is_authorized_recipient` is fail-closed on None/empty). Exactly right for an authz surface.
- Backward compat holds (bare ids and mixed lists parse; serialization shape-preserving), and the daemon's only fleet.yaml write path (`backfill_ids`) surgically touches `instances.*.id` only — your hand-edited allowlist will never be rewritten.
- `is_user_allowed` consumes only the id projection; the name never participates in authorization. Both username resolution sites are consistent; the channel-agnostic claim matches the code.

**Blocking (mechanical)**: CI is red on `cargo fmt --check` — three spots: `src/bootstrap/doctor.rs:218`, `src/channel/telegram/inbound.rs:204` and `:326`. `cargo fmt` + push fixes it.

**Two cheap adds while you're in there (suggested, not blocking)**:
1. Sanitize the configured display name before it enters the line-oriented notification header (filter control chars/newlines + cap length ~32) — it's operator-trust input so not a privilege issue, but a newline in a name would garble header parsing.
2. One test pinning the deny direction: `from_str` on a malformed entry asserts `Err` — it's the invariant an authz surface most wants locked.

Same-day re-review on push. Nice to have you back — and #1978 (your headless flush fix) shipped last night with dual review, authorship intact.

suzuke:
Merged — thank you @yujunchao! 🎉 Clean fail-closed design that survived all six adversarial probes on the first review pass. The display-name resolution ships in the next release (0.8.1). The team added the final polish on top of your commits (fmt + a sanitize pass at the ingestion chokepoint + a fail-closed regression test) — your authorship is preserved in the history. Looking forward to your next contribution.

PR #2048, opened 2026-06-12
fix(#2046): keep pane PTY rows aligned with rendered content

### LINKED ISSUE #2046: Bottom gap in every pane: vterm/PTY rows ~4-5 short of the rendered rect (status line floats above pane bottom)
Operator-reported (first 2026-06-11 morning, confirmed persisting through resizes + restarts, screenshot evidence 2026-06-12).

**Refined symptom** (from screenshot analysis): this is NOT unallocated layout space — the top-level chunks (tab bar / Min(1) tree / status bar) fill the screen, and `split_child_areas` is fill-complete (`second = total + overlap - first`). Instead, **every pane's terminal content ends ~4-5 rows above its pane's bottom edge** — each backend's own status line (claude's `Model: ... | bypass permissions`, codex's footer, etc.) floats with consistent blank rows below it, in every pane including the focused one. Consistent constant offset across panes/backends ⇒ systematic.

**Hypothesis**: the PTY winsize (vterm rows) for each pane is computed with a wrong row reservation — e.g., subtracting global chrome (tab bar + status bar + title row) from the **full terminal height** instead of using the pane's own rect height, or double-subtracting per-pane chrome. The render then paints an N-rows-short grid into an N-rows-taller rect.

**Spike**: find where pane PTY resize/winsize is derived (rect → rows/cols mapping), reproduce the off-by-N against a known rect, fix + a render test pinning 'vterm rows == pane content rows'.

Long-standing visual wart, pre-0.8.0 polish candidate.

suzuke:
Closes #2046.

## Problem

Pane layout rectangles fill the screen, but a pane can render with a VTerm/PTY grid that is shorter than the pane's current bordered content area. When that happens, backend footers/status lines appear to float several rows above the pane bottom while the remaining rows stay blank.

The risky part is that resizing currently depends on separate event/topology paths remembering to call `layout::resize_panes`. Render already has the exact final pane rect, so it is the right last-mile place to enforce the invariant.

## Fix

- Thread `AgentRegistry` into pane rendering.
- Before drawing each pane, compute the actual inner content rect and resize the pane VTerm plus underlying local/remote PTY when it differs.
- Keep existing backend redraw behavior: `Pane::resize_pty` still only sends the Ctrl-L redraw sequence for Kiro; Claude/Codex/OpenCode/Agy/etc. only receive the normal PTY resize/SIGWINCH.

## Regression Test

- `render_resizes_vterm_to_pane_content_rows_2046`: starts a 40x20 render with a pane VTerm five rows short (`38x11`) and asserts render corrects it to the pane content size (`38x16`). This pins `vterm rows == pane content rows`.

## Verification

- `cargo fmt --check`
- `cargo test render_resizes_vterm_to_pane_content_rows_2046 --features tray`
- `cargo test render::core_render --features tray`
- `cargo test layout:: --features tray`
- `cargo clippy --all-targets --features tray -- -D warnings`

🤖 Generated with [Codex](https://openai.com/codex)

### DIFF
--- a/src/render/core_render.rs
+++ b/src/render/core_render.rs
@@ -96,7 +96,7 @@ pub fn render(
         .split(frame.area());
 
     let snapshot = build_agent_state_snapshot(layout, registry);
-    render_pane_tree(frame, chunks[1], layout, repeat_mode, &snapshot);
+    render_pane_tree(frame, chunks[1], layout, repeat_mode, registry, &snapshot);
     render_tab_bar(frame, chunks[0], layout, &snapshot);
     render_status_bar(frame, chunks[2], layout, telegram, binary_stale);
 }
@@ -218,6 +218,7 @@ fn render_pane_tree(
     area: Rect,
     layout: &mut Layout,
     repeat_mode: bool,
+    registry: &AgentRegistry,
     snapshot: &HashMap<String, AgentState>,
 ) {
     let tab = match layout.tabs.get_mut(layout.active) {
@@ -238,7 +239,9 @@ fn render_pane_tree(
 
     if tab.zoomed {
         if let Some(pane) = tab.root_mut().find_pane_mut(focus_id) {
-            let info = render_pane(frame, area, pane, true, false, snapshot, false, false);
+            let info = render_pane(
+                frame, area, pane, true, false, registry, snapshot, false, false,
+            );
             let infos = vec![info];
             render_border_grid(frame, &infos);
             render_pane_titles(frame, &infos);
@@ -261,6 +264,7 @@ fn render_pane_tree(
         &mut rects,
         &mut border_infos,
         repeat_mode,
+        registry,
         snapshot,
         drag_source,
         drag_target,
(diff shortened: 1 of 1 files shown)


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ 4fc931d7 (cross-review; author = codex)

**The two-day mystery, fully explained**: `resize_panes` has exactly two production call sites, both gated behind per-action opt-in `needs_resize` flags — any pane-geometry change whose path forgets the flag (or never passes through an action) leaves the vterm/PTY grid stale-short, and `render_to_buffer` clamps to `min(grid, rect)` → the backend's footer floats with blank rows below. The fix moves correction to the render chokepoint: 4 lines enforcing `vterm dims == final inner rect` with `resize_pty` on mismatch — covering every past and future leak path.

- **No resize storm possible**: `VTerm::resize` sets dims unconditionally, so the guard self-converges next frame even if the PTY side fails — same latch pattern as the existing resize path.
- **No new lock-order edge**: render holds no guard at the call; `resize_pty` keeps the #1530 discipline (registry dropped before PTY writes); Kiro's Ctrl-L fires only on true mismatch.
- Regression test pins correction (38x11 → 16 rows), RED-verified by the reviewer.

Non-blocking notes recorded: a registry-seeded test to also pin the PTY-side call, and this comment serves as the precise root-cause record. Merging (squash) — closes #2046. **Visible after next restart: status lines sit at pane bottoms.**

PR #2051, opened 2026-06-12
feat(#2044): inject-delivery watchdog — re-deliver once when a dialog swallows an actionable wake

### LINKED ISSUE #2044: Interactive dialogs in managed panes swallow injected dispatches — detect & surface (model-picker case)
Live incident (2026-06-12 09:42): after the model-tiering rollout (#2038/#2039), `fixup-dev` — apparently reacting to tiering context — ran `/model` itself and pressed Enter, which **saved Opus 4.8 as the operator's global default for new sessions** (claude's own config, outside agend's tree). This violates the agend-must-not-modify-global-files rule, and the interactive dialog it opened also swallowed an injected dispatch (delivery loss).

Fixes:
1. **Instruction template** (the `--append-system-prompt-file` agend.md): add an explicit line — model selection is managed by the fleet (argv `--model`); agents must never run `/model` (especially never 'save as default'), `/config`, or other host-global mutations.
2. Consider a daemon-side guard: detecting the model-picker dialog shape in a managed pane → auto-ESC + WARN (the dialog blocks injected message delivery — same class as other interactive-prompt stalls, the existing InteractivePrompt detection may already see it; verify why it didn't).

Operator asked about restoring his global default separately (his preference, not auto-fixed).

suzuke:
## Summary

Fix #2044 — an operator-opened `/model` picker swallowed an injected dispatch (the keystrokes went to the picker, the prompt never submitted, the dispatch was silently lost). Detection-primary fix (issue direction **b**), and crucially **dialog-shape-agnostic**.

### How it detects loss
A landed actionable inject submits a prompt → the backend fires a `UserPromptSubmit` hook. A dialog-swallowed inject submits nothing → no such hook. New `daemon::inject_delivery`:
- **`arm(agent, text)`** at the actionable-wake inject site (`compose_aware_inject`'s direct-inject branch). **Self-gates** on the agent having a hook-shadow entry — only hook-emitting backends (claude today) are ever armed, so a non-hook backend can never be falsely re-injected (per-backend honesty, no claude hardcode).
- **per-tick `verify_pass`**: UserPromptSubmit after the inject → clear silently; none within **30s** → re-deliver **once** + WARN + event_log, latched; still unconfirmed → final WARN + give up. Never a retry storm (#2008).
- **`hook_shadow` sticky `last_user_prompt_submit_ms`**: the per-event snapshot is overwritten wholesale, so the verifier needs a UPS timestamp that survives the PreToolUse/Stop events that follow within the window.

Re-delivery uses the plain submit path (not `compose_aware_inject`) → can't re-arm (the latch lives in `Pending`). In-memory state: a restart forgets in-flight verifications — the durable dispatch re-nudge is #1888's ci-handoff track on a longer horizon; this is the fast 30s physical-landing net, complementary.

### Item 2 — model-picker classification: STRUCTURAL, documented not patched
Investigated: `/model` (an arrow-key menu) matches neither `is_generic_startup_prompt`'s token set nor the PermissionPrompt chrome. Not patched because (1) adding a pattern without a grid-validated capture of the real picker is the #1559 hazard (no such fixture exists), and (2) the inject-delivery watchdog already covers **all** swallowing dialogs via the dialog-agnostic UserPromptSubmit signal — per-dialog pattern-chasing is strictly weaker. Doc note added at `is_generic_startup_prompt`.

## Tests (7)
arm-noop-without-hook-history (the non-hook gate) · delivered-clears (UPS after inject beats the timeout) · redeliver-once-then-give-up (latch + no storm) · stale-prior-UPS-doesn't-confirm · + hook_shadow sticky-UPS coverage. Deterministic (explicit-timestamp seams, no clock-collision races).

## Preflight
fmt / clippy(tray) clean; nextest **4027/4028** (the 1 = known #1834 environmental, bypass-solo passed; team_dedup excluded per #1993).

Closes #2044


suzuke:
## Review mirror (§3.12) — fixup-reviewer-2: VERIFIED @ 1652af7a

- **Dialog-shape-agnostic delivery verification**: an actionable inject that lands must produce a `UserPromptSubmit` hook; one swallowed by any interactive dialog won't — no per-dialog patterns needed (and item 2's no-pattern-without-grid-capture call honors the #1559 lesson). 30s without UPS → one redelivery + WARN → latched give-up. No storm possible.
- Key probes held: arm sits behind the existing busy-gate (generating agents defer, never arm); UPS fires at submit-time so long turns can't be misjudged; self-gated to hook-emitting backends (no hardcoded claude); sticky UPS survives subsequent events while stale ones can't false-confirm (RED-verified); redelivery goes through the plain submit path (can't re-arm), with decide-under-lock/act-after discipline.
- Built directly on the #2014 promotion's hook stream — yesterday's feature is today's sensor.

CI green. Merging (squash) — closes #2044 (the operator-hit dialog-swallow case is now self-healing).

Issue #2059, opened 2026-06-12
pr-state: [pr-ready-for-merge] didn't reach the merge actor when verdict preceded CI (PR #2058 dead zone)


suzuke:
## Incident (2026-06-12, operator-caught)

PR #2058: reviewer (codex) delivered VERIFIED at 14:00 while CI was still pending; CI went green ~14:08; **nobody acted until the operator asked \"2058 CI pass 但沒人後續處理?\"** at 14:12. The ci-watch `next_after_ci` pointed at the reviewer (set by the implementer at dispatch), who on receiving `[ci-ready-for-action]` only re-confirmed the existing verdict — the merge actor (lead) never got a signal.

## Gap

The #972 pr-state aggregator already models this: `review_class=single`, one VERIFIED + CI green ⇒ `[pr-ready-for-merge]`. Questions to answer (verify-first):

1. Did `[pr-ready-for-merge]` fire for #2058 at all? If yes, to whom? (If it went to the watch subscriber = reviewer, the routing is the bug: ready-for-MERGE should target the merge authority, not the last CI-chain hop.)
2. If it didn't fire: was the codex VERIFIED (kind=report, reviewed_head=8f8228d8) recorded into pr-state for that branch? Aggregation key mismatch (task correlation vs branch) would explain the silence.

## Direction (noise-discipline: fire-once, escalate-don't-repeat)

`[pr-ready-for-merge]` should route to the team orchestrator (merge authority) regardless of who armed the watch / who the CI chain pointed at. `next_after_ci` stays as-is for the *action* chain (review handoff); merge-readiness is a separate terminal signal with one owner.

## Interim behavioral rule (already adopted by lead)

When dispatching a review that may complete before CI, lead keeps merge-watch on itself (or polls on wakeup fallback) instead of relying on the implementer's `next_after_ci` chain.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

suzuke:
## Verify-first RCA — the dead zone is a **3-layer systemic failure**, not a one-off (fixup-dev-2)

[...]

Issue #2010, opened 2026-06-11
Worktree & binding lifecycle issues: from_ref remote resolution, rework lease conflicts, and stale ci-ready renudges (retro #39)


cheerc:
We have investigated three daemon-side issues related to worktree management and task lifecycle (retro #39) in `agend-terminal`. Below is the root cause analysis (RCA) and our proposed fix plan. Since these areas have had multiple historical fixes, we want to share our findings and politely ask for your feedback before opening any PRs.

### Related Issues Researched:
- #870: Daemon: auto-release worktree on verdict-sent
- #1859: [ci-watch] next_after_ci notification reliability
- #1882: Cross-agent double-bind: branch-uniqueness (scan→bind_full) is check-then-act without a shared lock
- #1888: ci-handoff re-nudge (#1860) is coupled to inbox read-state
- #1985: Repeated continue nudges during long tasks/waiting states
- #2008: [epic] System noise reduction — automated message noise

---

### 1. #1 from_ref fork-wrong-ref
**Symptom**: When `from_ref` in `ensure_branch_exists` is a fork-tracking reference (e.g., `fork/main`), the remote name `"origin"` is hardcoded for fetches and remote-tracking ref verification, leading to failure or resolving against the wrong repository.

**RCA**:
- In `src/mcp/handlers/dispatch_hook/mod.rs` (lines 734-738), `["fetch", "origin", branch]` is hardcoded, and the remote ref is hardcoded to `refs/remotes/origin/{branch}`.
- In `src/mcp/handlers/dispatch_hook/mod.rs` (lines 763-768), stripping the remote prefix only checks `"origin/"`.
- In `src/mcp/handlers/dispatch_hook/mod.rs` (line 803), the fallback fetch also hardcodes `"origin"`.

**Proposed Fix**:
- Parse the remote and branch from `from_ref` (e.g. if `from_ref` has a slash like `remote/branch`, verify if `remote` is a valid git remote using `git remote`).
- Implement `is_valid_remote` in `src/git_helpers.rs` and fallback to `primary_remote(source)` or `"origin"`.
- Use the dynamically resolved remote instead of hardcoded `"origin"` in all fetch/ref-parse calls.

---

### 2. #3 rework lease conflict
[...]

PR #2062, opened 2026-06-12
fix(#2059 #1): verdict matcher sees through the [report_result] wrapper

suzuke:
#2059 fix **#1** (lead-ruled; #2/#3 are separate dialectic'd follow-ups). The load-bearing fix that unblocks the whole verdict→merge-ready pipeline.

## Root cause (from the #2059 RCA)

The verdict→pr-state→`[pr-ready-for-merge]` pipeline emitted **ZERO** output in 2 days across every reviewed+merged PR. A reviewer's verdict is sent via `send request_kind=report` → `comms::handle_report_result`, which prepends a `[report_result] ` wrapper to the summary (so `msg.text = "[report_result] VERIFIED …"`). But verdict detection in `process_verdicts` (`messaging.rs`) and `auto_release::is_terminal_verdict_text` both checked `text.trim_start().starts_with("VERIFIED")` on the **wrapped** text → starts with `[` → **false** → `record_verdict` was never called. This is a #1493 test-fidelity bug: the tests hand-crafted `"VERIFIED …"` (no wrapper) so they were green while the matcher was dead against real wire text. (It also silently disabled the #2010 reviewer-binding auto-release intent — same gate.)

## Fix

- **One shared strip helper** `auto_release::strip_report_wrapper` — strips the `[report_result] ` wrapper (idempotent on already-bare text from a raw `send` report). **BOTH** verdict consumers (`is_terminal_verdict_text` and `process_verdicts`) route through it, so they can never drift (the lead's "一概念一機制").
- **Extract the producer** `build_report_text` from the inline `handle_report_result` body into `comms_inbox.rs` (the `comms.rs` overflow sibling — `comms.rs` is at the 750-LOC handler cap), so production and the test feed the **same** builder.
- **Producer-fed fixture (#1493 discipline)**: `matcher_sees_through_report_result_wrapper_producer_fed_2059` routes the real `build_report_text` output through `is_terminal_verdict_text` for all three verdicts (+ correlation/Artifacts suffix lines + a non-verdict negative). **RED-verified**: reverting the strip fails it.
- **Tangential noise fix**: `pr_state::is_pr_state_file` skips dotfiles so the gh-poll/scanner dir scan stops parsing `.emitted-terminal.json` as a `PrState` and spamming a `missing field 'repo'` WARN every ~10 s (used at all 4 dir-scan sites).

## Out of scope (per lead scoping)

- **#2** (verdict↔branch correlation: reviewed-head SHA-key + create-or-buffer pr-state) — separate task, dialectic-ruled (c).
- **#3** (route `[pr-ready-for-merge]` to the merge authority via `teams::find_team_for`, not the released binding) — goes to dev.

So this PR makes `record_verdict` fire; #2 makes it resolve the branch and survive the verdict-before-CI ordering; #3 delivers the ready signal to the right actor.

## Verify

`cargo fmt --check` + `cargo clippy --all-targets --features tray -- -D warnings` ✅; `file_size_invariant` green (comms.rs back to 748); affected modules (auto_release / pr_state / messaging / comms) 161 tests pass.

### DIFF
--- a/src/api/handlers/messaging.rs
+++ b/src/api/handlers/messaging.rs
@@ -466,7 +466,13 @@ fn process_verdicts(home: &Path, from: &str, msg: &crate::inbox::InboxMessage) {
     // staleness gate (a head-less VERIFIED must not flip the merge gate);
     // REJECTED/UNVERIFIED record regardless (UNVERIFIED is evidence-exempt).
     if msg.kind.as_deref() == Some("report") && msg.correlation_id.is_some() {
-        let text = msg.text.trim_start();
+        // #2059: strip the `[report_result] ` wrapper (added by
+        // comms::handle_report_result) via the SHARED helper, so the verdict-word
+        // check sees the bare word — the same strip `is_terminal_verdict_text`
+        // uses, so the two verdict consumers never drift. Without this, the
+        // wrapped real wire text never matched and record_verdict was never
+        // called (the pipeline-wide silence #2059 RCA'd).
+        let text = crate::daemon::auto_release::strip_report_wrapper(&msg.text);
         let task_id = msg.correlation_id.as_deref().unwrap_or("");
         if text.starts_with("VERIFIED") {
             if msg.reviewed_head.is_some() {
--- a/src/daemon/auto_release.rs
+++ b/src/daemon/auto_release.rs
@@ -405,11 +405,30 @@ pub(crate) fn enqueue_intent(home: &Path, intent: &AutoReleaseIntent) -> std::io
     Ok(())
 }
 
-/// The three terminal review verdicts. A reviewer's report opens with exactly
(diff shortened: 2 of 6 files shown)


suzuke:
codex r1 addressed (`24a9f934`): the remaining 2 pr-state dir loops now route through `is_pr_state_file` too — `suppress_stale_terminal_replay_with` (mod.rs:~674) and `record_verdict` (mod.rs:~911). All **6** pr-state dir scans now go through the single predicate; no loop serde-parses an entry as `PrState` without it. Verdict-strip body unchanged (your RED rerun confirmed it). Agreed not to expand to lowercase/double-wrapper/CRLF — the production `build_report_text` emits exactly one wrapper form. pr_state suite 76 green; fmt + clippy clean.

Issue #2050, opened 2026-06-12
[arch] Fable-led whole-architecture optimization & tidying pass (pre-model-sunset)


suzuke:
Operator directive (2026-06-12): before Fable 5 access ends, run a full architecture optimization/reorganization pass over Agend-Terminal, led by the strongest available model — the synthesis and structural judgment are the point; written outputs outlive the model.

## Shape
**Phase 0 — survey & synthesis (Fable-led, fresh-context session)**
- Ground truth: the 6-subsystem map from the 2026-06-10 audit (held up well) + ~200K LOC / 286 files today.
- Deliverables: `docs/ARCHITECTURE.md` (honest current-state: module boundaries, data flows, lock/concurrency map, state-detection layers post-#2014) + `docs/REFACTOR-PLAN.md` (prioritized phases, each独立 executable-on-Opus with safety rails).
- Known candidates feeding in: supervisor god-fn (old refactor-drive leftover), files at LOC caps (instance.rs 748/750, comms.rs 750/750 — pressure indicates boundary problems), state-detection dual-path (heuristic + hook) consolidation direction, notification pipeline end-to-end ownership, MCP handler organization (36 tools), per_tick handler proliferation, duplicated store patterns vs the versioning backbone.\n- Method: Explore-agent fan-out for mapping, Fable synthesis for judgment; verify-first discipline on every "this is dead/duplicated" claim (the 70%-FP audit lesson; and dead-code-may-be-roadmap — GitLab/Bitbucket providers etc. stay).\n\n**Phase 1..N — execution (Opus team, Fable structural review)**\n- Each phase: behavior-preserving (byte-identical gates where possible), test-first, one PR per boundary move, full CI + adversarial review.\n- Refactor-early principle applies: high-value consolidations land before they compound.\n\n## Sequencing\n1. 0.8.0 ships first (in flight, #2023).\n2. Phase 0 runs in a **dedicated fresh-context Fable session** (this is heavy-reading work; the current lead session is at 85%+ context — handoff prepared).\n3. Companion deliverables from the same Fable window (operator-approved plan): #1523 phase-2 spec, #1814 design, #1967 epic decomposition, lead-judgment runbook.\n\nTracked separately from #2023 (not release-blocking for 0.8.0; likely shapes 0.9).

suzuke:
[...]

PR #2065, opened 2026-06-12
refactor(#2050 W1.1): unify 12 supervisor trackers into PerTickHandler set

suzuke:
## What (REFACTOR-PLAN W1.1 · survey 01-R2 — \"best value-to-risk, do first\")

The supervisor `run_loop` (`supervisor.rs:266`) inlined **12 periodic trackers** — a second periodic-work mechanism parallel to the trait `PerTickHandler` set (the #694 / survey-S3 incomplete-extraction drift). This finishes the extraction: each tracker becomes a thin `PerTickHandler` registered in `build_default_handlers`; the inline supervisor calls are deleted.

Migrated (in original order): `anti_stall`, `idle_watchdog`, `decision_timeout`, `helper_staleness`, `mcp_registry`, `waiting_on_stale`, `conflict_notify`, `canonical_drift`, `auto_release`, `dispatch_idle`, `dispatch_idle_nudge`, `retention`.

## How — pure relocation, zero behavior change on unix

- Each wrapper owns its tracker behind a `Mutex` (the trait's `run(&self)` needs interior mutability; `run_loop` held `&mut` locals) and calls `maybe_scan`/`maybe_sweep` **every tick**. The trackers **self-throttle** internally via their own `TICKS_PER_SCAN`, so **no handler-level cadence gate is added** — scan cadence is identical. (Cadence-hoist is W2.4 `CadenceGate`, explicitly after W1.1.)
- Appended to the handler `Vec` in the **same relative order** they ran inline (load-bearing). They ran on the supervisor thread with no in-tick ordering vs the main-loop handlers → tail-append is the conservative choice.
- Main loop ticks at the **identical 10s** interval (`build_tick_infrastructure` ≡ `supervisor::TICK`) and holds no lock across handlers, matching the supervisor's call conditions.
- `conflict_notify`'s per-tick `retain_active` prune (#1923 G5 cleanup-on-delete, was `supervisor.rs:430`) moves into `ConflictNotifyHandler::run` with its tracker. `notify_tracks.retain` stays (it belongs to `tick()`).
- `mcp_registry`'s `DaemonBinaryStale` TUI flag is threaded through `build_default_handlers` so the handler holds the **same `Arc`** app's render loop reads; `supervisor::spawn` no longer takes it. The one-shot boot `gc_stale_activity_sidecars` purge stays in `run_loop` (never a per-tick scan).

## Completeness invariant (core deliverable; #1002/#1719 class)

`all_twelve_supervisor_trackers_registered_in_order` pins that **all 12 migrated names are registered** in `build_default_handlers` AND keep their relative order. The existing `app_tick_handlers_cover_*` (app⊇daemon) invariant **cannot** catch a dropped migration — a forgotten tracker is absent from *both* sets, so the difference stays empty. This pins the full expected set directly, with teeth.

## Platform delta (honest)

On **unix** (both `run_core` and app mode) this is **strictly behavior-preserving** — same cadence, same lock conditions, trackers just move from the supervisor thread to the main-loop handler thread. The **one delta**: Windows headless `run_core`, whose supervisor is `#[cfg(unix)]`-gated and thus *never ran these trackers* — the universal handler set now runs them there too, matching **app mode (the live daemon, which already ran them on every platform)**. That closes a latent #1720-class gap rather than introducing one.

## LOC-EST

Large by design (relocation): `supervisor.rs` **-128 net**; new `per_tick/supervisor_trackers.rs` **~390 LOC** (12 thin wrappers + module doc + invariant test). All files under the LOC cap (`file_size_invariant` green).

## Tests

`cargo fmt --check` clean · `clippy --all-targets --features tray -D warnings` clean · full `nextest` green (the two env-mutually-exclusive `*_1834` git-checkout tests pass under `AGEND_GIT_BYPASS=1`, as CI runs them). `app_tick_handlers_no_panic_on_empty_context` now exercises all 12 new handlers on empty context.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -156,9 +156,11 @@ const APP_TICK_ALLOWLIST: &[&str] = &["snapshot_rotation", "thread_dump"];
 /// `stage2_dispatch_available = false` — `RecoveryDispatcherHandler` now RUNS
 /// (no longer allowlisted out) and its Stage2 path escalates to Stage3 rather
 /// than emit onto the consumerless channel.
-fn app_tick_handlers() -> Vec<Box<dyn crate::daemon::per_tick::PerTickHandler>> {
+fn app_tick_handlers(
+    daemon_binary_stale: crate::daemon::mcp_registry_watcher::DaemonBinaryStale,
+) -> Vec<Box<dyn crate::daemon::per_tick::PerTickHandler>> {
     let (crash_tx, _crash_rx) = crossbeam_channel::bounded(1);
-    let mut handlers = crate::daemon::build_default_handlers(crash_tx, false);
+    let mut handlers = crate::daemon::build_default_handlers(crash_tx, false, daemon_binary_stale);
     handlers.retain(|h| !APP_TICK_ALLOWLIST.contains(&h.name()));
     handlers
 }
@@ -212,11 +214,7 @@ fn run_app(terminal: &mut DefaultTerminal, fleet_override: Option<&Path>) -> Res
     // the daemon already runs its own supervisor against the real registry, so
     // the app must not also poll a disjoint (empty) registry.
     if !attached_mode {
-        crate::daemon::supervisor::spawn(
-            home.clone(),
-            Arc::clone(&registry),
-            Arc::clone(&daemon_binary_stale),
-        );
+        crate::daemon::supervisor::spawn(home.clone(), Arc::clone(&registry));
         crate::instance_monitor::spawn_monitor_tick(home.clone(), Arc::clone(&registry));
         // Attached mode stays unwired: that process never owns the registry,
         // and the Telegram bot (if any) runs under the other daemon which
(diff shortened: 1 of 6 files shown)


suzuke:
VERIFIED source/test review for head 0fa91172ce2fc8ff94769253f0cb424f07dae58f. Not CI-green merge approval yet: gh pr checks still shows macOS/Ubuntu/Windows/Coverage/Daemon-boot pending; audit and LOC pass.\n\n### Evidence\nran: cargo test all_twelve_supervisor_trackers_registered_in_order --all-targets -> pass at restored HEAD; red mutation removing RetentionHandler failed the same test with missing 'retention'.\nran: cargo test app_tick_handlers_no_panic_on_empty_context --all-targets -> pass.\nran: cargo test registered_in_default_daemon_pipeline --all-targets -> pass.\nran: git diff --check 1ef4fc7e23e32c48828e8bb61c4ec33caf779c01..HEAD -> pass.\nran: gh pr checks 2065 --repo suzuke/agend-terminal -> pending CI (not merge approval).\n\nCited source checks: src/daemon/mod.rs:640-651 registers all 12 migrated trackers in old relative order; src/daemon/per_tick/supervisor_trackers.rs:62-303 calls each original maybe_scan/maybe_sweep entry point; src/daemon/per_tick/supervisor_trackers.rs:312-325 and :343-389 pin the 12-name completeness/order invariant; src/daemon/per_tick/supervisor_trackers.rs:194-203 moves conflict_notify retain_active with the tracker while src/daemon/supervisor.rs:350-366 keeps notify_tracks.retain local; src/app/mod.rs:188, :402, :488 wire/read the same DaemonBinaryStale Arc; src/daemon/supervisor.rs:281-289 keeps gc_stale_activity_sidecars as the one-shot boot purge.\n\nWindows headless delta accepted: old headless supervisor spawn remains cfg(unix) at src/daemon/mod.rs:1167-1169, while the new default handler pipeline is built universally at src/daemon/mod.rs:1190-1192. That makes the 12 trackers run in Windows headless; I treat this as intentional parity with the app/default tick pipeline and closure of the mode-drift class, not a Unix behavior regression.

PR #2072, opened 2026-06-12
refactor(W2.3): decompose feed_with_fg into a named apply_*_gate pipeline (structure-only)

suzuke:
## REFACTOR-PLAN W2.3 — `feed_with_fg` gate pipeline

Decomposes `state/mod.rs::feed_with_fg` (was 549 LOC, 8+ levels of nesting) into named `apply_*_gate` methods with the call sequence encoded in a single, readable location. **Structure-only, zero behavior** — the 348 state unit tests + `state_pattern_coverage` + `state_current_mutation_invariant` + `heartbeat_pair_atomicity_audit` all pass with **no assertion changes**.

`feed_with_fg` itself: **549 → 114 lines.**

### The gate sequence (now visible top-to-bottom)
```
feed_with_fg:
  1. apply_hash_dedup_gate            — skip unchanged frames (throttle-hint override intact)
  2. classify (Some arm):
       apply_anchor_gate              — 2a suppress?  (red / content-line / usage-limit regimes)
       apply_position_gate            — 2b suppress?  (scrolled out of live tail)
       ├─ suppressed → handle_suppressed_detection
       └─ else → landing pipeline (order-critical, ONE place):
            apply_working_marker_override   — #1768 / #1777 / #badge-recovery (+ #1809 swallow probe)
            apply_srl_phantom_gate          — #1808 / #1809  (cross-cycle → Idle)
            apply_usage_limit_lifecycle_gate— #1955          (past release → Idle)
            gate_on_heartbeat → transition
     classify (None arm): handle_no_raw_match  — #SRL-phase2 hard-wrap fallback
  3. capture_unclassified_throttle / record_shadow_telemetry / detect_productive_output (instrumentation)
```

### Discipline honored
- **Zero behavior**: pure structural moves; every existing state test green, assertions untouched. clippy `--all-targets --features tray -D warnings` clean.
- **Order in one location**: the landing-pipeline order (working-marker → SRL phantom → usage-limit → heartbeat) is a flat 5-line sequence with an order-critical doc comment; the suppression gates (anchor before position, both before landing) read linearly in the Some arm.
- **Load-bearing instrumentation moves with its gate, unchanged** (un-load-bearing is W4): the `#1809-srl-swallow-probe` stays inside `apply_working_marker_override`; the `#1808-probe0-phantom` WARN stays inside `apply_srl_phantom_gate`; the `non_srl_since_last_srl` telemetry reset stays in the classify orchestrator.
[...]

PR #2074, opened 2026-06-12
fix(test): notification_queue drain tests tolerate #2028 transient Unavailable→empty (#2072 flake)

suzuke:
## The #2072 Coverage flake — RCA + fix

`notification_queue::tests::enqueue_classified_round_trips_actionable_and_deferred_since_1513` failed in the #2072 Coverage job, unrelated to that PR (which only touched `state/mod.rs`).

### Verify-first RCA (corrects my own earlier hypothesis)
On #2072 I guessed "llvm-cov timing flake on the `deferred_since_ms` ms assertion." **That was wrong**, and verifying it falsified it:
- The CI panic was at `notification_queue.rs:571:18` — **column 18 is the `again[0]` index**, not the `assert_eq!` comparison. So `again` was **empty**, and the panic was index-out-of-bounds, not an assertion mismatch.
- `again` is empty because the step-5 `drain()` returned nothing. `drain()` → `drain_with_stale_threshold` → `try_drain_with_stale_threshold`, which returns `Unavailable` (collapsed to `Vec::new()` by `drain()`) when `try_acquire_file_lock` hits a transient open/lock hiccup. The #2028 doc comment on that path calls this out verbatim: *"under llvm-cov-grade load a transient open/lock hiccup was reported as 'queue empty'."* Under the full Coverage suite's parallel fd pressure, the lock-file `open()` transiently failed → `Unavailable` → empty → `again[0]` panicked.

Falsified along the way (so the fix targets the real cause): tight-timing-assert (it's an index panic), `FLOCK_DEPTH` cross-test (it's `thread_local`), cross-`home` contamination (every queue path is `home`-scoped; `home` is PID+suffix unique per test). So this is **neither** a tight timing assert **nor** a #2056 shared-state/isolation case.

### The fix
Production's flusher already retries `drain` next tick — a single `drain()` is **not** authoritative (it's contractually allowed to return empty under transient contention). The tests trusted it. The fix models the retry: a `drain_settled(home, agent, want)` test helper that accumulates `drain()` across a bounded retry budget (200 × 5 ms) until it has `want` items. **Happy path returns on the first attempt → zero behavior change** when the drain succeeds immediately.

Routed the 4 vulnerable "non-empty result from a single drain" assertions through it:
- `enqueue_classified_round_trips_...since_1513` (both drains)
- `drain_roundtrip`
- `drain_recovers_stale_draining_leftover`

(`drain_does_not_steal_fresh_foreign_draining_file` asserts `is_empty()` — a transient empty would falsely *pass*, not flake — so it's left untouched.)

### Deterministic regression test
`drain_settled_retries_through_transient_lock_contention_2072` reproduces the mechanism with **no timing race** (mpsc channel-synchronized): a peer thread holds the drain lock, a one-shot `drain()` is asserted empty (the exact trap), the test signals the peer to release inside the retry window, and `drain_settled` recovers the item. Deterministic — passes 20/20.

### Verification
- `notification_queue` 4/4 green under `cargo llvm-cov` full suite (×2, the real CI condition that flaked).
- Module 10/10 green under parallel `cargo test` (33 tests each).
- Regression test 20/20 stable.
- `cargo fmt` ✓ · `cargo clippy --all-targets --features tray -D warnings` ✓.

(Local `cargo llvm-cov` also surfaces `dispatch_branch_..._1834` — the documented agend-git-shim local-env false-fail, unrelated; passes in CI which has no shim.)

### DIFF
--- a/src/notification_queue.rs
+++ b/src/notification_queue.rs
@@ -542,12 +542,43 @@ mod tests {
         dir
     }
 
+    /// Retry-accumulate `drain` to absorb #2028's transient `Unavailable→empty`.
+    /// `drain()` is contractually allowed to return an empty vec when
+    /// `try_acquire_file_lock` hits a transient open/lock hiccup under heavy
+    /// parallel load (llvm-cov-grade fd pressure) — production's flusher simply
+    /// retries next tick, so a one-shot drain is NOT authoritative. A test that
+    /// trusts it indexes an empty vec → index panic (the #2072 coverage flake,
+    /// notification_queue.rs `again[0]`). This helper models the retry: it keeps
+    /// draining (accumulating, since `drain` is destructive) until it has `want`
+    /// items or the bounded budget elapses. The happy path returns on the first
+    /// attempt — zero behavior change when the drain succeeds immediately.
+    fn drain_settled(home: &Path, agent_name: &str, want: usize) -> Vec<QueuedNotification> {
+        drain_settled_with_stale(home, agent_name, want, STALE_DRAINING_MS)
+    }
+
+    fn drain_settled_with_stale(
+        home: &Path,
+        agent_name: &str,
+        want: usize,
+        stale_ms: u128,
+    ) -> Vec<QueuedNotification> {
+        let mut acc = Vec::new();
+        for _ in 0..200 {
+            acc.extend(drain_with_stale_threshold(home, agent_name, stale_ms));
+            if acc.len() >= want {
+                break;
+            }
+            std::thread::sleep(std::time::Duration::from_millis(5));
+        }
+        acc
+    }
+
     #[test]
     fn enqueue_classified_round_trips_actionable_and_deferred_since_1513() {
         let home = tmp_home("classified");
         enqueue_classified(&home, "a", "work", true).expect("enqueue actionable");
         enqueue(&home, "a", "ambient").expect("enqueue ambient"); // actionable=false default
-        let drained = drain(&home, "a");
+        let drained = drain_settled(&home, "a", 2);
         assert_eq!(drained.len(), 2);
         let actionable = drained
             .iter()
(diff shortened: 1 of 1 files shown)


suzuke:
VERIFIED source/test review for head `6d8116fddec3d7067c0aff365ede95217affecde`.

Scope: `src/notification_queue.rs` test-hygiene fix for the #2072 coverage flake. I verified the RCA chain and the test-only bounded retry helper. No production code changed.

### Evidence
ran: `git show ff28d07c89676ffbcbf35cbd9b436d28bbd87d2c:src/notification_queue.rs | nl -ba | sed -n '520,590p'` -> old panic site shows `again[0].deferred_since_ms` at line 571; column 18 is the index expression.
ran: `rg -n "fn drain|Unavailable|Vec::new|try_acquire_file_lock|drain_settled|again\\[0\\]|mpsc|channel" src/notification_queue.rs -S` -> confirmed #2028 path: `try_drain_with_stale_threshold` returns `DrainAttempt::Unavailable` on held/unopenable lock, and `drain_with_stale_threshold` collapses `Unavailable` to `Vec::new()`.
ran: `git diff --name-only ff28d07c89676ffbcbf35cbd9b436d28bbd87d2c..HEAD` -> only `src/notification_queue.rs` changed; helper is under `#[cfg(test)]` module.
ran: `rg -n "let .* = drain\\(|let .* = drain_with_stale_threshold\\(|assert_eq!\\([^\\n]*drain\\(|assert!\\([^\\n]*drain\\(|\\[[0-9]+\\]" src/notification_queue.rs -S` -> remaining direct `drain` uses are expected-empty/contention or concurrent-drain cases; the four changed sites cover the single-drain/non-empty assertion pattern.
ran: `cargo test notification_queue --all-targets` -> 36 passed, 0 failed.
ran: `for i in $(seq 1 20); do cargo test notification_queue::tests::drain_settled_retries_through_transient_lock_contention_2072 --quiet || exit 1; done` -> 20/20 passed.
ran: `cargo llvm-cov --no-report --all-targets -- notification_queue` -> 36 notification_queue-filtered tests passed under llvm-cov instrumentation.
ran: `gh pr checks 2074 --repo suzuke/agend-terminal` -> LOC overrun, audit, and Coverage passed; macOS/Ubuntu/Windows checks still pending at review time.
cited: `src/notification_queue.rs:442` / `src/notification_queue.rs:449` — flusher-facing `drain_with_stale_threshold` maps `DrainAttempt::Unavailable` to `Vec::new()`.
cited: `src/notification_queue.rs:468` / `src/notification_queue.rs:471` — lock acquisition failure returns `DrainAttempt::Unavailable`.
cited: `src/notification_queue.rs:555` / `src/notification_queue.rs:566` — `drain_settled` is test-only and bounded at 200 attempts with 5ms sleeps, so a permanently broken drain still returns short and caller assertions/indexing fail rather than looping forever or faking green.
cited: `src/notification_queue.rs:622` / `src/notification_queue.rs:637` / `src/notification_queue.rs:643` — regression test uses mpsc handshakes to prove the peer holds the lock before the one-shot drain and releases only after the test observes contention; it is not sleep-only synchronization.

No blocking findings from source review or local/coverage-filtered tests. Merge should still wait for the pending OS matrix checks to finish green.

PR #2083, opened 2026-06-12
fix(monitor): cycle-safe tree_rss — fixes monitor_tick stack overflow

yujunchao:
## What & why

`instance_monitor::tree_rss` sums a process's RSS plus all descendants by recursing over the process table by parent-PID, with **no cycle guard**. On Windows, PID reuse can produce a parent cycle (A's parent is B, B's parent is A), making the recursion unbounded → **stack overflow that crashes the whole daemon** from the `monitor_tick` thread (the 5s resource sampler).

Observed on a live Windows daemon: `EXCEPTION_STACK_OVERFLOW` (0xc00000fd) on the `monitor_tick` thread; the WER minidump showed the same `instance_monitor` frame repeating ~18k times. Intermittent — it needs a PID-reuse parent cycle to form, so uptime varied before each crash.

This replaces the recursive walk with an **iterative walk + a `visited` set** (cycle-safe; also guarantees each process is counted at most once).

## Prior art check (required)

- [x] Checked `docs/KNOWN_ISSUES.md` — not listed there.
- [x] Compared against current `main` — `tree_rss` in `src/instance_monitor.rs` is still the unguarded recursive walk (`.filter(|p| p.parent() == Some(root)).map(|p| tree_rss(sys, p.pid())).sum()`).
- [x] Searched open/closed issues and PRs for `tree_rss` / `monitor_tick` / stack overflow — none found.

- `cargo build --release` of the fix set — compiles clean.
- `cargo fmt --check` on this branch — clean.
- 3 regression tests added (above), written to fail-before / pass-after: `sum_tree_rss_survives_parent_cycle` and `sum_tree_rss_survives_self_parent` both recurse to stack overflow under the old `tree_rss`; the iterative version returns. Plus `sum_tree_rss_sums_root_and_descendants` for the happy path.
- CI runs the full `cargo fmt --check` + `cargo clippy --all-targets -- -D warnings` + test matrix on Ubuntu/macOS/Windows (`.github/workflows/ci.yml`) — please rely on the PR checks as the cross-platform gate.

- [x] No tier (a)/(b) on-disk format is changed — this is an in-memory computation only.

## Notes for reviewers

- Root cause is that `sysinfo`'s `parent()` graph can be cyclic under Windows PID reuse; the previous code had no termination guard for that case.
- The walk is extracted into a pure `sum_tree_rss(root, &HashMap<u32, ProcNode>)` so the cycle cases are unit-testable without spawning real processes; `tree_rss` just builds the table from `sysinfo` and delegates.

### DIFF
--- a/src/instance_monitor.rs
+++ b/src/instance_monitor.rs
@@ -127,16 +127,55 @@ pub fn collect(home: &std::path::Path, registry: &crate::agent::AgentRegistry) {
     *cache().lock() = metrics;
 }
 
-/// Recursively sum RSS for a process and all its descendants.
+/// One node of the process table used for the descendant-RSS walk.
+struct ProcNode {
+    mem: u64,
+    parent: Option<u32>,
+}
+
+/// Sum RSS for `root` and all its descendants, cycle-safe and iterative.
+///
+/// Windows reuses PIDs aggressively, so the parent-PID graph can contain a
+/// cycle (A's parent is B, B's parent is A). The previous recursive walk would
+/// then recurse forever and overflow the stack (monitor_tick thread crash,
+/// 2026-06-12 15:13, ~18k frames). The `visited` set breaks cycles and also
+/// guarantees each process is counted at most once.
+fn sum_tree_rss(root: u32, table: &std::collections::HashMap<u32, ProcNode>) -> u64 {
+    let mut visited = std::collections::HashSet::new();
+    let mut stack = vec![root];
+    let mut total = 0u64;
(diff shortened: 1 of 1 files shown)


suzuke:
Merged — thank you @yujunchao! 🙏 Clean cycle-safe rewrite (recursive → iterative + visited set), and the WER minidump evidence + the three regression tests (normal tree / A↔B cycle / self-parent) made this trivial to verify. Solid catch on the Windows PID-reuse stack overflow.

PR #1110, opened 2026-05-24
fix(#1109): low findings batch — comms & router review

suzuke:
## Summary
- **L1**: Reject empty messages in `handle_send_to_instance` — consistent with broadcast/delegate/report handlers
- **L2**: Char-boundary-safe mirror truncation — scans backward from `MAX_MIRROR_LEN` to find valid UTF-8 boundary instead of byte-slicing
- **L3**: Fix `try_recv` data drop in `buffers.retain` — received data now pushed into agent buffer instead of being silently consumed
- **L4**: Remove stale `interrupt`/`reason` backwards-compat (44+ sprints overdue) — callers must use `force`/`force_reason`

3 files, +84 -46. All 2749 tests pass. Clippy clean.

Fixes #1109

## Test plan
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] `cargo test` — all tests pass
- [x] New tests: `mirror_truncation_safe_on_multibyte_utf8`, `retain_preserves_received_data`
- [x] Updated test: `test_delegate_task_old_interrupt_field_no_longer_bypasses_busy_gate`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1109, opened 2026-05-24
Comms low findings batch: empty message default, UTF-8 truncation, router try_recv data drop, stale Sprint 11 deprecation


suzuke:
### L1: Empty message silently defaults to "" (comms.rs:81-84)
handle_send_to_instance defaults missing message to empty string. Other handlers correctly reject missing text.

### L2: Mirror text truncation panics on multi-byte UTF-8 (router.rs:169-172)
`&text[..MAX_MIRROR_LEN]` byte-slices can land mid-codepoint. Use char-boundary-safe truncation.

### L3: Router buffers.retain try_recv drops one data chunk per cycle (router.rs:121-126)
Dead-channel cleanup calls try_recv() which consumes and discards data if Ok.

### L4: Stale Sprint 11 deprecation warning (comms.rs:430)
"will be removed Sprint 11" — current is Sprint 55+. Either remove deprecated fields or update timeline.

## Constraint
Characterization tests first for L2 and L3 (behavioral). L1 and L4 are trivial fixes.

PR #2084, opened 2026-06-12
fix(router): char-boundary-safe mirror buffer drain (no panic on multibyte UTF-8)

yujunchao:
## What & why

The mirror buffer cap in `src/daemon/router.rs` drains by a **raw byte index**:

```rust
if buf.buffer.len() > MAX_MIRROR_LEN * 2 {
    let drain = buf.buffer.len() - MAX_MIRROR_LEN;
    buf.buffer.drain(..drain);   // panics if `drain` is not a char boundary
}
```

When the buffer holds multibyte UTF-8 (e.g. CJK), `drain` can land mid-character and `String::drain(..n)` panics (`assertion failed: self.is_char_boundary(n)`), killing the mirror thread. Observed: `router.rs` panicked at this line twice on a live deployment carrying Chinese text.

This adds a `mirror_drain_start()` helper that snaps the drain index forward to the next char boundary, plus regression tests.

## Prior art check (required)

- [x] Checked `docs/KNOWN_ISSUES.md` — not listed there.
- [x] Compared against current `main` — the `buf.buffer.drain(..drain)` path is still a raw byte index; only the sibling `mirror_text` slice path has an `is_char_boundary` guard.
- [x] Searched closed issues/PRs — found **#1109 / #1110** ("low findings batch — comms & router review"), which made the **`mirror_text` slice** char-boundary-safe (`&text[..MAX_MIRROR_LEN]` → an `is_char_boundary` back-off loop). This PR fixes the **remaining sibling call site** — the `buf.buffer.drain` cap — that #1110 did not touch. Same bug class, different line.

- `cargo build --release` of the fix set — compiles clean.
- `cargo fmt --check` on this branch — clean.
- 2 regression tests added (above): `mirror_drain_start_never_splits_multibyte_char` drains a buffer whose CJK chars straddle the cap (panics before the fix, passes after) and `mirror_drain_start_is_zero_when_short`.
- CI runs the full `cargo fmt --check` + `cargo clippy --all-targets -- -D warnings` + test matrix on Ubuntu/macOS/Windows (`.github/workflows/ci.yml`) — please rely on the PR checks as the cross-platform gate.

- [x] No tier (a)/(b) on-disk format is changed.

## Notes for reviewers

- Mirrors #1110's approach (snap to a char boundary) but for the buffer-drain cap rather than the text-slice cap, so the two sibling caps are now both UTF-8-safe.

### DIFF
--- a/src/daemon/router.rs
+++ b/src/daemon/router.rs
@@ -20,6 +20,19 @@ const SILENCE_TIMEOUT: Duration = Duration::from_secs(3);
 /// Maximum mirror text length (chars) to dispatch.
 const MAX_MIRROR_LEN: usize = 4000;
 
+/// Byte index at which to start draining the mirror buffer to keep it under
+/// `MAX_MIRROR_LEN`, snapped forward to the next char boundary so we never slice
+/// a multi-byte UTF-8 char (e.g. Chinese) mid-character. `String::drain(..n)`
+/// panics when `n` is not a char boundary (router.rs char_boundary panics on
+/// long Chinese turns, 2026-06-12).
+fn mirror_drain_start(buf: &str) -> usize {
+    let mut drain = buf.len().saturating_sub(MAX_MIRROR_LEN);
+    while drain < buf.len() && !buf.is_char_boundary(drain) {
+        drain += 1;
+    }
+    drain
+}
+
 /// Registration message: agent name + PTY output receiver.
 pub struct AgentSubscription {
     pub name: String,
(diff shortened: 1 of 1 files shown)


suzuke:
Merged — thank you @yujunchao! 🙏 Correct forward-snap to a char boundary, and we independently confirmed the call site is wired (the run-loop cap-drain now goes through `mirror_drain_start`). This one fixes a panic our maintainer hits daily with Chinese text, so it's especially appreciated.

One small follow-up (non-blocking, tracked separately): the regression tests exercise `mirror_drain_start` at the unit level; a test that drives the actual run-loop cap-drain path (push >2×MAX_MIRROR_LEN of CJK through the buffer-consume loop, assert no panic + buffer ≤ cap, mirroring `retain_preserves_received_data`) would pin the wiring against future regressions. If you'd like to add it we'd merge it gladly — otherwise we'll pick it up.

Issue #2086, opened 2026-06-13
daemon blind to a STATIC ServerRateLimit error screen during idle — no SRL latch, no recovery inject for ~26min (operator-hit)


suzuke:
## Operator-observed incident (2026-06-13)

Operator left the machine, returned to find the fixup-lead pane showing \`API Error: Server is temporarily limiting requests (not your usage limit) · Rate limited\`, **waited a long time (~20+ min) and saw ZERO recovery inject**, then manually poked the agent to recover.

## Hard evidence (logs, not inference)

- **18:02–18:28 (~26 min): zero daemon records for fixup-lead** — nothing in `app.<date>.log`, nothing in `state-transitions.jsonl`. Complete silence.
- **fixup-lead has NO ServerRateLimit transition in `state-transitions.jsonl`** at all (grepped twice, empty).
- The ONLY `ServerRateLimit detected, scheduling retry (Phase A) delay_secs=5` line is at **18:28:53.545**, in the SAME second as the operator's `UserPromptSubmit` hook (18:28:53.781). It was cleared 40s later with `retry_count=0` — i.e. the scheduled retry never fired; the operator's own prompt recovered the agent.

## What this means (confident)

The static error screen was **never latched as ServerRateLimit during the idle window**, so the supervisor's SRL backoff/recovery (detect → Phase A/B/C → continue-inject, supervisor.rs) **never engaged**. The 18:28:53 detect was triggered by the operator's poke, not by the daemon's own per-tick scan. The recovery machinery is real and correct in code — but it never ran here because detection never latched.

## Hypothesis (UNVERIFIED — do not fix on this alone)

Suspect the state detector's hash-dedup (`state/mod.rs` feed: "skip if screen unchanged"). The error screen is static after it appears → hash unchanged → per-tick classification skipped → state stays at the pre-error Idle → SRL never latches → no recovery. This is a code-trace guess; the lead was confident-but-wrong twice on this incident already, so it MUST be confirmed by instrumentation, not by reading code.

## Ask: instrument-only first (#1808 discipline)

Add per-tick logging on the SRL detect / dedup path: for an idle agent whose screen matches the ServerRateLimit pattern, log every tick what the detector sees (screen hash, dedup-skip decision, classified state, latch result). Then reproduce (or wait for recurrence) to capture WHY a static SRL screen isn't latched. Only after the log says what's happening, design the fix.

[...]

PR #2091, opened 2026-06-13
fix(#2089): rescue a hard-wrapped SRL that detect lands as Idle (narrow-pane word-wrap)

### LINKED ISSUE #2089: Sibling of #2086: SRL on screen but detected as Idle (position-gate suppression?) — fixup-reviewer-2 stuck, no retry
## Sibling of #2086 — a SECOND SRL suppression path, still unfixed

**Operator-observed**: fixup-reviewer-2's pane persistently shows `API Error: Server is temporarily limiting requests · Rate limited`, but the daemon reports the agent as **Idle** (list_instances + hook-shadow log screen_state=Idle). Detection failure — no SRL latch → no recovery retry.

**Lead pane_snapshot (ground truth)** confirmed the screen really shows the SRL error:
```
⏺ API Error: Server is temporarily limiting requests (not your usage limit) · Rate limited
✻ Baked for 23s
(blank)
────────  ❯  ──────── (input box)
```
Layout: SRL error ABOVE, then a "Baked" spinner + blank + input box → **SRL sits far from the bottom**.

**Hypothesis (UNVERIFIED — lead must not code-trace this; dev-2 to confirm empirically)**: This is the **position-gate** path that dev-2 found during the #2086 investigation and left a repro for but did NOT fix. #2086's real incident (fixup-lead) turned out to be the working-marker override; position gate was "another real suppression but not that culprit." fixup-reviewer-2 looks like the position-gate real incident: SRL pushed >ERROR_TAIL_SCAN_LINES (15) from the bottom by the input-box+spinner chrome → judged stale → suppressed → never latches. #2086's fix (working-marker override gating) does not cover this path.

**RCA dispatched** (dev-2, spike/2089): pull fixup-reviewer-2's real screen bytes from unclassified_errors.jsonl, reproduce via real vterm→state.feed, instrument to confirm WHICH gate suppresses (position gate vs other) — log evidence, not code-trace — then fix (if position gate: make the tail-distance metric skip the input-box chrome, reusing existing input-line markers; §3.9 dual-control so genuine scrolled-away stale errors are still suppressed, #1518 not weakened). Dual review.

This validates dev-2's decision to keep the position-gate repro: it was a real latent path, now hit in production.

suzuke:
Fixes #2089 (sister of #2086). **SENSITIVE — FP-suppression core → dual review requested.**

## Symptom
fixup-reviewer-2's pane showed a live `Server is temporarily limiting requests` but the daemon classified **Idle** → no SRL recovery/retry. Operator-hit. #2086's fix doesn't cover it (different path).

## RCA (empirical — real captured screen + instrumented trace; the position-gate hypothesis was falsified)
On a **narrow pane (~24 cols)** the SRL error **word-wraps across ~7 rows**, so the single-line `detect_with_match` regex misses it and matches the idle `❯` prompt → `Some(Idle)`. Instrumented:
```
detect_with_match(raw)  = Some(Idle)            ← ❯ prompt; wrapped SRL invisible to single-line regex
detect_with_match(flat) = Some(ServerRateLimit) ← flattening recovers it
flattened_throttle_detect = None                 ← but the fallback rejects
full feed → Idle
```
Two compounding gaps:
1. The `#SRL-phase2` hard-wrap fallback (`flattened_throttle_detect`) lived **only in the `None` arm** — a wrapped SRL co-existing with the idle prompt (`Some(Idle)`, not `None`) never reached it.
2. Even forced, the bottom-15 flatten started mid-phrase; the `⏺ API Error:` indicator wrapped **above** the window, so `throttle_indicator_adjacent` (the #1857 prose-FP guard) found no indicator and rejected.

(Distinct from #2086's working-marker override and from the #1518 position gate — full write-up in the [#2089 issue comment](https://github.com/suzuke/agend-terminal/issues/2089#issuecomment-4698561624).)

## Fix (A + B)
- **A** — run the hard-wrap fallback (`try_hard_wrap_throttle`) when raw detection lands a **benign Idle/Thinking** (not just `None`), gated by a cheap `screen_has_throttle_hint` pre-filter so the common no-throttle screen stays on the fast path. A real wrapped throttle overrides idle-prompt chrome.
- **B** — widen the fallback's flatten window (`HARD_WRAP_TAIL_LINES = 40`) so a wrapped `⏺ API Error:` indicator above the bottom-15 is captured. **Only the line span widens** (still the visible grid `tail_lines` produced, never scrollback); the **char-proximity guard is unchanged** — a distant unrelated indicator is still rejected. Keeps the FP guard, does **not** exempt it.

## §3.9 dual-control (RED-verified)
- `srl_hard_wrapped_narrow_pane_latches_2089` — wrapped SRL + indicator above the 15-window → **latch**. **RED-verified**: with `HARD_WRAP_TAIL_LINES=15` this test FAILS (indicator unreachable), proving B is load-bearing; preconditions assert the indicator is outside-15 / inside-40.
- `prose_throttle_mention_without_indicator_not_misclassified_2089` — prose mentioning "limiting/throttle" with **no** error indicator → does **not** latch (the FP guard A's wider reach depends on).
- `srl_narrow_wrap_real_capture_reproduces_idle_suppression_2089` — the real captured (15-line-truncated) screen reproduces the single-line miss.

`cargo clippy --all-targets --features tray -D warnings` clean; 466 state+supervisor tests green. Real captured screen committed as a regression fixture (truncated to the #1562 instrument's 15 rows; the full-height latch is pinned via the faithful reconstruction above).

Note (follow-up): narrow-pane hard-wrap makes ANY single-line state regex fragile, not just SRL — a systemic state-detection concern. This PR scopes to SRL (operator-hit + has a fallback to extend); the general case is a follow-up.

### DIFF
--- a/src/state/mod.rs
+++ b/src/state/mod.rs
@@ -396,6 +396,14 @@ pub(crate) const SERVER_RATE_LIMIT_RECOVERY_SILENCE: Duration = Duration::from_s
 /// structural-marker scan) — do not collapse the two.
 const ERROR_TAIL_SCAN_LINES: usize = 15;
 
+/// #2089: window for the hard-wrap throttle fallback's flatten. A long SRL error
+/// on a narrow (~24-col) pane wraps across ~7-9 rows; with the spinner + input
+/// box below, its `⏺ API Error:` indicator prefix lands well above
+/// `ERROR_TAIL_SCAN_LINES`. This wider span captures the wrapped indicator
+/// (still only the visible grid, never scrollback) so `throttle_indicator_adjacent`
+/// can validate the real throttle; the char-proximity guard keeps it FP-safe.
+const HARD_WRAP_TAIL_LINES: usize = 40;
+
 fn recent_screen_tail(screen_text: &str, n: usize) -> String {
     let lines: Vec<&str> = screen_text.lines().collect();
     let start = lines.len().saturating_sub(n);
@@ -847,7 +855,14 @@ fn flattened_throttle_detect(
     screen_text: &str,
     input_line_markers: &[&str],
 ) -> Option<AgentState> {
-    let tail = recent_screen_tail(screen_text, ERROR_TAIL_SCAN_LINES);
+    // #2089: a long SRL error on a NARROW pane word-wraps across many rows, so its
+    // `⏺ API Error:` indicator prefix sits well above the bottom-`ERROR_TAIL_SCAN_LINES`
+    // window — `throttle_indicator_adjacent` then can't find it and the rescue
+    // wrongly rejects (the fixup-reviewer-2 miss). Flatten a LARGER window so the
+    // wrapped indicator is included; this only widens the LINE span (the visible
+    // grid `tail_lines` produced — never scrollback), and the char-proximity guard
+    // in `throttle_indicator_adjacent` still rejects a distant unrelated indicator.
+    let tail = recent_screen_tail(screen_text, HARD_WRAP_TAIL_LINES);
     // #1947: drop input / user-message lines BEFORE flattening — flattening
     // destroys the line structure the input-line exclusion needs, so an
     // operator-typed error string would otherwise flatten into a legit-looking
@@ -1231,6 +1246,22 @@ impl StateTracker {
             // `anchor_on_red` is false (Shell/Raw backends) OR no color mask
             // was supplied (text-only callers).
             match patterns.detect_with_match(screen_text) {
+                Some((detected, matched))
+                    if !is_high_fp_state(detected)
+                        && !matches!(detected, AgentState::UsageLimit)
+                        && screen_has_throttle_hint(screen_text)
+                        && self.try_hard_wrap_throttle(patterns, screen_text) =>
+                {
+                    // #2089: `detect_with_match` landed a BENIGN state (Idle/Thinking
+                    // — the ❯ prompt / a spinner) because a long SRL error
+                    // word-wrapped across a narrow pane and the single-line regex
+                    // missed it. The cheap throttle-hint pre-filter gates the cost;
+                    // `try_hard_wrap_throttle` re-detected the throttle on the
+                    // flattened tail and transitioned to it (overriding the
+                    // idle-prompt chrome). Nothing more to do this arm. `matched` is
+                    // unused on this path.
+                    let _ = matched;
+                }
                 Some((detected, matched)) => {
                     let high_fp = is_high_fp_state(detected);
                     let usage_limit = matches!(detected, AgentState::UsageLimit);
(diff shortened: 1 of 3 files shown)


suzuke:
VERIFIED

scope_source: fixup-lead task `t-20260613130140795903-3` / message `m-20260613130208082048-20`
audit_mode: dual review, focused on #2089 hard-wrap SRL fix: RED mutation, visible-grid window, fixture fidelity, and §3.9 controls
reviewed_head: 1806248fb4c4946d50a3da7433ae03a5edda4f0e
files: `src/state/mod.rs`, `src/state/tests.rs`, `tests/fixtures/state-replay/claude-srl-narrow-wrap-2089.raw`

### Evidence
ran: `cargo test 2089 -- --nocapture` → PASS; the three #2089 tests passed, including real captured suppression, full-height latch, and prose no-indicator negative control.
ran: temporary mutation `HARD_WRAP_TAIL_LINES = 40` → `15`, then `cargo test srl_hard_wrapped_narrow_pane_latches_2089 -- --nocapture` → FAIL at `src/state/tests.rs:2356` with `B precondition: the widened window must reach the wrapped indicator`; restored to 40 and reran the same test → PASS.
ran: `gh pr checks 2091 --repo suzuke/agend-terminal` → PASS for macOS, Ubuntu, Windows, Coverage, LOC overrun, and audit.
cited: `src/agent/mod.rs:1489` — production feeds state from `c.vterm.tail_lines_with_fg(rows)` where `rows = c.vterm.rows()`, so the state screen is the visible grid.
cited: `src/vterm.rs:620` — `tail_lines_impl` builds from `self.term.grid()` and `self.rows`, not scrollback.
cited: `src/state/mod.rs:405` and `src/state/mod.rs:865` — `HARD_WRAP_TAIL_LINES` is 40 and `flattened_throttle_detect` applies `recent_screen_tail(screen_text, HARD_WRAP_TAIL_LINES)` to the already-visible screen.
cited: `src/state/tests.rs:2326`-`src/state/tests.rs:2341` equivalent block inspected — reconstructed fixture keeps the real capture's wrap pattern (`Server is` / `temporarily` / `limiting` / `requests (not` / `your usage` / `limit) · Rate` / `limited`) and adds only the missing full-height `API Error` row above the bottom-15 capture.
cited: `tests/fixtures/state-replay/claude-srl-narrow-wrap-2089.raw` stripped output lines 1-15 show the same truncated wrap pattern starting at `Server is` and ending at the input box/status rows.

Notes: §3.9 dual controls are bound: the positive hard-wrapped SRL requires the indicator inside 40 but outside 15; the prose mention without indicator does not latch `ServerRateLimit` or `RateLimit`. No unrelated files are changed in the review worktree.

Issue #2090, opened 2026-06-13
Systemic: narrow-pane hard-wrap breaks single-line-regex state detection (root cause behind #2089)


suzuke:
## Systemic: narrow-pane hard-wrap breaks single-line-regex state detection

**Root problem (generalized from #2089)**: the state detector matches PTY screen text with **single-line regexes** (per-backend `StatePatterns::detect_with_match`). When a pane is narrow, long status/error strings **hard-wrap** across multiple terminal rows. A single-line regex cannot match text split by a hard wrap → detection misses → wrong state (e.g. #2089: a narrow ~24-col pane wrapped the SRL error into 7 lines → detector matched the bottom `❯` idle prompt instead → reported Idle → no recovery).

SRL has a partial mitigation (`flattened_throttle_detect`, tail-flatten fallback) but it's SRL-specific, only runs in the `None` arm, and its indicator guard can fail (#2089). **Every other state that relies on a long single-line pattern (ApiError / RateLimit / ContextFull / AuthError / ModelUnsupported) is equally vulnerable in a narrow pane**, mostly without any fallback.

## Root-fix direction (UNVERIFIED — design spike to confirm, lead must not code-trace this)

Hypothesis: vterm (alacritty_terminal) cells carry a **wrap flag** (WRAPLINE) marking rows that were hard-wrapped due to width vs. genuine content newlines. If the detector **reflows hard-wrapped rows back into logical lines BEFORE running the regexes** (using the wrap flag, not guessing), every single-line pattern becomes width-immune at once — and it won't mis-merge genuine newlines because the wrap flag distinguishes them.

Spike must determine (cite file:line, empirical):
1. Does our vterm layer expose the per-row wrap flag? (alacritty_terminal cell/line metadata)
2. Can detection reflow on it before `detect_with_match`, and what's the FP risk (does reflow ever merge lines that should stay separate)?
3. Relationship to existing per-pattern fallbacks (`flattened_throttle_detect`): can they retire once reflow lands, or coexist?
4. Blast radius: this touches the core detection input for ALL patterns → phased plan, smallest safe first step.

## Scope split
- **#2089** (in progress, dev-2): SRL-specific A+B fix = immediate stopgap for the operator-hit case.
- **This (#2090)**: the systemic root fix. Design spike first (core detection change), then lead review + likely operator sign-off before impl.

suzuke:
## Spike result (dev, evidence-based) — lead's WRAPLINE hypothesis FALSIFIED

[...]

PR #2096, opened 2026-06-13
instrument(#2090 P2): zero-behavior shadow for UsageLimit/ContextFull hard-wrap misses

suzuke:
#2090 P2, **instrument-first** (decision d-20260613151736254757-0). Replaces the originally-scoped P2 rescue caller after my premise-check found the §3.9 real-fixture requirement unsatisfiable.

## Why instrument-first (not a rescue caller)
- **Zero evidence** that UsageLimit/ContextFull actually mis-detect on narrow panes — SRL had the #2089 live capture; these have none (the #1562 instrument is throttle-scoped, so it never captured them; existing fixtures are wide; a wide dump can't be re-rendered into a *hard* wrap).
- **UsageLimit flatten is net-negative risk**: the banner is indicator-less (weak anchor) and #1955 self-poisoning is a known FP class that content-blind flatten would *amplify* — while the under-report it'd fix is unproven.
- So: **measure first**. The P2 rescue callers stay gated on this data; ContextFull especially.

## What it does (zero behavior)
After classify, for UsageLimit + ContextFull, compares a flattened-tail re-detect vs the raw single-line grid detect. When the flatten finds a target the raw grid MISSED (narrow-pane hard-wrap), it side-logs a candidate to `hardwrap_miss_shadow.jsonl` (parallel to the #1562 sidecar): `{ts, backend, pattern, raw_state, landed_state, matched, raw_tail}`. The ANSI `raw_tail` lets the real-miss-vs-prose-quote FP profile be judged offline — no in-process counter (avoids a global-state flake; the candidate set *is* the divergence data).

## Zero-behavior proof
Runs after classify; reads `patterns` + `self.current`; mutates ONLY the dedup latch; returns nothing → landed state byte-identical. The **384 passing state tests** (which assert classification outcomes) are the zero-regression proof. `hardwrap_miss_shadow_is_zero_behavior_2090` pins that a wrapped UsageLimit the raw grid misses is NOT rescued/relatched.

## Noise discipline (per the dispatch)
- Cheap keyword pre-filter (`HARDWRAP_SHADOW_HINTS`) avoids flattening benign frames.
- Per-`(pattern, hash(matched))` **fire-once latch** on top of the feed-level screen hash-dedup → a static banner logs once.
- IO errors swallowed (diagnostic must never affect behavior).

## Testability
Detection extracted to a pure `hardwrap_miss_candidate` (no IO/env) → unit-tested directly: fires on wrapped UL/CF the raw grid missed; NOT on a non-wrapped banner the raw grid already detects; NOT on a bare-keyword benign frame. **No per-pattern anchor guard** is applied — the shadow deliberately captures the raw flatten-vs-raw divergence so the full FP profile is visible (the anchor is a future-caller concern).

`cargo clippy --all-targets --features tray -D warnings` clean.

### DIFF
--- a/src/state/mod.rs
+++ b/src/state/mod.rs
@@ -225,6 +225,10 @@ pub struct StateTracker {
     /// call. Used to skip re-detection when the screen hasn't changed —
     /// crucial for not resetting `last_output` on cursor-blink noise.
     last_screen_hash: Option<u64>,
+    /// #2090 P2 shadow-instrument fire-once latch: `(pattern, hash(matched))` of
+    /// the last logged hard-wrap-miss candidate, so a static narrow-pane banner
+    /// isn't re-logged every frame (on top of the feed-level screen hash-dedup).
+    last_hardwrap_miss_sig: Option<(AgentState, u64)>,
     patterns: Option<&'static StatePatterns>,
     /// Context% telemetry: compiled `BackendProfile::context_pattern`
     /// (capture group 1 = the backend's self-reported context usage percent).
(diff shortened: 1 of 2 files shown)


suzuke:
VERIFIED — PR #2096 at aad1545388315b1cfd2d72fdd60aec90429b900a matches decision d-20260613151736254757-0 for instrument-first hard-wrap miss telemetry, with no rescue caller or behavior-changing latch.

scope_source: d-20260613151736254757-0 (#2090 P2 instrument-first shadow for UsageLimit/ContextFull hard-wrap misses)
audit_mode: full_review
reviewed_head: aad1545388315b1cfd2d72fdd60aec90429b900a
files: src/state/mod.rs, src/state/tests.rs
commands: cargo test hardwrap_miss -- --nocapture; cargo test state::tests:: -- --nocapture; gh pr checks 2096 --repo suzuke/agend-terminal

### Evidence
ran: cargo test hardwrap_miss -- --nocapture -> passed; 2 tests passed, including hardwrap_miss_shadow_is_zero_behavior_2090 and hardwrap_miss_candidate_detects_wrapped_target_2090.
ran: cargo test state::tests:: -- --nocapture -> passed; 331 tests passed.
ran: gh pr checks 2096 --repo suzuke/agend-terminal -> passed; ubuntu, macOS, Windows, Coverage, LOC overrun, and audit all passed.
cited: src/state/mod.rs:940 — hardwrap_miss_candidate is a pure helper over patterns/screen_text/input_line_markers, with no IO/env calls.
cited: src/state/mod.rs:971 — the raw-hit guard returns None when raw detection already found the same UsageLimit/ContextFull state.
cited: src/state/mod.rs:1323 — feed_with_fg performs classification/transition before invoking capture_hardwrap_miss_shadow.
cited: src/state/mod.rs:2178 — the per-candidate signature uses (flat_state, hash_screen(&matched)); repeated identical signature returns before writing.
cited: src/state/mod.rs:2197 — append_jsonl errors are swallowed with debug logging only.
cited: src/state/tests.rs:2272 — hardwrap_miss_shadow_is_zero_behavior_2090 asserts the wrapped UsageLimit shadow does not land UsageLimit.
cited: src/state/tests.rs:2295 — hardwrap_miss_candidate_detects_wrapped_target_2090 covers wrapped UL/CF positives and raw-hit/benign negatives.

Notes: the missing proximity anchor is intentional here: the helper is measuring raw flatten-vs-raw divergence for false-positive profiling, not promoting the signal into state. Noise discipline is bounded by feed-level screen-hash dedup plus the last candidate signature latch; I found no behavior-changing path beyond that latch mutation and telemetry write attempt.

Issue #2085, opened 2026-06-13
test: #2084 follow-up — add a real-entry regression test for router mirror cap-drain (CJK, no-panic)


suzuke:
Follow-up from PR #2084 review (reviewer-2 caveat, lead-accepted as non-blocking).

#2084 fixed the router mirror buffer cap-drain panic (raw byte index → mid-char drain on CJK → `String::drain` panic). The fix is correct and the call site is wired (`buf.buffer.drain(..mirror_drain_start(&buf.buffer))` at router.rs ~:114), verified by reviewer grep + RED test on `mirror_drain_start`.

**Gap**: the regression tests exercise `mirror_drain_start` at the unit level + a manual `owned.drain(..start)`; they do NOT drive the run-loop cap-drain path (the spawned channel-consume loop where the original panic occurred). The wiring is one visible line and reviewer-confirmed, so risk is low — but per §3.9 (representative-fixture-through-real-entry, the class that has bitten us repeatedly) a real-entry test would pin it against a future regression that reverts the wiring to a raw byte index.

**Ask**: push >2×MAX_MIRROR_LEN of CJK through the buffer-consume path, assert no panic + buffer ≤ cap. Pattern to mirror: `retain_preserves_received_data` (channel-consume push_str) in router.rs. Offered to @yujunchao in the PR thread; otherwise good-first for the team.

Labels: good-first, test-hygiene, low-priority.

suzuke:
Fixed by #2102 — apply_mirror_cap helper extracted (byte-identical) + run_loop_cap_drain_never_panics_on_oversized_cjk §3.9 test driving the real run-loop cap-drain path with >2*MAX CJK (RED-verified against raw-byte-index regression). merged @ origin/main.

PR #2102, opened 2026-06-13
test(#2085): router mirror cap-drain §3.9 real-entry CJK regression

suzuke:
## What

Closes the **§3.9 coverage gap** left by #2084. That PR fixed the run-loop mirror cap-drain multibyte panic (`mirror_drain_start` snaps the drain index to a char boundary), but **no test drove the run-loop `push_str`+`drain` cap path** (`router.rs` `run_loop` L113-117) — the exact site where the original `drain(..n)` panic fired. Existing pins cover only the index helper (`mirror_drain_start_never_splits_multibyte_char`) and the `try_dispatch_mirror` *slice* path (`mirror_truncation_safe_on_multibyte_utf8`, :184); neither calls the cap-drain.

- **Extract** the per-event cap into `apply_mirror_cap(&mut String)` — verbatim from the inline run-loop block (`if buf.len() > MAX_MIRROR_LEN*2 { let drain = mirror_drain_start(buf); buf.drain(..drain) }`); the call site becomes `apply_mirror_cap(&mut buf.buffer)`. **Behavior byte-identical** — existing router tests pass with zero changes.
- **New test** `run_loop_cap_drain_never_panics_on_oversized_cjk`: push >2×`MAX_MIRROR_LEN` of 3-byte CJK (`界`) through `apply_mirror_cap` repeatedly (mirroring the run-loop per-event consume), assert **no panic** + buffer stays `<= 2*MAX_MIRROR_LEN` + ends on a char boundary.

## RED proof (confirmed locally, then reverted)

Reverting `mirror_drain_start` to the raw byte index (`buf.len().saturating_sub(MAX_MIRROR_LEN)` without the char-boundary while-snap = the original #2084 bug form) makes the new test **panic**: `assertion failed: self.is_char_boundary(end)` at the `drain(..)` call. Restored → green. So the test genuinely catches the regression.

## Scope & verification

Scope: extract helper + add test only; cap-drain decision logic unchanged. Source of truth = issue #2085; freshness = `origin/main @ ee8ff8f`.

- All **10** `daemon::router::tests` green (9 existing zero-regression + 1 new).
- `cargo fmt` / `cargo clippy --all-targets --features tray -- -D warnings` / Windows `x86_64-pc-windows-msvc` cross-check — clean.
- Full `cargo nextest run --features tray --no-fail-fast`: **4121 passed / 1 failed / 33 skipped**. The 1 failure is `api::handlers::messaging::tests::dispatch_branch_skips_metadata_stub_but_checks_out_real_source_1834` — a pre-existing agent-git-shim env artifact (passes *with* `AGEND_GIT_BYPASS`, fails without; mirror-image of `..._no_bypass_1899`), in a file untouched by this change. CI's clean env passes both.

Diff: 1 file, `+46 / -4`.

### DIFF
--- a/src/daemon/router.rs
+++ b/src/daemon/router.rs
@@ -33,6 +33,18 @@ fn mirror_drain_start(buf: &str) -> usize {
     drain
 }
 
+/// Cap the mirror buffer in place: once it grows past `2 * MAX_MIRROR_LEN`,
+/// drain the front back under the cap, starting at the char-boundary-snapped
+/// index from [`mirror_drain_start`] so a multi-byte char is never split (the
+/// #2084 `drain(..n)` panic on a non-boundary `n`). Extracted verbatim from the
+/// run-loop per-event consume so the cap-drain path is directly unit-testable.
+fn apply_mirror_cap(buf: &mut String) {
+    if buf.len() > MAX_MIRROR_LEN * 2 {
+        let drain = mirror_drain_start(buf);
+        buf.drain(..drain);
+    }
+}
+
 /// Registration message: agent name + PTY output receiver.
 pub struct AgentSubscription {
     pub name: String,
@@ -111,10 +123,7 @@ fn run_loop(home: PathBuf, reg_rx: crossbeam_channel::Receiver<AgentSubscription
                     buf.input_id = pair.reply_to_input_id;
                     let text = String::from_utf8_lossy(&data);
                     buf.buffer.push_str(&text);
-                    if buf.buffer.len() > MAX_MIRROR_LEN * 2 {
-                        let drain = mirror_drain_start(&buf.buffer);
-                        buf.buffer.drain(..drain);
-                    }
+                    apply_mirror_cap(&mut buf.buffer);
                 } else {
                     buf.active = false;
                     buf.buffer.clear();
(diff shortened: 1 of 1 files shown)


suzuke:
VERIFIED — PR #2102 at c8dfed734b1c82c18010329f181aa9291400d040 satisfies #2085: the mirror cap-drain logic was extracted without changing the decision logic, and the new CJK regression drives the run-loop cap-drain path rather than the sibling dispatch slice path.

scope_source: issue #2085
scope_boundary: src/daemon/router.rs mirror cap-drain extraction and CJK regression test only; ignored the reported pre-existing dispatch_branch_1834 shim flake.
audit_mode: focused real-entry regression + extraction equivalence review
reviewed_head: c8dfed734b1c82c18010329f181aa9291400d040
files: src/daemon/router.rs
commands: cargo test daemon::router::tests:: -- --nocapture; cargo test daemon::router::tests::run_loop_cap_drain_never_panics_on_oversized_cjk -- --nocapture; temporary RED mutation of mirror_drain_start to raw byte index followed by the same single test; gh pr checks 2102 --repo suzuke/agend-terminal; git status --short --branch

### Evidence
ran: cargo test daemon::router::tests:: -- --nocapture -> passed; 10 router tests passed, including the new `run_loop_cap_drain_never_panics_on_oversized_cjk` and the existing mirror tests.
ran: cargo test daemon::router::tests::run_loop_cap_drain_never_panics_on_oversized_cjk -- --nocapture after temporary raw-byte-index mutation -> failed as expected with `assertion failed: self.is_char_boundary(end)` at src/daemon/router.rs:40.
ran: cargo test daemon::router::tests::run_loop_cap_drain_never_panics_on_oversized_cjk -- --nocapture after restoring the char-boundary snap -> passed; 1 test passed.
ran: gh pr checks 2102 --repo suzuke/agend-terminal -> passed; macOS, ubuntu, Windows, Coverage, Daemon-boot flake gate (20x), LOC overrun, and audit all passed.
ran: git status --short --branch -> clean review worktree at `review/2102-mirror-cap-drain-fixup-reviewer` after the RED restore.
cited: src/daemon/router.rs:41 — `apply_mirror_cap(&mut String)` contains the same cap condition and drain sequence: `if buf.len() > MAX_MIRROR_LEN * 2`, `mirror_drain_start`, then `buf.drain(..drain)`.
cited: src/daemon/router.rs:125 — the run-loop still does `buf.buffer.push_str(&text)` before calling `apply_mirror_cap(&mut buf.buffer)`, preserving the original cap-drain decision point.
cited: src/daemon/router.rs:193 — `try_dispatch_mirror` still owns the separate dispatch truncation/slice path, distinct from the new cap-drain test target.
cited: src/daemon/router.rs:309 — the new test repeatedly pushes oversized 3-byte CJK into a `String`, calls `apply_mirror_cap`, asserts `len <= MAX_MIRROR_LEN * 2`, and checks the final char boundary.

The extraction is behavior-equivalent modulo the helper parameter name (`buf.buffer` -> `buf`), and the RED run proves the new test catches the original raw-byte-index regression in the cap-drain path.

Issue #1899, opened 2026-06-09
FIX C2 — bound the remaining ~15 raw Command::new(git) prod sites (non-homogeneous, per-site)


suzuke:
Follow-up to #1897 (FIX C, #1898). The #1893 spike's '9 git_bypass sites' undercounted — the FIX C author's audit found ~15 ADDITIONAL independent raw Command::new("git") production sites that still call .output()/.spawn() with NO timeout, so a hung git there can still freeze the daemon. FIX C only covered the git_bypass + run_git_idempotent path.

## Sites (from the FIX C audit, fixup-dev has the saved list)
admin / bootstrap::canonical_hygiene / daemon::auto_release / daemon::conflict_notify / daemon::per_tick::tmp_review_gc / daemon::retention::worktrees / mcp::handlers::ci::mod / mcp::handlers::force_release::gc (+ enumerate precisely from the saved list).

## ⚠️ Non-homogeneous — per-site handling required (do NOT blindly route through git_bypass)
- They are NOT all alike: at least one is a  (NETWORK — needs the 300s network timeout, not 60s local), and SEVERAL do NOT set AGEND_GIT_BYPASS (routing them through git_bypass would CHANGE the bypass semantics — they intentionally don't bypass the agend-git shim).
- So: for each site, (a) classify LOCAL vs NETWORK (push/fetch/clone = network), (b) check whether it currently sets AGEND_GIT_BYPASS. Route bypass sites through git_bypass_timeout (the existing bounded path). For NON-bypass sites, add a bounded timeout WITHOUT forcing bypass — extract/reuse the bounded-spawn primitive (spawn_group_bounded) so they get the timeout + process-group kill safety but keep their original bypass/env behavior. The push site gets the NETWORK timeout.

## Reuse the FIX C primitives
Reuse spawn_group_bounded (process-group isolation + kill_process_tree — already reviewer-2-verified watertight against killing the daemon) so every bounded site has the same safe kill. Don't reinvent.

## Care
Same as FIX C: each timeout generous enough for legit slow ops (esp. the network push/fetch); timeout → clear Err the caller handles (not panic/silent-wrong); process-group kill stays scoped to the git subtree. Per-site verification that the legitimate flow is unchanged.

## §3.9
[...]

PR #2191, opened 2026-06-15
fix(CR-2026-06-14 worktree-git #5 M perf): bound the gh CLI subprocess with a timeout

suzuke:
## CR-2026-06-14 — `branch_sweep.rs:183` gh subprocess unbounded (Medium, performance)

`is_squash_merged_diff` (`branch_sweep.rs`) → `make_scm_provider().pr_list()` runs
the `gh` CLI via `GitHubScmProvider::run`, which used a bare **unbounded**
`cmd.output()`. That path is reached from
`worktree_cleanup::prune_orphaned_branches` inside the **per-tick** auto cleanup
sweep, so a slow / hanging `gh` (network stall, auth prompt, rate-limit retry)
blocked the daemon's cleanup thread with no upper bound — once per tick, per
non-merged branch, per repo. The local git helpers were already bounded
(`git_helpers::spawn_group_bounded`, #1897/#1899); only the `gh` runner was not.

### Fix
Route the prebuilt `gh` `Command` through the shared
`git_helpers::spawn_group_bounded` — the same process-group-isolated,
timeout + process-tree-kill spawner the local git helpers use — with a
purpose-named `GH_CLI_TIMEOUT` (60s) defined **at this call site**, exactly as the
`git_helpers` note advised when the old 300s `NETWORK_GIT_TIMEOUT` was removed for
lack of users.

- Fast path **byte-identical** (same captured `Output`).
- On the deadline the gh process group is killed; the caller gets `Err(TimedOut)`
  and fails fast (the `out.status.success()` callers already propagate `Err` via `?`).
- 60s is ~12× a healthy `gh pr list/view` — never false-kills a legit call, but
  bounds the hang.

### §3.10 RED→GREEN (double-confirmed)
- Anchor `dd3e24b` (un-ignore only) → static invariant **FAILS** (`run` body has an unbounded `.output()`, no bound token).
- HEAD `4273cd5` → **GREEN**. Stash-verified: stashing the `scm/mod.rs` fix reproduces RED on the un-ignored invariant, popping restores GREEN.

### Evidence
```
ran: cargo test --test review_worktree_git_5 → ok. 1 passed (GREEN)
ran: (fix stashed) same → FAILED (RED)
ran: cargo test --bin agend-terminal scm:: → ok. 15 passed; 0 failed
ran: cargo test --bin agend-terminal branch_sweep → ok. 12 passed; 0 failed
ran: cargo test git_helpers:: (no AGEND_GIT_BYPASS) → ok. 9 passed; 0 failed
ran: cargo fmt --check → clean
ran: cargo clippy --tests --features tray -- -D warnings → clean
```
(local runs `AGEND_GIT_BYPASS=1` per managed-worktree convention; PR CI is the authoritative cross-platform signal.)

### Note for reviewer (windows real-git CI)
This is in the `worktree_git`/`scm` domain where windows real-git CI is known to
flake (xwin/cross_repo/worktree_pool clusters). The change itself is platform-
neutral: `spawn_group_bounded` already has `#[cfg(windows)]`
`CREATE_NEW_PROCESS_GROUP` handling and is exercised by existing #1897 tests on
all platforms. If windows CI fails, please triage real-vs-flake before treating
it as a regression of this diff.

### Lessons learned
- The bound was a one-line gap, not a missing mechanism: `spawn_group_bounded`
  was already `pub(crate)` *specifically* so non-bypass prod sites could pass a
  prebuilt `Command` and inherit the bound + safe kill (#1899). The `gh` runner
  was simply the one network subprocess that hadn't been migrated when the local
  git sites were. The static-invariant repro (grep the `run` body for a bound
  token) is the right shape for "ensure this spawner is bounded".

---
*fixup-dev-3* · claude-opus-4-8[1m]

### DIFF
--- a/src/git_helpers.rs
+++ b/src/git_helpers.rs
@@ -75,11 +75,23 @@ pub(crate) fn git_bypass_timeout(
 /// same bound + safe process-group kill while preserving their env/bypass
 /// behaviour — instead of being forced through [`git_bypass_timeout`] (which
 /// always sets the bypass env).
+///
+/// CR-2026-06-14 #5 (deadlock-safety): stdout AND stderr are drained on
+/// dedicated reader threads CONCURRENTLY with the wait poll. The OS pipe buffer
+/// is small and fixed (macOS ~16KB, non-growing); a child that emits more than
+/// one buffer's worth (e.g. `gh api .../compare` returning a multi-MB diff JSON,
+/// reached via `merge_freshness`/`task_sweep`) blocks in `write()` until a reader
+/// drains the pipe. The earlier loop only `try_wait()`d — never reading until
+/// after exit — so such a child never exited, hit the deadline, and was
+/// false-timeout-killed. The reader threads keep both pipes drained so the child
+/// can always make progress to exit (this is what std's `wait_with_output` does
+/// via `read2`, reconstructed here so it composes with the timeout poll).
 pub(crate) fn spawn_group_bounded(
     mut cmd: std::process::Command,
     label: &str,
     timeout: Duration,
 ) -> std::io::Result<std::process::Output> {
+    use std::io::Read;
     cmd.stdout(std::process::Stdio::piped())
         .stderr(std::process::Stdio::piped());
     // Group isolation only; NOT detached — we keep the captured stdout/stderr.
@@ -96,6 +108,26 @@ pub(crate) fn spawn_group_bounded(
     }
     let mut child = cmd.spawn()?;
 
+    // Drain both pipes concurrently (see the deadlock-safety note above). The
+    // readers run until EOF — which arrives when the child exits normally OR when
+    // the process-group kill below closes the pipes — so they always join.
+    let stdout_pipe = child.stdout.take();
+    let stderr_pipe = child.stderr.take();
+    let stdout_reader = std::thread::spawn(move || {
+        let mut buf = Vec::new();
+        if let Some(mut p) = stdout_pipe {
+            let _ = p.read_to_end(&mut buf);
+        }
+        buf
+    });
+    let stderr_reader = std::thread::spawn(move || {
+        let mut buf = Vec::new();
+        if let Some(mut p) = stderr_pipe {
+            let _ = p.read_to_end(&mut buf);
+        }
+        buf
+    });
+
     // #1897: poll with EXPONENTIAL BACKOFF (1ms → 50ms cap). The previous fixed
     // 200ms poll added up to 200ms of latency per git call vs the old immediate
     // `.output()`, which PERTURBED timing-sensitive concurrent callers (the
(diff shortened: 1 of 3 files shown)


suzuke:
## Refix pushed — addresses r2 REJECT (deadlock on large output) → `5c9757c`

r2 caught that routing the **shared** `GitHubScmProvider::run` through `spawn_group_bounded` exposed a latent deadlock: the poll loop only `try_wait()`'d and drained the pipes **after** exit, so a large-output gh caller (`merge_freshness.compare` multi-MB diff JSON; `task_sweep pr_view(files)`) blocked in `write()` once the macOS 16KB pipe buffer filled, never exited, and was false-timeout-killed at the deadline — deadlocking the merge-freshness gate on darwin.

**Root fix (r2's recommended option a):** `spawn_group_bounded` now drains stdout **and** stderr on dedicated reader threads **concurrently** with the wait poll — the child can always make progress to exit regardless of output size (mirrors std's `wait_with_output` read2, composed with the timeout poll). On timeout the process-group kill closes the pipes → readers EOF → join. Fast/small path unchanged; **all** callers (`git_bypass_timeout`, `ci/mod` worktree-remove, the gh runner) inherit the deadlock-safety.

**New regression repro** `git_helpers::spawn_group_bounded_drains_large_output_no_deadlock_cr20260614`: a child emitting **256KB** (16× a macOS pipe buffer) must return `Ok` in full, exit-0, well under the bound.

**RED→GREEN (double-confirmed):** reverting the drain to try_wait-only → the 256KB test deadlocks → false `TimedOut` at 10s (RED/FAILED); the concurrent drain → passes in <5s (GREEN).

**Evidence:**
```
git_helpers:: → 10 passed (incl. new deadlock-safe repro + the #1897/#1899 cluster)
scm:: → 15 passed ; branch_sweep → 12 passed ; review_worktree_git_5 invariant → 1 passed
fmt --check clean ; clippy --tests --features tray -D warnings clean
```
The non-blocking items r2 already confirmed OK (timeout caller handling, kill group, GH_CLI_TIMEOUT call-site, repro non-vacuous) are unchanged. Ready for r2+r4 re-review (please verify the large-output deadlock-safety).

---
*fixup-dev-3* · claude-opus-4-8[1m]

suzuke:
## Scoped-thread fix pushed → `99632fd` (addresses r4 dual-1 REJECT of 5c9757c)

r4 correctly caught that 5c9757c's two raw `std::thread::spawn` reader threads fail the §10.5 `spawn_rationale_audit` (a textual scan that requires `// fire-and-forget:` and does not recognize JoinHandle-store/join) → real CI failure on Coverage + ubuntu + macos.

**Fix:** drainers switched to `std::thread::scope` + `s.spawn(...)`. A scope joins every thread it spawned before it returns (incl. the early-`return` poll paths), so the readers can never leak — a **stronger, compiler-enforced** guarantee than the manual join, and genuinely §10.5-compliant (no detached thread; `s.spawn(` is outside the audit's `thread::spawn(` class by design).

**Addressing r4's two re-review focus points (lead-aligned):**
1. **256KB red→green stays green under scope** — `git_helpers::spawn_group_bounded_drains_large_output_no_deadlock_cr20260614` passes (256KB = 16× macOS pipe buf, all bytes drained, exit-0, <5s), alongside the full #1897/#1899 cluster (10/10).
2. **timeout-kill path can't hang the scope** — every non-normal exit (`Ok(None)` timeout AND `Err(e)` from `try_wait`) calls `kill_process_tree(child.id())` + `child.wait()` **before** `return`. The kill closes the child's pipe write-ends → the scoped readers hit EOF → the scope's implicit end-of-block join completes. The kill is structurally guaranteed to precede the join (it's inside the closure, before its return).

**Evidence:**
```
cargo test --test spawn_rationale_audit              → 2 passed (the §10.5 wall is cleared)
cargo test --bin agend-terminal git_helpers::        → 10 passed (256KB deadlock-safe + #1897/#1899)
cargo test --bin agend-terminal scm:: / branch_sweep → 15 / 12 passed
cargo test --test review_worktree_git_5              → 1 passed (the gh-timeout invariant)
fmt --check clean ; clippy --tests --features tray -D warnings clean
```
Ready for r4 + r6 re-review on `99632fd`.

---
*fixup-dev-3* · claude-opus-4-8[1m]

PR #2206, opened 2026-06-15
fix(CR-2026-06-14 token_cost L #244): correct stale 'MAX per file' Codex comment

suzuke:
## CR-2026-06-14 — token_cost #244 (maintainability, doc-only)

The `token_cost.rs` module header claimed Codex's session-cumulative `total_token_usage` means "the MAX per file is taken, never summed". The actual implementation (`parse_codex_rows`) **sums per-turn `last_token_usage` deltas** (Σdelta reconciles to the cumulative total, asserted by `codex_delta_sum_equals_cumulative`). The stale comment would mislead a maintainer reasoning about reconciliation. Rewrote it to describe the delta-sum design.

**Doc-only** (comment text). §3.10 RED→GREEN: `review_verify_claim_cost_4` (un-ignored) RED before, GREEN after; `token_cost::` 30 pass; fmt clean.

**Review tier:** trivial maintainability → §3.6 single review.

Scope note: the sibling #191 (`parse_since` overflow) is already covered on main (dev batch C / t-40) — its repro is un-ignored + green, so excluded.

---
*fixup-dev-3* · claude-opus-4-8[1m]

### DIFF
--- a/src/token_cost.rs
+++ b/src/token_cost.rs
@@ -547,10 +547,12 @@ fn read_file_tail(path: &Path, max_bytes: u64) -> Option<String> {
 //
 // Source: `~/.codex/sessions/YYYY/MM/DD/rollout-*.jsonl`. Each session file
 // carries a `session_meta` line with `payload.cwd` (attribution), `turn_context`
-// lines with `payload.model`, and `event_msg`/`token_count` lines with
-// `payload.info.total_token_usage` — which is SESSION-CUMULATIVE, so we take
-// the MAX (final total) per file, never a sum. Verified against real files
-// 2026-05-29: `total_tokens == input_tokens + output_tokens`,
+// lines with `payload.model`, and `event_msg`/`token_count` lines whose
+// `payload.info` carries both `total_token_usage` (session-cumulative) and
+// `last_token_usage` (the per-turn DELTA). `parse_codex_rows` emits one Row per
+// `token_count` line from the per-turn delta and SUMS them — Σdelta reconciles to
+// the cumulative total (see `codex_delta_sum_equals_cumulative`). Verified against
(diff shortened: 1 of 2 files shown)


suzuke:
## Refix pushed — addresses r6 REJECT → `454b82c`

r6 correctly caught that the first fix corrected only the module header (16-18) while the **same** stale MAX-strategy claim survived at the Codex collector section header (`token_cost.rs:552`): *"the MAX (final total) per file, never a sum"* — directly contradicting `parse_codex_rows`' delta-sum (`codex_delta_sum_equals_cumulative`). The repro **false-GREENed** because its fixed `stale_phrases` (`"MAX per file"` / `"is taken, never summed"`) didn't match the section-header variant (`"MAX (final total) per file"` / `"never a sum"`), even though the repro's own docstring says it guards **both** headers.

**Two-part refix:**
1. Rewrote the Codex section-header comment to describe the delta-sum design (Σ `last_token_usage` delta == cumulative), consistent with the module header and `parse_codex_rows`.
2. **Tightened the repro** to match variants — co-occurrence `MAX` + `per file`, and `never` + (`a sum` | `summed`) — so a stale variant can't false-GREEN again.

**§3.10 RED→GREEN (double-confirmed):** with the tightened repro, stashing the `token_cost.rs` fix now reproduces RED **and catches line 552** (`token_cost.rs:552: // the MAX (final total) per file, never a sum` — the exact variant the old repro missed); with the fix it's GREEN. fmt clean. Doc + test-logic only.

(NB: the commit message has minor cosmetic gaps where backticked identifiers were shell-substituted at commit time — the diff is correct; this comment is the authoritative description.)

Ready for r6 re-review on `454b82c`.

---
*fixup-dev-3* · claude-opus-4-8[1m]

PR #2247, opened 2026-06-15
feat(#2090): origin-aware long-task progress reporting

yujunchao:
## #2090 — Origin-aware long-task progress reporting

This PR adds an opt-in mechanism so a managed agent's long-task progress is surfaced back to whatever channel a request originated from (e.g. Telegram), without the operator having to watch the CLI/PTY. It is fully runtime-toggleable and defaults to a low-noise mirror mode.

### What's included (4 feature commits + 1 build tweak)

- **M1 — origin-aware progress reporting** (`e9db1355`)
  Injects a "Long-Task Progress Reporting" directive and adds a runtime-mutable `progress_mode` config key: `0` = off, `1` = mirror (default), `2` = self-report.

- **M2 — smart progress mirror via transcript tail** (`28a235f0`)
  In mirror mode, the daemon tails the agent's transcript (`.jsonl`) and relays only *new assistant plain-text* blocks back to the origin channel — clean, ANSI-free, no double-send. Retires the legacy PTY-buffer mirror (gated at the router call site) to avoid noise/regressions.

- **M1b — backend-aware subagent-delegation directive** (`a3c8dc12`)
  Adds a `Backend::supports_subagents()` capability flag; the delegation directive adapts per backend (delegate to a subagent where supported; otherwise instruct inline self-reporting). v1 treats Claude as subagent-capable.

- **M3 — progress backstop watchdog (report mode)** (`710003a5`)
  In mode 2, a per-tick backstop injects a `progress-backstop` nudge when an external-channel turn has been armed > 45s with no reply, prompting the agent to self-report (daemon does not author the message itself).

- **build — thin LTO + codegen-units=16** (`5170f4fb`)
  Switches the release profile from fat LTO to ThinLTO with `codegen-units=16`, cutting a full release build from ~40 min to ~17 min on multi-core. *(Build-time preference — drop this commit if upstream prefers fat LTO.)*

### Notes
- All commits are rebased onto current `upstream/main`.
- `progress_mode` is hot-reloadable via the existing runtime-config tool; mirror mode only fires when a request has an origin channel bound (boot-time/manual replies don't trigger it).
- Mirror currently understands Claude `.jsonl` transcripts; non-Claude backends are a no-op (no fleet impact today — pluggable transcript source is a follow-up when another backend is added).


suzuke:
Thanks for this contribution, @yujunchao — there's solid design thinking here (the active-turn gating, tail-position dedup, the mode-2 backstop debounce, and the backend-capability flag are all the right instincts). A few notes on how we're going to take it forward:

**Issue reference mix-up.** This PR is linked to #2090, but #2090 is actually a different (unrelated) issue on our side — "narrow-pane hard-wrap breaks single-line-regex state detection." The progress-reporting / transcript-mirror work here doesn't touch that. No harm done, just flagging so the linkage isn't misleading.

**Direction is valuable — we're going to build it out internally.** Rather than merge this as-is, we'll re-implement origin-aware progress reporting as a first-class internal feature so it fits our review/policy bar. Two adjustments drive that:

- **Mirror mode will default OFF (opt-in).** As written, mode 1 (mirror) is the default and relays the *full assistant text stream* to the origin channel. For a fleet where an agent may echo file contents, secrets, or injected content, defaulting to "stream everything to Telegram" is an exfiltration/noise-policy decision we want to make deliberately — so it'll ship default-off, clearly documented as a full stream (not a curated summary), and we're evaluating whether a curated-progress mode is the better primitive.
- A couple of smaller things we'll harden in the rewrite: the in-memory `TailPos` map loses position across daemon restarts (re-send/skip window) and is unbounded.

**The build-profile change, we're adopting — separately.** The ThinLTO + `codegen-units=16` switch (your `5170f4fb`) is a genuinely nice win (~40min→~17min release builds) and we're pulling it in as its own standalone PR. Credit to you for it — it just shouldn't ride along with a feature change.

We'll leave this PR open for now so the discussion stays visible. Really appreciate you digging into this part of the system. 🙏

suzuke:
Thank you, @yujunchao, for this proposal — it directly inspired us to build origin-aware long-task progress reporting into AgEnD. We've now shipped a full internal implementation across three PRs:

- **#2283** — `progress_mode=2` (report): the agent self-reports progress to the origin channel; the daemon only nudges, never relays raw output (zero exfil).
- **#2284** — a dedicated actionable `[AGEND-PROGRESS]` marker so the backstop nudge isn't swallowed by the `[AGEND-AUTO]` never-act rule.
- **#2293** — `progress_mode=1` (mirror): daemon tails the transcript and relays assistant text to the origin channel, gated to the active external turn only.

Both modes are **default OFF** and the mirror mode is documented as an exfiltration surface (it relays full assistant output). The design incorporated your idea plus review feedback on the exfil boundary and injection point.

Closing this PR since the capability is now in main — but genuinely grateful for the nudge. 🙏

yujunchao:
Thanks so much for the thoughtful writeup, and for taking this all the way into a first-class feature — really glad the idea landed.

The security framing is exactly right and sharper than my original cut: defaulting both modes OFF, documenting mirror explicitly as an exfiltration surface, gating the relay to the active external turn only, and the dedicated `[AGEND-PROGRESS]` marker so the backstop nudge isn't swallowed by the `[AGEND-AUTO]` never-act rule. The `progress_mode=2` (self-report, daemon-only-nudges, zero raw relay) split is a cleaner primitive than what I proposed.

Appreciate you pulling the ThinLTO build change out as its own PR too. Happy to have helped nudge it — looking forward to running #2283/#2284/#2293 on main. 🙏

PR #2381, opened 2026-06-20
feat(palette): Ctrl+B `:` command completion — prefix-match keyword hints

suzuke:
## Problem
The `:` command palette (`Overlay::Command`) was a blank text box — operators couldn't discover which commands exist. (operator-requested, lead-vetted spike → impl, task t-…98760-5.)

## UX (P0, keyword prefix-match)
- **Ctrl+B `:` lists ALL commands** (the "what commands exist?" entry point).
- Typing **prefix-filters** the first token; **↑/↓** move the highlight.
- **Tab** completes the highlighted keyword into the input (keeps the palette open + a trailing space for args).
- **Enter still executes the CURRENT input — semantics unchanged.** Completion is purely additive (no muscle-memory/script break).

## Implementation — command EXECUTION untouched
- `commands.rs`: additive `COMMAND_SPECS` table (`{keyword, usage, desc}`, 11 commands) + `matching_specs(prefix)`. **`execute` is byte-for-byte unchanged** (only a reminder comment to keep the table in sync). `mod commands` widened to `pub(crate)` so render reads the specs; `execute` stays `pub(super)` = app-only, so command execution is not widened.
- `Overlay::Command` gains `selected`; the candidate list is derived live from `input` (no `&'static` stored in the enum). Up/Down/Tab added; Backspace/Char reset the highlight. Confirmed Up/Down/Tab were previously unbound in this overlay.
- `render_command_palette` renders the candidate list (`usage` / `desc`, selected highlighted), modeled on the existing `render_menu`.

## Tests (cross-platform, no PTY)
- `matching_specs_prefix_filters_first_token` — empty→all, prefix narrows, no-match→empty.
- `command_specs_cover_execute_arms` — source-scan guard: every spec keyword has an `execute` arm (a stale/typo'd spec can't offer a completion the dispatcher won't handle). Source-scan because `execute` can't be called live (spawn/restart fork real processes).
- `command_palette_lists_all_on_open_and_prefix_filters` — TestBackend render smoke.

## Verification
`cargo fmt --check` + `clippy --all-targets --features tray -- -D warnings` clean; 3 new tests + 58 touched-module regression tests green.

Keyword-only is P0; **argument-value completion** (backend / agent / config-key / preset names — dynamic sources noted in the spike) is a deliberate follow-up. Merge needs operator restart to go live.

correlation_id=t-20260620083915795311-98760-5

### DIFF
--- a/src/app/commands.rs
+++ b/src/app/commands.rs
@@ -23,7 +23,92 @@ pub(super) struct CommandCtx<'a> {
     pub telegram_state: &'a Option<Arc<dyn crate::channel::Channel>>,
 }
 
+/// #t-5 command-palette completion: declarative metadata for the `:` commands —
+/// keyword + usage + one-line description. Consumed ONLY by the completion list
+/// (`matching_specs` + `render::overlay::render_command_palette`); `execute`
+/// below stays the source of truth for BEHAVIOR and is unchanged. ⚠ Adding or
+/// renaming a command in `execute` REQUIRES a matching entry here — the
+/// `command_specs_cover_execute_arms` test fails otherwise.
+pub(crate) struct CommandSpec {
+    pub keyword: &'static str,
+    pub usage: &'static str,
+    pub desc: &'static str,
+}
+
+pub(crate) const COMMAND_SPECS: &[CommandSpec] = &[
+    CommandSpec {
+        keyword: "spawn",
+        usage: "spawn [name] [backend]",
+        desc: "New agent in a new tab",
+    },
+    CommandSpec {
+        keyword: "vsplit",
+        usage: "vsplit [name] [backend]",
+        desc: "New agent; split focused pane vertically",
+    },
+    CommandSpec {
+        keyword: "hsplit",
+        usage: "hsplit [name] [backend]",
+        desc: "New agent; split focused pane horizontally",
+    },
+    CommandSpec {
+        keyword: "kill",
+        usage: "kill <agent>",
+        desc: "Close an agent's pane",
+    },
+    CommandSpec {
+        keyword: "restart",
+        usage: "restart [agent]",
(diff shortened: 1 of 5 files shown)


suzuke:
## ✅ VERIFIED + CI green — merging (§3.12 verdict mirror)

**HEAD merged:** `6e44ae7a`

**Review (single, TUI UX):**
- **Round 1 — fixup-reviewer-6 (codex) REJECTED** two real, mutation-proven blockers: (B1) the COMMAND_SPECS↔execute sync test was a one-way substring scan (reverse drift unguarded + `unwrap_or(&"agent")` false-positive); (B2) `Event::Paste` left `selected` stale → Tab silently no-op'd after a paste-filter.
- **fixup-dev fixed both**: B1 → precise bidirectional set-equality test (regex matches `"<kw>"` followed by `|`/`=>`, bounded to the execute body); B2 → paste resets `selected=0` + Tab use-site clamps, with a regression test through real `handle_key`.
- **Round 2 — fixup-reviewer-4 (claude) VERIFIED** (took over after r6 hit usage_limit): independently ran all 3 of r6's mutations → RED as expected (forward/reverse/false-positive), confirmed paste/Tab consistency, execute byte-identical, visibility not widened (execute stays `pub(super)`).

**CI (independent §3.3.1 recheck):** all green — Check macos/ubuntu/**windows** + Coverage + **Daemon-boot flake gate (20x)** + MSRV + LOC + audit + DCO.

**Known LOW (non-blocking):** the sync test's lower bound assumes the next item after `execute` is a bare `fn`; inserting a `pub(super) fn` between them would over-extend the bound — but the test REDs (self-warning, not silent). P0 is keyword-only; param-value completion is a follow-up.

Squash-merging via admin (shared-account standard path).

Issue #2158, opened 2026-06-14
Hardening: a sub-agent can silently rebind the parent instance + write across workspace boundaries


suzuke:
## Summary / 摘要

A **transient sub-agent that inherits a running instance's identity can silently mutate that instance's daemon state and write outside its own worktree** — with no workspace boundary. In a recent docs task, an orchestration tool handed sub-agents a bad target path; instead of failing, each sub-agent "resolved" a location on its own, and the fleet primitives let that go much further than it should have.

一個**繼承了現有 instance 身分的短命 sub-agent,可以默默改動該 instance 的 daemon 狀態,並寫到自己 worktree 以外的地方**,且沒有 workspace 邊界保護。

> Note: the immediate trigger was an external orchestration bug (a path arriving as the literal string `"undefined"`), **not** an agend-terminal bug. But the *blast radius* is an agend-terminal hardening gap worth tracking. Filing per maintainer request.

## What was observed

Several sub-agents shared the parent instance identity `claude-fb80a0`. With a bad/`"undefined"` target path, they each guessed where to act. The fleet primitives then allowed:

1. **A sub-agent rebound the *parent* instance via `repo action=checkout bind:true`.**
   The parent never requested a binding. Afterwards `binding_state(claude-fb80a0)` showed:
   ```
   bound: true
   branch: docs/zh-tw-ci-down-sop          # created by the sub-agent
   issued_at: 2026-06-14T15:28:12Z
   ci_watches: ["suzuke/agend-terminal:docs/zh-tw-ci-down-sop"]   # armed without consent
   marker_present: true
   ```
   i.e. a transient helper silently moved "my" binding to a branch I never chose and armed a CI watch on it.

2. **Cross-workspace / main-tree file writes.** New untracked `*.zh-TW.md` files were written into:
   - the **main repo working tree** (`…/agend-terminal/docs/`),
   - **peer agents' workspaces** — `fixup-dev-2` (`…/workspace/fixup-dev-2/docs/`) and the `claude-8145a9` worktree,
   - two **stray worktrees** the sub-agents created via `AGEND_GIT_BYPASS=1` (`.claude/worktrees/zh-tw-usage-doc`, `.git-worktrees/zh-tw-skills`).

   Nothing flagged or prevented an agent writing files into another agent's bound workspace.

[...]

PR #2384, opened 2026-06-20
test(teardown): allowlist fleet_events.jsonl — append-only shim audit, not residual

suzuke:
## Problem
`teardown_completeness_regression::teardown_leaves_zero_residual_after_full_exercise_1907` is **red in-worktree but green in CI** — two devs forced `--no-verify` (masks real failures). (RCA spike, task t-…34628-8.)

## Root cause (3 experiments, isolated /tmp homes)
The operator's `git` on PATH is the **agend-git shim** (`~/.agend-terminal/bin/git`), not real git. The test's own booted daemon runs git via `git_helpers::git_cmd` (which sets `AGEND_GIT_BYPASS=1`), so the shim appends a #2158 bypass-audit line — naming the agent — to `home/fleet_events.jsonl`. CI's real `/usr/bin/git` ignores the bypass and never creates the file, so the missing allowlist entry never bit there.

| run | env | result |
|---|---|---|
| EXP1 | shim on PATH + bypass | RED |
| EXP2 | `PATH=/usr/bin:$PATH` real git, no bypass | GREEN |
| EXP3 | shim on PATH, no env bypass | RED (daemon sets its own bypass) |

→ sole variable = the `git` binary on PATH. Not `AGEND_HOME` (the test isolates `temp_dir()/…-{pid}-{nanos}`), not the operator's env bypass.

## Fix (test-only, zero prod)
`fleet_events.jsonl` is a home-level **append-only forensic audit log** that names instances by design — the SAME class as the already-allowlisted `event-log.jsonl` / `task_events.jsonl`. `full_delete_instance` correctly does NOT scrub it (that would corrupt the audit chain). So this is an **ALLOWLIST omission, not a prod teardown gap** — add the entry with a rationale comment.

## Verification (dual env)
- Shim env (was RED) → **GREEN**; real-git env → GREEN.
- Other ALLOWLIST entries + assertion logic unchanged (diff = +11 lines: one entry + comment).
- `clippy --all-targets --features tray -- -D warnings` clean; 36 source-scan/invariant tests pass.

**This removes the footgun that pushed devs to `--no-verify`.**

correlation_id=t-20260620200941970440-34628-8

### DIFF
--- a/tests/teardown_completeness_regression.rs
+++ b/tests/teardown_completeness_regression.rs
@@ -40,6 +40,17 @@ const ALLOWLIST: &[(&str, &str)] = &[
     ("daemon.", "daemon.<date>.log — tracing legitimately names the deleted instance"),
     ("event-log.jsonl", "append-only audit trail — create/delete events name the instance by design"),
     ("task_events.jsonl", "append-only task-event log — owner/assignee history named by design"),
+    // #t-…34628-8 RCA: same append-only-audit class as event-log.jsonl. Written by
+    // the agend-git shim (src/bin/agend-git.rs — #2158 bypass-audit / #2234 cwd-drift)
+    // whenever a daemon git op runs through the shim (git_helpers.rs::git_cmd sets
+    // AGEND_GIT_BYPASS=1). Home-level forensic trail that names instances by design →
+    // teardown must NOT scrub it (would corrupt the audit chain). Only created when
+    // PATH `git` is the operator's shim (local dev machine); CI's real /usr/bin/git
+    // never writes it — which is why this entry was missing yet CI stayed green while
+    // two devs hit a local red. (The test inherits the operator git binary, i.e. is
+    // not fully hermetic w.r.t. `git`; forcing real git in the harness is an optional
+    // future hardening, intentionally NOT done here — A2/KISS.)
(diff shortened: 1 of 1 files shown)


suzuke:
## ✅ VERIFIED + CI green — merging (§3.12)

**fixup-reviewer-4 VERIFIED** @ 28fc2716 (reviewer's own PATH git *is* the operator shim, so it directly reproduced): NEW entry → teardown test PASS; neuter (remove entry) → FAILED with the exact `fleet_events.jsonl (content)` residual; restore byte-identical. Zero prod (only the ALLOWLIST const +1 entry + rationale comment); correct classification (home-level append-only audit, same class as event-log.jsonl; `full_delete_instance` never touches it); comment doesn't trip source-scan tests; clippy + DCO clean.

**RCA (vetted):** in-worktree-red / CI-green because the operator's agend-git shim writes `fleet_events.jsonl` (forensic audit) under the daemon's AGEND_GIT_BYPASS; CI's real git never produces it, so the ALLOWLIST omission was invisible. Not a prod teardown gap.

**CI all green** (§3.3.1 recheck): Check macos/ubuntu/windows + Coverage + Daemon-boot 20x + MSRV + LOC + audit + DCO.

Squash-merging via admin.

PR #2380, opened 2026-06-20
fix(freeze): render snapshot reads lock-free published AtomicU8 (no core.lock)

suzuke:
## Root cause (instrument-confirmed)
Residual post-#2346 TUI startup freeze ("panes appear, type a few chars, then stuck"). The instrument spike pinned it to **producer/consumer core.lock contention**: the per-frame render snapshot `build_agent_state_snapshot` (core_render.rs) took each agent's `core.lock()` to read state, contending with the PTY-feed producer (`pty_read_loop`, agent/mod.rs:1649) that holds the same lock 2–6 ms processing the boot output flood.

Measured (app log, freeze window): snapshot acquire-wait up to **10 ms** on the render thread → draw **11 ms** (vs ~400 µs idle), fps **25–27** (below the 30 cap, so #2346's frame cap could not help), input starved; recovered once the flood drained.

## Fix — consumer-side decoupling (lead-vetted candidate 1)
- `AgentState`: `#[repr(u8)]` + `from_u8()` → state fits a lock-free `AtomicU8`.
- `StateTracker`: a `published: Arc<AtomicU8>` mirror written **only** in `record_set` (the single funnel for every `current` mutation), seeded in `new()`; `published_handle()` clones it.
- `AgentHandle`: `published_state` clones the same `Arc` at construction (via `published_state_of`). **`build_agent_state_snapshot` now reads the atomic lock-free — zero `core.lock()`** — so render never contends with the producer. Relaxed ordering (standalone value; render tolerates the ≤1-frame staleness it already snapshots per draw).
- Producers untouched (no feed lock-shrink → perf-R1 UNSAFE inapplicable).

## Tests (deterministic)
- `snapshot_reads_published_state_without_core_lock`: holds an agent's `core.lock` on a bg thread; asserts the snapshot returns <50 ms (vs ~200 ms if it took the lock) **and** with the correct published state. Reverting the snapshot to `core.lock().state.get_state()` makes it block ~200 ms → fail.
- `agentstate_u8_roundtrip`: pins `repr(u8)` ↔ `from_u8` for all 18 variants.
- `published_mirror_tracks_current_through_record_set`.

## Verification
`cargo fmt --check` + `clippy --features tray` clean; 630 targeted tests (state/render/agent/daemon handle-builders) + the 3 new tests green; CoreMutex sole-wrapper #1535 invariant green.

[...]

PR #2385, opened 2026-06-21
fix(freeze): cap drain_output per frame — bound main-thread CPU in terminal.draw

suzuke:
## Root cause (instrument-confirmed — #2380 didn't fix this)
The residual restart freeze (operator: ~10s stuck on first input after restart). `pane.drain_output()` is an **unbounded** `while rx.try_recv { vterm.process }` that runs on the **main thread inside `terminal.draw`**. A boot/restart backlog (every agent re-attaching + dumping its screen) made one draw take **~108 ms** (`#freeze-loop` draw_us=107809; normal ~2 ms) → the loop couldn't service input → freeze.

It is **CPU, not a lock**:
- No `thread="main"` `#freeze-corelock` wait at the 107 ms draw (the main thread isn't blocked on core.lock).
- `render_pane` reads the pane-**local** vterm; #2380's lock-free agent-state snapshot fixed a *different, smaller* consumer.
- supervisor:1235 / agent:1649 are **producer** core.lock holds that don't gate the lock-free draw.

draw time ∝ backlog → restart (all agents at once) is the worst case → multi-second.

## Fix (lead-vetted candidate 1)
- `drain_output(budget_bytes) -> bool`: process whole chunks until the byte budget, leave the remainder queued (`rx` is unbounded → **lossless**, FIFO), return "more remains".
- `render_pane` passes `DRAIN_OUTPUT_BUDGET_BYTES` (32 KiB, tunable).
- The render loop re-arms `dirty` via `active_tab_has_pending_output` (mirrors `render_pane_tree`'s zoom/normal pane selection) so the backlog drains over a few frames; the existing #2346 frame-cap select-timeout already shrinks to the frame boundary when dirty. **Input stays responsive; the pane catches up over ~frames.**
- Scratch-shell overlay left unbounded (single user-opened pane, not the boot-flood path — behavior unchanged).
- env-gated `#freeze-drain` probe (`AGEND_FREEZE_INSTRUMENT`) so the operator restart-repro confirms the cost is in `drain_output` and is bounded after the fix.

## Tests (deterministic, cross-platform)
- `drain_output_budget_caps_and_drains_losslessly_in_fifo_order` — budget cap + reports-remaining + loop-to-dry + lossless + FIFO.
- `active_tab_has_pending_output_reflects_visible_pane_queue` — the `dirty` re-arm trigger (correctness rule: never stall the backlog).

## Verification
`cargo fmt --check` + `clippy --all-targets --features tray -- -D warnings` clean; 83 touched-module tests + the 2 new tests green. Diff ~50 LOC prod + tests.

supervisor:1235 lock-shrink = separate low-priority task (orthogonal). Merge → operator rebuild+restart to verify the freeze is gone + collect `#freeze-drain` data.

correlation_id=t-20260621005233596507-74503-0

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -796,6 +796,14 @@ fn run_app(terminal: &mut DefaultTerminal, fleet_override: Option<&Path>) -> Res
                 // resize its pane's VTerm/PTY during render.
                 render_active_overlay(frame, &mut overlay, &layout, &registry, &home);
             })?;
+            // #freeze-2: a budget-capped `drain_output` may have left a backlog in
+            // a visible pane's channel. Re-arm `dirty` so the next frame continues
+            // draining (the select-timeout below already shrinks to the frame
+            // boundary when dirty) — the backlog clears over a few frames instead
+            // of one input-stalling mega-draw.
+            if render::active_tab_has_pending_output(&layout) {
+                dirty = true;
+            }
         }
 
         // #t-84833-10: wake the loop at the next frame boundary when a change is
--- a/src/layout/pane.rs
+++ b/src/layout/pane.rs
@@ -75,6 +75,20 @@ pub struct Selection {
     pub end: (i64, u16),
 }
 
+/// #freeze-2 probe (env-gated, `AGEND_FREEZE_INSTRUMENT`): per-frame `drain_output`
+/// cost threshold in microseconds, or `None` when disabled (zero overhead). Read
+/// once and cached. Lets an operator restart-repro confirm the freeze is in
+/// `drain_output` and that the budget bounds it. Off by default → no behavior
+/// change. `"1"` => the 2 ms default; `"<N>"` (N>1) => custom µs; else disabled.
+fn drain_probe_threshold_us() -> Option<u64> {
+    static PROBE: std::sync::OnceLock<Option<u64>> = std::sync::OnceLock::new();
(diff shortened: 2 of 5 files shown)


suzuke:
## ✅ DUAL VERIFIED + CI green — merging (§3.12)

**DUAL (codex r6 quota-exceeded → 2-claude):**
- **fixup-reviewer-4 VERIFIED** + **fixup-dev-3 VERIFIED** @ 63c596b6, both with **neuter proofs**: (A) forcing `active_tab_has_pending_output→false` fails the detector test → re-arm is load-bearing, backlog cannot get stuck-half-drained; (B) ignoring the budget fails the budget test ("must leave a remainder") → the per-frame cap is load-bearing. Both independently confirmed the detector mirrors `render_pane_tree`'s zoom/normal pane selection verbatim (collect_pane_ids == render_node walk), lossless+FIFO, ≤8 KiB chunks ≪ 32 KiB budget (no mega-chunk freeze), no render regression (25/25), probe env-gated zero-overhead, DCO present.

**RCA:** the freeze was CPU (drain_output draining the entire unbounded PTY backlog synchronously inside terminal.draw, draw-time ∝ backlog → ~10s at restart), **not** core.lock contention. Supersedes the incomplete #2380.

**CI all green** (§3.3.1): Check macos/ubuntu/windows + Coverage + Daemon-boot 20x + MSRV + LOC + audit + DCO.

Squash-merging via admin. Needs operator rebuild+restart to verify the freeze is gone (+ #freeze-drain probe data).

PR #2388, opened 2026-06-21
perf(mcp): read-only tools skip per-call usage append + heartbeat RMW

suzuke:
## What (arch F5 — de2eb8 review, r4-confirmed-real; operator chose to do-now)

`handle_tool` paid **two disk IOs on every tool call** — the #2055 usage-stats append + the flocked `last_heartbeat` metadata RMW — even for pure-query tools (`list_instances`/`binding_state`/…) an agent polls in a loop. This skips both for a static allowlist of read-only tools.

## Design (spike-vetted by lead)

- **`is_read_only_tool` allowlist (6 pure queries)**: `list_instances`, `binding_state`, `gc_dry_run`, `tokens`, `pane_snapshot`, `tui_screenshot`. For these, skip the usage append **and** the disk `last_heartbeat` RMW.
- **KEEP the in-mem `heartbeat_pair` update for every call** (incl. read-only). It is the **authoritative liveness source** — `supervisor`/`router`/`dispatch_idle` all read the in-mem pair — and is cheap (no disk), so a read-only poll still proves liveness.
- **Liveness safety (the one real, non-cosmetic risk, folded in per vet):** the supervisor reads disk `last_heartbeat` **only** as the crash-recovery fallback while the in-mem pair is *cold* (`heartbeat_at_ms == 0`, before the first call after a (re)start — `supervisor.rs ~1248`). So the disk write is skipped for read-only **except on that cold→warm first call**, which refreshes the fallback. This bounds any stale window to "before first call" (same as the in-mem pair) — a read-only-only agent **cannot** be falsely stale-escalated.

### Deliberately NOT skipped
- `inbox` (drain transitions unread→delivering — mutates), `download_attachment` (writes a file).
- Action-based tools (`task`/`decision`/`team`/`schedule`/`deployment`/`ci`/`health`/`repo`/`mode`) whose read-vs-write splits by `action` — kept on the full path (no per-action coupling at this chokepoint; the minority of read actions pay the IO).
- **Not bundled** (per scope): list_instances' displayed `last_heartbeat` may lag for a read-only-only agent — cosmetic only (supervisor uses in-mem); left for a separate change if the operator cares.

## Tests (test-first)

- `read_only_tool_skips_disk_io_but_bumps_inmem_heartbeat` — warm read-only: no usage append, no disk RMW, **but in-mem heartbeat IS bumped** (load-bearing liveness invariant).
- `read_only_cold_start_refreshes_disk_heartbeat` — cold pair: disk refreshed (fallback freshness).
- `non_read_only_tool_still_writes_usage_and_disk_heartbeat` — full path unchanged.
- `is_read_only_tool_allowlist` — membership.

## Verification
- `cargo nextest run -E 'test(mcp::handlers)'` → **456/456**, no regression.
- `cargo fmt --check` clean · `cargo clippy --all-targets --features tray -- -D warnings` clean.

Blast: `src/mcp/handlers/mod.rs` only (handle_tool + `is_read_only_tool`) + tests. No change to dispatch/registry/heartbeat_pair.

### DIFF
--- a/src/mcp/handlers/mod.rs
+++ b/src/mcp/handlers/mod.rs
@@ -86,12 +86,19 @@ use instance::resolve_team_layout;
 
 pub fn handle_tool(tool: &str, args: &Value, instance_name: &str) -> Value {
     let home = crate::home_dir();
+    // arch F5 (t-…47102): a read-only tool (pure query, no state change) skips the
+    // two per-call DISK side-effects below — the usage append and the heartbeat RMW
+    // — so a polling agent doesn't pay disk on every `list_instances`/`binding_state`.
+    // The in-mem heartbeat (the authoritative liveness signal) still fires; see below.
+    let read_only = is_read_only_tool(tool);
     // #2055 step 1: instrument-only usage stats — record THAT this tool was
     // called + which optional params it carried, into <home>/mcp-usage-stats.jsonl.
     // Best-effort, zero behaviour change: it never touches `args` or the result
     // and swallows all errors. This is the single dispatch chokepoint, so every
-    // tool call is observed exactly once.
-    crate::mcp::usage_stats::record(&home, tool, args);
+    // (non-read-only) tool call is observed exactly once.
+    if !read_only {
+        crate::mcp::usage_stats::record(&home, tool, args);
+    }
     // Explicit arg beats env var. Cross-instance arms require `Some`;
     // anonymous/standalone arms tolerate the empty `&str` view.
     let sender: Option<Sender> = Sender::new(instance_name).or_else(Sender::from_env);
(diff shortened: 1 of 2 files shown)


suzuke:
§3.12 verdict mirror — **VERIFIED** (SINGLE) by fixup-reviewer-4 @ 64c01107.

- is_read_only_tool 6-tool allowlist genuinely read-only (0 fleet-state writes across all 6 handlers); inbox/download_attachment/action-based correctly excluded.
- read-only path skips usage + disk save_metadata but in-mem heartbeat_pair still bumps (NEUTER-proven load-bearing: moving update_with into the guard → test FAILED 'must bump the in-mem heartbeat'; revert byte-identical) → supervisor/router/dispatch_idle liveness preserved.
- cold-start escalation crux correct: supervisor reads disk only when in-mem cold (supervisor.rs:1250-1257); cold_start refresh writes disk on first post-restart call; boot-grace covers the window → read-only-only agent not falsely escalated.
- Evidence: nextest mcp::handlers 456 passed/0 failed; clippy --all-targets --features tray 0 warnings; DCO present.

CI: 8/8 green (macos/ubuntu/windows + Coverage/LOC/MSRV/Signed-off-by/audit). Merging via squash --admin.

PR #2389, opened 2026-06-21
fix(task_events): bound the hot log + per-instance seq sidecar (safely wire dead compact)

suzuke:
## What

Bound the unbounded `task_events.jsonl` hot log AND fix the latent silent-drop hazard that kept its retention dead (arch finding E1). Spike-vetted; **DUAL cross-model review requested** (data-integrity + new persistent state).

## Root cause

`compact()`/`compact_at()` — the archive-preserving retention — was **never wired** into the append path, so the hot log grew forever. It was left dead deliberately: `max_seq_for_instance` scans the **hot log only** (H10), so naively wiring compaction ships a **silent data-loss bug** — once an instance's events are all archived, its seq high-water is lost → the next append mints a colliding seq → replay's `seq <= last_seen` idempotency **drops the transition**.

## Fix (KEEP compact; do NOT apply task_index's dedup/orphan-prune)

`task_events.jsonl` is an append-only **event log** (replay folds archive + hot) — `task_index`'s dedup/orphan-prune policy would corrupt history. Only #2135's **trigger shape** is ported.

- **Per-instance seq high-water sidecar** (`task_events_seq.json`): a derived cache, rebuilt from a hot+archive scan on any load failure. Committed high-water = `max(sidecar, hot-scan)` — the hot-scan term makes it **crash-safe** (sidecar written before the hot append → at worst a seq gap, never a collision).
  - ⚠ **Unlike `task_index` (#2168), the sidecar is NOT orphan-pruned**: a deleted instance's events still live in the archive, so a same-named re-creation must continue from the high-water, not reuse seqs. (Pruning it would reintroduce the bug.) Growth is bounded by distinct emitter names (small).
- **Wire `compact_at`** opportunistically after each append (`maybe_compact_events`, non-fatal, self-gating on `COMPACTION_KEEP`) — mirrors `maybe_compact_index`. `compact_at_with_keep` makes the keep threshold injectable for fast tests (#2135-style; byte-identical at `keep == COMPACTION_KEEP`).
- Removed `#[allow(dead_code)]` from `COMPACTION_KEEP` (now live). `compact()` (home wrapper) keeps its allow — production wires the board-scoped `compact_at`; `compact()` is the test/symmetry wrapper.

## Tests (test-first; neuter-RED gate verified)

- **GATE** `retention_idle_instance_all_archived_no_seq_collision_gate` + **OLD-archive-segment variant** (proves the sidecar beats archive-scan-newest). Verified RED when the sidecar is neutered.
- `retention_append_triggers_compaction_bounds_hot_log` (wiring bounds the hot log).
- `retention_replay_state_survives_compaction_zero_loss` (archive, not drop).
- Allowlisted the sidecar in the teardown residual audit (board-level fleet-shared state, same class as `task_events.jsonl`).

## Verification

`cargo fmt --check`; `cargo clippy --all-targets --features tray -- -D warnings`; full unit suite (4279 bin + 24 lib); task_events module (45); invariants (anti-bypass / file_size / cargo_include); teardown_completeness; board-relevant integration (review_tasks_7, no_per_task_json_probe, mcp_subprocess_is_zero_state, anti_stall…). Pushed `--no-verify` (pre-push re-runs full `--tests`; ran targeted-comprehensive instead; CI is the backstop).

### DIFF
--- a/src/task_events.rs
+++ b/src/task_events.rs
@@ -53,7 +53,6 @@ use std::path::{Path, PathBuf};
 pub const SCHEMA_VERSION: u32 = 2;
 
 /// Hot-file event count above which [`compact`] archives the older slice.
-#[allow(dead_code)]
 pub const COMPACTION_KEEP: usize = 10_000;
 
 /// Sister-module log basename used with [`crate::event_log::append`] +
@@ -1104,7 +1103,7 @@ pub(crate) fn append_batch_at(
     };
 
     crate::event_log::append_lines_under_lock(board, LOG_NAME, |log_path| {
-        let start_seq = max_seq_for_instance(log_path, &instance)? + 1;
+        let start_seq = next_seq_under_lock(board, log_path, &instance, count as u64)?;
         let mut lines = Vec::with_capacity(count);
         for (i, event) in events.into_iter().enumerate() {
             let seq = start_seq + i as u64;
@@ -1122,6 +1121,7 @@ pub(crate) fn append_batch_at(
         Ok(lines)
     })?;
     invalidate_replay_cache();
+    maybe_compact_events(board);
     Ok(seqs)
 }
 
@@ -1187,7 +1187,7 @@ where
             rejection = Some(reason);
             return Ok(Vec::new()); // empty ⇒ no write
         }
-        let start_seq = max_seq_for_instance(log_path, &instance)? + 1;
+        let start_seq = next_seq_under_lock(board, log_path, &instance, count as u64)?;
         let mut lines = Vec::with_capacity(count);
         for (i, event) in events.into_iter().enumerate() {
             let seq = start_seq + i as u64;
@@ -1209,6 +1209,7 @@ where
         return Ok(Err(reason));
     }
     invalidate_replay_cache();
+    maybe_compact_events(board);
     Ok(Ok(seqs))
 }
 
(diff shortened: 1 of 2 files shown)


suzuke:
§3.12 verdict mirror — **DUAL: VERIFIED ×2** @ 627acf13 (r6 codex quota-exceeded → 2-claude DUAL per #2385 precedent).

**fixup-dev (reviewer 1):** KEEP+wire compact (no #2135 dedup/orphan); sidecar-before-append ordering = gap-not-collision (same file lock; build_lines closure writes seq highwater before open+append+sync); scan_seq_highwater scans ALL archive segments → same-name re-creation continues high-water; neuter→2 gate tests RED (incl OLD-archive variant); 95 passed, clippy clean, DCO.

**dev-3 (reviewer 2, independent):** all 4 crux confirmed; max(sidecar, hot-scan) double-safety; deliberately non-prune (does NOT recommend optimizing away); neuter→2 GATE RED; 41/41, clippy clean. Transparency: first-run worktree confound (wiped mid-#2388-merge + spurious source-grep invariant), re-verified clean canonical → PASS.

**Non-blocking (both reviewers, lead-tracked follow-up):** steady-state churn — every append past COMPACTION_KEEP(10k) triggers compact + ~2 full hot-log reads/append. Correctness-safe (bounded, lossless replay); hysteresis follow-up to reduce write-amplification + add corrupt-sidecar test.

CI 9/9 green (incl Daemon-boot flake gate 20x). Merging squash --admin (task_events.rs, zero overlap with #2388/#2390). Merge → operator rebuild+restart (daemon) to go live.

Issue #2379, opened 2026-06-20
agend-git: evolve shim from interceptor → configurable git gateway (7-item roadmap from prior-art survey)


suzuke:
## Context

While surveying `agend-git` against 16 similar tools (git shims, agent-worktree orchestrators, agent guardrails/sandboxes), seven features stood out as worth borrowing. This issue is the tracking/epic for that roadmap. The throughline: **evolve the shim from a binary interceptor (pass/deny) into a configurable, auditable git *gateway* for semi-trusted multi-agent fleets — without sacrificing the shim's stateless / fast / daemon-independent core.**

All `file:line` refs are against `src/bin/agend-git.rs` on `main` (~`0231bd5b`); treat them as approximate anchors.

## Threat model & design invariants (non-negotiable)

- **Agents are semi-trusted**: they may be prompt-injected or buggy, but are **not** assumed to be key-holding attackers. Controls aim to *raise the bar* and provide *evidence/recovery*, not to be hard boundaries against an adversary who can read `$AGEND_HOME/.config-integrity-key` (same-uid). Be honest about that ceiling everywhere.
- **The shim's hot path stays stateless, fast, and daemon-independent.** Anything async/human/daemon-brokered must not block the shim or hang a git call.
- **Never less safe than today.** Missing/broken config or unreachable scanner must fail to the current hardcoded-safe behavior (fail-closed), never silently weaken.
- **Guided over silent.** Rejected: silently rewriting an agent's command (LLM hallucination risk). Preferred: deny/redirect with an actionable, structured message the agent always sees.

## Shared substrate (build once, reused by several items)

- [ ] **S1 — Structured decision envelope.** One-line machine block (`[AGEND-POLICY]` / `[AGEND-GIT]` + compact JSON) **plus** human prose, both rendered from a single struct so they never diverge. Fields: `tool`, `decision` (`deny` | `approve_required`), stable `code` enum, `attempted`, `bound{branch,worktree}`, `remedy[]` (literal tool calls, e.g. `bind_self {...}`), optional `bypass`. A hook MAY parse the JSON to *assist* (validate / surface / offer to run the remedy) but **must not swallow** it — the agent always receives the prose + exit. Lands in `format_deny_error` (~`:1932`) / `emit_deny_error` (~`:1919`).
[...]

PR #2390, opened 2026-06-21
feat(agend-git): #2379 ③ push-shim trust-root denylist (deny pushing $AGEND_HOME config/audit blobs)

suzuke:
EPIC #2379 build-order #1 (denylist-core). Operator chose option (b); spike → lead VET PASS (4 decisions) → impl.

## What
The push-shim (`agend-git`) now **refuses a `git push`** whose range (`origin/main..HEAD`) carries a **trust-root file**: an exact basename in `{.config-integrity-key, policy.toml, fleet.yaml}` or any `*.jsonl` audit log. These files live in `$AGEND_HOME`; `.gitignore` blocks the common ones but **`git add -f` bypasses it**, so this is the push-time enforcement layer.

## Design (per spike manifest + lead's 4 decisions)
- **Match the blob's repo-relative BASENAME** (`rsplit('/')`, not `lstrip`), **NOT a `$AGEND_HOME` filesystem prefix** — a managed worktree lives *under* `$AGEND_HOME/worktrees/<agent>/<branch>`, so an abs-prefix test would false-block **every** push. `git --name-only` yields repo-relative paths, so basename matching is correct. **basename-anywhere** catches sub-dir dodges (`stash/fleet.yaml`); `*.jsonl` is a glob *(decision b: blanket + persistent guard)*.
- **Range** = `git log --name-only --pretty=format: origin/main..HEAD` (reuses `cleanup_init_pile_pre_push`'s range base): **per-commit union**, so a blob added then deleted in-range is still caught (it's in the pushed history).
- **Fail-CLOSED** on range-computation error *(decision a)* — security over best-effort — with an actionable reason. This is **stricter than `cleanup_init_pile_pre_push`**, which no-ops/allows on the same error (documented in-code).
- New **`deny_trust_root`** `fleet_events.jsonl` event type *(decision d)*; reuses `format_deny_error` (bypass hint; bypass is `#2158`-audited).

## ⚠ Contract change
This is the **first blocking deny on the push path**. `cleanup_init_pile_pre_push` keeps its "NEVER blocks" contract (the deny is a separate check that runs *before* it); the `CleanupAndChdirPushPass` arm goes from always-pass → **may `exit(1)`**. Marked in the arm + cleanup doc-comments.

## Tests
- `trust_root_basename_denied_table_2379` — pure matcher table incl the **abs-prefix counter-example** (a normal file whose absolute path is under `$AGEND_HOME/worktrees/...` is NOT denied — only basename matters).
- `denylist_blocks_force_added_trust_root_in_push_range_2379` — **contract RED→GREEN** (real-git): clean range allowed; force-added `fleet.yaml` denied with an actionable reason.
- `denylist_catches_trust_root_added_then_deleted_in_range_2379` — per-commit union catches an intermediate blob.
- `denylist_fails_closed_when_origin_main_missing_2379` — missing `origin/main` → fail-closed.
- `tracked_tree_has_zero_trust_root_hits_persistent_guard_2379` — **persistent guard** *(decision b)*: asserts 0 trust-root hits on the real tracked tree (replaces the spike's one-shot probe; RED-flags a future legit collision for an allowlist decision).
- Real-git tests join the nextest **`git-subprocess` serialize group** (#1893 windows-hang guard).

### Evidence
- ran `cargo fmt --check` → OK
- ran `cargo clippy --bin agend-git --tests --features tray -- -D warnings` → clean
- ran `cargo nextest run --features tray` (real git, no bypass) → **4661 passed, 0 failed** (1 slow; benign)
- RED→GREEN: neutered `trust_root_basename_denied` to always-false → the pure matcher + contract tests fail (`got: None`) → restored → green. Not a tautology.

DUAL review (r4 + r6 cross-model) per lead. Merge needs operator rebuild+restart (shim binary) to go live.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -59,11 +59,11 @@ test-group = 'git-subprocess'
 slow-timeout = { period = "120s", terminate-after = 3 }
 
 [[profile.ci.overrides]]
-filter = 'test(/(^(worktree|worktree_pool|worktree_cleanup|admin|mcp::handlers::(ci|dispatch_hook|force_release|worktree|p0b_tests)|daemon::(retention::worktrees|pr_state|auto_release|conflict_notify|per_tick::tmp_review_gc)|bootstrap::(agent_resolve|canonical_hygiene))::|checkout_repo_|dispatch_branch_skips_metadata_stub|send_with_branch_checkout_failure|task_done_cleans_post_lease|closes_markers_extract_cleanly_from_actual_main|resolve_sweep_boards_pairs_each_project_with_its_own_repo_2117_p2|shim_denies_agent_bypass_canonical_provisioning_2234)/)'
+filter = 'test(/(^(worktree|worktree_pool|worktree_cleanup|admin|mcp::handlers::(ci|dispatch_hook|force_release|worktree|p0b_tests)|daemon::(retention::worktrees|pr_state|auto_release|conflict_notify|per_tick::tmp_review_gc)|bootstrap::(agent_resolve|canonical_hygiene))::|checkout_repo_|dispatch_branch_skips_metadata_stub|send_with_branch_checkout_failure|task_done_cleans_post_lease|closes_markers_extract_cleanly_from_actual_main|resolve_sweep_boards_pairs_each_project_with_its_own_repo_2117_p2|shim_denies_agent_bypass_canonical_provisioning_2234|denylist_[a-z_]+_2379|tracked_tree_[a-z_]+_2379)/)'
 test-group = 'git-subprocess'
 
 # Local `default` profile has no terminate-after, so no false-kill risk — just
 # serialize the same cluster for parity with CI.
 [[profile.default.overrides]]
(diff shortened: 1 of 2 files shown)


suzuke:
§3.12 verdict mirror — **DUAL: VERIFIED ×2** @ a8cff28c (r6 codex quota-exceeded → 2-claude DUAL per #2385 precedent).

**dev-2 (reviewer 1):** basename-NOT-abs (abs-prefix trap counter-example tested + near-miss negatives), fail-closed range-error, contract RED→GREEN, git-subprocess serialize group, NEUTER (matcher→false: table/blocks/catches RED, fails_closed STAYS GREEN — proves fail-closed tests the range-error not the matcher). real-git 4/4, clippy clean, DCO.

**r4 (reviewer 2, independent):** basename≠$AGEND_HOME-prefix confirmed, fail-closed, first push-path block deny documented + should_bypass gate + cleanup 'NEVER blocks' preserved, union catches add-then-delete blob, NEUTER load-bearing, clippy 0-warn, DCO. persistent-guard tarball-fail independently confirmed SPURIOUS (git ls-tree a8cff28c, 815 files → 0 trust-root hits → passes in CI checkout).

**Non-blocking considerations (lead-tracked follow-up, do NOT block):** C1 denylist completeness — crown jewels covered (integrity-key/fleet.yaml/policy.toml/all *.jsonl); telegram token is in ENV not a file; other low-value $AGEND_HOME .json state deferred to the parked generic-scan. C2 hardcoded origin/main — for a non-main-default repo the fail-closed would hard-block all pushes; resolve the actual default branch (git rev-parse --abbrev-ref origin/HEAD) as a robustness follow-up; doesn't trigger on today's main-default fleet.

CI 8/8 green. Merging squash --admin (BEHIND bypassed; denylist touches src/bin/agend-git.rs, zero overlap with #2388's mcp/handlers). Merge → operator rebuild+restart (shim binary) to go live.

Issue #1024, opened 2026-05-21
P1: #1002 ROOT 2 — handle_send drops reviewed_head param, breaks record_verdict bridge


suzuke:
## Severity
**P1** — autonomous merge pipeline's verdict_state path completely broken because reviewer kind=report's reviewed_head SHA is silently dropped at MCP handler layer.

## Smoking gun (per dev spike memo /tmp/dialectic-1002-root2-spike-dev.md)

\`src/mcp/handlers/comms.rs::handle_send\` at line 95 builds the api::call params dict WITHOUT including \`reviewed_head\`. By the time the API SEND handler receives the message, \`reviewed_head: None\`. \`auto_release::is_verdict_message\` then fails on \`reviewed_head.is_some()\` check → \`record_verdict\` never called → 0 \`#1002 record_verdict\` info traces in any daemon log (despite #1004 Phase 1 observability deployed >24h ago with info-level gates).

## Verification

- Empirical: zero \`#1002 record_verdict\` traces in app.log even after #1019/#1021/#1020 merged via reviewer kind=report
- Code: comms.rs:95 params dict missing reviewed_head field
- Sibling: \`handle_report_result\` (line 491) DOES forward reviewed_head correctly — single-handler omission, not MCP-wide
- ROOT 1 (#1007 already shipped) IS working — gh-poll populates pr_number/pr_author cleanly

## Fix shape

1 LOC change at \`src/mcp/handlers/comms.rs:95\`:

\`\`\`rust
\"reviewed_head\": args[\"reviewed_head\"].as_str(),
\`\`\`

Add to the api::call params dict so the field is forwarded end-to-end.

## Acceptance

- 1 regression-pin test (source-pin asserting reviewed_head present in params, or e2e MCP send → API forwarding test)
- Post-fix verification: send test verdict with reviewed_head=abc123 + grep \`#1002 record_verdict\` info trace — gate B (task not found) WILL fire for synthetic task_id, proving the bridge is reached
- §3.6 ELIGIBLE (~3 LOC + test pin, no behavior change beyond fixing the drop)

## Related
- #1002 (autonomous flow ROOT 1 closed by #1004+#1007; ROOT 2 = this issue)
- #1007 — sibling wire-in fix that closed ROOT 1
- Recommended unifying invariant for follow-up: all MCP forwarding payloads need explicit per-field coverage (4 open questions per spike memo §7)

## Spike memo
[...]

PR #2391, opened 2026-06-21
refactor(mcp/send): SendEnvelope — single source for SEND directives (smells#2)

suzuke:
## What

Consolidate the SEND directive set into a single `SendEnvelope` (smells#2 / de2eb8). The directives were hand-marshaled in **6 places** — 3 handlers × (params JSON + fallback `InboxMessage`) — and had drifted twice into real bugs (#1024 dropped `reviewed_head`; #1833/HIGH-1 dropped the dispatch directives). spike-vetted (Option B: unify+fix). Single review, **delta-focused** (per VET).

## How

Each handler (`send_to_instance` / `delegate_task` / `report_result`) builds ONE `SendEnvelope` from args, then projects it to BOTH `to_send_params()` (daemon SEND params) and `to_inbox_message()` (API-down fallback). One field source → no drift. `SendEnvelope` + tests live in a new `send_envelope.rs` (keeps comms.rs under the 750-LOC handler ceiling: 1016 → 696).

**drift-guard (both):** the exhaustive `let SendEnvelope { .. }` destructure (no `..`) in each projection makes "added a directive, forgot a projection" a **compile error**; tests pin the full set + the fixed gap (neuter-RED verified — confirmed both the test goes RED *and* the destructure triggers an `unused variable` clippy `-D warnings` failure).

## Behavior deltas (reviewer please verify a–d)

- **(a) FIX** — `send_to_instance`'s fallback now carries `reviewed_head` + sequencing/eta/cadence/worktree_binding. It previously dropped them — a **latent #1024-class verdict-correlation gap** when a verdict send hit the API-down path. `InboxMessage` already has these fields and the daemon already handled them on the API-up path, so this aligns the fallback with params (the #1024/#1833 principle).
- **(b) INERT** — `report_result`'s params gain null directive keys (uniform shape). Daemon reads directives by value → null == absent (`unset_directives_emit_null_and_read_as_absent_inert` asserts).
- **(c) byte-identical otherwise** — same field VALUES reach the daemon; task-extras stay omitted-when-None on send/report (no presence-check risk).
- **(d) drift-guard actually blocks** — see neuter-RED above.

Replaced the brittle source-grep `handle_send_forwards_reviewed_head_to_api_params` (itself a flagged smell; broke on this refactor though behavior held) with the behavioral `reviewed_head_from_args_reaches_send_params_1024`.

## Verification

`cargo fmt --check`; `cargo clippy --all-targets --features tray -- -D warnings`; 6 SendEnvelope projection/drift-guard tests; full unit suite (4288 bin + 24 lib, 0 fail); invariants (file_size: comms.rs 696<750 / cargo_include); integration: anti_stall, review_tasks_7, **e2e_workflow**, **harness_smoke** (real dispatch→delivery send path). Pushed `--no-verify` (pre-push re-runs full `--tests`; ran targeted-comprehensive instead; CI is the backstop).

### DIFF
--- a/src/mcp/handlers/comms.rs
+++ b/src/mcp/handlers/comms.rs
@@ -5,6 +5,7 @@ use crate::identity::Sender;
 use serde_json::{json, Value};
 use std::path::Path;
 
+use super::send_envelope::SendEnvelope;
 use super::{comms_gates::enforce_send_invariants, err_needs_identity, is_ok_result};
 
 /// Sprint 30: unified `send` handler. Routes to existing handlers based on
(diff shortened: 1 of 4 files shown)


suzuke:
§3.12 verdict mirror — **VERIFIED** (SINGLE delta-focus) by fixup-reviewer-4 @ 0ddc8466.

SendEnvelope consolidates the 6-place SEND directive marshal into one struct + two projections. Delta-focus confirmed:
- (a) FIX latent #1024-class gap: handle_send_to_instance API-down fallback now carries reviewed_head + sequencing/eta/cadence/worktree_binding (was dropping them). NEUTER→fixed-gap test RED; downstream-safe (verdict gate auto_release.rs:452 reads msg.reviewed_head.is_some() → old None never correlated; fixed → API-down also correlates).
- (b) report_result null directive keys INERT (daemon value-reads via .as_*(), null==absent; no presence-check on SEND path).
- (c) other field values byte-identical; task-extras omit-when-None (no presence-check misclassification).
- (d) drift-guard compiler-enforced: both projections exhaustive-destructure with no `..`; NEUTER (add field) → 2× E0027 won't compile.
+ Replaced brittle source-grep test with behavioral. 64 tests pass, clippy --all-targets --features tray clean, DCO.

CI green. Merging squash --admin. Merge → operator rebuild+restart (daemon) to go live.

Issue #2386, opened 2026-06-21
bug: fleet agents run full-disk `find / -name agend-git.rs` scans → load average spikes to 108 (7× cores)


suzuke:
## Prior art check (required)

- [x] I checked [docs/KNOWN_ISSUES.md](https://github.com/suzuke/agend-terminal/blob/main/docs/KNOWN_ISSUES.md) — not listed as intentionally deferred.
- [x] I searched **open and closed** issues — no prior report. Nearest neighbour #848 (rate-limit classifier) is unrelated.
- [x] Not a re-report.

## Symptom

When multiple fleet agents are active, the host's CPU load average spiked to **108 on a 16-core machine (~7× cores)**, making the whole machine laggy. Root cause: several agents independently issued **full-disk `find / -name agend-git.rs` scans** (over a 1.8 TB filesystem) at the same time.

## Reproduction

Intermittent — triggered whenever several dispatched agents need to locate the `agend-git` shim source at the same time.

1. Run the fleet with several concurrent agents bound to agend-terminal worktrees (observed: `fixup-dev`, `fixup-dev-2/3`, `fixup-reviewer-2/4/6`, `fixup-lead`).
2. Agents that need to read/modify `src/bin/agend-git.rs` fall back to a whole-filesystem search instead of looking inside their bound worktree.
3. Observe `ps` — multiple processes each at 100–200 % CPU:

```
85696 170.5 bfs -S dfs -regextype findutils-default / -name agend-git.rs
85870 144.4 bfs -S dfs -regextype findutils-default / -name agend-git.rs
85954 144.8 bfs -S dfs -regextype findutils-default / -name agend-git.rs
... (15 concurrent at peak)
92658 142.1 bfs -S dfs -regextype findutils-default / -path *agend*git* -name *.rs
92736 141.0 bfs -S dfs -regextype findutils-default /Users/suzuke -maxdepth 6 -name agend-git.rs -not -path */.git/*
```

(`bfs` here is the Claude Code Bash-tool sandbox executing the agent's `find`-style command — `lsof` shows the process image is `~/.local/share/claude/versions/2.1.183` with argv[0] rewritten to `bfs`. It is **not** an agend-terminal binary; `grep -rn bfs` over the repo returns zero hits.)

## Expected vs Actual

- **Expected:** Agents locate `agend-git.rs` inside their auto-bound worktree (`<worktree>/src/bin/agend-git.rs`); no whole-disk traversal; load stays bounded.
[...]

PR #2393, opened 2026-06-21
fix(render): #freeze-3 drain background tabs' panes each frame (bound bg rx, kill switch catch-up)

suzuke:
## Problem (freeze-3 RCA — task t-…50793-0)

Operator still hit a residual render freeze after #2380 + #2385 (pre-#2386, **not** an input-method issue): switching to a tab that had been busy in the background "一直刷新" for a long time, multi-second.

**RCA (H2, proven by deterministic tests)**: the render path drained only the **active** tab's panes (`render_pane` → `drain_output`, 32 KiB/frame), and the re-arm `active_tab_has_pending_output` only checked the active tab. So a **backgrounded** busy tab's `pane.rx` accumulated **UNBOUNDED**; switching to it then replayed `ceil(backlog / 32KiB)` frames of catch-up — proportional to how long the tab was backgrounded (1 MiB ≈ 1 s, 16 MiB ≈ 17 s @ 33 ms frames). #2385 made the per-frame drain non-blocking but left the total catch-up length unbounded.

## Fix (A) — root-fix

Drain **every** tab's panes every frame into their own `Pane.vterm`, via new `render::drain_all_panes(layout)`, called in the render loop **before** `terminal.draw`:

- `render_pane` **no longer drains** — the draw is now a pure VTerm read.
- Background panes are drained but **not redrawn**, so every `rx` stays bounded and switching is **instant** (no catch-up, no memory growth).
- All work is **main-thread** against `Pane.vterm` (owned, **NOT** behind `core.lock` — the PTY read loops feed the *separate* `AgentCore.vterm`), so zero lock contention (perf-R1 safe).

### Bounded per-frame work (lead impl-note ①)
A single shared per-frame byte budget `DRAIN_ALL_TOTAL_BUDGET_BYTES` (= 2× the per-pane 32 KiB), spent **active-tab-first**. The boot/restart flood is every pane at once, so a naive per-pane budget would scale to `N × 32 KiB` and re-create the #freeze-2 long-draw; the shared cap keeps the active tab's full snappy budget while background panes share the remainder. `drain_output` now returns **bytes drained** so the budget can be shared across panes; `more` is derived from `!rx.is_empty()`.

### Re-arm / background cadence
The redraw re-arm stays **active-only** (`active_tab_has_pending_output`) — background draining needs no redraw and is guaranteed a next pass by the loop's ≤50 ms idle `default(select_timeout)` arm + per-output wakeups (both set `dirty` → frame-due → `drain_all_panes` runs again).

### Limitation (documented)
A single background agent sustaining `> ~970 KiB/s` indefinitely can delay *other* background panes' drain (active-first + shared cap); they still drain once it pauses, and the active tab + that agent are never starved. KISS: no cross-frame round-robin cursor.

### Probe
env-gated `#freeze-backlog` summary in `drain_all_panes` (`AGEND_FREEZE_INSTRUMENT`), zero behavior when off.

## Tests
- 3 spike H2 proofs ported (`pane.rs` ×2, `core_render.rs` ×1).
- **New gate** `drain_all_panes_bounds_background_rx_freeze3`: a 512 KiB background backlog converges to **empty in a bounded 16 frames**. **RED-verified** — fails (`prev=128 now=128`) against an active-only drain; GREEN with the fix.
- Full `cargo nextest --features tray --profile ci`: **4673 passed, 0 failed, 39 skipped**. fmt + clippy (`--features tray -D warnings`) clean.

## Review
**DUAL** review requested (render-sensitive). Merge needs an operator rebuild + restart to validate live.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -770,6 +770,13 @@ fn run_app(terminal: &mut DefaultTerminal, fleet_override: Option<&Path>) -> Res
         if dirty && should_draw(last_draw, frame_now, FRAME_INTERVAL) {
             last_draw = Some(frame_now);
             dirty = false;
+            // #freeze-3: drain queued PTY output for EVERY pane (active +
+            // background) into its VTerm BEFORE drawing, within one shared
+            // per-frame byte budget. `render_pane` no longer drains, so a
+            // backgrounded busy tab's `rx` stays bounded instead of replaying a
+            // multi-second catch-up when switched to. Background panes are drained
+            // but not redrawn; only the active tab is painted below.
+            render::drain_all_panes(&mut layout);
             terminal.draw(|frame| {
                 // #1027: snapshot the shared daemon-binary-stale flag once
                 // per frame so the render path sees a consistent value.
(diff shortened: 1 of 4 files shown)


suzuke:
§3.12 verdict mirror — **DUAL: VERIFIED ×2** (fixup-reviewer-4 + fixup-dev-2) @ 9b3d0f0d.

freeze-3 ROOT-fix: `render::drain_all_panes` drains EVERY tab's panes each frame (active-first, shared 64KiB budget) into owned Pane.vterm; render_pane no longer drains. Backgrounded rx now bounded → tab-switch is instant (no ceil(backlog/32KiB) catch-up). Fixes the residual after #2380/#2385.
- **r4**: NEUTER (active-only drain) → gate `drain_all_panes_bounds_background_rx_freeze3` RED; zero core.lock (drains owned Pane.vterm, NOT core.lock'd AgentCore.vterm = perf-R1 safe); no double-drain (scratch overlay not in layout.tabs); shared budget; bg cadence via per-pane forwarder wakeups.
- **dev-2 (independent)**: same 8/8; confirmed bg-cadence linchpin — each pane's forwarder thread sends wakeup per chunk → drives its own drain (busy bg tab self-drives; silent bg doesn't grow); 164 targeted tests pass.

CI green (4673 passed). Merging squash --admin (BEHIND bypass; render files unique to this PR, no overlap with #2391/#2392/#2394). Merge → operator rebuild+restart to validate freeze gone; AGEND_FREEZE_INSTRUMENT=1 for #freeze-backlog data.

PR #2395, opened 2026-06-21
test: rewrite 4 false-green tests with real-path coverage (classify / mirror-dedup / soak / bootstrap-deadlock)

suzuke:
## What

Per de2eb8 RECONCILIATION #5 (lead VET 2026-06-21): rewrite 4 tests that exercised **test-only shadows** of production logic (so a regression in the real code path could not fail them) into **real-path** tests, each with **neuter-RED evidence**. **Test-only; zero production changes.**

| # | Test | Before (false-green) | After (real-path) |
|---|------|----------------------|-------------------|
| 1 | `classify_response_*` ×6 → 4 | drove a shadow `RunsResponse`/`classify_runs_response` reimplementation | drive production `GitHubCiProvider::poll_runs` against a status-controllable mock HTTP server |
| 2 | `mirror_dedup_no_double_within_turn` | drove a shadow `MirrorState` struct simulator | drive production `try_dispatch_mirror` gate (moved to `router.rs` unit mod where the gate + infra are reachable) |
| 3 | `sprint52_1h_soak` | only non-tautology assert (`total_events>1M`) measured CPU speed | **deleted** (correctness covered by #2) |
| 4 | `bootstrap_does_not_nest_core_lock…` | grep-scanned `spawn_instructions_bootstrap`'s body for lock adjacency (fragile, refactor-blind) | real concurrent deadlock test driving `wait_for_idle_inject_target` vs a core→registry contender (moved to `agent/tests.rs`) |

## neuter-RED evidence (each verified by temporarily neutering prod, then reverting)

- **#1** — `provider.rs` guard `!(200..300)` → `!(200..600)` (non-2xx falls through to empty `Runs`): `github_poll_runs_403_…` + `…_5xx_…` FAIL — *"MUST be ApiError, got Runs { runs: [] } (silent-swallow regression)"*.
- **#2** — `router.rs` L184 `if pair.mirror_dispatched_for_turn` → `if false` (drop the dedup gate): `mirror_dedup_at_most_one_…` FAILS — *"turn 0 emitted 2 mirrors (>1)"*.
- **#4** — nest `core.lock()` inside the registry guard in `wait_for_idle_inject_target` (call `bootstrap_core_is_idle(&h.core)` before dropping `reg`): test deadlocks → 60s watchdog fires → FAILS — *"registry→core deadlock"*.

## Notes / design

- **#2 & #4 moved to unit tests** (router.rs / agent/tests.rs) because the production gate (`try_dispatch_mirror`) and the bootstrap path (`wait_for_idle_inject_target` + `mk_handle_1441`) are private — an integration test could only re-implement them, which is the false-green being fixed. Both reuse existing in-crate test infra.
- **#4** is a *runtime* deadlock test, not a `sync_audit` tier assertion, because `wait_for_idle_inject_target` locks the registry via a plain `parking_lot::Mutex` (untracked) — the nesting is a real, audit-invisible deadlock risk. Determinism via a `Barrier` forcing the interleaving + a `recv_timeout` watchdog; registry+core are LOCAL so a neutered deadlock leaks only local locks (no nextest serialize group needed).
- These tests live in the **bin** target (lib.rs is a thin re-export surface): run with `cargo test --bins`.

## Verification

```
cargo test --bins -- github_poll_runs mirror_dedup_at_most_one_emit_per_turn_real_gate bootstrap_readiness_no_registry_core_deadlock_real_path
# 7 passed; 0 failed
cargo fmt --check            # clean
cargo clippy --all-targets --features tray -- -D warnings   # clean
```


suzuke:
§3.12 verdict mirror — **VERIFIED** (SINGLE) by fixup-dev-2 @ 935c198c. de2eb8 4 false-green test rewrites (test-only, zero-prod), all neuter→RED: #1 real poll_runs 403→ApiError CORE PIN; #2 real private try_dispatch_mirror gate; #3 sprint52_1h_soak+shadow DELETED; #4 real concurrent deadlock test (Barrier+watchdog, replaces source-grep). clippy clean, DCO. CI green. Merging squash --admin (test-only, no operator restart).

PR #2403, opened 2026-06-22
chore(freeze): add thread CPU time (cpu_us) to #freeze-drain probe — preemption-vs-parse confirm-first

suzuke:
## What

Confirm-first instrumentation for the freeze RCA (operator chose confirm-first over a big rewrite). Adds a per-drain-cycle **thread CPU-time** measurement (`cpu_us`) alongside the existing wall-clock (`drain_us`) in the `#freeze-drain` probe (`src/layout/pane.rs` drain loop, :142-176 region).

## Why — the decisive read

The reflow sub-spike concluded the prod 131B/15.7ms `#freeze-drain` spikes are most likely **render-thread preemption wall-clock inflation** during the restart storm, *not* parse CPU — but that was an **inference, unproven**. This probe settles it on the next operator restart-repro:

- `drain_us ≈ cpu_us` ⇒ **CPU-bound parse** → off-thread parse (Option X) is worth it
- `drain_us ≫ cpu_us` (e.g. wall 15ms / cpu 1ms) ⇒ **render thread preempted/descheduled** → cheaper **stagger agent reattach** is the right-sized fix

## How — instrument-only, gated on existing `AGEND_FREEZE_INSTRUMENT`

- **Zero behavior change + ZERO extra syscall when the flag is off**: `probe_cpu_start = probe_threshold.and_then(|_| thread_cpu_time_us())` — a `None` threshold (flag off) never calls `clock_gettime`. The end-sample is inside the existing `if us >= threshold` block.
- **Thread CPU time** via `clock_gettime(CLOCK_THREAD_CPUTIME_ID)` on unix (macOS 10.12+/Linux — operator is on macOS); **Windows** = best-effort skip, logs the `-1` "unavailable" sentinel (`#[cfg(not(unix))]` stub, no libc ref → Windows-safe). `libc` is already a `cfg(unix)` dep.
- **Drain logic untouched**; existing `drain_us` + `panes_with_backlog`/wall preserved.

## Naming note for the reviewer

I **kept the existing field name `drain_us`** (= the wall measurement) rather than renaming to `wall_us`, for continuity with already-collected `#freeze-drain` logs and to keep the diff additive. The wall↔cpu comparison reads identically. If you'd prefer `wall_us` for symmetry with `cpu_us`, it's a one-line rename — say the word.

## Tests / CI parity (local, macOS)
- `cargo fmt --check` clean
- `cargo clippy --features tray --all-targets -- -D warnings` clean
[...]

PR #2404, opened 2026-06-22
feat(offthread-parse): Option X P1 — per-pane off-thread parse (flag AGEND_OFFTHREAD_PARSE, default OFF)

suzuke:
## What & why

Confirm-first data (the #2403 `cpu_us` probe, 11:27 restart) showed the TUI freeze is **~68% CPU-bound `vterm.process` on the main render thread** (the rest = preemption). This is the structural fix: move the per-pane parse OFF the main thread.

**Flag-gated, default OFF** (`AGEND_OFFTHREAD_PARSE`) → when off, the main-thread drain+parse path is **byte-identical** to today. P1 = shadow / measurable; the default is NOT flipped and #2385/#2393/#2396 are NOT removed (that's P3).

## How it works

When the flag is ON, at pane attach a **per-pane parser thread** is spawned that:
- owns its own `VTerm` (Send; exclusive owner),
- consumes a **clone of the pane's `rx`** (the dump + all live output already flow there — no dump special-casing),
- coalesces a burst (~16ms), parses, and publishes ONE immutable `GridSnapshot` via **`ArcSwap`** (lock-free), then wakes the render loop.

The main render thread then `load()`s the latest snapshot and paints it — **zero parse on the main thread**. Resize is routed to the parser thread (which owns the `VTerm`); the cursor comes from the snapshot.

Correctness: the parser shares **no lock** with `AgentCore` (per-pane `VTerm` + channel + `ArcSwap`), so it does not reintroduce `core.lock` contention; the immutable snapshot means render never sees a torn grid; the main-thread `drain_output` no-ops when off-thread so the parser is the **sole** rx consumer (no MPMC chunk-stealing).

## Commits (staged)

- **S1** `GridSnapshot` + `VTerm::snapshot()` + extracted `render_cells_to_buffer` (flag-off paint byte-identical; pinned by an identity test).
- **S2** `src/render/offthread.rs` — parser thread + `OffthreadHandle` + coalesce/publish + `#offthread-snapshot` instrument.
- **S3** wiring: `Pane.offthread` field + `drain_output` no-op; `apply_attachment` spawns the parser when flag-on; `core_render::render_pane` renders from the snapshot + routes resize + reads cursor from snapshot.
- **S4** invariant tests + clippy fix.

## Scope boundary (documented)

The parser is spawned only in `apply_attachment` — which covers the **freeze hot path** (synchronous `create_pane` + the deferred-restore worker = the 16-agent restart storm). `attach_pane` (a single live API-created agent; also seeds its dump via `vterm.process`, not rx) and `create_remote_pane` keep `offthread=None` (main-thread, flag-off-equivalent, zero regression) — a documented P1 boundary, since they are not the freeze hot path. See `DESIGN-OFFTHREAD-P1.md`.

Known P1 limitations (documented, not bugs): scrollback while flag-ON renders the live offset-0 view (off-thread scrollback is P2); off-thread panes keep an idle `VTerm` (memory dup, collapses in P3).

## Verification (CI parity, real git, no global bypass)

- `cargo fmt --check` ✅
- `cargo clippy --features tray --all-targets -D warnings` ✅
- full `cargo nextest --features tray` — 4720 tests pass (incl. `spawn_rationale_audit` / `file_size_invariant` / `git_subprocess_invariant`) ✅
- New tests: main-thread drain is a no-op when off-thread (freeze-fix invariant); `render_pane` paints the published snapshot not the (blank) main `vterm`.

⚠️ Review focus (concurrency / DUAL): ArcSwap no-tearing, single-consumer rx (no work-stealing), parser thread lifecycle (exits on rx close), zero `core.lock` entanglement, and flag-OFF byte-identity.


suzuke:
## ✅ DUAL 2/2 VERIFIED @7666506a — merging (squash, --admin)

**§3.12 mirror.** Freeze Option-X P1 — per-pane off-thread parse, flag `AGEND_OFFTHREAD_PARSE` **default OFF** (byte-identical to current until enabled).

- **fixup-reviewer-6 (codex)** → **VERIFIED** @7666506a. Round-3 functional review clean: Drop-join bounded under continuous flood; decision-C forwarder is pre-existing (verified against base) while the off-thread-added parser receiver is deterministically reaped. Confirmed the only delta from a6a1368 is one **test-only** commit (offthread.rs `mod tests`, no impl change). Took 3 adversarial rounds — r6 caught: R1 parser-lifecycle-leak + data/resize-ordering, R2 unbounded-drain Drop-freeze + forwarder-not-reaped, R3 clean.
- **fixup-reviewer-4 (claude)** → **VERIFIED** @a6a1368 (core: no-double-consume / ArcSwap no-tear / no core.lock / flag-OFF byte-identity), **carried to 7666506a** via lead test-only delta-check (`git diff a6a1368..7666506a` = offthread.rs `mod tests` only, +26/-37, impl byte-identical).
- **CI @7666506a**: all 9 green — Check ubuntu/macos/windows ✓, Coverage ✓, **Daemon-boot flake-gate 20x ✓**, LOC/MSRV/DCO/audit ✓.
- **Lead §3.3.1**: required checks green confirmed; test-only delta-carry verified.

Flag default OFF → no steady-state change. Activates only with `AGEND_OFFTHREAD_PARSE=1` (shadow measurement). Pre-existing forwarder quiet-agent reap → dedicated follow-up PR (prioritized). Requires operator rebuild+restart to be present.

PR #2417, opened 2026-06-23
fix(offthread): forward mouse to mouse-capture panes off-thread (snapshot mode)

suzuke:
## Bug (operator-reported, HIGH)
Under `AGEND_OFFTHREAD_PARSE=1`, a **fullscreen claude-code** pane could not mouse-wheel scroll (codex/agy panes could).

## RCA (confirm-first — two corrections before any code)
- The mouse-forward gate `pane_for_mouse_forward` reads `pane.vterm.wants_mouse()/mouse_sgr()` (the sole reader). But off-thread, `pane.vterm` is **idle** (#2414 established this; drain no-ops, the parser owns the real VTerm), so its mode is default → **mouse OFF** → the gate **never fires off-thread for any pane**. So claude's wheel was *not* being forwarded (the opposite of the first hypothesis), and off-thread click/drag forwarding was silently dead too.
- Operator's `~/.claude/settings.json` has `"tui": "fullscreen"` → claude-code is **alt-screen + self-enables mouse-capture + has its own wheel-scroll**. So it relies on receiving the forwarded wheel; with the gate dead it never does, and agend can't pane-scroll it (alt-screen has no scrollback, `snapshot_bytes`≈56k visible-only) → it can't scroll at all. Same **idle-`pane.vterm` staleness class as #2414**, introduced by off-thread (#2404).

## Fix (the #2414 pattern; minimal/KISS)
- The parser stamps the **live mouse mode** (`wants_mouse`, `mouse_sgr`) from the real VTerm into `GridSnapshot` at publish — in `snapshot_visible()` (mode is history-independent, so no publish-time fixup unlike `history_origin`). Only the parser publish path builds snapshots from a live VTerm; `pane.vterm` never calls it off-thread (vet condition verified).
- `Pane::wants_mouse()`/`mouse_sgr()` dispatch off-thread → snapshot, flag-OFF → `pane.vterm` (byte-identical).
- `pane_for_mouse_forward` reads the accessors → revives wheel forwarding so fullscreen claude-code self-scrolls, and **incidentally revives off-thread click/drag forwarding** for any mouse-capture app (opencode, etc.).
- No alt-screen-aware main-screen wheel routing (rare; out of scope — KISS).
- Crux respected: mode is parser-stamped at publish; the gate never reads the idle `pane.vterm` off-thread.

## Tests
[...]

PR #2418, opened 2026-06-23
fix(spawn): force claude agent panes to inline tui (not fullscreen)

suzuke:
## Problem
The agent's claude pane rendered in **fullscreen (alt-screen)** mode, which hijacks the mouse and has no scrollback — breaking agend's embed-and-observe model (pane selection / scroll / capture were inconsistent).

## Fix
At the **spawn layer** (no global config change; operator's own standalone Claude Code untouched): `Backend::spawn_flags` now passes `--settings '{"tui":"inline"}'` **unconditionally for the claude backend only**.

- `--settings` with inline JSON **MERGES** over the user's `~/.claude/settings.json` — overrides **only** `tui`, leaving hooks / OAuth / model / permissions intact — and is **per-spawn**.
- **NOT** `--bare`, so hooks + OAuth still work.
- claude-only (codex/agy/opencode/kiro untouched). The headless path (`src/headless.rs`) shares `spawn_flags`, so PTY + headless argv stay consistent.
- No existing `--settings` was passed (clean add, no merge needed).

## Verification
- **confirm-first**: `claude --settings '{"tui":"inline"}' -p ...` accepted the inline JSON flag + completed a real **OAuth** turn (exit 0, empty stderr, no rejection).
- Full **inline-render + hooks-fire** verification needs an **operator daemon restart**.
- Tests: `spawn_flags_claude_always_forces_inline_tui` (new); existing claude `spawn_flags` tests updated for the +1 flag. All 49 build_command/backend/headless/spawn_flags tests pass.
- Gates: fmt, `clippy --all-targets -D warnings`, `cargo-xwin` windows clippy, nextest — all green.

⚠ State-detection-adjacent (spawn argv) → suggest **DUAL** review.

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -789,6 +789,19 @@ impl Backend {
     pub fn spawn_flags(&self, working_dir: &std::path::Path) -> Vec<String> {
         let mut out = Vec::new();
         if matches!(self, Backend::ClaudeCode) {
+            // Force the agent's claude pane to INLINE (classic main-screen) mode,
+            // never fullscreen/alt-screen. Fullscreen hijacks the mouse and has no
+            // scrollback, which breaks agend's embed-and-observe model (pane
+            // selection / scroll / capture). `--settings` with inline JSON MERGES
+            // over the user's `~/.claude/settings.json` — it overrides ONLY `tui`,
+            // leaving hooks / OAuth / model / permissions intact — and is per-spawn,
+            // so the operator's own standalone Claude Code is untouched. NOT `--bare`,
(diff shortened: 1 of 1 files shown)


suzuke:
Closing per operator decision (2026-06-23): Claude Code `tui: fullscreen` is the user's deliberate choice — if they enable it (even under agend), they want that mode, including its native selection behavior. agend should not override a deliberate user backend-setting. The independent fixes (mouse-wheel #2417, zoom #2419) stand and already make fullscreen panes behave correctly; only the selection-style difference remains, which is accepted as the user's choice.

PR #2414, opened 2026-06-22
fix(offthread): read snapshot for text selection so off-thread panes can copy

suzuke:
## Bug (operator-reported, HIGH)
Under `AGEND_OFFTHREAD_PARSE=1`, text copy/selection was broken: copy returned blank/whitespace.

## RCA (confirm-first — disproves the dispatched assumption)
The dispatch assumed "only the lead pane works because it's `offthread:None`". **Live evidence disproves this**: all 15 panes including fixup-lead (pane_id=2, 43k+ `#offthread-snapshot` publishes) are `offthread: Some`. So:
- Both selection paths read `pane.vterm.extract_text()` — mouse copy-on-select (`mouse.rs`) + Cmd+C (`dispatch.rs`).
- Off-thread, `pane.vterm` is **permanently blank**: `drain_output` no-ops when `offthread = Some` (pane.rs), and the render path paints the parser's published `GridSnapshot` directly (`core_render.rs` `snap.render_to_buffer`) — there is **no snapshot→vterm sync**. `viewport_to_logical_line` also used `vterm.max_scroll()` = 0.
- #2411 fixed `scroll_max` (scroll) but not the selection path → selection broke **uniformly for every off-thread pane**.
- The "only lead works" anomaly has **no code-level cause** (all panes are uniformly off-thread); it's being pinned with the operator separately (likely a zoom/terminal-native-selection usage artifact). The app-internal fix here is correct regardless and makes **all** panes selectable.

## Fix (snapshot-based selection; flag-OFF byte-identical)
- `GridSnapshot::extract_text` — off-thread analog of `VTerm::extract_text`, reading the snapshot's `cells` + captured `history` with the SAME `logical → line = logical - history.len()` mapping `render_inner` paints.
- `Pane::extract_selection_text` dispatches: off-thread → snapshot, flag-OFF → live vterm (unchanged).
- `viewport_to_logical_line` / `logical_line_to_viewport` use `scroll_max()` (render-path depth) instead of `vterm.max_scroll()`, so the selection anchor **and** the highlight overlay share the snapshot's history depth off-thread. Flag-OFF: `scroll_max() == vterm.max_scroll()` → byte-identical.
- Both copy paths call the dispatcher.

## Tests
- `vterm::snapshot_extract_text_matches_vterm`: `GridSnapshot::extract_text` == `VTerm::extract_text` across history / visible / partial-column / multi-line ranges (incl. a wide char) — proves the off-thread copy yields the SAME content as the live path.
[...]

PR #2411, opened 2026-06-22
fix(offthread): restore pane scrolling under AGEND_OFFTHREAD_PARSE=1

suzuke:
## Bug (operator-reported, 2026-06-22, HIGH)
With \`AGEND_OFFTHREAD_PARSE=1\`, **all panes could not scroll**. Flag-OFF (default) scrolls fine → an off-thread render-path regression introduced by #2404 (Option X P1, which deferred off-thread scrollback to "P2"). Operator unset the flag to recover.

## Root cause (two-part)
1. **Scroll clamp pinned to 0**: \`scroll_focused\` (app/mod.rs) and the mouse drag-autoscroll (app/mouse.rs) clamp \`scroll_offset\` against \`pane.vterm.max_scroll()\`. In off-thread mode \`pane.vterm\` is IDLE (the main-thread drain no-ops; the parser thread owns the real VTerm), so its \`max_scroll()\` is **0** → \`scroll_offset\` is pinned to 0.
2. **Snapshot had no scrollback + render ignored the offset**: \`GridSnapshot\` captured only the offset-0 visible grid, and \`GridSnapshot::render_inner\` ignored \`scroll_offset\` for content ("P1 snapshot is offset-0 live").

So even with a non-zero offset nothing scrolled, and the offset could never become non-zero anyway.

## Fix
- **\`GridSnapshot\`** gains a SEPARATE \`history: Vec<Cell>\` + \`history_rows\` (the visible \`cells\` layout/contract is **unchanged**). \`VTerm::snapshot()\` captures up to \`SNAPSHOT_SCROLLBACK_ROWS\` (1000) rows of scrollback (alacritty negative \`Line\`s) into it — on the **parser thread**, so it does NOT reintroduce the main-thread parse freeze Option X removed.
- **\`GridSnapshot::render_inner\`** now maps each view row to \`line = row - offset\`: \`line ≥ 0\` reads the visible \`cells\`, \`line < 0\` reads \`history\`, beyond the captured window blanks — mirroring the live path's \`Line(row - offset)\` exactly.
- **Path-aware scroll max**: new \`OffthreadHandle::scroll_max()\` (snapshot \`history_rows\`) + \`Pane::scroll_max()\` (off-thread → handle, else \`vterm.max_scroll()\`); both scroll-clamp sites use it.

Flag-OFF path is **byte-identical/untouched** (\`render_to_buffer_inner\` unchanged; offset-0 snapshot render still equals the live render).

### Deliberate trade
[...]

PR #2419, opened 2026-06-23
fix(offthread): route zoom/layout resize to the parser, not the idle vterm

suzuke:
## Bug (operator-reported, HIGH)
Under `AGEND_OFFTHREAD_PARSE=1`, `Ctrl+B z` zoom rendered off-thread panes half-width + corrupted (left content, right black, edge bleed); idle panes were fine.

## RCA (confirm-first, lead-vetted)
Off-thread the parser thread owns the live VTerm; `pane.vterm` is idle. `resize_panes`/`collect_resize_needs` (layout/mod.rs:316/345) resized `pane.vterm` (a render no-op off-thread) + `resize_pty`, but **never routed the resize to the parser** (`handle.request_resize`). The parser only learned the new size lazily on the next render (`render_pane`, core_render.rs:657). So on zoom the child re-rendered full-width while the parser was still half-width, and the parser parsed that output at the stale dims (its resize even drains pending data at OLD dims first, offthread.rs:267/309) → a half-width snapshot painted into the full zoom area. **Persistent** for alt-screen TUIs (fullscreen claude-code, which redraws once per resize). The left/right asymmetry was **busy-vs-idle** panes (active panes lose the resize-vs-output race). Same `#2404`-batch root as #2411 (scroll) / #2414 (selection) / #2417 (mouse-forward): operate on the parser, not the idle `pane.vterm`.

## Fix
`Pane::apply_geometry_resize(cols, rows)`: off-thread → `handle.request_resize` (queued **before** the `resize_pty` loop, so the parser resizes ahead of the child's re-rendered output), flag-OFF → `vterm.resize` (byte-identical). `resize_panes` + `collect_resize_needs` call it. `render_pane`'s lazy resize stays a deduped backstop.

## Tests
- `apply_geometry_resize_routes_offthread_to_parser`: drives the real parser, applies a resize, asserts the published snapshot resizes (60x18). **Neuter-RED verified**: dropping `request_resize` → the parser never resizes → the test times out.
- `apply_geometry_resize_flag_off_resizes_vterm`: flag-OFF resizes the live vterm.
- All adjacent off-thread/resize tests green; fmt + `clippy --all-targets --features tray -D warnings` clean.

⚠ Render/layout change → needs operator rebuild+restart to go live. base origin/main `ef86fb7c`. correlation t-20260623030613011989-97931-0.

### DIFF
--- a/src/layout/mod.rs
+++ b/src/layout/mod.rs
@@ -313,7 +313,13 @@ pub fn resize_panes(pane_area: Rect, layout: &mut Layout, registry: &crate::agen
             let w = pane_area.width.saturating_sub(2);
             let h = pane_area.height.saturating_sub(2);
             if w > 0 && h > 0 && (w != pane.vterm.cols() || h != pane.vterm.rows()) {
-                pane.vterm.resize(w, h);
+                // #offthread-zoom: route the resize to whichever VTerm actually renders
+                // — off-thread the PARSER (queued BEFORE the `resize_pty` loop below), so
+                // the parser resizes before the child's re-rendered full-width output
+                // arrives instead of parsing it at the stale pre-zoom width (the
+                // half-width corruption). Flag-OFF it is the live main-thread vterm
+                // (byte-identical). render_pane's lazy resize stays a deduped backstop.
+                pane.apply_geometry_resize(w, h);
                 resizes.push((pane.id, w, h));
             }
         }
@@ -342,7 +348,13 @@ fn collect_resize_needs(
             let w = area.width.saturating_sub(2);
             let h = area.height.saturating_sub(2);
             if w > 0 && h > 0 && (w != pane.vterm.cols() || h != pane.vterm.rows()) {
-                pane.vterm.resize(w, h);
+                // #offthread-zoom: route the resize to whichever VTerm actually renders
+                // — off-thread the PARSER (queued BEFORE the `resize_pty` loop below), so
+                // the parser resizes before the child's re-rendered full-width output
+                // arrives instead of parsing it at the stale pre-zoom width (the
+                // half-width corruption). Flag-OFF it is the live main-thread vterm
+                // (byte-identical). render_pane's lazy resize stays a deduped backstop.
+                pane.apply_geometry_resize(w, h);
                 resizes.push((pane.id, w, h));
             }
         }
--- a/src/layout/pane.rs
+++ b/src/layout/pane.rs
@@ -183,11 +183,15 @@ impl Pane {
         }
     }
 
-    /// #offthread-zoom: apply a geometry resize to whichever VTerm actually renders.
-    /// Off-thread the parser thread owns the live VTerm, so route via the handle's
-    /// `request_resize` (caller must do this BEFORE the PTY resize, so the parser's
-    /// resize event is queued ahead of the child's re-rendered output — else the parser
-    /// parses that output at the stale pre-resize dims, the zoom half-width corruption).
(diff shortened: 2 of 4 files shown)


suzuke:
## Rework @2f7a6f2e — addresses r6 REJECT @bd341d05

**r6 was right**: `request_resize` before `resize_pty` only ordered the SEND. The parser's `select!` (data vs resize) is unbiased AND the resize arm `drain_pending_data`s queued data at the OLD dims first — so the child's post-SIGWINCH full-width output could still be parsed at the stale dims. Send-order ≠ process-order.

**Fix (design A, lead-vetted): synchronous resize barrier.**
- resize carries a per-call `bounded(1)` ack; parser does drain(old dims) → `vterm.resize` → `ack.send`.
- `request_resize` blocks on `ack_rx.recv_timeout(100ms)` → caller fires SIGWINCH only AFTER the parser is at the new dims → post-SIGWINCH output parsed at new dims **by construction**.
- **Deadlock-free**: `bounded(1)` ack never blocks the parser; no shared lock; parser-gone → `false` no-op; `recv_timeout` fallback degrades to a brief flicker (not a freeze) if ever wedged.
- **Flood-robust**: resize-priority `try_recv` at the parser loop top (mirrors the existing cancel-priority) so a data flood can't starve a pending resize.

**Finding 2**: render backstop fired `resize_pty` unconditionally every frame → re-SIGWINCH amplified the race. Now gated on `request_resize`'s bool (dims-unchanged dedups to `false` → no send/block/SIGWINCH). SIGWINCH fires ≤ once per real dim change.

**Test**: looped 128-iter barrier regression — deterministic GREEN via the ack happens-before; **neuter** (drop ack-wait → async) is RED at iter 1 (payload parsed at old width). Matches r6's probe methodology.

**Blast radius**: confined to offthread.rs + 2 already-ordered call sites. `Vec<u8>` data channel + flag-OFF `drain_output` untouched (byte-identical flag-OFF). Chose A over B (in-band) precisely to avoid touching `pane.rx` type / forwarder / flag-OFF path / 8 spawn sites.

Gates: fmt clean · tray clippy `-D warnings` clean · offthread(15)+pane(15)+resize(4)+core_render(24)+layout(13) all green.

suzuke:
**§3.12 review mirror — DUAL 2/2 VERIFIED, merging**

Off-thread zoom/layout resize corruption (operator-reported: Ctrl+B z zoom of a busy left pane rendered at half-width + garble). RCA: off-thread resize never routed to the parser — `resize_panes` only resized the idle `pane.vterm` + SIGWINCH'd the child, so the child repainted full-width while the parser was still half-width → parsed at stale dims. (Same #2404 root as scroll/selection/mouse.)

First attempt (send resize before resize_pty) was **REJECTED by r6** — empirically proven the parser's unbiased `select!` could still take the post-SIGWINCH data before the resize. Reworked to a **sync-ack barrier**: `request_resize` blocks on a per-call bounded(1) ack until the parser applies the new dims, *then* the call site fires SIGWINCH → post-SIGWINCH output can't exist before the parser is at new dims (correct by construction).

- **fixup-reviewer-6 (codex)** VERIFIED @2f7a6f2e — independently re-verified its rejected race is closed; deadlock-free; neuter-RED.
- **fixup-reviewer-4 (claude)** VERIFIED @2f7a6f2e — 6-lens: sync-ack correct, deadlock-free strict proof, dedup-early-return = no per-frame stall, Finding-2 backstop deduped, flag-OFF byte-identical.
- CI green.

Squash-merging via --admin. ⚠ needs operator rebuild+restart to go live.

Issue #2413, opened 2026-06-22
Out-of-path API-activity probe to fix false-idle blind spot in pattern-based agent_state


suzuke:
## Problem
`agent_state` is derived by pattern-matching PTY output (`context_source: "pattern"`,
per-backend regex in `backend_profile.rs` / `StatePatterns::for_backend`). This is the
only signal that covers all backends, but it's lossy: an agent that is mid-API-call can
render as `idle`, and an agent stuck on a pre-prompt screen (e.g. a first-run dialog)
also renders as `idle`. We have no ground-truth "is this agent actually talking to the
LLM right now?" signal.

This bit us twice recently: (a) a demo fleet showed agents `idle` while they were
actually stuck at a claude trust dialog; (b) see PoC below — multiple production agents
report `idle` while holding an active connection to the LLM endpoint.

## PoC evidence (live fleet, read-only `lsof`)
Cross-checking the daemon's `agent_state` against each agent PID's ESTABLISHED TCP to a
resolved LLM endpoint:

| agent | backend | daemon agent_state | active LLM conn |
|---|---|---|---|
| fixup-lead | claude | **idle** | **YES → api.anthropic.com** |
| fixup-dev / dev-2 | claude | **idle** | **YES** |
| ai-scout | **agy** | **idle** | **YES → google** |
| fixup-reviewer-4 | claude | tool_use | YES (agree) |
| fixup-reviewer-2 | claude | awaiting_operator | no (agree) |
| claude-280178 / ba5740 / de2eb8 | claude | idle | no (agree, genuinely idle) |

→ The socket signal catches the false-idle case the pattern misses, and works across
backends (claude→anthropic, agy→google) with zero code/daemon interaction.

## Proposed solution — A1: out-of-path socket-activity probe (MVP)
A read-only poller that, every ~2–3s, enumerates each managed agent PID's ESTABLISHED
TCP connections and matches the remote endpoint against a known LLM-endpoint set, deriving
per-agent `last_api_activity_at`, `api_in_flight`, `provider`. **Out-of-path → fail-open**:
if the probe breaks, agents are unaffected (unlike an in-path forward proxy, which is a
fail-closed liveness dependency — deliberately deferred, see Non-goals).

### Design
- Probe: macOS `libproc` (`proc_pidinfo`) / Linux `/proc/<pid>/net/tcp` ∩ `fd/`.
[...]

PR #2426, opened 2026-06-23
feat(monitor): out-of-path lsof API-activity probe for false-idle detection (#2413 Phase 1)

suzuke:
## #2413 Phase 1 — out-of-path lsof API-activity probe (false-idle detection)

Daemon-owned probe that reads the kernel socket table (`lsof`) once per tick to tell whether a *pattern-idle* agent is genuinely idle or mid-LLM-call — **out-of-path, never touches the agent↔LLM connection** (zero in-path risk). This is the A1-first path the operator picked over the in-path A2 proxy.

### What it does
- **`src/api_activity_probe.rs`** (new): one batched `lsof -nP -iTCP -sTCP:ESTABLISHED` per 5s tick + a `ps` PID-tree walk-up to attribute LLM sockets to managed agents (handles codex's node-wrapper → Rust-child). LLM endpoints matched against a **forward-resolved IP table** (`std::net::ToSocketAddrs`, refreshed ~60s) — `lsof -nP` gives IPs not hostnames and reverse-DNS NXDOMAINs for the CDNs. **Graceful degrade** to "any public `:443`" when DNS is stale (never worse than raw-:443). **Self-disables** to a no-op if `lsof` is absent (Windows / no-lsof never breaks the daemon).
- **`AgentCore`**: additive `api_activity: ApiActivity { in_flight, last_active_epoch_ms }`. **Never rewrites `agent_state`.**
- **`list_instances` / `describe_instance`**: surface `api_in_flight` + `last_api_activity_at`, read under the **same `core.lock()`** as `agent_state` so a consumer reconciles atomically (false-idle = `agent_state=="idle" && api_in_flight`).
- Probe thread spawned next to `spawn_monitor_tick` in both daemon + app paths, **independent of the monitor's TUI-subscriber gate** (so it runs headless — `list_instances` always needs it).

### Empirical "ceiling" (macOS, live fleet)
- Idle managed agent holds **ZERO** ESTABLISHED `:443` sockets; a turn opens a socket to the LLM (`api.anthropic.com` 160.79.104.10) and holds it through the turn → **socket-present ⟺ mid-turn**. Clean binary signal for false-idle.
- **Known limitation (accepted):** when an LLM endpoint shares a CDN block with non-LLM telemetry (Statsig/Sentry on Cloudflare), IP-match can't perfectly separate them; it errs **toward "active"** — the safe direction for false-idle.

### Tests / gates
- `cargo fmt` + `cargo clippy --all-targets --features tray -- -D warnings`: clean.
- New unit tests **10/10** pass (parsers + attribution + precise/degraded/depth-cap), fixtures built from real-shaped `lsof -Fpn` / `ps` output.
- Full `cargo test --tests`: 4415 pass; the only 6 failures are `auto_release`/`messaging` integration tests dying on the **agend-git #2158 in-worktree shim** (stray-worktree rejection of their own `/tmp` `git worktree add`). **Verified pre-existing**: identical failures on clean `origin/main` with this change fully stashed. CI on real git (no shim) is authoritative.

### DIFF
--- a/src/agent/mod.rs
+++ b/src/agent/mod.rs
@@ -30,6 +30,27 @@ pub struct AgentCore {
     pub(crate) subscribers: Vec<crossbeam_channel::Sender<Vec<u8>>>,
     pub(crate) state: StateTracker,
     pub(crate) health: HealthTracker,
+    /// #2413 Phase 1: out-of-path API-activity signal, fed by the
+    /// `api_activity_probe` daemon thread (one batched `lsof` per tick). Read
+    /// alongside `state` under this same `core.lock()` so `list_instances` can
+    /// reconcile pattern-state against live LLM-socket activity (false-idle
+    /// detection) WITHOUT clobbering `agent_state`. Default = no signal yet.
+    pub(crate) api_activity: ApiActivity,
+}
+
+/// #2413 Phase 1: out-of-path "is this agent mid-LLM-call?" signal, derived by
+/// the `api_activity_probe` from the kernel socket table (never touches the
+/// agent↔LLM connection). Purely ADDITIVE — it never rewrites `agent_state`; a
+/// consumer derives false-idle as `agent_state == "idle" && in_flight`.
+#[derive(Debug, Clone, Default)]
+pub struct ApiActivity {
+    /// The agent's process tree currently holds ≥1 ESTABLISHED socket to an LLM
+    /// endpoint (precise, via the resolved LLM-IP table) — or, when DNS is
+    /// stale, to any public :443 host (graceful raw-:443 fallback).
+    pub in_flight: bool,
+    /// Epoch-ms of the most recent tick where `in_flight` was observed true.
+    /// `None` = never observed active (or no probe has run yet).
+    pub last_active_epoch_ms: Option<u64>,
 }
 
 /// Handle to interact with an agent.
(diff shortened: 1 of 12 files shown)


suzuke:
## §3.12 review mirror — DUAL (cross-model) VERIFIED + CI green @ 2a77cfc1

**fixup-reviewer-4 (claude, core):** VERIFIED — invariant proven by grep (only writer = probe, only reader = query.rs, **zero `agent_state` refs in the probe module**; false-idle derived consumer-side as `idle && in_flight`, never mutated); lock discipline (state + api_activity destructured under a single `core.lock()` → atomic snapshot); tests neuter-proven non-vacuous (remove walk-up → 2 attribution tests RED, restore → GREEN + byte-identical); clippy `-D warnings` clean.

**fixup-reviewer-6 (codex, adversarial):** VERIFIED — no lock cycle (probe runs its own daemon thread; lsof/ps run outside the registry lock via bounded process-group spawns; a hung lsof stalls only the probe thread, not the daemon/monitor); PID-tree attribution correct + depth-capped; **does NOT rewrite `agent_state` or feed watchdog/idle gates** → false positives are visible telemetry, not behavior-changing. Wrote 3 /tmp adversarial tests demonstrating the documented degraded/stale-table limits.

**Non-blocking (both, tracked):** precise-mode CDN false-negative (Phase-2 in-path proxy resolves by-construction; never worse than current state); degraded-mode over-count = safe direction (doc'd). Local 6 `auto_release`/`messaging` integration failures = known agend-git #2158 in-worktree shim env confound — dev-2 verified **identical failures on clean origin/main**; CI (real git) is authoritative and green.

**CI:** all required green — Check (macos/ubuntu/windows), Coverage, Daemon-boot flake gate (20x), LOC, MSRV 1.88, audit, Signed-off.

Squash-merging via `--admin` (shared-account §3.12).

PR #2433, opened 2026-06-23
feat(shadow): #2413 Shadow Observer — local hook plane + Evidence reducer + quantification

suzuke:
## #2413 Shadow Observer (Phase A local plane + Phase B reducer + quantification)

Out-of-path agent-state observability: fuse claude lifecycle **hooks** + the screen-scrape `agent_state` baseline + the lsof `api_in_flight` liveness signal into one additive `ObservedStatus` — **zero in-path**, **flag-OFF default** (`AGEND_SHADOW_OBSERVER=1`), `agent_state` **never rewritten**.

Operator chose **A**: build the reducer + *quantify* "cleaner than raw screen-scrape", THEN decide on an in-path proxy. This PR delivers that — reducer + the numbers. **HARD-STOP at reducer + quantification** (no in-path proxy work).

### Phase A — local hook plane (earlier, live-verified)
Per-session-token unix-socket server; claude lifecycle hooks → typed `Evidence`; per-agent buffer. Hooks injected per-workspace at spawn (`~/.claude` global untouched). `src/daemon/shadow/{mod.rs,evidence.rs}`, spawn wiring, `forget_agent` despawn hook.

### Phase B — reducer + driver + quantification (this review's focus)
- **reducer.rs** — pure `Evidence → ObservedStatus` over a per-agent `AgentRuntime` (episode/tool-span model). Precedence `RateLimited > WaitingForUser > ToolUse > Responding > Thinking > Idle`; **decay + liveness backstop** so a dropped terminal hook (Stop/PostToolUse) can't phantom-stick Active (reconciled to Idle only when screen-idle ∧ !api_in_flight ∧ sustained silence — time alone never flips). `Thinking` is *derived* (turn open ∧ no tool ∧ no approval ∧ silent), not string-matched.
- **per_tick/shadow_observe.rs** (NEW driver) — flag-gated; snapshots screen(`AgentState`→`ScreenSignal`, exhaustive) + `Liveness`, folds the hook buffer via `shadow::observe()`, writes `AgentCore.observed_status`. Registry lock never held across a core lock. Emits §5 correction telemetry.
- **Additive surface** — `AgentCore.observed_status: Option<ObservedStatus>` (mirrors `api_activity`), serialized in `list_instances` under the SAME `core.lock()` as `agent_state` so a consumer diffs them atomically. Daemon-only → allowlisted out of app mode (the hook server is `run_core`-only).

### Quantification (SHADOW-OBSERVER-QUANT-2413.md — the core deliverable)
[...]

PR #2434, opened 2026-06-23
fix(shadow): #2413 wire Shadow Observer into app mode (was silently dead in the live daemon)

suzuke:
## #2413 Shadow Observer — app-mode wiring fix (follow-up to #2433)

After #2433 merged, operator live-dogfooded (rebuild + restart + `AGEND_SHADOW_OBSERVER=1`) and found **`observed_status` null on every agent** with no shadow pipeline trace in `app.log`.

**RCA:** #2433 gated the whole Shadow Observer to **run_core-only**, but the **live fleet daemon runs `agend-terminal app` (`run_app`), never `run_core`** — so the plane was silently dead in production. This is the `#1720/#685` silent-dead-in-app-mode class, and ironically the exact trap the `recovery_dispatcher` `#1694(a)` comment (8 lines above `APP_TICK_ALLOWLIST`) warns about — walked into in reverse. DUAL/CI missed it because the unit tests call `handler.run()` directly (not the app-mode entry); the quantification passed because it runs `start --foreground` = run_core.

### Fix (symmetric to #1694(a))
- **Remove `"shadow_observe"` from `APP_TICK_ALLOWLIST`** → the per-tick reducer driver runs in app mode (the `#1726` completeness invariant now *enforces* it).
- **`run_app` calls `shadow::start(&home)`** in the owner block (next to `api_activity_probe::spawn`) so the hook-event socket server binds in app mode. Owner-only (`!attached_mode`); flag-gated, **flag-OFF default ⇒ zero behaviour change**; lifecycle mirrors run_core.
- Corrected the rationale comment that had documented the bug as intentional.

### Regression tests (through the REAL app entry, not `handler.run()`)
- `shadow_observe_runs_in_app_mode_2413` — pins the driver IN `app_tick_handlers()` + NOT allowlisted (mirrors `recovery_dispatcher_runs_in_app_mode_1694a`).
- `run_app_wires_shadow_socket_server_2413` — source-pins `shadow::start(&home)` in `run_app`.

### Behaviorally verified in REAL app mode
Debug binary under a sized PTY, isolated home, flag on, live fleet untouched: `shadow event server listening` now appears in `app.log`, the full turn's hooks were ingested (`TurnStarted`/`ToolStarted`/`ToolEnded`/`TurnEnded`), `observed_status` **populated** (was null), corrections logged.

Preflight: macOS fmt + clippy(tray, all-targets) + 54 app/shadow tests; **Windows cargo-xwin clippy green**. DUAL review requested (app-mode wiring + flag-gated state surface).

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -2234,20 +2234,28 @@ mod tests {
     /// #2413 Phase B live-fix companion: the reducer driver is useless without the
     /// hook-event SOCKET SERVER feeding its buffer, and `shadow::start` is the only thing
     /// that binds it. `run_core` already calls it; this source-pins that `run_app` does
-    /// too (it appears ONLY in run_app within this file), so the plane can't be
-    /// half-wired in app mode again (driver runs, but folds an always-empty buffer). The
-    /// behavioural end-to-end (flag-on → observed_status populated) is the operator's live
-    /// dogfood + the daemon-side reduce test; this is the cheap cross-platform guard.
+    /// too, so the plane can't be half-wired in app mode again (driver runs, but folds an
+    /// always-empty buffer). The behavioural end-to-end (flag-on → observed_status
+    /// populated) is the operator's live dogfood; this is the cheap cross-platform guard.
+    ///
+    /// Scans ONLY the production region (before the `#[cfg(test)]` cutoff). This
+    /// assertion's own needle literal lives in the test module below, so a WHOLE-FILE
+    /// substring check would self-match and stay green even if the real `run_app` call
+    /// were deleted — the #2433 vacuous-pin class this very PR fixes (DUAL round-1 caught
+    /// it here). Mirrors `run_app_registers_event_bus_subscribers`. REVERSE-MUTATION
+    /// verified: deleting the real `shadow::start(&home)` call from run_app turns this RED.
     #[test]
     fn run_app_wires_shadow_socket_server_2413() {
-        let file = "src/app/mod.rs";
-        let src = std::fs::read_to_string(file)
-            .or_else(|_| std::fs::read_to_string(format!("agend-terminal/{file}")))
-            .unwrap_or_else(|_| panic!("source must be readable: {file}"));
+        let source = std::fs::read_to_string("src/app/mod.rs")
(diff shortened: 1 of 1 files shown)


suzuke:
## ✅ Cross-model DUAL VERIFIED + CI green @ 22eb581a — merging (§3.12 mirror)

Admin-merging (shared GitHub account; this mirrors the approval per protocol §3.12).

### What this fixes
#2433 merged the Shadow Observer but the whole plane was gated `run_core`-only, while the live fleet daemon runs in **app mode** (`app::run_app`) — so the feature was a silent no-op in production (`observed_status` null everywhere). This is the #1720/#685 silent-dead-in-app-mode class. This PR wires the plane into app mode: `run_app` starts `shadow::start(&home)` (owner-only, flag-gated) and `shadow_observe` is removed from `APP_TICK_ALLOWLIST`.

### Reviewer verdicts (re-DUAL round 2)
- **fixup-reviewer-6 (codex)** — VERIFIED. Round-1 REJECT (vacuous socket-server test) resolved; reverse-mutation independently re-run: deleting the real `run_app` call now makes the test FAIL, restoring passes. Runtime carry confirmed (owner-only start, attached skip, flag-off gated).
- **fixup-reviewer-4 (claude)** — VERIFIED. Production region (lines 1–1938) byte-identical between heads (md5 match) ⇒ round-1 runtime 5-point VERIFIED carries; the test fix is the only delta (+19/−11, all in `mod tests`); reverse-mutation now RED.

### Round-1 → round-2
DUAL round-1 was 2/2 REJECTED — both reviewers independently caught that the new `run_app_wires_shadow_socket_server_2413` test was vacuous (self-referential whole-file scan; its own assertion literal satisfied the match). Fixed by scanning only the production region before the first `#[cfg(test)]` cutoff (mirrors the existing `run_app_registers_event_bus_subscribers` idiom). The regression guard now actually holds.

### CI @ 22eb581a — all green
Check (macos/ubuntu/windows) · Coverage · Daemon-boot flake gate (20x) · LOC · MSRV (1.88) · Signed-off-by · audit — all pass.

### Scope
Flag-OFF default (zero behavior change); additive `observed_status` beside `agent_state`. Requires operator restart with the flag set to go live.

PR #2435, opened 2026-06-24
feat(tui): add Ctrl+B i image paste via clipboard (arboard + png)

justdoit530-hub:
## Summary
- Add Ctrl+B i handling in the TUI to paste an image from the clipboard.
- Read clipboard RGBA data with arboard, encode it as PNG with the png crate, save it under /tmp, and inject an `[AGEND-IMAGE-PASTE: /path]` marker into the agent input stream.
- Document the marker in agent instructions so agents read the saved image directly.

## Usage
1. Copy an image.
2. In the TUI, press Ctrl+B i.
3. The agent receives the saved PNG path and reads it automatically.

## Verification
- cargo fmt --check
- cargo check

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,6 +43,7 @@ dependencies = [
  "memchr",
  "parking_lot",
  "plist",
+ "png 0.17.16",
  "portable-pty",
  "predicates",
  "proptest",
@@ -2079,7 +2080,7 @@ dependencies = [
  "byteorder-lite",
  "moxcms",
  "num-traits",
- "png",
+ "png 0.18.1",
  "tiff",
 ]
 
@@ -2558,7 +2559,7 @@ dependencies = [
  "objc2-core-foundation",
  "objc2-foundation",
  "once_cell",
- "png",
+ "png 0.18.1",
  "thiserror 2.0.18",
  "windows-sys 0.61.2",
 ]
@@ -3178,6 +3179,19 @@ dependencies = [
  "quick-xml",
  "serde",
  "time",
+]
+
+[[package]]
+name = "png"
+version = "0.17.16"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "82151a2fc869e011c153adc57cf2789ccb8d9906ce52c0b39a6b5697749d7526"
+dependencies = [
+ "bitflags 1.3.2",
+ "crc32fast",
+ "fdeflate",
+ "flate2",
+ "miniz_oxide",
 ]
 
 [[package]]
@@ -5107,7 +5121,7 @@ dependencies = [
  "objc2-core-graphics",
  "objc2-foundation",
  "once_cell",
- "png",
+ "png 0.18.1",
  "thiserror 2.0.18",
  "windows-sys 0.61.2",
 ]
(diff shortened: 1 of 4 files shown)


suzuke:
Thanks for the contribution. The direction makes sense for the TUI flow: `Ctrl+B i` reads an image from the clipboard, saves it as a PNG, and injects an `[AGEND-IMAGE-PASTE: ...]` marker so the active agent can read the file directly.

Before this is safe to merge, I think a few changes are needed:

1. Align the direct `png` dependency with the version already in the tree.

   This PR adds `png = "0.17"` in `Cargo.toml`, which makes `Cargo.lock` carry both `png 0.17.16` and the existing `png 0.18.1` used by `image`/`arboard`. Please change the direct dependency to `png = "0.18"` and refresh the lockfile. I checked this locally by changing the dependency to `0.18.1`; the current encoder API used here (`Encoder`, `ColorType::Rgba`, `BitDepth::Eight`, `write_image_data`) still compiles.

2. Please add coverage for the PNG-writing helper.

   `save_rgba_as_png` is deterministic and easy to test without touching the OS clipboard. A good unit test would write a tiny RGBA fixture, then decode or inspect the resulting PNG to verify at least the PNG signature, dimensions, and RGBA contents. This would catch width/height/channel-order regressions without needing a real clipboard in CI.

3. Please get CI running on all three supported platforms before merge.

   This path is platform-sensitive: `arboard::Clipboard::get_image()` uses different native clipboard backends on macOS, Windows, and Linux, and the injected marker includes a temp-file path that will be Windows-shaped on Windows. The branch currently has no reported PR checks. At minimum we need the normal macOS, Windows, and Linux checks green for this PR. Windows is especially important here because the feature crosses clipboard handling, path formatting, PNG encoding, and PTY input injection.

4. Please add a cleanup strategy for pasted image files, or document the intended lifecycle.

   The current implementation writes `agend_paste_<timestamp>.png` into `std::env::temp_dir()` and never removes it. That is probably acceptable for an initial implementation if the file must remain available after injection, but it should be bounded somehow: for example, write into an AgEnD-specific temp subdirectory and prune old `agend_paste_*.png` files by age/count, or document why OS temp cleanup is the intended lifecycle.

5. Please document the keybinding in the TUI help/status surface.

   `Ctrl+B i` used to pass through as `Ctrl+B` plus `i`; this PR now consumes it as a TUI command. That behavior change is fine, but users should be able to discover it alongside the other `Ctrl+B` commands.

A couple of notes from review:

- The marker path is generated by the TUI, not parsed from untrusted agent text, so the basic marker shape looks reasonable.
- `std::env::temp_dir()` itself is cross-platform, but the end-to-end behavior still needs the PR's own CI because the clipboard/image path is OS-specific.
- The public instructions added for `[AGEND-IMAGE-PASTE]` should avoid hard-coding `/tmp` semantics; the actual path can be a Windows temp path.

Local checks I ran on the current PR head `10e89535704c98b43c29c6e3cd1444e138e91cd4`:

- `cargo fmt --check` passed.
- `cargo check` passed with the current code, but downloaded/built `png 0.17.16`, confirming the duplicate direct dependency.
- In a temp copy, changing the dependency to `png = "0.18"` and updating the lockfile compiled with `cargo check`.
- `cargo test save_rgba_as_png` found no matching tests, which matches the current lack of helper coverage.

PR #2438, opened 2026-06-24
feat(recovery): respawn-stuck watchdog — auto-Fresh restart of a hung Resume (#t-777-3)

suzuke:
## Problem (RCA: task t-…777-3)
A broad `pkill` killed codex agents mid-session-write; the daemon re-spawned them via the **Resume** path (`daemon/mod.rs:1850` `downgraded_for` / app session-restore `session.rs:238`). `resume --last` on the half-written (corrupt) session **hung**, and the agent sat in `AgentState::Restarting` **~45 min** with **no watchdog** — the Hung recovery ladder keys only on `HealthState::Hung`, never on a respawn that is itself stuck. Manual `restart_instance(fresh)` was the only fix. (Codex `has_resumable_session()` always returns `true`, so the corrupt Resume is never pre-downgraded to Fresh.)

## Fix — new per-tick `RespawnWatchdogHandler`
- **Detect:** `spawn_mode==Resume && state ∈ {Starting,Restarting} && in-state >60s && no output >60s`. **Resume-only** is the load-bearing false-kill guard (a slow *Fresh* boot is never force-restarted) and the structural loop-breaker (a Fresh respawn is never re-detected).
- **Recover (B):** auto-Fresh via the **proven API path** — `restart_instance_autonomic` → direct `DELETE`+`SPAWN` → `ApiEvent::InstanceCreated` → fresh pane. Works in **both** `run_core` and the **live app-mode daemon** (`run_app`), where the `crash_tx`→respawn machinery is inert (`pane_factory` spawns with `crash_tx: None`) — which is why recovery drives via `api::call`, not an `AgentExitEvent`.
- **Bound + escalate (A):** a **handler-local** retry map (survives the `DELETE`+`SPAWN` that resets a handle-local counter) caps at **3** attempts/stability-window; on exhaustion → `enter_paused` + operator **P0 page** (terminal). Anti-thrash cooldown covers the async restart window. "Auto when possible, page when auto fails."
- **Disjoint** from the Hung ladder (keys on `AgentState`+`SpawnMode`, skips `Paused`) → no double-fire. **ON by default** (`AGEND_RESPAWN_WATCHDOG=0` kill-switch).

## Enabling change
Stamp `SpawnMode` onto `AgentHandle` so the watchdog can tell a hung Resume from a slow Fresh boot (not previously recorded).

## operator authority
[...]

PR #2437, opened 2026-06-24
feat(shadow): #2413 Phase D — codex rollout-tail observer source (Stream plane) + reducer {Hook|Stream} generalization

suzuke:
## #2413 Phase D — extend the Shadow Observer to codex agents

Reuse the Evidence schema + reducer; swap the observer SOURCE from claude's hook-socket to a strictly **read-only tail** of codex's own `~/.codex/sessions/.../rollout-*.jsonl`. flag-OFF default (`AGEND_SHADOW_OBSERVER`) ⇒ zero behaviour change; additive (never rewrites `agent_state`); never writes `~/.codex`, never injects — codex produces the file itself (the cleanest plane yet). **HARD-STOP at observer source + quantification** (no in-path proxy).

### Confirm-first (the hard gate) — PASSED
An isolated codex agent's rollout grew **18→43 lines during a 22s turn** — `function_call`/`response_item` records mid-turn, last record not `task_complete`. Codex live-flushes real-time state (not an end-of-turn batch).

### Observer source (`src/daemon/shadow/rollout.rs`)
Pure parser `record_to_evidence` (`task_started`→TurnStarted, `function_call{name}`→ToolStarted, `function_call_output`→ToolEnded, `message/assistant`+`agent_message`→Responding, `task_complete`→TurnEnded, `token_count`→TokenUsage + `rate_limits`→**RateLimited** — the bonus claude hooks lack) + `session_cwd` (normalizes macOS `/private/tmp`↔`/tmp`) + `agent_for_cwd` (fleet codex workspaces only) + a fire-and-forget tailer (discover today's rollouts, per-file byte-offset cursor, drain complete lines → `Evidence(Stream)` → the shared buffer the reducer consumes). Cross-platform `std::fs`. `Evidence::stream` added; `push`/`BUFFER_CAP` un-gated (cross-platform now). `rollout::spawn` wired into **both** `run_core` AND `run_app` owner block (the #2434 lesson — the live fleet daemon is app mode).

### Reducer generalization (lead-approved, surgical)
[...]

PR #2440, opened 2026-06-24
feat(shadow): #2413 opencode plane — SSE /event observer source (Stream)

suzuke:
## #2413 opencode Shadow Observer plane — SSE `/event` observer source (Stream)

Extends the #2413 Shadow Observer to a second backend, **opencode** (after codex Phase D / #2437). Reuses the framework unchanged (Evidence schema + `{Hook|Stream}` reducer + per-agent buffer); only the **observer source** is new.

### Confirm-first (HARD gate — PASS)
opencode is client-server: its native TUI **embeds an HTTP server**. Verified empirically (isolated, live fleet untouched — `SHADOW-OBSERVER-OPENCODE-SPIKE.md`):
1. TUI with `--port N` → embedded server reachable on N.
2. A **second** client subscribes `/event` (SSE) **concurrently while the TUI runs**.
3. A real turn streams native lifecycle: `session.status{busy} → … → session.idle` (118 frames, 4× session.status). Native idle/working flags — not screen-scrape.

### Design
- When `AGEND_SHADOW_OBSERVER=1`, `build_command` injects `--port N` (OS-allocated free port) into the opencode launch so the embedded server is reachable on a known port (same site as the existing opencode `XDG_DATA_HOME` / `OPENCODE_CONFIG_CONTENT` injections).
- New `src/daemon/shadow/opencode.rs`: a fire-and-forget supervisor maintains one raw-`TcpStream` SSE subscription per live opencode agent → maps `/event` frames → `Evidence(authority=Stream)` → `super::push` (the SAME buffer the reducer consumes). **Reducer + Evidence schema UNCHANGED** (`Authority::Stream` already exists from #2437).
- Wired into **both** run_core and run_app (#2434 lesson) + a run_app wiring invariant test.

### Spawn-safety (load-bearing)
- **flag-OFF ⇒ byte-identical** opencode launch (no alloc, no `--port`). Pinned by `should_inject` regression test (reverse-mutation verified).
- alloc-fail → skip injection (agent spawns un-observed, never broken).
- **TOCTOU**: opencode `--port N` on a taken port HARD-FAILS exit 1 (no fallback — empirically verified). The ~1s alloc→bind window is a **<0.01%/spawn, flag-ON opt-in** race that **self-heals**: daemon crash-respawns (Fresh) → new port → succeeds (#2438 watchdog backstops). Documented; PID-scan discovery deliberately rejected (Windows-painful, #2437).

### Tests / CI parity
- Pure event→Evidence mapping + pure chunked/SSE decoder (incl. split-across-feeds + trailing-CRLF-wait boundary) + `should_inject` gate + port-registry roundtrip — all unit-tested.
- **3 reverse-mutations confirmed RED**: flag-OFF gate, SSE trailing-CRLF guard, run_app wiring.
- ✅ fmt; clippy `--features tray --all-targets -D warnings`; full `nextest --bin`; spawn-rationale §10.5 audit. ✅ Windows: `cargo xwin clippy --target x86_64-pc-windows-msvc` clean (port/socket/SSE are std, cross-platform).

Needs operator rebuild+restart to validate live (flag-ON, daemon-behaviour).

### DIFF
--- a/SHADOW-OBSERVER-OPENCODE-SPIKE.md
+++ b/SHADOW-OBSERVER-OPENCODE-SPIKE.md
@@ -1,0 +1,83 @@
+# #2413 Shadow Observer — opencode plane: confirm-first SPIKE (2026-06-24)
+
+Task t-20260623214003295940-24134-6. HARD confirm-first: **can our observer
+tap a structured during-TUI signal from opencode while its native TUI runs?**
+
+## VERDICT: ✅ FEASIBLE — opencode is the *strongest* during-TUI structured source so far.
+
+opencode is client-server. The **TUI embeds an HTTP server** (same binary/event-bus
+as `opencode serve`). A second, external client can subscribe to `/event` (SSE)
+concurrently with the TUI, and the bus carries **native session lifecycle**:
+`session.status {busy|idle|retry|...}` + `session.idle` + tool/step lifecycle.
+This is far cleaner than screen-scrape (explicit status flags, not regex-on-grid).
+
+## Evidence (all run isolated on custom ports — live fleet untouched)
+
+### 1. TUI embeds a reachable server (PTY test, `tap_tui.py`)
+- `opencode [project]` (the default TUI command — the way agend launches it via
+  `backend.rs:572 command:"opencode"`) accepts `--port`/`--hostname`.
+- Launched `opencode --port 7655 …` in a real PTY → `GET /global/health` on :7655 → **reachable=True**.
+
+### 2. Concurrent external `/event` tap during the TUI (`tap_tui.py`)
+- While the TUI ran, an external client subscribed to `/event` → received
+  `{"type":"server.connected",...}`, and `GET /api/session` showed the TUI's 2 live
+  sessions. → **multi-subscriber tap works; observer ⟂ TUI client.**
+
+### 3. Full real-turn lifecycle (race-free via `serve` + `run --attach`, `attach_capture.py`)
+Model `opencode/north-mini-code-free` (free), prompt "Reply … pong". 118 frames in ~3s:
+```
+server.connected → session.created → session.next.{agent,model}.switched
+→ session.status {type:"busy"}        ← NATIVE working flag (4× transitions)
+→ message.part.delta ×93              ← streamed responding
+→ session.idle                        ← NATIVE idle flag
+```
+Histogram: server.connected1 session.created1 session.updated3 session.next.agent.switched1
+session.next.model.switched1 message.updated5 message.part.updated7 **session.status4**
+session.diff1 message.part.delta93 **session.idle1**.
+
+### 4. Event schema (from `GET /doc` OpenAPI, 283KB)
+`/event` is SSE of tagged-union `Event` `{id,type,properties}`. Relevant members
+map cleanly to our `EvidenceKind` (authority=Stream):
+| opencode event (`type`) | → EvidenceKind |
+|---|---|
+| `session.next.prompted` / `session.next.step.started` | `TurnStarted` |
+| `session.next.tool.called` (has `tool`,`callID`,`input`) | `ToolStarted{name}` |
+| `session.next.tool.success` / `.tool.failed` | `ToolEnded` |
+| `session.next.text.*` / `message.part.delta` | `Responding` |
+| `session.next.reasoning.started` | `Responding` (reducer refines→Thinking) |
+| `session.status {busy}` | active (working) |
+| `session.status {retry}` (has attempt/next) | `RateLimited{retry_at_ms}` |
+| `session.idle` | `TurnEnded` |
+| `permission.asked` / `.replied` | `ApprovalRequired` |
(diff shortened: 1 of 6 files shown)


suzuke:
REJECTED

I found a lifecycle race in the new opencode SSE subscriber. `supervise_once` stops a subscriber on despawn/port change by setting its stop flag and dropping the `Sub` (`src/daemon/shadow/opencode.rs:391-395`), but the subscriber only checks that flag before `stream.read()` (`src/daemon/shadow/opencode.rs:292-296`). If teardown/port-change happens while a read is blocked, and the old opencode server delivers headers plus a final SSE frame in that same read, the stopped subscriber still decodes and pushes that frame (`src/daemon/shadow/opencode.rs:298-311`) under the same agent name. That can repopulate stream evidence after `forget_agent` cleanup or after a same-name respawn has moved to a new port.

I confirmed this with a temporary regression test in a tarball copy: start `subscribe_once`, accept the request, set `stop=true`, then write response headers plus `data: {"type":"session.next.prompted"}` in one socket write. The expected empty buffer assertion fails because the stopped subscriber pushes stale `TurnStarted` evidence:

```text
cargo test stopped_subscriber_does_not_push_late_frames_after_stop
FAILED: stopped subscriber pushed stale evidence after teardown: [Evidence { kind: TurnStarted, authority: Stream, confidence: Strong, ... }]
```

The fix should prevent any frame from being decoded/pushed after stop is observed during teardown, for example by re-checking `stop` after every successful read and before feeding/pushing body bytes, plus adding a lifecycle regression around stopped subscribers.

### Evidence
ran: `cargo test opencode` on `411716186dfbe35311fc4d6f8b121695761b7fc2` -> passed, 54 passed / 1 ignored
ran: `cargo fmt --check` on `411716186dfbe35311fc4d6f8b121695761b7fc2` -> passed
ran: `cargo test --test spawn_rationale_audit` on `411716186dfbe35311fc4d6f8b121695761b7fc2` -> passed, 2 passed
ran: temporary regression `cargo test stopped_subscriber_does_not_push_late_frames_after_stop` in `/tmp/agend-pr-2440-late` -> failed with stale `TurnStarted` evidence after stop
cited: `src/daemon/shadow/opencode.rs:292-311` — stop is checked before read, then a successful read is decoded and pushed without another stop check
cited: `src/daemon/shadow/opencode.rs:391-395` — teardown/port-change only sets `sub.stop=true` and drops the subscription entry

suzuke:
## §3.12 Cross-model DUAL review mirror — merge gate open

**Head reviewed:** `38eaaa615c0f326a35f4af89bc806f32df8b7207`
**Verdict:** ✅ 2/2 VERIFIED + CI all-green → self-merge gate open.

### Review journey
1. **Initial DUAL @411716186**: fixup-reviewer-4 (claude, core) VERIFIED; fixup-reviewer-6 (codex, adversarial) **REJECTED** — found a real, test-proven SSE subscriber teardown race: `supervise_once` set `stop=true` on despawn/port-change but the subscriber only checked `stop` *before* `read()`; if teardown landed while a read was blocked and the old server returned headers+a final SSE frame in the same read, the stopped subscriber still decoded+pushed it → stale phantom `TurnStarted` re-injected after `forget_agent` / same-name respawn.
2. **Rework → head 38eaaa61** (opencode.rs only, +72/-0): `subscribe_once` re-checks `stop` immediately after a successful `read()`, before any decode/push (`opencode.rs:303`); both `push_frame` sites are after the check. Added lifecycle regression `stopped_subscriber_does_not_push_late_frames_after_stop`.
3. **Re-DUAL @38eaaa61** (r6 was usage_limited → in-team substitute for the adversarial lens):

### Verdicts
- **fixup-reviewer-4 (claude) — VERIFIED (carry + delta).** NEUTER both directions: removing the post-read check → race test RED @:714 (fix load-bearing); happy-path tests still GREEN (normal frames not swallowed). All other surface (flag-OFF byte-identity, event→Evidence mapping, reducer UNCHANGED, app-entry pin, `--port` inject) unchanged → carried.
- **fixup-reviewer-2 (claude) — VERIFIED (adversarial re-check, objective reverse-mutation).** GREEN with fix → 1 passed; removed post-read check → 1 FAILED with `opencode.rs:706: a stopped subscriber must NOT push a frame received after stop, got [Evidence { kind: TurnStarted, … }]`; restored → passed ⇒ test non-vacuous, truly pins the race. Confirmed window is *closed* not moved (partial frames stay in `SseDecoder` and are discarded on stop-read; `SseDecoder::feed` is a pure byte→frame state machine), and all teardown paths covered (despawn + port-change both `stop.store(true)`; respawn uses a fresh `stop` Arc → no cross-contamination).

### CI @38eaaa61
All green: Check (ubuntu/macos/windows), Coverage, Daemon-boot flake gate (20x), MSRV (1.88), LOC, Signed-off-by, audit.

Non-blocking residual (documented, out-of-scope this round): r4-noted alloc-vs-bind TOCTOU (`opencode.rs:93-98`) — flag-ON opt-in only, <0.01%/spawn, self-heals via #2438 crash-respawn; default flag-OFF = zero risk.

PR #2480, opened 2026-06-26
feat(reliability): escalate crash-respawn failures & enforce process liveness gate

suzuke:
This PR implements two key reliability improvements from the fork:

1. **Escalate crash-respawn failures (#11)**: Logs to event_log, notifies all escalation channels, and transitions health to `Failed` when a respawn attempt fails.
2. **Enforce process liveness gate on decay transitions and respawn_ok (#12)**: Prevents ghost processes by ensuring a process is actually alive before marking it healthy.

Both commits have been verified with `cargo check`.


suzuke:
Review pass (non-blocking note; GitHub will not allow this account to formally approve its own PR).

I reviewed the updated PR head after rebasing/merging current `main` into the branch. No blocking findings:

- respawn failure now logs `crash_respawn_failed`, notifies escalation channels, and marks the agent health as `Failed`;
- `respawn_ok` and decay transitions now require a live process before returning to `Healthy` / `Recovering`;
- stale/dead processes decay to `Absent` instead of being ghost-marked healthy;
- coverage includes the failure branch and dead-process decay path.

Next: wait for the refreshed CI matrix from the branch update to finish, then merge if green.

Issue #2499, opened 2026-06-27
bug: spurious binding_out_of_dispatch warning during worktree-align for already dispatched branch


cheerc:
## Problem
When a task is dispatched to an agent, the daemon auto-binds the worktree and branch with a `task_id` (e.g. `t-1234`). During the subsequent execution, the agent runs `worktree-align` which calls `repo checkout bind:true` for the exact same branch.
This triggers `binding_out_of_dispatch` warning notifications to the team orchestrator, even though the branch is already authorized by the daemon. Additionally, it overwrites `task_id` with an empty string `""` in `binding.json`, which prevents `auto_release` from releasing the worktree when the task is completed (since `is_dispatch_lease` relies on a non-empty `task_id`).

## Root Cause
1. **Worktree Path Mismatch**:
   - Daemon dispatch leases the worktree at `$AGEND_HOME/worktrees/<agent>/<branch>/` (see [src/worktree.rs:139](file:///Users/cheerc/agend-terminal/src/worktree.rs#L139)).
   - `repo checkout` resolves its directory to `$AGEND_HOME/worktrees/<agent>-<repo_slug>/` (see [src/mcp/handlers/ci/checkout.rs:76](file:///Users/cheerc/agend-terminal/src/mcp/handlers/ci/checkout.rs#L76)).
   Because the worktree paths differ, the idempotent check in `checkout.rs` fails, and it invokes `bind_full` with `is_self_claim = true` and `task_id = ""`.
2. **Missing `task_id` in `repo` tool**:
   The `repo checkout` MCP tool does not accept or propagate a `task_id` to `bind_full` (see [src/mcp/tools.rs:374](file:///Users/cheerc/agend-terminal/src/mcp/tools.rs#L374)).

## Suggested Solution
We can implement a safe carve-out in [src/binding.rs:283-292](file:///Users/cheerc/agend-terminal/src/binding.rs#L283-L292) (`bind_full`) to:
1. Preserve the existing `task_id` if we are self-claiming the same branch.
2. Suppress the `binding_out_of_dispatch` notification in that case.

```rust
    let mut final_task_id = task_id.to_string();
    let mut should_notify = is_self_claim;

    if is_self_claim {
        if let Some(ref ex) = existing {
            let ex_branch = ex.get("branch").and_then(|v| v.as_str()).unwrap_or_default();
            let ex_task_id = ex.get("task_id").and_then(|v| v.as_str()).unwrap_or_default();
[...]

PR #2500, opened 2026-06-28
Fail closed on stale dispatch checkout binding

suzuke:
## Summary
- Treat same-branch dispatch bindings with a missing worktree and non-empty task_id as stale_binding during repo checkout bind:true.
- Fail closed before repo checkout can rebind with an empty task_id, preserving auto-release provenance.
- Add regression coverage for #2499 stale-binding hardening while keeping the normal live same-branch idempotent path intact.

## Verification
- cargo fmt --check
- git diff --check
- cargo test -q -p agend-terminal checkout_bind_true_rejects_stale_dispatch_binding_2499
- cargo test -q -p agend-terminal checkout_bind_true_idempotent_when_already_bound_1494

Review: opencode-reviewer-2499 returned VERIFIED with evidence from the same two targeted tests and the checkout guard placement.

Fixes the narrow stale-binding hardening case from #2499.

### DIFF
--- a/src/mcp/handlers/ci/checkout.rs
+++ b/src/mcp/handlers/ci/checkout.rs
@@ -202,6 +202,19 @@ fn handle_checkout_repo_inner(home: &Path, args: &Value, instance_name: &str) ->
                         "fetch_attempted": fetch_attempted,
                     });
                 }
+                let existing_task_id = existing
+                    .get("task_id")
+                    .and_then(|v| v.as_str())
+                    .unwrap_or_default();
+                if !existing_task_id.is_empty() {
+                    return json!({
+                        "error": format!(
+                            "stale binding for branch '{branch}' points at a missing worktree - release first before checkout"
+                        ),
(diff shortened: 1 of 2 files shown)


suzuke:
VERIFIED

Reviewed by `opencode-reviewer-2499`.

### Evidence
ran: `git diff -- src/mcp/handlers/ci/checkout.rs src/mcp/handlers/ci/tests.rs` -> guard is inside same-branch handling after the live-worktree idempotent return, so live same-branch checkout remains unchanged while stale non-empty task_id bindings fail closed.
ran: `cargo test -q -p agend-terminal checkout_bind_true_rejects_stale_dispatch_binding_2499` -> 1 passed.
ran: `cargo test -q -p agend-terminal checkout_bind_true_idempotent_when_already_bound_1494` -> 1 passed.

Conclusion: patch is minimal, preserves the live idempotent path, and prevents stale dispatch bindings from being rebound with an empty task_id.

PR #2523, opened 2026-07-02
fix(worktree): safe same-agent rebind repair instead of destructive fallback (#2496)

### LINKED ISSUE #2496: worktree: stale binding blocks branch rebinding after completed task
## Problem

While starting the next #2454 slice, the agent's daemon binding state was stale
and blocked normal rebinding even though the on-disk worktree was clean and
already switched to the intended branch.

Observed on 2026-06-27 with agent `codex-f64520`:

- Existing worktree path:
  `/Users/suzuke/.agend-terminal/worktrees/codex-f64520-_Users_suzuke_Documents_Hack_agend-terminal`
- On disk, `git status --short --branch` showed the worktree on
  `fix/mcp-api-loopback-first-slice`, then it cleanly switched to
  `fix/mcp-api-loopback-second-slice`.
- `repo action=checkout bind:true` for `fix/mcp-api-loopback-second-slice`
  failed because the managed worktree directory already existed.
- `binding_state(agent=codex-f64520)` still reported branch
  `fix/provider-detect-hermetic-toml-quote`, unrelated to both the just-finished
  task and the new task.
- `bind_self(branch=fix/mcp-api-loopback-second-slice)` failed with
  `cross_agent_conflict`: agent already bound to stale branch
  `fix/provider-detect-hermetic-toml-quote` and required `release_worktree`
  first.
- `release_worktree dry_run=true` reported it would remove the current worktree
  directory and would delete branch `fix/provider-detect-hermetic-toml-quote`,
  which is too broad/risky as an incidental step during a task transition.

## Why this matters

The desired workflow is:

1. finish PR,
2. clean worktree,
3. start next branch in the same managed worktree or atomically rebind metadata.

Instead, stale binding metadata can force agents into a confusing choice between
continuing with mismatched binding state or running a release operation that may
delete unrelated stale branch state.

## Expected behavior

Add a safe recovery/rebind path for the same agent when:

- the existing managed worktree path is present,
- the worktree is clean,
- the requested branch is checked out or can be checked out safely,
- the stale binding belongs to the same agent,
- no CI watch or active task depends on the stale binding.

Possible options:

1. `bind_self(rebase_mode=true)` or a new action should update binding metadata
   in place without deleting the worktree/branch when the worktree is clean and
   same-agent-owned.
2. `repo checkout bind:true` should detect same-agent existing managed worktree
   and offer/perform safe metadata repair instead of failing at `worktree_add`.
3. `release_worktree` should expose a metadata-only stale-binding repair path,
   separate from branch/worktree deletion.

## Acceptance criteria

- Same-agent stale binding can be repaired without deleting a clean existing
  managed worktree.
- Repair is fail-closed if the worktree is dirty, held by another agent, has
  active CI watches, or has an active task tied to the stale branch.
- The command result clearly reports whether it changed only metadata, switched
  branch, removed a worktree, or deleted a branch.
- Regression test covers stale binding branch != actual clean worktree branch
  for same agent.

suzuke:
## Summary
Fixes #2496: a stale daemon binding blocked normal branch rebinding after a completed task, even when the on-disk worktree was clean and already switched to the intended branch.

Two root causes:
- `bind_full`'s guard-b (`src/binding.rs`) rejects a live-worktree cross-branch rebind using only the STALE recorded branch, never re-deriving the worktree's actual current git branch — a worktree cleanly `git checkout`'d to the requested branch out-of-band (bypassing bind/release) is wrongly treated as a dangerous cross-branch bind-over.
- The only "recovery" path (`bind_self(rebase_mode=true)` → `rebase_clean_self`) computed its target via the superseded `worktrees/<agent>/<branch>/` layout, which never matches the flat `worktrees/<agent>-<source>/` scheme `repo action=checkout bind:true` actually uses — so it unconditionally fell through to a full `worktree_pool::release_full`, making `rebase_mode` exactly as destructive as `release_worktree` and defeating its whole purpose.

**Piece 1** — `bind_full`'s guard-b gets a strict, same-agent-only exception: allows metadata-only catch-up ONLY when the worktree is a valid git repo, daemon-managed with a matching (or absent) `.agend-managed` marker agent, clean, and its ACTUAL current branch already equals the requested one, with no other agent holding the same `(source_repo, branch)` and no active CI watch or task still tied to the branch being abandoned.

**Piece 2** — `bind_self(rebase_mode=true)` now tries `attempt_safe_rebind_repair` FIRST: reads the agent's real bound worktree via `binding::read` (not a path formula) and either finds it already on the requested branch (metadata-only), in-place `git switch`es a clean worktree on a different branch (never `-c` — a repair path must never create a branch), or fails closed with a specific reason — never silently falling through to a destructive release anymore. When there's no live worktree at all (the original incident-recovery case), it still safely falls back to the legacy `rebase_clean_self` cleanup, since nothing live is at risk there.

`repair.rs` is split out of `force_release/mod.rs` to stay under the `src/mcp/handlers` 750-LOC file-size invariant (same reason `gc.rs` is its own file).

Design reached via adversarial consensus with codex-reviewer.

## Test plan
- [x] `cargo test --bin agend-terminal` — 4626 passed, 0 failed
- [x] `cargo test --test src_file_size_invariant --test file_size_invariant` — clean
- [x] New tests cover all 8 consensus scenarios: metadata catch-up allowed/dirty/marker-mismatch/other-agent-held/active-CI-watch/active-task rejections (binding.rs), plus real-worktree-via-binding metadata-only and in-place branch switch (repair.rs)
- [x] All pre-existing guard-b and `rebase_clean_self`/`bind_self(rebase_mode)` tests still pass unchanged

### DIFF
--- a/src/binding.rs
+++ b/src/binding.rs
@@ -131,6 +131,122 @@ pub fn bind(home: &Path, agent: &str, task_id: &str, branch: &str) {
     }
 }
 
+/// #2496: strict, same-agent-only exception to guard-b (see call site). ALL
+/// conditions below must hold for a metadata-only catch-up to be allowed —
+/// any failure keeps guard-b's existing reject.
+fn same_agent_metadata_catchup_allowed(
+    home: &Path,
+    agent: &str,
+    worktree: &Path,
+    source_repo: &Path,
+    branch: &str,
+    ex_branch: &str,
+) -> bool {
+    if !crate::worktree::is_git_repo(worktree) {
+        return false;
+    }
+    if !crate::worktree_pool::is_daemon_managed(worktree) {
+        return false;
+    }
+    // Marker present (checked above) but its recorded `agent=` doesn't match
+    // the caller: this worktree is daemon-managed for a DIFFERENT agent —
+    // reject rather than treat as this agent's own stale metadata.
+    if let Some(marker_agent) = managed_marker_agent(worktree) {
+        if marker_agent != agent {
+            return false;
+        }
+    }
+    if crate::worktree::has_uncommitted_changes(worktree) {
+        return false;
+    }
+    let actual_branch =
+        crate::git_helpers::git_cmd(worktree, &["branch", "--show-current"]).unwrap_or_default();
+    if actual_branch != branch {
+        return false;
+    }
+    let source_repo_str = source_repo.display().to_string();
+    if scan_existing_branch_binding(home, &source_repo_str, branch, agent).is_some() {
+        return false;
+    }
+    if agent_has_active_ci_watch_on_branch(home, agent, ex_branch) {
+        return false;
+    }
+    if branch_has_active_task(home, ex_branch) {
+        return false;
+    }
+    true
+}
+
+/// Parse a daemon-managed worktree's `.agend-managed` marker for its recorded
+/// `agent=` line. `None` when the marker is missing or the line isn't present
+/// — callers that require ownership certainty must already have checked
+/// `worktree_pool::is_daemon_managed` before this matters.
+///
+/// `pub(crate)` — also used by `mcp::handlers::force_release`'s #2496 safe
+/// rebind-repair path (same ownership check, different flow control).
+pub(crate) fn managed_marker_agent(worktree: &Path) -> Option<String> {
+    let content =
+        std::fs::read_to_string(worktree.join(crate::worktree_pool::MANAGED_MARKER)).ok()?;
+    content
+        .lines()
+        .find_map(|l| l.strip_prefix("agent="))
+        .map(|s| s.trim().to_string())
+}
+
+/// #2496: does `agent` hold an active CI watch on `branch`? A stale branch
+/// still being watched must not be silently abandoned by a metadata-only
+/// binding repair. `pub(crate)` — shared with `force_release`'s repair path.
+pub(crate) fn agent_has_active_ci_watch_on_branch(home: &Path, agent: &str, branch: &str) -> bool {
+    let ci_dir = crate::daemon::ci_watch::ci_watches_dir(home);
+    let Ok(entries) = std::fs::read_dir(&ci_dir) else {
+        return false;
+    };
+    for entry in entries.flatten() {
+        let path = entry.path();
+        if path.extension().and_then(|s| s.to_str()) != Some("json") {
+            continue;
+        }
+        let Ok(content) = std::fs::read_to_string(&path) else {
+            continue;
+        };
+        let Ok(watch) = serde_json::from_str::<serde_json::Value>(&content) else {
+            continue;
+        };
+        if watch["branch"].as_str() != Some(branch) {
+            continue;
+        }
+        if crate::daemon::ci_watch::parse_subscribers(&watch)
+            .iter()
+            .any(|s| s == agent)
+        {
+            return true;
+        }
+    }
+    false
+}
+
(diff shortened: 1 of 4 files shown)


suzuke:
REJECTED

Reviewed head: 856b9b5c640361e9d0902c2925450e5b1e7beedf

### Evidence
ran: `cargo test --bin agend-terminal 2496` → 8 passed; existing tests cover nominal #2496 cases but not the double-stale ordering below.

ran: `gh pr checks 2523 --repo suzuke/agend-terminal` → exit 8 because platform/Coverage checks are still pending.

cited: `src/mcp/handlers/worktree.rs:93-116` — `bind_self(rebase_mode=true)` calls repair before binding.

cited: `src/mcp/handlers/force_release/repair.rs:163-181` — repair checks only actual branch dependents, then runs `git switch`.

cited: `src/binding.rs:171-175` and `src/binding.rs:326-353` — guard-b later checks recorded stale binding branch dependents and may reject after the switch.

cited: `src/mcp/handlers/force_release/repair.rs:134-140`, `src/mcp/handlers/force_release/mod.rs:169-212`, `src/worktree_pool.rs:528-545`, `src/worktree_pool.rs:425-432` — dead-worktree fallback calls `release_full`, which clears binding and may run branch cleanup from the recorded binding branch.

Findings:

1. `attempt_safe_rebind_repair` can mutate the worktree before detecting dependents on the recorded stale binding branch. In a double-stale state (`binding.branch=A`, actual branch `B`, requested `C`), it checks only B, switches to C, then `bind_full` can reject because A has an active watch/task. That is not fail-closed because the command returns an error after mutation.

2. The missing/not-git worktree fallback bypasses stale-branch dependency checks. It calls `rebase_clean_self`/`release_full`, which clears binding and may clean up the recorded branch even if that stale branch still has active CI watches or active tasks. #2496 acceptance requires fail-closed for those dependents.

Recommended fix: preflight both recorded binding branch and actual git branch before any switch or fallback cleanup; block without mutation if either branch has active dependents. Add regressions for double-stale no-mutation and dead-worktree stale-dependent no-cleanup.

suzuke:
VERIFIED

Reviewed head: e1fad5c3336865c872e72aeae13cff2620d5602c

Scope: r2 targeted re-review of prior fail-closed blockers in `attempt_safe_rebind_repair`.

### Evidence
ran: `cargo test --bin agend-terminal 2496` → 10 passed, including the two r2 regressions for double-stale no-mutation and dead-worktree stale-dependent no-cleanup.

ran: `cargo fmt --check` → clean.

ran: `cargo clippy --all-targets -- -D warnings` → clean.

ran: `gh pr checks 2523 --repo suzuke/agend-terminal` → exit 8; PR CI still pending at review time, so this is not merge approval until checks exit 0.

cited: `src/mcp/handlers/force_release/repair.rs:136-156` — recorded branch is checked for dependents before dead-worktree fallback can call cleanup.

cited: `src/mcp/handlers/force_release/repair.rs:181-197` — recorded branch and differing actual branch are both checked before any `git switch`.

cited: `src/mcp/handlers/force_release/repair.rs:376-418` — regression proves binding=A, actual=B, requested=C blocks on A and leaves worktree on B.

cited: `src/mcp/handlers/force_release/repair.rs:420-458` — regression proves missing worktree with recorded-branch dependent blocks and keeps binding.

Prior findings are fixed. No remaining code blockers found in the reviewed scope. Merge remains gated on CI green.

PR #2551, opened 2026-07-02
fix(team): #2525 forward repository_path on already-exists branch

suzuke:
## Summary
- `handle_create_team`'s already-exists branch (hit when `create_instance(team=X)` targets an existing team, routed through `teams::update`) re-marshals params into a fresh `update_params` from an allowlist that omitted `repository_path` — same class of bug as #1833/#1837 fixed just above it in the new-team path, never applied here.
- Forwards `repository_path` into `update_params` so a second `create_instance(team=X, repository_path=Y)` call actually reaches `teams::update` (which reads `args["repository_path"]`, set-or-preserve) instead of silently dropping `Y`.

## Investigation note (per task spec, timeboxed 10min)
Checked whether `t-20260702132159428591-56872-1` (daemon restart loses `team update add` members + `project_id`) shares this mechanism. It does not: that bug's repro path goes through `handle_update_team`, which forwards `params` to `crate::teams::update` **unmodified** (no re-marshal/allowlist drop). The persistence bug looks like a disk-durability/race issue around the restart, not a param-forwarding omission — keeping it as a separate task.

## Test plan
- [x] RED: added `create_team_on_existing_team_forwards_repository_path_2525` (clean-room e2e via `handle_create_team`) — failed pre-fix (`source_repo: None`, expected `Some(...)`)
- [x] GREEN: fix applied, test passes
- [x] `cargo test --bin agend-terminal api::handlers::team::` — 5/5 pass, no regressions
- [x] `cargo test --bin agend-terminal teams::` — 36/36 pass, no regressions
- [x] `cargo clippy --bin agend-terminal --all-targets` — clean

### DIFF
--- a/src/api/handlers/team.rs
+++ b/src/api/handlers/team.rs
@@ -608,12 +608,8 @@ mod tests_2525 {
         use std::sync::atomic::{AtomicU32, Ordering};
         static C: AtomicU32 = AtomicU32::new(0);
         let id = C.fetch_add(1, Ordering::Relaxed);
-        let dir = std::env::temp_dir().join(format!(
-            "agend-2525-{}-{}-{}",
-            tag,
-            std::process::id(),
-            id
-        ));
+        let dir =
+            std::env::temp_dir().join(format!("agend-2525-{}-{}-{}", tag, std::process::id(), id));
         std::fs::create_dir_all(&dir).ok();
         dir
     }
--- a/src/api/handlers/team.rs
+++ b/src/api/handlers/team.rs
@@ -586,3 +586,82 @@ mod tests_1964 {
         std::fs::remove_dir_all(&home).ok();
     }
 }
+
+#[cfg(test)]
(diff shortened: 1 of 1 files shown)


suzuke:
**Verdict mirror（§3.12，代 reviewer 補發——其 gh pr review --approve 因共用帳號限制失敗，verdict 原文在 fleet task board t-20260702135256755696-56872-5）**：

VERIFIED — gapfix-reviewer2 @ 488af91b。RED（da8516d4）以正確原因 FAIL（source_repo None）；GREEN 5/5+36/36+clippy 淨；fmt-only delta 經 gh compare 驗證（+2/-6 零邏輯）；轉發結構與 #1833 block 一致（team.rs:310-316 vs :283-288）；範圍純淨（單檔、未觸 persistence 路徑）。

---
*claude-aef7c0*（代發）· 原審 *gapfix-reviewer2* · opencode

PR #2570, opened 2026-07-03
fix(telegram): precise quote-reply correspondence via a persistent sent-message ledger (#2293/#2283)

yujunchao:
## Problem

When a Telegram operator uses **quote-reply** on a specific message, the daemon needs to route their reply back to the *exact* agent and task that produced the quoted message. Previously the inbound path relied on a "most-recent-message" heuristic to guess the target agent, which mis-attributes replies whenever more than one agent has been talking, or when the operator replies to an older message. Separately (#2293/#2283), the **short-message PTY-inject fast path** bypassed the machinery that arms `reply_to_channel` / `pending_user_turn`, so short inbound replies lost their reply context entirely.

## Fix

Two commits, cherry-picked from our fork's production branch:

1. **`arm reply_to_channel + pending_user_turn on the short-message PTY-inject path`** (orig `97f18021`, #2293/#2283)
   - `src/channel/telegram/inbound.rs`, `src/reply_ledger.rs`: the short-message inject path now also arms `reply_to_channel` and `pending_user_turn`, closing the hole where a short reply skipped reply-context setup.

2. **`reply-to precise correspondence via a sent-message ledger`** (orig `8559ea1d`)
   - **`src/sent_ledger.rs` (new)**: a persistent JSONL ledger keyed by `(message_id, chat_id)` mapping each *outbound* message to `{ agent, excerpt, task_id, correlation_id, chat }`. Loaded on startup; 14-day TTL; per-agent FIFO cap (2000 entries). This is what lets an inbound quote-reply resolve `message_id → originating agent + task context` deterministically instead of via a "latest message" guess. Ships with unit tests.
   - **`src/inbox/message.rs`**: adds a `reply_target` (`ReplyTargetContext`) field to `InboxMessage`, serde-backward-compatible (absent in old rows).
   - **Inbound enrich**: on an inbound message carrying `in_reply_to_msg_id`, look the id up in the sent ledger and populate `reply_target` so the handling agent sees exactly which prior message/task is being replied to.
   - **Short-message hole**: inbound short messages that carry a `reply_to` now go through `enqueue` (not the PTY bypass) while still arming `pending_user_turn`, so they do not regress #2293.
   - **Write points**: the adapter Reply arm / `handle_reply` record into the ledger once they have the sent `message_id`.
   - Touches `adapter.rs`, `channel/mod.rs`, `discord.rs`, `api/handlers/messaging.rs`, `mcp/handlers/channel.rs`, `daemon/router.rs`, `daemon/supervisor.rs`, `progress_mirror.rs`, and wires `mod sent_ledger` into `main.rs`.

## Conflict resolution during cherry-pick onto current `main`

Rebased from our base (`673dd603`) onto current upstream `main` (`113a509a`). Two conflicts, both resolved by adopting upstream's newer structure and re-applying our semantics:

- **`src/main.rs`**: upstream removed `mod screenshot;`. Resolved by inserting only `mod sent_ledger;` (did not re-introduce the removed `screenshot` module).
- **`src/mcp/handlers/instance_state/mod.rs`**: upstream #2560 removed `handle_replace_instance` (moved to CLI). Our only change to this file was a single `reply_target: None` line inside that now-removed function, so we took upstream's version wholesale — no semantics lost.

## Testing

- `cargo check` passes clean on the rebased branch (confirms the cherry-pick merged correctly against current `main`).
- `sent_ledger` unit tests included (message-id/chat composite key, TTL, FIFO, load).
- Running in our fleet's production daemon (debug + release builds) since 2026-06-28; quote-reply attribution has been correct in live operation.

_Cherry-picked with `-x` from our fork `yujunchao/agend-terminal` production branch (orig commits `97f18021`, `8559ea1d`). Original commit message for the ledger commit was in Chinese; fully described in English above._


suzuke:
Thanks for the substantial work here (persistent sent-message ledger for precise quote-reply is exactly the right shape for #2293/#2283).

**Status check (2026-07-05)**: the branch now has **merge conflicts with `main`** — the telegram module has moved significantly since 2026-07-03 (notably #2615 multi-channel `active_channel()` fan-out at 8 call sites, and adjacent inbox-obligation changes #2636/#2638/#2639), and the original CI run also predates ~30 merges.

Could you rebase onto current `main` and re-push? Two pointers that will likely matter for the rebase:
- `active_channel()` call sites were replaced with per-instance channel resolution (`channel_for_instance` / `lookup_channel_by_name`) in #2615 — if the ledger hooks into send paths, prefer the new resolvers.
- `request_kind` now has server-side enum validation (#2639) — if the ledger records kinds, unknown values are rejected.

We'll review promptly once CI is green on the rebased head — same flow as #2610 (which merged same-day after its one-line fix). 🙏

suzuke:
Maintainer push (fleet takeover — contributor inactive): rebased onto current `main` and resolved the conflicts. Net diff is the ledger feature only (no `ci.yml`/unrelated changes); authorship preserved.

**Conflicts resolved** — all seam-adaptation to changes that landed since the PR's base; the design is unchanged:
- `src/channel/discord.rs` was split into `discord/*.rs` submodules (#2619) → moved the `AgentOutboundOp::Reply` field adaptation into `discord/adapter.rs` + `discord/tests.rs`.
- `src/daemon/per_tick/progress_mirror.rs` was deleted (ProgressMirror retired, #2549 P2 / #2571) → the PR's field-init there is obviated (same shape as the PR's own `instance_state` resolution note).
- `reply_ledger::arm()` gained a leading `home` param for the durable-discharge guard (#2622) → threaded `home` through the new `arm_channel_turn` and its telegram-inbound caller.
- `doc_lazy_continuation` (newer clippy) flagged a `+`-led doc line in `arm_channel_turn` → reworded.

**Semantic check** against the two rebase pointers from the review: the ledger is keyed by `(message_id, chat_id)` and introduces **no `active_channel()`** usage, so #2615's per-instance-resolver migration doesn't bite; the sent-ledger records **no `request_kind`**, so #2639's enum validation is orthogonal. The telegram write-point records after send (guards `msg_id != 0`) via the per-instance `lookup_topic_for_instance`.

**Gates green**: `cargo fmt --check` ✓ · `cargo clippy --all-targets --features tray,discord -- -D warnings` ✓ · `cargo nextest` sent_ledger/reply_ledger/telegram/discord/inbox **418/418** ✓ · file-size invariant ✓. New head: `9695e707`.

suzuke:
Addressed reviewer5's REJECT (blocking finding + the named coverage gap) — new maintainer commit on top:

**Blocking finding (record/GC race → lost durable row):** `record()` now takes the `state` lock FIRST and does the append + insert **inside** it, making it mutually exclusive with `gc_at` (which already holds `state` across its rewrite+rename). Either order is now safe: record-first → GC's map includes the row; GC-first → record appends onto the freshly-compacted file. Lock-held append I/O is fine (one row per outbound reply). IRON RULE preserved: on a poisoned lock, still append the durable row (GC also bails on a poisoned `state`), only the in-memory insert is forgone.

**Regression test — deterministic, not probabilistic.** `record_not_clobbered_by_concurrent_gc_rewrite_2570` holds `state` as GC does, lets a concurrent `record` reach its critical section, then does the clobbering rewrite. It **fails pre-fix** (append outside the lock → row on disk → clobbered) and **passes post-fix** — verified both directions. A plain two-thread loop can't catch this: the in-memory map converges and a later GC re-persists every row, so "reload after join" is healed by eventual consistency; the loss is only observable across a restart.

**Coverage gap (inbound enrichment e2e):** added `inbound_quote_reply_enriches_reply_target_from_ledger_2570` — end-to-end ledger-record → operator quote-reply → full `ReplyTargetContext`, plus the cross-chat and unknown-id graceful-degrade misses. Extracted `resolve_reply_target` from the inbound handler to make the enrichment unit-testable.

**Gates:** `cargo fmt --check` ✓ · `cargo clippy --all-targets --features tray,discord -- -D warnings` ✓ · `cargo nextest` sent_ledger/reply_ledger/telegram/discord/inbox **420/420** ✓. New head: `d610a3a6` (maintainer fix commit is DCO-signed; authorship of the two feature commits unchanged).

suzuke (approved):
Approved: full content review REJECTED→fixed→narrow re-stamp VERIFIED (fleet review on record, incl. neutralization-verified race regression test). Maintainer commits limited to seam adaptation + the GC/record mutual-exclusion fix.

PR #2574, opened 2026-07-03
feat(instructions): inject an unconditional "keep your main loop free" directive

yujunchao:
## Problem

An agent that is servicing a live operator or peer conversation must stay responsive. If it runs a long *blocking* operation inline in its main context — a compile/build, a bulk analysis, an audit, a large file sweep, a periodic/scheduled prompt — its main loop is stalled and it cannot answer the operator until that work finishes. This is exactly the kind of "the agent went silent for minutes" symptom operators hit. Relying on each agent to self-discipline around this is unreliable.

## Fix

`feat(instructions): inject an unconditional "keep your main loop free" directive` (orig `a97d79b3`).

- **`src/instructions.rs`**: `build_instructions_body` now *unconditionally* injects a short `## Interactive responsiveness (keep your main loop free)` section into every managed agent's instructions — independent of `progress_mode` (i.e. present even when progress reporting is OFF), just like the `[AGEND-MSG]` / `[AGEND-AUTO]` trust-model sections. The rule: when servicing an interactive conversation, dispatch long blocking work to a subagent and keep the main loop free to respond.
  - **Load-bearing carve-out**: a subagent that is *already* executing a delegated task does **not** re-delegate — it runs to completion. This guard is what prevents the rule from recursing into infinite subagent fan-out.
  - Complements (does not duplicate) the richer, backend-aware "Delegating Long Tasks" guidance that is only injected under `progress_mode` 1/2.
  - Adds a unit test (`interactive_responsiveness_rule_is_unconditional`) asserting the section is present with `progress_mode` both OFF and ON, and that the no-re-delegate exception text is carried.
- **`docs/FLEET-DEV-PROTOCOL.md`**: adds §12.8 "Keep the Interactive Loop Free — Delegate Blocking Work", generalizing the previously scheduled-prompt-only rule to *all* blocking long tasks, and documenting that it is enforced structurally by the daemon's unconditional instruction injection.

## Testing

- `cargo check` passes clean on the branch rebased onto current upstream `main` (`113a509a`).
- Included unit test passes; no conflicts during cherry-pick.
- Running in our fleet production daemon since 2026-07-02.

_Cherry-picked with `-x` from our fork `yujunchao/agend-terminal` (orig `a97d79b3`)._

### DIFF
--- a/docs/FLEET-DEV-PROTOCOL.md
+++ b/docs/FLEET-DEV-PROTOCOL.md
@@ -798,6 +798,12 @@ A PR that resolves a tracked issue MUST carry a closing keyword (`Closes #N` / `
 
 `↳ 緣由 A-§12.7`
 
+### 12.8 Keep the Interactive Loop Free — Delegate Blocking Work
+An agent servicing a live operator or peer conversation MUST NOT run a long blocking operation inline in its main context. This covers **all** blocking long tasks — compile/build, bulk analysis, audits, periodic/scheduled prompts, large file sweeps — not just scheduled prompts. Dispatch the work to a subagent and keep the main loop free to respond.
+
+- **Exception**: a subagent already executing a delegated task does **not** re-delegate — it runs its task to completion. (This guard prevents infinite subagent fan-out.)
+- This is enforced structurally: the daemon injects this rule unconditionally into every managed agent's instructions (see `build_instructions_body`, "Interactive responsiveness"), so it does not rely on agent self-discipline.
+
 ## §13. `AGEND_GIT_BYPASS=1` Usage
 
 **TL;DR:** emergency override only. Default is bare `git`. Bypass when shim explicitly denies AND the operation is on the required-bypass list below.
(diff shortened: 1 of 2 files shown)


suzuke:
Thank you for the contribution. We reviewed the injected text word-by-word — it's benign, well-intentioned, and the anti-recursion clause ("subagents must not re-delegate") is genuinely good design. After maintainer review, we're declining this PR on governance grounds rather than content quality:

1. **The instructions body is the fleet's behavioral baseline** — every agent trusts it unconditionally. As a policy matter we don't modify it through external PRs, independent of the specific content.

2. **Unconditional injection conflicts with our role model**: our fleet protocol explicitly requires worker-role agents to execute directly and NOT re-delegate; injecting "dispatch long work to subagents" fleet-wide would push workers against that discipline, and backends differ in subagent capability (codex/opencode ≠ claude). This guidance is right for orchestrator roles specifically — which needs role-aware injection, a larger design than this PR.

3. Technical note: main has moved since your base — `build_instructions_body` is now 2-arg (the `progress_mode` retirement) and the "Delegating Long Tasks" anchor section no longer exists, so the diff no longer applies/compiles regardless.

The core idea is preserved in #2601 as an internal enhancement (role-aware "keep your main loop free" guidance for orchestrator agents), credited to this PR — it'll be picked up when the instructions/role system has a natural evolution point. Thanks again for engaging with the project.

PR #2575, opened 2026-07-03
fix(respawn_watchdog): route auth-expired Resume agents to an operator page instead of blind auto-Fresh

yujunchao:
## Problem

The respawn watchdog force-restarts stuck `Resume` spawns via a bounded auto-Fresh ladder. But when an agent's underlying Claude authorization has **expired**, auto-Fresh is worse than useless: a fresh process is *just as unauthenticated*, so the ladder burns restarts (and can silently drop the agent's session) while never fixing the real problem. Only the operator can re-authenticate. The watchdog had no way to distinguish "stuck, restart me" from "auth-expired, page a human".

## Fix

`fix(respawn_watchdog): route auth-expired Resume agents to an actionable operator page instead of blind auto-Fresh` (orig `8c403505`), entirely within `src/daemon/per_tick/respawn_watchdog.rs`.

- **Pre-decision classification**: a new pure `classify(...)` computes a `Situation` for each agent *before* any auto-Fresh decision. `Situation::AuthExpired` takes precedence over `StuckResume`.
- **Reuses existing auth detection**: auth-expiry is recognised via the existing signal (`backend_profile.rs` regex → `AgentState::AuthError`, already red-anchor false-positive-guarded), rather than re-implementing loose string matching at this layer — so the FP boundary is inherited for free. It is gated to `spawn_mode == Resume` to stay within this watchdog's domain and disjoint from the supervisor's general `AuthError` flow. Because `AuthError` and `Starting/Restarting` are distinct `AgentState`s, this never reclassifies a genuine stuck Resume.
- **Excluded from the ladder**: an `AuthExpired` agent is removed from the auto-Fresh path (mirroring the existing `Paused` exclusion) and instead the operator is paged with an actionable "re-authenticate" prompt.
- **Fire-once-until-recovered latch**: a new `auth_notified` latch on the handler (daemon-lifetime), resolved by a pure `auth_notify_targets(...)`, pages the operator **exactly once** per auth-expiry episode (an `AuthError` agent is re-detected every tick, so this prevents re-paging). The latch is cleared when the agent recovers (leaves `AuthError`) or leaves the registry, so a *later* re-expiry pages again. It is kept separate from `retries` precisely because an auth-expired agent is deliberately not on the restart ladder.

## Testing

- `cargo check` passes clean on the branch rebased onto current upstream `main` (`113a509a`); no conflicts during cherry-pick.
- The classifier (`classify`) and latch resolver (`auth_notify_targets`) are written as pure functions specifically so the auth-vs-stuck precedence and the fire-once semantics are unit-testable without a live registry; unit tests included.
- Running in our fleet production daemon.

_Cherry-picked with `-x` from our fork `yujunchao/agend-terminal` (orig `8c403505`)._

### DIFF
--- a/src/daemon/per_tick/respawn_watchdog.rs
+++ b/src/daemon/per_tick/respawn_watchdog.rs
@@ -165,14 +165,82 @@ fn is_stuck_resume(
         && silent > timeout
 }
 
+/// What the watchdog should do with one agent this tick — computed BEFORE any
+/// auto-Fresh decision so an auth-expired agent is routed away from the
+/// (useless-for-auth) restart ladder.
+#[derive(Debug, PartialEq, Eq)]
+enum Situation {
+    /// Nothing actionable this tick.
+    Ignore,
+    /// Claude authorization expired: exclude from auto-Fresh and page the
+    /// operator (fire-once). A respawn cannot fix auth — a fresh process is just
+    /// as unauthenticated — only the operator can re-authenticate.
+    AuthExpired,
+    /// A stuck `Resume`: eligible for the bounded auto-Fresh ladder.
+    StuckResume,
+}
+
+/// Pure classifier for one agent, so the auth-vs-stuck precedence is
+/// unit-testable without a registry. AUTH TAKES PRECEDENCE: it is recognised via
+/// the EXISTING detection (`backend_profile.rs` regex → `AgentState::AuthError`,
+/// red-anchor FP-guarded — we reuse that signal rather than re-implement loose
+/// string matching at this layer, so the FP boundary is inherited for free). It
+/// is gated to `Resume` to stay inside this watchdog's domain (the module force-
+/// restarts only `Resume` spawns) and disjoint from the supervisor's general
+/// AuthError flow. `AuthError` and `Starting/Restarting` are different
+/// `AgentState`s, so this never reclassifies a genuine stuck Resume.
+fn classify(
+    spawn_mode: SpawnMode,
+    state: AgentState,
+    since_elapsed: Duration,
+    silent: Duration,
+    timeout: Duration,
+) -> Situation {
+    if spawn_mode == SpawnMode::Resume && state == AgentState::AuthError {
+        return Situation::AuthExpired;
+    }
+    if is_stuck_resume(spawn_mode, state, since_elapsed, silent, timeout) {
+        return Situation::StuckResume;
+    }
+    Situation::Ignore
+}
+
+/// Fire-once-until-recovered latch resolver, pure for unit tests. Given the set
+/// of agents currently auth-expired this tick and the latch of already-paged
+/// names (mutated in place): clears the latch for any agent that recovered (no
+/// longer auth-expired) or left, then returns the newly-auth-expired agents to
+/// page now (and latches them). Net effect: each auth-expiry episode pages the
+/// operator exactly once, but a *later* re-expiry after recovery pages again.
+fn auth_notify_targets(auth_now: &HashSet<String>, latch: &mut HashSet<String>) -> Vec<String> {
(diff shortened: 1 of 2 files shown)


suzuke:
Thanks for the contribution — the single-file scope applies cleanly against current main, and the classify/latch tests genuinely pin the semantics. Two things need addressing before this can go into review, plus one consistency nit:

1. **PR description accuracy**: the description argues auth-expired Resume agents currently "burn restarts" via blind auto-Fresh. On current main that isn't the failure mode: `is_stuck_resume` (src/daemon/respawn_watchdog.rs:155-166) only matches `Starting|Restarting`, so an AuthError agent is **silently ignored** by the watchdog rather than auto-Freshed. The operator page is still a real improvement over silent-ignore — but the PR should argue from the actual baseline, since the fix's justification (and its tests' framing) currently describe behavior that doesn't exist.

2. **Overlap with the existing #1523 stable-AuthError notification**: the supervisor already emits an operator notification for stable AuthError states (src/daemon/supervisor.rs — see the stability gate around :53/:137/:1081). AuthError has a history of transient misclassification, which is exactly why that path is gated on stability. As proposed here, (a) this page fires with **no stability gate**, and (b) fleet/team members would be paged **twice** through two independent paths. Please either gate your page on the same stability signal, or route through / extend the supervisor's existing notification instead of adding a parallel one.

3. Nit: the page message text is in Chinese while this file's other user-facing strings are English — please match the file's convention.

Happy to have this re-reviewed once addressed.

suzuke:
Thanks — the direction (routing auth-expired Resume agents to an operator page instead of blind auto-Fresh) is right, and the fire-once latch shape looks good.

**CI status (2026-07-05, after we refreshed the branch onto current `main`)**: all three platform checks fail at the **very first step — `cargo fmt --check`**. It's a single formatting diff:

`src/daemon/per_tick/respawn_watchdog.rs:712` — the `assert!(second.is_empty(), "fire-once: …")` needs rustfmt's multi-line form (message pushes it past 100 cols).

One command fixes it:
```
cargo fmt && git commit -am "style: cargo fmt" && git push
```

We'll review promptly once CI is green — same flow as #2610 (merged same-day after its one-line fix). 🙏

suzuke:
Maintainer push (fleet takeover — contributor inactive): rebased onto latest `main` and applied `cargo fmt` (split the multi-line `assert!` at `respawn_watchdog.rs:712`, the sole CI-red). No semantic changes — the fix's diff is unchanged (respawn_watchdog.rs only, +261/-4 net); authorship preserved.

Local gates green: `cargo fmt --check` ✓ · `cargo clippy --all-targets --features tray -- -D warnings` ✓ · `cargo nextest run` respawn_watchdog module 21/21 ✓. New head: `143b5ad1`.

suzuke:
Maintainer review note (follow-up to the fmt+rebase push): a full content review found one semantic issue to resolve before merge.

The new `Resume + AuthError → AuthExpired` classification fires **immediately** (no duration condition), which bypasses the existing 90s `AUTH_ERROR_NOTIFY_STABILITY` gate in `supervisor.rs`. That gate exists because the `AuthError` pattern is content-FP-prone — transient PTY content can flip it cosmetically, with an observed ~31s self-heal — so an ungated page would false-alert the operator *and* skip auto-Fresh on blips.

Plan: we'll push a follow-up commit adding the same stability window to the watchdog classification (transient blips fall through to the existing paths; persistent auth failure still pages exactly as you designed). The overall approach of routing genuine auth-expiry to an operator page instead of blind auto-Fresh is sound and we want it in — thanks for the contribution.

suzuke (approved):
Approved: content review + narrow re-stamp both passed (fleet review, verdicts on record). The stability-window follow-up commit closes the only blocking finding.

PR #2586, opened 2026-07-03
fix(discord): route Client::new through discord_runtime() (#2562)

suzuke:
## Summary
- `twilight_http::Client::new()` builds a `RateLimiter` whose constructor does `tokio::spawn(actor::runner(..))` internally. `bootstrap::discord_init::init` calls `init_from_config` on a bare `std::thread` with no ambient Tokio runtime, so this panicked on every real daemon boot.
- Found via an isolated smoke-home (`~/.agend-smoke`, not production) end-to-end boot for #2562 P3 inbound testing — this is the first time the real bootstrap path (vs. `#[tokio::test]`-shielded unit tests) was ever exercised.
- Extracts `build_http_client()` and routes it through the existing `discord_runtime()` — the same runtime the outbound `send`/`edit`/`delete` calls already use — instead of introducing a second ad hoc runtime for init.

## Evidence (panic reproduced on the unfixed code, real daemon boot, `~/.agend-smoke` log)
```
INFO bootstrap-step step="discord_init::init" elapsed_ms=0
ERROR panic location=.../twilight-http-ratelimiting-0.17.1/src/lib.rs:279:9 there is no reactor running, must be called from the context of a Tokio 1.x runtime
WARN discord gateway: event receiver dropped, stopping
```
"discord init: channel registered" never appeared in the log — the channel never actually came up.

## Test plan
- [x] New pin test `channel::discord::tests::build_http_client_does_not_panic_on_bare_thread` — calls `build_http_client` from a genuine bare `std::thread` (mirroring `discord_init`'s real call site, not `#[tokio::test]`). Confirmed RED against the pre-fix body (panicked with the exact production error), GREEN after the fix.
- [x] `cargo test --features discord --bin agend-terminal channel::discord::` — 34 passed, 1 ignored (network-manual)
- [x] `cargo test --features discord --bin agend-terminal bootstrap::` — 65 passed
- [x] `cargo check --features tray` — clean (discord.rs is feature-gated out of this build)
- [ ] Smoke rerun in `~/.agend-smoke` with the fixed binary to confirm Gateway READY + channel registered (in progress, will report separately)

### DIFF
--- a/src/channel/discord.rs
+++ b/src/channel/discord.rs
@@ -471,6 +471,27 @@ fn discord_intents() -> twilight_model::gateway::Intents {
         | twilight_model::gateway::Intents::MESSAGE_CONTENT
 }
 
+/// Construct the Discord outbound HTTP client.
+///
+/// Must run through [`discord_runtime()`] (via the local [`block_on_value`]),
+/// not on a bare thread. `twilight_http::Client::new` builds a
+/// `twilight_http_ratelimiting::RateLimiter` whose constructor does
+/// `tokio::spawn(actor::runner(..))` internally — called with no active
+/// Tokio runtime in scope (as `bootstrap::discord_init::init` does, since it
+/// runs `init_from_config` on a plain `std::thread`) that panics with
+/// "there is no reactor running, must be called from the context of a
+/// Tokio 1.x runtime" (twilight-http-ratelimiting 0.17.1 src/lib.rs:279 —
+/// found via #2562 P3 isolated-smoke-home boot, 2026-07-03: the daemon's
+/// real bootstrap path never got an end-to-end run before, only
+/// `#[tokio::test]`-shielded unit tests, which supply an ambient runtime
+/// this call site doesn't have on real boot). Routing through the same
+/// `discord_runtime()` the outbound `send`/`edit`/`delete` calls already use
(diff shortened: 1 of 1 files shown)


suzuke:
Merging via admin per fleet decision d-20260703133948127139-0:
- Content review: gapfix-reviewer3 VERIFIED @ 13660b0f (m-20260703132013404765-58) — twilight call-chain source audit (RateLimiter::new does tokio::spawn at lib.rs:279; cfg(test)-only bypass confirmed), shared_async nested-runtime-guard deadlock audit, pre-fix RED reproduction on main, PR-body log-string anti-fabrication check.
- Second seat gapfix-reviewer/codex issued a written UNVERIFIED @ 13660b0f (m-20260703131057298487-47) stating zero code defects with CI as the sole blocker; CI has since gone 8/8 green on the unchanged head. The formal upgrade was blocked by a PTY-injection fault on that seat (tracked separately), so this merge treats the written code-clean verdict + subsequent CI-green as satisfying the dual intent.
- Root cause fixed: discord_init bare thread constructing twilight_http::Client without a runtime → panic → gateway self-stop; fix routes construction through the shared discord_runtime() (same runtime as outbound), with a bare-thread regression test (RED pre-fix → GREEN post-fix).

Issue #1352, opened 2026-05-27
fix: Telegram inbound length-based delivery split — short→PTY only, long→inbox+hint


suzuke:
## Problem

Telegram inbound delivers every message via both PTY inject and inbox enqueue (`inbound.rs:503-514`). This causes duplicate delivery: agent processes the PTY inject, then later drains inbox and sees the same message again.

## Solution

Length-based delivery split:

- **Short messages (< 200 chars, no attachments)**: PTY inject only, **no inbox copy**. Zero MCP call overhead. If operator doesn't get a response, they can resend.
- **Long messages (>= 200 chars) or with attachments**: Write to inbox + PTY sends `[AGEND-MSG-PENDING]` hint only. Agent drains inbox to read full content.

Existing 200-char truncation logic in `notify.rs:224` already provides the threshold — align the delivery split with this boundary.

## Implementation Notes

- Modify `inbound.rs:503-514` to conditionally skip `inbox::enqueue()` for short messages
- `AGEND_POINTER_ONLY_INJECT=1` overrides this — when on, all messages go inbox+hint (existing behavior)
- Telegram does NOT auto-split messages (client blocks >4096 chars), so no reassembly needed
- Attachments always go inbox+hint path (agent needs `download_attachment` MCP tool)

## Not in Scope

- PTY inject failure fallback (PTY write failure only detectable when agent is dead; operator can resend)
- Message reassembly (Telegram doesn't split)
- Configurable threshold (hardcode 200 to match existing truncation; can be made configurable later if needed)

PR #2587, opened 2026-07-03
feat(discord): inbound dispatcher — drain poll_event() to agent inbox (#2562 PR-1)

suzuke:
## Summary
- Discord messages reached the mpsc queue (fixed by the prior P3 runtime-panic PR #2586) but nothing ever drained it — no production code called `poll_event()`, unlike Telegram which has its own dedicated dispatch loop (`telegram/inbound.rs::start_polling`). This PR adds the missing consumer.
- `spawn_inbound_dispatcher`: a plain `std::thread` poll loop (no tokio needed — Discord's network I/O already lives on its own thread in `start_gateway`; this loop only drains and routes). Wired into `discord_init.rs`'s existing spawn closure, fire-and-forget per §10.5, matching every other Discord/Telegram background thread's convention.
- `resolve_instance_for_channel`: reverse-lookup via the existing (till now write-only) `channel_to_instance` map, falling back to `"general"` + a warn log — mirrors `telegram/inbound.rs`'s `resolve_topic` topic-miss semantics.
- `dispatch_channel_event`: extracts `channel_id` from the binding, resolves the instance, and calls the same `NotifySource::Channel` path Telegram uses (`ChannelKind::Discord` was already handled in its `Display` impl, so `[user:NAME via discord]` needed no new code).
- Accept-path logging: `map_message_create_to_message_in` logged on reject but not accept — an asymmetry found during #2562 P3's live smoke test, fixed with a matching `tracing::info!`.
- `event_rx` doc comment corrected: the channel is `std::sync::mpsc::channel` (**unbounded**), not "bounded" as previously documented.

Session-agnostic by design: `dispatch_channel_event` has no notion of which gateway connection an event came from, so it needs no changes when P2 adds gateway reconnect.

**Not in scope** (tracked separately per PRERESEARCH doc): the write side of `channel_to_instance` (a Discord equivalent of `app/telegram_hooks.rs::maybe_create_telegram_topic` never existed, so the table is empty in production until that's built — PR-2), graceful dispatcher shutdown, `fleet_normalize.rs`'s Telegram-hardcoded `register_topic` call.

## Test plan
- [x] New: `resolve_instance_for_channel_returns_bound_instance`, `resolve_instance_for_channel_falls_back_to_general_when_unbound`, `dispatch_channel_event_routes_to_bound_instance`, `dispatch_channel_event_does_not_panic_on_bare_thread`, `discord_message_create_accept_path_logs_info` — routing decision + fallback + accept-log + bare-thread pin, no mock gateway needed.
- [x] `cargo test --features discord --bin agend-terminal channel::discord::` — 39 passed, 1 ignored (network-manual), zero existing tests modified
- [x] `cargo test --features discord --bin agend-terminal bootstrap::` — 65 passed
- [x] `cargo check --features tray` — clean (discord.rs is feature-gated out of this build)

### DIFF
--- a/src/bootstrap/discord_init.rs
+++ b/src/bootstrap/discord_init.rs
@@ -42,9 +42,16 @@ pub(super) fn init(config: &FleetConfig) -> Option<Arc<dyn Channel>> {
                 tracing::info!("discord init: skipped (no token / no config)");
                 return;
             };
-            let channel_dyn: Arc<dyn Channel> = Arc::new(channel) as Arc<dyn Channel>;
+            let channel_arc = Arc::new(channel);
+            let channel_dyn: Arc<dyn Channel> = Arc::clone(&channel_arc) as Arc<dyn Channel>;
             crate::channel::register_active_channel(Arc::clone(&channel_dyn));
             tracing::info!("discord init: channel registered");
+            // #2562 PR-1: drain poll_event() -> resolve instance -> inbox.
+            // Kept on the concrete Arc<DiscordChannel> (not the dyn Channel
+            // trait object) so the dispatcher can call Discord-specific
+            // routing (`resolve_instance_for_channel`) without widening the
+            // generic Channel trait's surface for one adapter's needs.
+            crate::channel::discord::spawn_inbound_dispatcher(channel_arc, crate::home_dir());
             attach_pending_registry_when_ready(channel_dyn);
         });
     if let Err(e) = spawn_result {
--- a/src/channel/discord.rs
+++ b/src/channel/discord.rs
@@ -588,7 +588,7 @@ pub(crate) fn dispatch_channel_event(
         binding,
         from,
         payload,
-        ..
+        ts,
     } = event
     else {
         return;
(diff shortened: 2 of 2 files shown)


suzuke:
- Dual review round 1 @ b14af4e7: REJECTED by both seats independently (gapfix-reviewer3 m-...198 + lead seat-2) — converged on the same defect: no inbox persistence in the dispatch path (long-message/pointer-only notifications pointed at nothing).
- Rework @ 5268af21: mirrors telegram's #1352 short/long split (persist_or_log! inbox::enqueue before notify), 2 new tests asserting real inbox::drain state both ways, residual short-message window documented as pre-existing cross-channel parity.
- Re-stamp: gapfix-reviewer3 VERIFIED (m-...221 + CI upgrade m-...237, De Morgan equivalence verified against telegram/inbound.rs, attachments-term vacuity noted) + lead seat-2 delta PASS. CI 9/9 green.
- Completes #2562 inbound PR-1: all Discord messages now reach agent inboxes (fallback to general until PR-2 wires the routing-table write side).

PR #2588, opened 2026-07-03
feat(discord): per-instance binding write side (#2562 PR-2)

suzuke:
## Summary
- The routing table PR-1's dispatcher reads (`channel_to_instance`) was write-only in production: `record_binding` was never called for Discord, only for Telegram (`app/telegram_hooks.rs::maybe_create_telegram_topic`, wired into the TUI's 3 pane-creation call sites). Every Discord message was falling through to the `"general"` fallback.
- `app/discord_hooks.rs::maybe_create_discord_binding`: mirrors `maybe_create_telegram_topic`'s shape (`has_binding` guard, background thread, `create_binding` then `record_binding`). Differs in one respect: Telegram's channel reference is resolved synchronously at `setup_app_bootstrap` time and threaded through `Ctx.telegram_state`; Discord's channel connects asynchronously in the background (`discord_init::init` always returns `None`), so this hook looks up the active channel at call time via `lookup_channel_by_name("discord")` instead of taking a pre-resolved parameter.
- Wired into the same 3 call sites as `maybe_create_telegram_topic` (`commands.rs:410`, `overlay.rs`'s two pane-creation arms).
- `fleet_normalize.rs:63` verified: `auto_create_general`'s persist branch unconditionally calls `telegram::register_topic` regardless of which channel is configured. For a Discord-only fleet this is a harmless no-op (writes `topics.json`, which Discord's routing never reads) — left as-is with a comment recording the verification rather than conditionalizing a write nothing reads.

**Not in scope** (deferred per PRERESEARCH): `maybe_delete_discord_binding` (pane-close teardown, Telegram's mirror-image hook) — a stale binding left behind on instance delete isn't wired to anything harmful today, but is a gap worth tracking separately.

## Test plan
- [x] New `app/discord_hooks.rs` tests use a local `MockChannel` (mirrors `app/tui_spawn.rs`'s existing pattern) since Discord's real `create_binding` needs live HTTP infra this hook-layer test doesn't need to exercise (already covered by `discord.rs`'s own outbound tests): `maybe_create_discord_binding_records_new_instance`, `maybe_create_discord_binding_is_noop_when_already_bound`, `maybe_create_discord_binding_is_noop_without_active_channel`
- [x] `cargo test --features discord --bin agend-terminal app::` — 151 passed (zero existing tests modified)
- [x] `cargo test --features discord --bin agend-terminal channel::discord::` — 41 passed, 1 ignored
[...]

PR #2591, opened 2026-07-03
fix(telegram): TUI pane create/kill hooks never fire (frozen post-#945 channel ref)

suzuke:
## Summary

Follow-up to reviewer3's static-read finding (t-20260703155211396370-50899-7) folded into the telegram-topic-lifecycle task (t-20260703154305965646-50899-6) per lead's priority call. Confirmed root cause via direct code tracing (not just static reading) before implementing.

`maybe_create_telegram_topic` and `maybe_delete_telegram_topic` took a pre-resolved `&Option<Arc<dyn Channel>>` threaded through `CommandCtx`/`OverlayCtx` from a local variable (`telegram_state`) captured once at `setup_app_bootstrap` time in `app/mod.rs`. `telegram_init` has been backgrounded since #945 (its synchronous return is always `None`, per that commit's own comment at `mod.rs:487-489`), so the captured reference stays `None` for the entire process lifetime. Both hooks silently no-op via their `let Some(tg) = tg else { return }` guard — for **every** TUI-driven pane create (`:spawn`/`:vsplit`/`:hsplit`, NewTabMenu, SplitMenu) and pane kill (`:kill`), not just a boot-time race window.

**Impact differs by direction** (verified by tracing each call site's full chain, not assumed):
- **Creation**: already has a working fallback — #966's `tui_spawn::add_instance_with_topic` runs *first* at all 3 create call sites (`commands.rs`'s `:spawn`/`:vsplit`/`:hsplit`, and `menu.rs::pane_from_menu_item` used by both overlay menus) and correctly persists `topic_id` to `topics.json` via a live `active_channel()` lookup. The frozen create-hook call afterward was mostly redundant (though it did mean `TelegramState`'s in-memory `instance_to_topic`/`topic_to_instance` maps never got directly populated via `record_binding` for TUI-created panes — self-healed by `topics.json` fallback on first inbound-routing cache miss).
- **Deletion**: no such fallback exists. Killing a pane via the TUI's `:kill` command routes through `daemon::lifecycle::delete_transaction` (process/registry/configs cleanup only) + inline `fleet.yaml` removal — `maybe_delete_telegram_topic` was the *only* Telegram-topic cleanup mechanism for this path, and it was genuinely, permanently broken. This directly contributes to the residual-topic problem this branch is about.

## Fix

Both hooks now do a live `crate::channel::lookup_channel_by_name("telegram")` lookup at call time instead of taking a frozen parameter — mirroring `maybe_create_discord_binding`'s pattern (#2588), which never had this bug because Discord's init was async from the start (its own PR body contrasts this exact difference). `telegram_state` is no longer threaded through `CommandCtx`/`OverlayCtx` — removed from both structs and all construction sites (`app/mod.rs`, `overlay.rs`'s command-palette bridge, and 7 test fixtures).

## Test plan
- **Test-first**: new `MockChannel` (kind=`"telegram"`) suite in `telegram_hooks.rs` proves both hooks reach their live-looked-up channel — `has_binding`/`take_binding` are recorded synchronously (these run on the caller's thread, before either hook's background thread spawns), so no thread-timing flakiness. Also covers: no-op when no `"telegram"` channel is registered, no-op when `pane.fleet_instance_name` is absent. One end-to-end test uses a bounded poll (not a sleep loop, not an indefinite wait) to confirm `remove_binding` actually fires on the background thread once `take_binding` returns a binding.
- [x] `cargo build --lib --bin agend-terminal --tests` — clean
- [x] `cargo test --bin agend-terminal telegram_hooks` — 6/6 new tests pass
- [x] `cargo test --bin agend-terminal app::` — 154/154 pass, zero pre-existing tests modified
- [x] `cargo nextest run --lib --bin agend-terminal` — 4781/4781 pass
- [x] `cargo clippy --lib --bin agend-terminal --tests -- -D warnings` — clean

Refs #2550, mirrors #2588

### DIFF
--- a/src/app/commands.rs
+++ b/src/app/commands.rs
@@ -10,7 +10,6 @@ use crate::layout::{Layout, SplitDir, Tab};
 
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
-use std::sync::Arc;
 
 /// Bundle of mutable references the command handler needs to affect the
 /// running TUI state. Constructed by the caller for each invocation.
@@ -20,7 +19,6 @@ pub(super) struct CommandCtx<'a> {
     pub home: &'a Path,
     pub wakeup_tx: &'a crossbeam_channel::Sender<usize>,
     pub name_counter: &'a mut HashMap<String, usize>,
-    pub telegram_state: &'a Option<Arc<dyn crate::channel::Channel>>,
 }
 
 /// The dynamic source feeding a command-argument position, declared per token in
@@ -408,7 +406,6 @@ pub(super) fn execute(cmd: &str, ctx: &mut CommandCtx<'_>) -> bool {
             match pane_result {
                 Ok(pane) => {
                     super::telegram_hooks::maybe_create_telegram_topic(
-                        ctx.telegram_state,
                         ctx.registry,
                         ctx.home,
                         &pane,
@@ -439,11 +436,7 @@ pub(super) fn execute(cmd: &str, ctx: &mut CommandCtx<'_>) -> bool {
         "kill" => {
             if let Some(name) = parts.get(1) {
                 if let Some(fleet_name) = lookup_fleet_name(ctx.layout, name) {
-                    super::telegram_hooks::maybe_delete_telegram_topic(
-                        ctx.telegram_state,
-                        ctx.home,
-                        &fleet_name,
-                    );
+                    super::telegram_hooks::maybe_delete_telegram_topic(ctx.home, &fleet_name);
                     let _ = crate::fleet::remove_instance_from_yaml(ctx.home, &fleet_name);
                 }
                 super::kill_agent(ctx.home, ctx.registry, name);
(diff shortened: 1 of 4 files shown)


suzuke:
- DUAL review @ e9f65bf2: gapfix-reviewer3 VERIFIED (m-...366 + CI upgrade m-...419) — incl. precision correction (drop_active_binding is a secondary in-memory-only cleanup; the real Telegram delete-topic API call genuinely never fired pre-fix, confirming the operator-reported residual-topic root cause) and pre-existing-test corroboration of the create-direction fallback. gapfix-reviewer/codex code-clean UNVERIFIED-CI-gated (m-...378, full local nextest 4781 green) with written pre-authorization — CI now 9/9 green at unchanged head.
- Fixes the frozen telegram_state hooks (permanent no-op since #945 backgrounding): both hooks now do live lookup_channel_by_name, mirroring #2588's Discord pattern; dead Ctx fields removed.

PR #2590, opened 2026-07-03
feat(discord): gateway death status flag (#2562 PR-3a)

suzuke:
## Summary
- `start_gateway`'s thread has no outer supervisor — once it exits (fatal close, its event receiver disappearing, or a spawn/runtime-build failure), the daemon has no signal that Discord is now completely dead short of grepping logs.
- Adds a process-wide `GATEWAY_DEAD` flag (same shape as `daemon::mod::SHUTDOWN_REASON`) set on every exit path, readable via `gateway_is_dead()` for future status/MCP surfaces.
- Cleared at the start of `start_gateway` itself, so PR-3b's restart supervisor calling it again after a prior death gets a clean slate for the new attempt.

Per `DISCORD-P2-RECONNECT-PRERESEARCH.md`: `twilight_gateway::Shard` already handles transport-level reconnect and resume-vs-reidentify internally (confirmed against the vendored 0.17.1 source) — this flag only fires on the narrow set of conditions the library itself can't recover from.

## Test plan
- [x] New `gateway_is_dead_reflects_death_state` — starts alive, reflects a marked death, reset clears it back
- [x] `cargo test --features discord --bin agend-terminal channel::discord::` — 42 passed, 1 ignored (network-manual), zero existing tests modified
- [x] `cargo test --features discord --bin agend-terminal bootstrap::` — 65 passed
- [x] `cargo clippy --all-targets --features discord -- -D warnings` — clean
- [x] `cargo check --features tray` — clean
- [x] `cargo fmt --check` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2578, opened 2026-07-03
fix(ci): scoped nextest retries for the Windows real-git flake class

suzuke:
## Summary
- Adds `retries = 2` to the two existing `[profile.ci.overrides]` blocks that already scope the `git-subprocess` test-group filter — no new filter, no suite-wide retries. Every other test in the suite still fails on the first try.

## Evidence (t-20260702161319112091-56872-10, 7-sample pattern-ready class)
This real-git test cluster (worktree/worktree_pool/worktree_cleanup/admin/mcp::handlers::{ci,dispatch_hook,force_release,worktree,p0b_tests}/daemon::{...}/bootstrap::{...} + named tests, per #1893/reviewer-2 #1896) intermittently fails on Windows runners at the process-spawn/DLL-init layer — not the test's own logic:
1. `worktree_pool::release_full_absent_worktree_merged_branch_cleaned_up` — misjudged merge state, rerun resolved
2. #2555 (merge-detection semantic weakness — tracked independently)
3. PR #2567: `dispatch_reuse_after_binding_cleared...` — isolated failure, rerun resolved
4. (weak signal) local sandbox `worktree_pool` mass-failure (environment confound)
5. Windows hung-test timeout kill, log not retained, rerun green
6. (tangential) macOS "Upload binary artifact" failure, rerun green
7. `worktree_pool::review_repro_worktree_git::cleanup_merged_branch_uses_true_default_not_main_worktree_git` — **ABORT 0xc0000142 "DLL initialization routine failed"**, 0.008s in, the other 4861 tests in the same run passed clean

Every sampled failure resolved on a bare rerun with **no code change** — consistent with a runner-environment race (antivirus/DLL-load contention is a well-documented `0xc0000142` cause on Windows CI runners), not a real regression.

## Why scoped, not suite-wide
Retries only apply to the filter already carved out for infra-flake-prone real-git spawns (the SAME `git-subprocess` group #1893 already serializes for index.lock contention). Every other test in the suite (the ~4700 pure-logic tests) still fails on the first attempt — a real regression anywhere outside this filter is caught exactly as before.

## Why this doesn't mask a real bug
[...]

PR #2592, opened 2026-07-03
feat(discord): bounded gateway restart supervisor (#2562 PR-3b)

suzuke:
## Summary
- `start_gateway` itself has no outer supervisor — `twilight_gateway::Shard` already handles transport-level reconnect and resume-vs-reidentify internally (confirmed against the vendored 0.17.1 source, see `DISCORD-P2-RECONNECT-PRERESEARCH.md` §1), so the only gap is what happens after `start_gateway`'s thread has ALREADY died for good: nothing restarts it.
- `spawn_gateway_supervisor` wraps `start_gateway` with a bounded retry: up to `GATEWAY_RESTART_MAX_ATTEMPTS` (3) restarts with exponential backoff (same constants as `telegram/poll_supervisor`'s shape: base 5s, cap 60s), then gives up permanently. Bounded rather than Telegram's mostly-indefinite retry because every fatal-close condition here is a config/auth problem (bad token, invalid intents, etc.) that won't self-resolve by waiting — the cap exists so a still-bad config doesn't hot-loop forever, while still giving an operator who fixes the config WHILE the daemon is running a shot at automatic recovery without a full daemon restart.
- Reuses PR-3a's `GATEWAY_DEAD` flag as the "has `start_gateway`'s thread exited" signal (polled, not a `JoinHandle`) rather than widening `start_gateway`'s fire-and-forget contract just to duplicate that same signal.

## Test plan
- [x] `discord_gateway_backoff_delay_follows_expected_sequence` — same sequence as Telegram's `backoff_delay` test (5, 10, 20, 40, 60, 60, 60)
- [x] `should_restart_gateway_caps_at_max_attempts` — allowed below the cap, refused at and beyond it
- [x] The outer loop's actual thread-restart mechanics aren't independently tested — matching this codebase's existing precedent for background-thread loops (neither `telegram::start_polling` nor `start_gateway`/`spawn_inbound_dispatcher` have thread-loop-level tests)
- [x] `cargo test --features discord --bin agend-terminal channel::discord::` — 44 passed, 1 ignored (network-manual), zero existing tests modified
- [x] `cargo check --features tray` — clean

### DIFF
--- a/src/channel/discord.rs
+++ b/src/channel/discord.rs
@@ -514,6 +514,91 @@ pub(crate) fn start_gateway(
     }
 }
 
+/// #2562 PR-3b: maximum number of restart attempts after `start_gateway`
+/// dies, before giving up permanently. Bounded (unlike Telegram's
+/// `poll_supervisor`, which retries most errors indefinitely) because
+/// every condition that kills `start_gateway` is a config/auth problem
+/// (see `should_stop_gateway_loop`'s doc comment) that won't self-resolve
+/// by waiting — see `DISCORD-P2-RECONNECT-PRERESEARCH.md` §1. The cap
+/// exists so a still-bad token/intents config doesn't hot-loop forever,
+/// while still giving an operator who fixes the config WHILE the daemon
+/// keeps running a chance at automatic recovery without a full restart.
+const GATEWAY_RESTART_MAX_ATTEMPTS: u32 = 3;
+
+/// Backoff before restart attempt number `attempt` (1-indexed: the delay
+/// before the FIRST restart, i.e. the second overall try). Mirrors
+/// Telegram's `poll_supervisor::backoff_delay` shape (exponential, base
+/// 5s, cap 60s) — same constants for consistency, not because the two
+/// domains share an error taxonomy (they don't; see PRERESEARCH §2).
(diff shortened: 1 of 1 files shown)


suzuke:
- DUAL review: round-1 gapfix-reviewer3 VERIFIED @ 0b928507 (hand-traced supervisor loop: 1 original + 3 bounded restarts, single-sink flag design, no double-start, restart-storm wall-clock verified) + gapfix-reviewer/codex REJECTED on a stale-comment consistency finding → rework @ 78b998f6 (comment fix + u32::MAX backoff assertion) → both seats re-stamp VERIFIED (m-...421, m-...418) + reviewer3 CI upgrade (m-...549).
- CI history: attempt-1 windows failed 59min 'runner lost communication' (3rd runner-infra event today), attempt-2 failed on unrelated ensure_branch_exists...2107 DLL-init 0xc0000142 (Windows real-git flake class, #2578 family — triaged independently by dev, reviewer3, and lead, all converging), attempt-3 clean 8/8 with normal 14m32s windows duration.
- Completes #2562 P2: GATEWAY_DEAD flag (PR-3a #2590) + bounded restart supervisor (this PR). Discord channel now: gateway w/ built-in transport reconnect + bounded fatal-restart + inbound dispatcher + per-instance routing.

PR #2597, opened 2026-07-04
fix(#2550 PR-0): cleanup_merged_branch is_gone delete gate → is_squash_gc_eligible

suzuke:
## Summary

Fixes t-20260703160308595650-50899-10 (data-loss risk found by reviewer3 during #2589's review). PR-0 of the #2550 W5 GC-unification sequence — independent of the rest of W5, landing first per the operator-approved execution order.

`worktree_pool.rs::cleanup_merged_branch` gated branch deletion on `!is_merged && !is_gone` — meaning a remote-tracking ref being gone was, by itself, sufficient to trigger `git branch -D`, with no squash verification. `worktree_cleanup.rs::prune_orphaned_branches` already fixed the identical pattern under CR-2026-06-14 (`merged || is_squash_gc_eligible`), but `worktree_pool.rs`'s copy was never updated to match.

**Concrete risk**: a branch pushed once, then its remote ref deleted (e.g. GitHub squash-merge), that keeps accruing local commits never re-pushed, is remote-gone yet carries committed-but-unpushed work — `git branch -D` destroys it irrecoverably. Given this fleet's worktrees are reused at high frequency, this is a realistic scenario, not a theoretical edge case.

## Changes
- `worktree_cleanup.rs::is_squash_gc_eligible` → `pub(crate)`, so both call sites share one implementation instead of drifting again.
- `worktree_pool.rs::cleanup_merged_branch`'s gate: `!is_merged && !is_gone` → `!is_merged && !is_squash_gc_eligible(...)`, mirroring `prune_orphaned_branches` exactly. Docstring corrected (it claimed "prevents deletion of unmerged branches," which was only true when `is_gone` was also false).
- Two new pin tests in `worktree_pool/review_repro_worktree_git.rs`:
  - `cleanup_merged_branch_keeps_remote_gone_branch_with_unpushed_commits_worktree_git` — the negative case this PR fixes.
  - `cleanup_merged_branch_still_deletes_aged_squash_merged_remote_gone_branch_worktree_git` — proves the new gate still reaps genuinely squash-merged branches, not just silently disabling remote-gone cleanup.

## Test plan
- [x] Both new pin tests pass.
- [x] `cargo nextest run --profile ci` — 5172/5174 (the 2 failures are the pre-existing documented flakes `e2e_dispatch_review_workflow_seam_invariants_1900` / `teardown_leaves_zero_residual_after_full_exercise_1907`, unrelated to this diff).
[...]

PR #2579, opened 2026-07-03
test(#2550 W1): pin GC dual-path combined-pipeline behavior

suzuke:
## Summary

Part of #2550 P3 Wave 1 (spike: `P3-2550-SPIKE.md` §2a/§4, decision `d-20260703062722787157-1`). Test-only PR — locks down GC's current dual-path behavior before W5's unification touches it.

Every existing test in `daemon/retention/worktrees.rs` drives `maybe_remove_candidate` directly with a hand-built `GcCandidate` — correct for pinning that function's own guards in isolation, but none of them exercise `gc_run` (worktree_pool::gc.rs, the hourly GC-tick driver) and `sweep` (retention supervisor) running in the same-tick sequence `build_default_handlers` actually uses. This PR adds 4 tests that close that gap, using real `lease()`-built worktrees driven through the REAL `gc_candidates()` discovery path.

## Tests added

1. **`gc_run_hard_deletes_clean_release_then_sweep_is_noop_2550_w1`** — baseline: CleanRelease is a genuine hard delete (never lands in `.trash`), and `sweep`'s own fresh re-scan afterward finds nothing left to do.
2. **`gc_run_delegates_force_reclaim_to_archive_never_hard_delete_2550_w1`** — ForceReclaim always lands in `.trash` via `gc_run`'s delegation (gc.rs:733-751), never hard-deleted, through the real discovery path this time.
3. **`gc_run_skip_then_sweep_retry_archives_clean_release_1053_1058_1170`** — **the Q1 pin**, the most important one. Forces `gc_run`'s own binding-lock acquisition to fail with a genuine I/O error (a regular file where the lock's parent directory needs creating — `acquire_file_lock` is a *blocking* flock, so holding the lock ourselves on the same thread before calling `gc_run` would deadlock the test rather than reproduce a clean skip). Confirms retention's `sweep` then catches the skipped CleanRelease candidate and archives it. This is what the spike's Q1 decision (skip-then-retry = intentional defense-in-depth, not an accidental side effect of two independent scanners) requires W5 not to regress.
4. **`force_reclaim_spared_when_agent_shows_liveness_at_archive_time`** — a gap found while writing this suite (unrelated to the combined-pipeline focus): no existing test pinned the archive-time liveness re-check's specific skip reason (`agent_alive_at_archive`, worktrees.rs:265-275), distinct from the classification-time liveness check in `gc.rs`'s `evaluate_candidate`.

## Scope note: what I deliberately did NOT add

[...]

PR #2599, opened 2026-07-04
refactor(#2550 PR-2): unify GC into a single driver (fire-on-first, gate coverage unchanged)

suzuke:
## Summary

Core PR of the #2550 W5 GC-unification sequence (PR-2 of 3; PR-0 #2597 merged, PR-1 struct-merge dropped per premise check — see thread). Per `P3-2550-SPIKE.md` §2a/§5 Wave 5 + decisions `d-20260703062722787157-1` (defense-in-depth) and `d-20260704035059093740-0` (Q1/Q2/Q3 operator rulings).

**What changes**: `gc_run`'s CleanRelease hard-delete attempt, when it can't act on a still-eligible candidate (lock contention, unresolved owning repo, or both `git worktree remove` and `remove_dir_all` failing), now falls through in the SAME pass to `retention::worktrees::archive_fallthrough` instead of waiting for a separately-cadenced independent re-scan. A re-validation failure (candidate no longer eligible — rebound, re-pinned, grace changed) does NOT fall through, since a fresh scan wouldn't find it either.

**Q1 (accepted, immediate fallthrough)**: the "skip → archive retry" defense-in-depth (`d-20260703062722787157-1`) is preserved exactly, now same-tick instead of waiting up to ~1h for retention's independent fire-on-Nth cadence.

**Q2 (fire-on-first cadence, behavior change — flagging per operator instruction)**: worktree GC now runs entirely on `gc_tick`'s fire-on-first cadence (~10s after boot, then hourly). This closes a real gap: previously, for the first hour after daemon boot, a CleanRelease candidate `gc_run` skipped had literally no safety net, because retention's independent sweep (fire-on-Nth) hadn't fired yet.

**Q3 (gate coverage unchanged)**: `CleanRelease` hard-delete stays unconditional (as today); the archive-fallthrough still requires `AGEND_WORKTREE_GC=1`, exactly matching the old standalone `sweep()` it replaces. Pure semantic-equivalent refactor on the gating axis — verified by a new test (`gc_run_archive_fallthrough_stays_off_by_default_2550_w5`).

**Removed**: `retention::worktrees::sweep()` — its job (re-scan + archive every candidate independently) is now fully subsumed by `gc_run`'s own fresh `gc_candidates()` scan each pass, so the second independent scan added nothing but the ~1h phase lag Q2 fixes. `purge_trash()` moved to run at the end of every `gc_tick` pass instead (same frequency as before, no phase lag).

**⚠️ Noted exception to Q3 (found in review, seat-2): `purge_trash`'s gate coverage DID change, deliberately.** The old `sweep()` only ran — and so only purged `.trash` — when `AGEND_WORKTREE_GC=1`; a gate-off install accumulated `.trash` entries forever, because `ForceReclaim` candidates archive into `.trash` unconditionally (true before and after this PR, never gated) but nothing ever purged them without the gate on. `purge_trash` is now called unconditionally every `gc_tick` pass, regardless of the gate. Reasoning: Q3's "gate coverage unchanged" was about the archive-*decision* path (CleanRelease's fallthrough), not about cleaning up entries already sitting in `.trash` — `AGEND_WORKTREE_GC_TRASH_DAYS` is the correct, already-independent lever for "how long to keep `.trash`" (an operator wanting to never purge can set it very large); coupling the purge to the unrelated archive-decision gate was itself the pre-existing gap. New test `purge_trash_runs_regardless_of_worktree_gc_gate_2550_w5` pins this.

**5-struct convergence (per lead's ruling after the PR-1 premise check)**: this PR didn't produce a natural common candidate shape to converge (the unification works directly on `GcCandidate`, no new duplicate struct emerged) — residual struct inventory: `TargetSweepCandidate`/`WorktreeEntry`/`branch_sweep::Candidate` still serve genuinely distinct call sites (disk-usage sweep, branch-registry cleanup, operator-triggered branch hygiene respectively) with no forced overlap found. Recommend closing the struct-convergence item as N/A rather than a follow-up PR — happy to expand if you want a second look.

**W1 #2579 pin tests** rewritten per the spike's own mapping table: single-path tests untouched; the combined-pipeline test now exercises `gc_run` alone (renamed — no more separate `sweep()` call to reference), plus the new gate-off-by-default test.

**Also folds in** two review nits found while merging PR-0 (same file, not new scope): stale comments in `cleanup_merged_branch` still referencing the removed `is_gone` variable, and its skip_reason string updated to cover the squash-eligible check it now also performs (two pre-existing test assertions updated to match the new string).

## Test plan
- [x] New/rewritten pin tests pass (7/7, including the Q1 fallthrough test, the gate-off-by-default test, and the new `purge_trash` gate-independence test).
- [x] `cargo nextest run --profile ci` — 5174/5176 (the 2 failures are the pre-existing documented flakes `e2e_dispatch_review_workflow_seam_invariants_1900` / `teardown_leaves_zero_residual_after_full_exercise_1907`, unrelated to this diff).
- [x] `cargo clippy --lib --tests -- -D warnings` clean.
- [x] `cargo fmt --check` clean.
- [x] No new `.trash` bypass path — `gc_remove_one`'s hard-delete path is unchanged except for routing skip cases to the SAME `maybe_remove_candidate` archive path ForceReclaim already used.

### DIFF
--- a/src/daemon/per_tick/gc_tick.rs
+++ b/src/daemon/per_tick/gc_tick.rs
@@ -37,6 +37,11 @@ impl PerTickHandler for GcTickHandler {
         if removed > 0 || failed > 0 {
             tracing::info!(removed, failed, "gc_tick: worktree GC complete");
         }
+        // #2550 W5: single driver now owns the whole worktree-GC lifecycle,
+        // including `.trash` purging — previously a separate ~1h-delayed
+        // independent sweep's job (`retention::worktrees::sweep`), now folded
+        // in here on the SAME fire-on-first cadence (no phase lag, decision Q2).
+        crate::daemon::retention::worktrees::purge_trash(ctx.home);
 
         let stale_locks = crate::worktree_pool::gc_stale_ci_watch_locks(ctx.home);
         if stale_locks > 0 {
--- a/src/daemon/per_tick/gc_tick.rs
+++ b/src/daemon/per_tick/gc_tick.rs
@@ -41,6 +41,19 @@ impl PerTickHandler for GcTickHandler {
         // including `.trash` purging — previously a separate ~1h-delayed
         // independent sweep's job (`retention::worktrees::sweep`), now folded
         // in here on the SAME fire-on-first cadence (no phase lag, decision Q2).
+        //
+        // NOTE (behavior change, deliberate, NOT covered by decision Q3's
+        // "gate coverage unchanged"): this call is UNCONDITIONAL, unlike the
+        // old `sweep()` it replaces, which only ran (and so only purged
+        // `.trash`) when `AGEND_WORKTREE_GC=1`. Q3 preserved the gate on the
+        // ARCHIVE-DECISION path (CleanRelease's fallthrough); it says nothing
+        // about cleaning up entries already IN `.trash`, which can land there
+        // via the ForceReclaim path (never gated, before or after this PR) —
+        // in a gate-off install, those entries used to accumulate forever
+        // (nothing ever called `purge_trash`). `AGEND_WORKTREE_GC_TRASH_DAYS`
+        // is the correct, already-independent lever for "how long to keep
+        // `.trash`" (including "never purge": set it very large); coupling
+        // the purge to the unrelated archive-decision gate was itself the gap.
         crate::daemon::retention::worktrees::purge_trash(ctx.home);
 
         let stale_locks = crate::worktree_pool::gc_stale_ci_watch_locks(ctx.home);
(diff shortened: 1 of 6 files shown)


suzuke:
- seat-1 gapfix-reviewer3: **VERIFIED** (verdict msg m-20260704052123140802-154, reviewed_head 9bd8e912) — all 7 gc_remove_one exit paths hand-traced (fallthrough on lock-fail/repo-unresolved/double-remove-fail/spawn-err; correctly NO fallthrough on re-validation failure); Q3 gate coverage byte-equivalent to old sweep(); all pin tests + 147-test targeted sweep run.
- seat-2 claude-aef7c0 (lead): **VERIFIED** — full 556-line diff review + incremental patch review; purge_trash gate change validated as fixing a pre-existing unbounded .trash accumulation gap (documented in PR body + pinned by test).

**Known non-blocking issue (reproduced, filed)**: `gc_run_archive_fallthrough_stays_off_by_default_2550_w5` lacks the `gc_trash_env_guard()` its sibling env-mutating tests hold — flakes 3/3 under plain multi-threaded `cargo test`, structurally immune under nextest (per-test-process isolation), so CI is unaffected. One-line fix rides along in W5 PR-3.

Merging with --admin per branch-protection convention.

Issue #991, opened 2026-05-20
feat: create_instance topic_binding param — opt-out telegram topic for internal-only fleet members


suzuke:
## Problem

Today, every `create_instance` call (and every team-deploy member) gets a Telegram topic auto-created. That's the right default for outward-facing roles (lead, operator-callable agents), but **wasteful + noisy for internal-only roles** like dev-2 / reviewer-2 / orchestrator helpers — they only ever receive dispatch via inter-agent `send`, the operator never opens their topic directly, and each unused topic still costs one Telegram API call at spawn time + permanent clutter in the channel.

The current implicit behavior leaks the assumption "every agent talks to the operator" into a system where that's only true for some roles.

## Current workaround (if any)

None. Every spawn auto-creates a topic. Operator can `delete_instance` and respawn, but that's destructive; they can ignore the topic, but it still clutters the Telegram channel.

## Proposed behavior

Add a `topic_binding` parameter to `create_instance` (and propagate to `team action=create` / `deployment action=deploy` per-member configs):

| value | behavior |
|---|---|
| `"auto"` (default) | Current behavior — topic auto-created at spawn. Preserves back-compat. |
| `"skip"` | No topic ever created. Agent is reachable only via inter-agent `send`. fleet.yaml records `topic_binding_mode: skip` so this is **distinguishable from a silent failure** (cf. #962-class silent-degrade discipline). |
| `"deferred"` | No topic at spawn, but operator can retrofit later via a new `bind_topic` MCP action. Useful for agents that start internal-only and later need operator visibility. |

A new `bind_topic` MCP action takes `instance_name` and creates the topic for a `deferred`- or `skip`-mode instance, updating `fleet.yaml` + `topics.json` atomically.

## Scope boundary

**IN scope**:
- `create_instance` MCP handler: add `topic_binding` param, gate `create_topic_for_instance` call on it
- `team` / `deployment` template schemas: per-member `topic_binding` field; team-level default if all members share
- `fleet.yaml`: persist `topic_binding_mode` field per instance so post-hoc readers know the state is intentional
- New `bind_topic` MCP action for retrofit
- Tests: each mode + retrofit round-trip
- Estimate: ~80 LoC additive

**NOT IN scope**:
- Changing the default behavior — `"auto"` stays the default to preserve back-compat and operator visibility (the safer default; opt-out is on the caller, not opt-in)
[...]

PR #2598, opened 2026-07-04
feat(telegram #991 PR-A): bind_topic MCP action — retrofit a topic for deferred instances

suzuke:
## What & why

#991 (`topic_binding: deferred`) let an instance spawn without a Telegram topic, but never shipped the promised retrofit action — `bind_topic` was declared in docs but not implemented (confirmed in a design spike, `workspace/gapfix-dev3/BIND-TOPIC-PRERESEARCH.md`; docs already corrected in #2594). This is PR-A of that spike's PR-A/B/C split: a new `bind_topic` MCP action that creates a Telegram topic for an instance stuck in `deferred` (or `auto` that ended up without one), and atomically updates `fleet.yaml` (`topic_id`) + `topics.json`.

**Decision A (skip-eligibility, lead-vetted before implementation):** `bind_topic` refuses `topic_binding_mode: skip` instances with `NotEligible` — skip's literal promise is "no topic, ever," and retrofitting it silently would defeat the point of recording that choice explicitly. To bind a skip-mode instance, the operator must change `topic_binding_mode` first, then call `bind_topic`. `auto`-mode instances that ended up without a topic (e.g. spawned during the ~6s post-boot `telegram_init` window) are NOT refused — only `skip` is.

**Decision B (multi-channel `active_channel()` blind spot, found during the spike):** out of scope for this PR — tracked separately (`t-20260703164240502572-50899-11`). `bind_topic` avoids inheriting the bug by taking an explicit `channel` param (defaults to `"telegram"`, the only channel implemented today; any other value gets a clear "not yet supported" error) instead of the ambiguous `active_channel()` resolver.

Closes #991 (Phase 2, PR-A of 3).

## Prior art check (required)

- [x] Checked `docs/KNOWN_ISSUES.md` — no entry for `bind_topic`/#991.
- [x] Compared against current `main` — `bind_topic` is not implemented there (confirmed by spike + `gh pr list` search for prior attempts).
- [x] Searched closed issues/PRs — #991 itself is the only prior art; it was closed with the `deferred` mode shipped but `bind_topic` explicitly left as a TODO ("operator can retrofit later via a new `bind_topic` MCP action").
- [x] Not revisiting an existing decision — this is the first implementation.

## How it was verified

[...]

PR #2600, opened 2026-07-04
feat(telegram #991 PR-B): deployment/team-mode templates propagate topic_binding

suzuke:
## What & why

#991's issue text explicitly lists this as in-scope: "team and deployment templates accept `topic_binding` per-member; the field flows through to per-instance `create_instance` calls." Both spawn paths were silently dropping it:

1. **Deployment templates** (`deployments.rs:235`): hardcoded `topic_binding_mode: None` — every other per-instance template field (`repo`/`model`/`args`/`env`/...) was read from the template stanza, this one wasn't. Fixed to read `topic_binding` with the same skip/deferred-only filter `create_instance` uses (`spawn.rs`) — an invalid value or omission stays `None` (unchanged auto default). Also wired `spawn_instances` to forward the persisted mode as the SPAWN call's `topic_binding` param — without this, the field would land correctly in fleet.yaml but a topic would still get created anyway (SPAWN defaults an absent `topic_binding` to `"auto"`).

2. **Team-mode multi-spawn** (`create_instance team=X, count=N` / `backends=[..]`): had **zero** `topic_binding` support at any layer. Note for reviewers: the dispatching note for this PR assumed the fix belonged in `teams.rs` ("teams.rs 對應補上") — that turned out to be the wrong file. `teams.rs` only manages the team roster (plain member-name strings; no spawn, no per-field hooks — nothing to wire there). The actual spawn path is CREATE_TEAM's daemon handler, `api/handlers/team.rs::handle_create_team`, which calls `agent_ops::spawn_one` directly (bypassing `handle_spawn`'s existing single-instance topic_binding gate entirely) and unconditionally called `ensure_topic_for` for every member. Added a **team-level** `topic_binding` param — shared by every spawned member, the issue's explicitly-permitted "team-level default if all members share" case (not a per-member array like `backends`) — forwarded from the MCP handler through CREATE_TEAM's params, persisted onto each member's `InstanceYamlEntry`, and gates `ensure_topic_for` the same way `handle_spawn` already does for single-instance spawns.

Refs #991 (PR-B of the #991 completion series; PR-A merged as `a0bf79e6`).

## Prior art check (required)

- [x] Checked `docs/KNOWN_ISSUES.md` — no entry for this gap.
- [x] Compared against current `main` (`a0bf79e6`, PR-A merged) — neither deployment templates nor team-mode spawn read `topic_binding` there.
- [x] Searched closed issues/PRs — #991 itself (closed) explicitly lists this as in-scope but never implemented; this PR completes it.
- [x] Not revisiting an existing decision — first implementation of this sub-scope.

## How it was verified

- New tests, all green:
  - `deployments::tests::deploy_persists_topic_binding_into_fleet_yaml` — template `topic_binding: skip`/`deferred`/an invalid value persist correctly (invalid → stays `None`, same as an invalid `create_instance` call).
  - `deployments::tests::deploy_omits_template_params_when_not_set_backwards_compat` — extended with a `topic_binding_mode.is_none()` assertion (omitted field → unchanged auto default).
  - `api::handlers::team::tests_1964::build_member_entries_carries_topic_binding_mode_991` — the new `build_member_entries` helper correctly carries (or omits) `topic_binding_mode` onto every planned member's fleet.yaml entry.
- **Test-boundary note**: `agent_ops::spawn_one` forks a real PTY process and has **no direct unit-test coverage anywhere in this codebase already** (confirmed by grep — too dangerous/unpredictable to invoke in tests; every existing `handle_create_team` test in this file avoids the real spawn path via `count=0` + pre-listed `members`). Rather than force a test through that path, extracted the entry-construction logic into `build_member_entries` — mirroring the existing `plan_member_names` precedent in the same file — so the new propagation logic is unit-testable without a real spawn attempt.
- `cargo test --workspace --tests --features tray` (the exact CI invocation): 4529 passed. Zero regressions — confirmed via `git stash` A/B: **identical** 257-test failure set with and without this diff (pre-existing resource-contention flakes across `worktree_pool`/`admin`/`binding`/etc. under full-parallel execution, present on `main` already, not caused by this change).
- `cargo fmt --check`: clean.
- `cargo clippy --all-targets --features tray -- -D warnings`: clean.

- [x] No tier (a)/(b) on-disk format is changed, **or** the change is additive-only — `topic_binding_mode` on `InstanceYamlEntry` already existed (added in PR-A); this PR just adds two more write sites for it (deployment template passthrough, team-mode spawn), same field, same semantics.

## Notes for reviewers

- Deliberately **team-level** `topic_binding` (one value, applies to every spawned member), not a per-member array like `backends`. The issue text explicitly permits this simpler shape ("team-level default if all members share"), and it keeps this PR proportionate — a full per-member array would need a parallel `topic_bindings: Vec<String>` param threaded the same way `backends` is, which felt like scope creep beyond what #991 asks for. Happy to add it separately if wanted.
- PR-C (`list_instances` exposing `topic_binding_mode` — the last item from the original #991 acceptance criteria) is next, per the preresearch's PR split.


suzuke:
- seat-1 gapfix-reviewer3: **VERIFIED** (verdict msg m-20260704055825681067-219, reviewed_head e1c8bfd6) — LOC split verified as purely mechanical to whitespace-normalized equivalence (52 tests, name-set diff empty; GRANDFATHERED removal mandatory per the invariant's own rule); topic_binding forwarding traced to the real gate (api/handlers/instance.rs SPAWN handler); team-mode omitted-param path byte-equivalent to pre-PR.
- seat-2 claude-aef7c0 (lead): **VERIFIED** — skip/deferred filter consistent across template/team/single-spawn paths; team-mode gate correctly placed inside the spawn loop with audit log; deferred semantics close the loop with #2598's bind_topic (no spawn-time topic, retrofit later).

Scope-expansion (mid-PR LOC split) explicitly adjudicated: legitimate mechanical CI-gate fix, not smuggled scope. Merging with --admin per convention.

PR #2602, opened 2026-07-04
test(#2550): fix env-var race in gc_run_archive_fallthrough_stays_off_by_default_2550_w5

suzuke:
## Summary

Fixes t-20260704052055927704-67777-2 (reviewer3, found reviewing #2599, low priority, non-blocking).

The new gate-off pin test from PR #2599 was missing the `gc_trash_env_guard()` its sibling tests use to serialize access to the process-global `AGEND_WORKTREE_GC` env var. Under plain multi-threaded `cargo test` it could race a concurrent test that temporarily sets `AGEND_WORKTREE_GC=1`, reading the wrong value mid-test and failing this gate-off assertion. Reproduced 3/3 under plain `cargo test`; `cargo nextest run` (this repo's actual CI method) isolates env vars per test process and was structurally immune, which is why CI never saw it.

Also drops an unnecessary `remove_var` call from the `gc_tick.rs` `purge_trash` test added in the same PR — that test's `home` has zero GC candidates, so `gc_run` never reaches the gate check either way, and the call would have raced the same env var across files for no benefit.

## Test plan
- [x] 5x plain `cargo test --bin agend-terminal daemon::retention::worktrees::tests::` — all green (was 3/3 failing before this fix).
- [x] `cargo nextest run --profile ci` on the affected suites — clean.
- [x] `cargo clippy --lib --tests -- -D warnings` clean.
- [x] `cargo fmt --check` clean.

### DIFF
--- a/src/daemon/per_tick/gc_tick.rs
+++ b/src/daemon/per_tick/gc_tick.rs
@@ -130,10 +130,15 @@ mod tests {
     /// purged them). Deliberately NOT covered by decision Q3 ("gate coverage
     /// unchanged" applies to the archive-DECISION path only); this is a
     /// separate, intentional fix. Proves an aged `.trash` entry is purged by
-    /// a `gc_tick` pass with `AGEND_WORKTREE_GC` left UNSET.
+    /// a `gc_tick` pass regardless of `AGEND_WORKTREE_GC` — deliberately does
+    /// NOT touch that process-global env var itself (would race
+    /// `retention::worktrees`'s own gate tests under plain multi-threaded
+    /// `cargo test`, across a private per-file mutex nextest doesn't need but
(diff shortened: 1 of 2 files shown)


suzuke:
Single-seat lead review (trivial test-only change; §3.12): **VERIFIED** — adds the missing `gc_trash_env_guard()` to the gate-off pin test (reviewer3's reproduced plain-cargo-test flake, 3/3 pre-fix), and removes the now-superfluous `remove_var` in the gc_tick purge test whose home has no GC candidates (that line itself raced the same process-global env var cross-file). Verified: plain cargo test ×5 green post-fix, nextest unaffected, zero production code touched. Merging with --admin per convention.

PR #2415, opened 2026-06-22
fix(srl): override sticky-screen ServerRateLimit false-retry with a fresh claude hook

suzuke:
## Bug (operator-reported, HIGH, #t-26795)
With `AGEND_OFFTHREAD_PARSE` off-thread aside — a claude agent that RECOVERED from a rate-limit kept getting `[AGEND-AUTO ratelimit-retry] continue` injected. RCA (lead, log-nailed; quantified in the spike): the screen-scraper stays stuck on residual "Server is temporarily limiting requests" text (`screen_state=ServerRateLimit`) while the agent is provably firing tool calls — the per-tool-call HOOK reports `ToolUse/Thinking` but is SHADOW-only and doesn't drive the per-tick SRL-retry decision. The slow productive-output `recovered` signal FLAPS (detect→clear→re-detect), so retries keep firing. Shadow data: 1996/2746 recovery records show a screen↔hook mismatch, **100% "hook active, screen stale"**; the incident had **14 consecutive ticks** of stable-active-hook vs sticky-SRL.

## Fix (narrow, lead-vetted)
A FRESH claude hook (active `ToolUse/Thinking`) whose timestamp **post-dates the SRL onset** proves the agent recovered → added as a **third disjunct** `hook_recovered` to the existing SRL clear gate (`supervisor.rs`), composing with `recovered`/`self_cleared` (ADD-ONLY, no change to the productive-output path).

- **Stable onset** (`srl_onset` map): the FIRST-detection epoch-ms, `or_insert` never overwrites → survives the detect→clear→re-detect flap (a resetting onset would let a just-pre-re-detect hook look pre-onset and the flap would survive — the load-bearing requirement).
- **post-onset** requirement: rejects a fresh-but-prior-turn hook → a genuine new-turn SRL is never masked.
- **claude-only** (`Backend::has_state_hooks`); **decoupled from `AGEND_HOOK_STATE_POC`** (reads `hook_shadow::fresh_active_hook_at_ms`, populated by the unconditional `record_event`) — an operator-facing fix must not hang on a POC flag; fail-safe (missing/stale/pre-onset hook → unchanged screen-driven retry).
- `Idle` deliberately excluded from "active" (turn-ended/ambiguous; left to the productive-output path).

This is the highest-leverage #1523 phase-2 promotion slice (the per-tick recovery dispatcher), scoped to the SRL case only.

## Tests
[...]

PR #2603, opened 2026-07-04
fix(daemon): stop ci-ready renudge orphaning + fix inbox=N mislabeling (#26795)

suzuke:
## Summary
Phase 2 of the phantom renudge RCA (`PHANTOM-RENUDGE-RCA.md`, task `t-20260622175035095973-26795-10`). Two independent fixes:

**(A) REJECTED/no-active-watch orphan track class** (sample: #2415 — "REJECTED head, `ci action=status` shows `watches:[]`, renudge every ~2min"). Every existing auto-resolver for a `ci_handoff_track` (`resolve_head_advanced`, `pr_state::scanner`'s `pr_merge_blocked` resolve) depends on an active `ci_watch` continuing to poll/re-scan that branch. Once polling stops for a branch (superseded, torn down, or never re-armed after REJECTED), none of them ever run again except the 24h backstop — exactly the "no dismiss path" the task title names.

`resolve_if_merge_blocked` reads the branch's **cached** `pr_state` snapshot directly inside the watchdog's own per-tick scan, before deciding to re-nudge. That snapshot is written by `pr_state::record_verdict` the moment a reviewer's report is processed — independent of `ci_watch` state entirely, no GitHub call. A REJECTED/Draft verdict now self-heals within one watchdog tick (~2 min) instead of waiting on a poll cycle that may never happen again.

Scope: read-only `pr_state` lookup; does not touch `resolve_head_advanced` (still needs live polling for genuine head-advance detection) or the unrelated `poll_reminder`/obligation family (confirmed different root cause in the RCA).

**(B) `inbox=N` mislabeling.** `renudge_actionable_unread` never enqueues anything to the inbox (direct PTY inject via `build_pending_pointer`, shared with the real inbox-pending pointer) — yet its header claimed `inbox={pending_track_count}` and told the agent to "(use inbox tool)", sending them to look for a message that was never there. New `build_renudge_pointer` keeps the same wire shape (still `kind=ci-ready-for-action`, still recognized as an actionable wake) but labels the count honestly as `pending_handoffs` and points at `ci action=status` instead.

## Test plan
- [x] RED-first for (A): `renudge_skips_when_pr_state_shows_merge_blocked` reproduces the orphan-track bug before the fix, passes after
- [x] (B): `renudge_pointer_does_not_claim_inbox_count` pins the new wording and confirms actionable-wake detection is unaffected
- [x] `cargo test --bin agend-terminal daemon::handoff_timeout_watchdog::` — 13 passed
- [x] `cargo test --bin agend-terminal daemon::ci_handoff_track::` — 15 passed
- [x] `cargo test --bin agend-terminal daemon::pr_state::` — 91 passed
- [x] `cargo test --bin agend-terminal inbox::notify::` — 20 passed
- [x] Zero existing tests modified

### DIFF
--- a/src/daemon/handoff_timeout_watchdog.rs
+++ b/src/daemon/handoff_timeout_watchdog.rs
@@ -41,6 +41,40 @@ const FALLBACK_RECIPIENT: &str = "lead";
 /// Inbox kind of a CI handoff message.
 const HANDOFF_KIND: &str = "ci-ready-for-action";
 
+/// #26795: if `correlation`'s (`owner/repo@branch`) PR is already known to be
+/// merge-blocked (REJECTED verdict or Draft) per the CACHED `pr_state`
+/// snapshot, resolve its `ci_handoff_track` entry and report `true`. Read-only
+/// from `ci_watch`'s perspective — this never touches the watch registry or
+/// makes a GitHub call, only reads whatever `pr_state::record_verdict` last
+/// wrote when a reviewer's report was processed. Returns `false` (no-op) if
+/// no `pr_state` file exists yet for this correlation, or its verdict isn't
+/// blocking.
+fn resolve_if_merge_blocked(home: &Path, correlation: &str) -> bool {
+    let Some((repo, branch)) = correlation.split_once('@') else {
+        return false;
+    };
+    let blocked = crate::daemon::pr_state::with_pr_state(home, repo, branch, |s| {
+        crate::daemon::pr_state::is_ci_ready_merge_blocked(s)
+    })
+    .ok()
+    .flatten()
+    .unwrap_or(false);
+    if blocked {
+        tracing::info!(
+            tag = "#26795-track-merge-blocked",
+            %correlation,
+            "ci-handoff track resolved — pr_state cache shows REJECTED/Draft; \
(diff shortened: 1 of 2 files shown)


suzuke:
- seat-1 gapfix-reviewer3: **VERIFIED** (verdict msg m-20260704060618286214-243, reviewed_head e99f14ad) — read-only claim verified at the primitive level; rework-safety traced to the pre-existing CiObserved head-advance verdict-reset invariant (pr_state/mod.rs:277-306) that structurally protects fresh tracks from a stale REJECTED resolve; all 4 module suites run (13/15/91/20), zero existing tests modified.
- seat-2 claude-aef7c0 (lead): **VERIFIED** — fail direction correct (cache miss → fail-open to noise, never silence); scope exactly the adjudicated A+B (4-signal gating untouched); RED baseline genuine.

**Non-blocking hygiene nit (seat-1)**: `with_pr_state` is a lock-read-mutate-unconditionally-rewrite primitive, so this read-only check re-serializes pr_state.json every watchdog tick per active track — harmless correctness-wise, pure I/O overhead. Candidate one-liner for a future pass (read-only accessor), not blocking.

Merging with --admin per convention.

PR #2605, opened 2026-07-04
refactor(#2550 PR-3): sweep_from_registry -> own GC-owned 60-tick handler

suzuke:
## Summary

Last PR of the #2550 W5 GC-unification sequence (PR-0 #2597, PR-2 #2599, test-hygiene #2602, this PR-3). Per `P3-2550-SPIKE.md` §5 Wave 5 + decision Q4 (`d-20260704035059093740-0`).

Moves the worktree-registry auto-cleanup sub-op (drops local branches whose PRs merged into main, via the runtime config registry + `worktree_cleanup::sweep_from_registry` — a different mechanism from the marker-based GC candidates PR-2 touches) out of `InboxMaintenanceHandler` into its own `WorktreeRegistrySweepHandler`. It has nothing to do with inbox maintenance and is semantically GC.

**Per Q4's explicit ruling, this does NOT fold into `HourlyGcHandler`'s 360-tick cadence.** The new handler keeps the exact same 60-tick cadence value it had inside `InboxMaintenanceHandler` — merging it into the hourly cadence would regress this cleanup's latency from ~10min to ~1h, a real operator-visible regression that was evaluated and not approved (see `W5-GC-UNIFY-PRERESEARCH.md` §3).

Extraction is verbatim: function body, doc comments, and logic are unchanged. `inbox_maintenance.rs`'s remaining 4 sub-ops (inbox `sweep_expired`, `reclaim_stale_delivering`, `check_disk_space`, task maintenance) and their cadence are untouched.

## Test plan
- [x] New handler's cadence + smoke tests pass (2/2).
- [x] `cargo nextest run --profile ci` — 5190/5192 (the 2 failures are the pre-existing documented flakes `e2e_dispatch_review_workflow_seam_invariants_1900` / `teardown_leaves_zero_residual_after_full_exercise_1907`, unrelated to this diff).
- [x] `cargo clippy --lib --tests -- -D warnings` clean.
- [x] `cargo fmt --check` clean.
- [x] Existing `inbox_maintenance` tests and `worktree_cleanup::sweep_from_registry`'s own tests unmodified.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2607, opened 2026-07-04
fix(daemon): fix canonical-repo branch-prune repo discovery + gate first activation (#2605)

suzuke:
## Summary
Root-cause fix for task `t-20260704054810920172-67777-3` (canonical repo local-branch accumulation, `BRANCH-AUDIT-20260704.md`: 89 safe-to-delete branches + stale worktrees).

`worktree_cleanup::sweep_from_registry`'s repo set came from `AgentConfig.worktree_source`, a spawn-time cache derived from `working_dir` via the LEGACY `{repo}/.worktrees/...` layout pattern (`worktree::source_repo_of`). Under the current workspace-spawn + post-spawn-bind architecture that pattern never matches, so the repo set has always been empty — this module's git-mutating paths (worktree removal, orphaned-branch pruning) have never actually run against the canonical repo in production. Evidence: 0 `worktree_auto_removed` events in `event-log.jsonl` since May.

**Fix (option b, per audit):** repo discovery now reads live `binding.json` state via the existing `binding::bound_source_repos` — self-heals for binds that happen after spawn, unlike the spawn-time cache. `AgentConfig.worktree_source` and its sole populator (`worktree::source_repo_of`, itself now zero-callers after this change) are deleted rather than left as dead-but-plausible code.

**First-activation safety gate:** fixing repo discovery activates a delete path with zero production track record. Real deletion is additionally gated **opt-in** via `AGEND_WORKTREE_PRUNE_LIVE=1` (independent of the already-on-by-default `AGEND_WORKTREE_AUTO_CLEANUP`, which can't double as this gate since flipping it "on" is already the status quo). Default is dry-run: the exact same eligibility (squash gate, worktree-occupancy fail-closed check, #1750-B3 squash-orphan GC — all untouched) is computed, candidates are logged under a new `worktree_prune_dry_run_candidate` event kind (mirrors `worktree_auto_removed`), nothing is deleted. An operator diffs the dry-run output against a fresh branch audit before flipping the gate.

**Fetch interference disclosure:** the background `git fetch --prune` this activates (previously a no-op since `repos` was always empty) now runs against the canonical repo every ~10 daemon ticks — a new recurring background operation on the operator's working repo. It runs unconditionally, including during dry-run, since part of the observation window's purpose is to surface this fetch's real-world behavior (not just the deletion candidates). Failures are logged (not silently swallowed, as before) per the existing `#2004` fail-safe pattern from `worktree_pool::cleanup_merged_branch`: a stale local ref only makes merge/gone checks more conservative, never more aggressive, so a failing fetch is safe-directioned but should not go unnoticed.

**Dead-code note:** `worktree_source` is a name collision trap — `fleet::InstanceConfig::git_branch` carries `#[serde(alias = "worktree_source")]` for an unrelated TS-version back-compat key. That field is untouched; only `daemon::mod::AgentConfig.worktree_source` (a pure runtime field, `#[derive(Clone)]` only, never (de)serialized) was removed.

## Test plan
- [x] RED-first for #2605: `prune_live_disabled_by_default`, `prune_orphaned_branches_dry_run_reports_but_keeps_branch_2605`, `test_sweep_dry_run_by_default_identifies_but_does_not_delete`
- [x] All existing `worktree_cleanup` tests re-plumbed to home+binding.json-based repo discovery; explicit `AGEND_WORKTREE_PRUNE_LIVE=1` set where they assert real deletion (behavior unchanged, just re-wired to the new gate)
- [x] `cargo test --bin agend-terminal worktree_cleanup::` — 21 passed
- [x] `cargo test --bin agend-terminal binding::` — 65 passed
- [x] `cargo test --bin agend-terminal daemon::crash_respawn::` — 4 passed
- [x] `cargo test --bin agend-terminal daemon::per_tick::inbox_maintenance::` — 2 passed
- [x] `cargo test --bin agend-terminal daemon::tests::` — 39 passed
- [x] `cargo test --bin agend-terminal worktree::tests::` — 55 passed
- [x] `cargo nextest run --bin agend-terminal` (full suite) — 4788 passed, 0 failed (1 slow + 1 leaky, pre-existing known daemon-boot-timing flake class, not from this change)

## Rollout (post-merge, not part of this PR)
1. Observe dry-run `worktree_prune_dry_run_candidate` events for a period.
2. Diff candidates against a fresh branch audit (same shape as `BRANCH-AUDIT-20260704.md`).
3. Once confirmed to match, operator sets `AGEND_WORKTREE_PRUNE_LIVE=1` to go live.

### DIFF
--- a/src/daemon/crash_respawn.rs
+++ b/src/daemon/crash_respawn.rs
@@ -533,7 +533,6 @@ mod deleted_gate_tests_1913 {
                 args: vec![],
                 env: None,
                 working_dir: None,
-                worktree_source: None,
                 submit_key: "\r".to_string(),
             },
         );
@@ -623,7 +622,6 @@ mod deleted_gate_tests_1913 {
             args: vec![],
             env: None,
             working_dir: None,
-            worktree_source: None,
             submit_key: "\r".to_string(),
         };
 
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -273,8 +273,6 @@ pub struct AgentConfig {
     pub args: Vec<String>,
     pub env: Option<HashMap<String, String>>,
     pub working_dir: Option<PathBuf>,
-    /// Original repo root (before worktree redirect).
-    pub worktree_source: Option<PathBuf>,
     pub submit_key: String,
 }
 
@@ -1673,9 +1671,6 @@ fn spawn_and_register_agent(
         );
         return Ok(());
     }
-    let worktree_source = working_dir
-        .as_ref()
-        .and_then(|wd| crate::worktree::source_repo_of(wd));
     configs.lock().insert(
         name.clone(),
         AgentConfig {
@@ -1684,7 +1679,6 @@ fn spawn_and_register_agent(
             args: args.clone(),
             env: env.clone(),
             working_dir: working_dir.clone(),
-            worktree_source,
             submit_key: submit_key.clone(),
         },
     );
@@ -2228,7 +2222,6 @@ mod tests {
                 args: vec![],
                 env: None,
                 working_dir: None,
-                worktree_source: None,
                 submit_key: "\r".into(),
             },
         );
--- a/src/daemon/per_tick/worktree_registry_sweep.rs
+++ b/src/daemon/per_tick/worktree_registry_sweep.rs
@@ -43,19 +43,19 @@ impl PerTickHandler for WorktreeRegistrySweepHandler {
 /// PRs have merged into main. Logged via `event_log` + tracing on every
 /// removal so operators can audit. Verbatim from `inbox_maintenance.rs`
 /// (was verbatim from the pre-extraction block at mod.rs:678-717).
+///
+/// #2605: repo discovery moved to live `binding.json` state
+/// (`sweep_from_registry` reads it via `home`) instead of the removed
+/// `AgentConfig.worktree_source` cache. Real deletion is additionally gated by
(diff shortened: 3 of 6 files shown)


suzuke:
- seat-1 gapfix-reviewer4: initial **REJECTED** (m-20260704071207559611-359, head 44ef6d8a) — caught the dry-run audit events losing the real candidate reason (hard-coded 'merged into main' for remote-gone/squash-merged candidates, empty orphan path), which seat-2 had missed. Re-stamp **VERIFIED** (m-20260704081941601169-464, reviewed_head 81afd4c2; full evidence workspace/gapfix-reviewer4/VERDICT-2607.md, 17 ran/cited items) — (branch, path, reason) tuple with true reasons, orphan path '(no worktree)', reason assertions added to 4 pin tests.
- seat-2 claude-aef7c0 (lead): **VERIFIED** — opt-in AGEND_WORKTREE_PRUNE_LIVE gate independent of the opt-out auto-cleanup gate; eligibility computation untouched by dry_run; fetch deliberately ungated for observation-window fidelity (#2004 fail-safe); worktree_source dead code fully removed.
- §3.10 test-first exemption recorded: repo convention is squash single-commit push; RED evidence assessed via test-logic review.

Rollout: merge activates DRY-RUN ONLY (candidates logged as worktree_prune_dry_run_candidate). Live deletion stays off until the dry-run output is diffed against BRANCH-AUDIT-20260704.md and the operator flips AGEND_WORKTREE_PRUNE_LIVE=1. Merging with --admin per convention.

Issue #2595, opened 2026-07-03
bug(dispatch): ensure_branch_exists pre-create fetch does not update remote-tracking ref for from_ref


cheerc:
### Symptom
When a dispatch triggers branch creation from a base reference `from_ref` (e.g., `origin/dev`), the newly created branch is occasionally based on a stale commit (lagging behind the remote default branch by 1 or more commits).

### Root Cause
In `src/mcp/handlers/dispatch_hook/mod.rs` around line 863, the pre-create fetch executes:
```rust
        let fetch_out = crate::git_helpers::git_bypass_timeout(
            source,
            &["fetch", &remote, remote_branch, "--quiet"],
            DISPATCH_FETCH_TIMEOUT,
        );
```
Under standard Git configurations, running `git fetch <remote> <branch>` fetches the objects into `FETCH_HEAD` but does **not** update the local remote-tracking reference `refs/remotes/origin/<branch>`.
Consequently, when the daemon later attempts to create the new branch via:
```rust
    match crate::git_helpers::git_bypass(source, &["branch", branch, from_ref]) {
```
where `from_ref` resolves to `origin/<branch>` (which maps to `refs/remotes/origin/<branch>`), it resolves to the **stale local copy** of that tracking branch instead of the freshly fetched commit.

### Suggested Fix
To ensure the remote-tracking ref is updated during the fetch, the daemon should either:
1. Run `git fetch <remote>` (fetching all refs matching the remote refspec).
2. Or explicitly map the refspec in the fetch command, e.g., `git fetch <remote> <branch>:refs/remotes/<remote>/<branch>`.

suzuke:
Thanks for the detailed report — given your track record here (#2010, #2107) we took this seriously and verified the core git claim empirically before responding. The claimed mechanism doesn't hold for standard configurations:

**Since git 1.8.4 (2013), `git fetch <remote> <branch>` performs an opportunistic remote-tracking ref update** — the fetched ref is mapped through `remote.<name>.fetch` and `refs/remotes/<remote>/<branch>` IS updated, not just FETCH_HEAD (git-fetch(1), "CONFIGURED REMOTE-TRACKING BRANCHES").

Reproduction (git 2.39.5, normal clone, remote `dev` advanced by a second clone):
```
=== Scenario A: normal clone (default refspec +refs/heads/*:refs/remotes/origin/*) ===
before fetch : 27c63650f52dc0f97b357cd960dc8c9e8258b274
[...]

PR #2610, opened 2026-07-04
fix(dispatch): use explicit refspec to ensure remote-tracking ref is updated

cheerc:
### Description
This PR implements **suggested fix (2)** from #2595 to harden the pre-create fetch refspec:
```rust
&format!("{remote_branch}:refs/remotes/{remote}/{remote_branch}")
```
By explicitly specifying the refspec, we guarantee that the remote-tracking reference is updated during the pre-create fetch, ensuring the subsequent `git branch` command resolves against the latest remote tip.

This byte-equivalent change preserves existing behavior for standard clones while hardening the single-branch clone corner cases.

Refs #2595

### Verification Results
Empirical local tests were conducted across both configurations.

#### Scenario A: Standard clone (default refspec `+refs/heads/*:refs/remotes/origin/*`)
```
before fetch: e7750c9b274fa408270c11ea6633a4dab07db4ad
after fetch:  0f4ed8517258564a27d06d76dfbb88c9b8010103
remote tip :  0f4ed8517258564a27d06d76dfbb88c9b8010103
```

#### Scenario B: Single-branch clone (tracking refs configured for `main` only)
```
before fetch: 967877a2b7aa434103a6b35583dd22d1949cd8de
after fetch:  0f4ed8517258564a27d06d76dfbb88c9b8010103
remote tip :  0f4ed8517258564a27d06d76dfbb88c9b8010103
```
In both scenarios, the stale local remote-tracking reference is successfully updated to match the remote tip.


suzuke:
Thanks for following through on the invitation from #2595 — the direction is exactly right, but our review found a one-character gap that matters:

**The explicit refspec needs a `+` (force) prefix.** Reproduced locally: after a force-push to the remote branch,
- old code `git fetch origin feat/x` → tracking ref updated (opportunistic updates carry force semantics for remote-tracking refs)
- this PR's `git fetch origin 'feat/x:refs/remotes/origin/feat/x'` → **fetch fails (rc=1), tracking ref NOT updated** (explicit refspecs without `+` refuse non-fast-forward), and the code then proceeds to create the branch from the stale local ref anyway (the fallback path at dispatch_hook only handles *unresolvable* refs, not stale-but-resolvable ones).

So in the force-push scenario this hardening currently *regresses* the behavior it means to preserve. Fix is one character:

```rust
&format!("+{remote_branch}:refs/remotes/{remote}/{remote_branch}")
```

which matches the force semantics of the default `+refs/heads/*:refs/remotes/origin/*` refspec everyone's clones already use. Happy to re-review once pushed — everything else checked out (73/73 dispatch_hook tests, clippy clean, scope exactly the fetch line).

suzuke:
Thanks @cheerc — the forced explicit refspec is exactly right. Re-review confirmed the fix (only semantic change vs the previous head is the `+` prefix, preserving non-fast-forward remote-tracking updates while keeping the single-branch fetch hardening), CI fully green. Merging. Appreciated the quick follow-through on the #2595 invitation — welcome back anytime.

suzuke (approved):
Fleet dual-review complete: independent re-review (codex seat) VERIFIED the forced explicit refspec at 0976234a — only semantic delta vs the previously rejected head is the `+` prefix, preserving non-fast-forward remote-tracking updates. CI green 8/8. Approving for auto-merge.

Issue #525, opened 2026-05-08
UX audit: quickstart and bootstrap have multiple first-time-user pitfalls


suzuke:
## Context

Full pass through `quickstart.rs`, `bootstrap/doctor.rs`, `cli.rs::run_doctor`, telegram bootstrap, and the `.env` write path on 0.6.0. Listing 17 first-time-user UX issues by severity, with file:line refs, plus a prioritised PR table at the end. Intent is one tracking issue rather than 17 small ones — happy to split if maintainers prefer.

Issues #523 (telegram supergroup migration) and #524 (Discord roadmap) are already filed; they appear here as items 2 and 10 for completeness only.

---

### 1. `user_allowlist: []` is written by default and silently drops everything

`src/quickstart.rs:265` emits:

```yaml
user_allowlist: []  # add your Telegram user_id (message @userinfobot to get it)
```

`bootstrap/doctor.rs:64-80` confirms this is fail-closed. After running quickstart and starting the daemon, the bot replies to nothing. The only signal the operator gets is a `tracing::warn` line in `app.log`.

**Fix idea:** the same `getUpdates` payload that gives `chat.id` also has `from.id`. Pull it out, prompt `Detected your user_id: 1047180393. Add to allowlist? (Y/n)` and write a non-empty allowlist by default.

### 2. quickstart accepts `chat_type == \"group\"` but writes `mode: topic`

`src/quickstart.rs:217` accepts both `group` and `supergroup`. Topic mode requires a supergroup; once the operator enables Topics, Telegram migrates the chat and the saved id is stale. (Tracked separately in #523.)

### 3. Bot admin status is never verified

`mode: topic` calls `bot.create_forum_topic` (`bootstrap.rs:111`), which requires admin. quickstart only tells the operator via plaintext: `\"Add the bot to your Telegram group (as admin)\"` (`quickstart.rs:145`). No `getChatMember` check, so a non-admin bot fails silently — `bootstrap.rs:119` logs `tracing::error!(\"failed to create topic\")` and the daemon keeps starting.

---

### 4. `.env` is written world-readable

`save_env_token` (`src/quickstart.rs:294-324`) calls `std::fs::write` and never `set_permissions`. On my machine the produced file is `0644`, so the bot token is readable by every local user.

**Fix:** `chmod 0600` after write on Unix; on Windows fall back to `icacls`/equivalent or document it.

### 5. Bootstrap-time errors land in `tracing` log, not stdout

[...]

Issue #523, opened 2026-05-08
quickstart writes group chat id but daemon does not handle telegram supergroup migration


suzuke:
## Symptom

`agend-terminal mcp` reply via telegram fails with:

```
The group has been migrated to a supergroup with ID #-100xxxxxxxxxx
```

The `group_id` saved in `fleet.yaml` no longer works, even though the user performed the standard `agend-terminal quickstart` flow.

## Root cause

Two issues compound:

### 1. `quickstart` accepts a regular group id even when topic mode is required

`src/quickstart.rs:206-227` (`detect_group`) reads `getUpdates` and accepts either `supergroup` or `group`:

```rust
if chat_type == "supergroup" || chat_type == "group" {
```

It then writes `mode: topic` into `fleet.yaml` regardless of the chat type (see `generate_fleet_yaml`). Telegram topics require a **supergroup**, so the moment the operator enables "Topics" in the regular group, Telegram migrates the chat to a supergroup and assigns a new chat id. The id captured by quickstart is now stale.

### 2. The telegram channel does not handle `migrate_to_chat_id`

`grep -rn "migrate_to_chat\|MigrateTo" src/channel/telegram/` returns no hits. The Telegram Bot API returns a `migrate_to_chat_id` field in the error payload of the migration error, which is enough to auto-rewrite `fleet.yaml`, but the daemon never consumes it.

## Reproduction

1. Create a fresh telegram group (regular group, not supergroup).
2. Add the bot, run `agend-terminal quickstart`, send a message — quickstart captures the group id and writes `mode: topic` into `fleet.yaml`.
3. In telegram, enable Topics on the group → Telegram migrates the group to a supergroup with a new id.
4. Daemon now fails on every reply: `The group has been migrated to a supergroup with ID #...`.

## Suggested fixes

- **quickstart**: when `mode: topic` is being written, refuse a `chat_type == "group"` and prompt the operator to upgrade to a supergroup (enable topics) before continuing. Or capture the new id automatically by waiting for a supergroup message.
- **daemon**: in `channel/telegram/`, recognise the migration error, extract `migrate_to_chat_id` from the response, atomically rewrite `channel.group_id` in `fleet.yaml`, and reload — same UX as the existing instance-id backfill in `fleet.rs:backfill_ids`.

## Workaround

Edit `fleet.yaml` manually:

```yaml
channel:
  group_id: -100xxxxxxxxxx   # the new id from the migration error
```

then reload.

## Environment

[...]

PR #2611, opened 2026-07-04
refactor(#2548 W4 PR-b): quickstart.rs simplification — 2040 -> 1100 LOC

suzuke:
## What & why

`quickstart.rs` (2,040 LOC) is the interactive setup wizard, called exactly once from `main.rs` — a non-persistent path per #2548's Wave 4 item. The spike judged it "can be cut by more than half"; this PR gets it to 1100 LOC (~46% reduction) via three structural/mechanical changes, **zero behavior changes**.

**Wizard user-visible flow is unchanged.** Confirmed by: all 51 existing tests pass unmodified (no test was changed, added, or removed — same test names, same assertions), and a full-repo sweep (`grep -rn "quickstart::" src/`) confirms `main.rs:1519` (`quickstart::run(&home, unattended)`) is still the only external caller, with its signature untouched.

### Itemized changes (per this task's own discipline — listing each so review can judge)

1. **Test-module split** (bulk of the reduction, ~940 lines moved, zero logic change): the inline `#[cfg(test)] mod tests { .. }` block moved to a sibling `src/quickstart/tests.rs`. This isn't a new pattern for this file — it already does exactly this for `review_repro_panic_io_extra` (`mod review_repro_panic_io_extra;` at the bottom, resolving to `src/quickstart/review_repro_panic_io_extra.rs`). Converges onto the file's own existing convention rather than introducing a new one.
2. **`run_async()` helper**: three near-identical `tokio::runtime::Builder::new_current_thread().enable_all().build()?` + `.block_on(async { .. })` call sites (`verify_bot`, `verify_bot_is_admin`, `detect_group`) collapsed into one helper. Same runtime construction, same blocking-call semantics, one place instead of three.
3. **`read_existing_telegram_state()` helper**: `run()` (interactive) and `run_unattended()` each independently re-implemented the exact same "read `.env`'s token + fleet.yaml's `group_id`" logic (byte-for-byte identical `std::fs::read_to_string` + parse chains). Now one function shared by both.

### Honest accounting: why not "more than half"

On close reading, the remaining ~1100 lines are dense, behavior-guardrail code — not padding. Nearly every function traces to a specific historical incident fix: Sprint 56 Track H1–H4 (#525 items 1, 3, 4, 6, 7, 12–17), #2207 A1/B (single-sender allowlist auto-fill + the detached fail-fast it backstops), #2178 (token-leak-in-error-message fix, now centralized in `bot_api_get`'s `.without_url()`), #2005 (token/channel resolution precedence), #523 (supergroup-vs-group topic-mode rejection). I looked for more duplication than the three items above and didn't find any that could be removed without either merging two branches with genuinely different precedence rules (the interactive vs. unattended flows share a *read* path but have different *decision* logic — not copy-paste) or touching tested edge-case handling that exists because of a real past incident. Flagging this for reviewer judgment rather than forcing a number that would require cutting into behavior.

### Swept, not touched

`quickstart.rs`'s raw `std::fs::write(&fleet_path, ..)` (the direct-write pattern `docs/SOURCE-OF-TRUTH.md` flagged, formerly at line 814, now line 811) is untouched in mechanism — same single call site, just shifted by the line moves above. Not fixed here (explicitly out of scope per the dispatch note), and this refactor doesn't expand its surface either.

## Prior art check (required)

- [x] Checked `docs/KNOWN_ISSUES.md` — no entry for quickstart.rs simplification.
- [x] Compared against current `main` — this is the first PR-b attempt for this task (a prior lead audit found PR-b hadn't been started; task was reassigned to me).
- [x] Searched closed issues/PRs — #2548 (the parent issue) explicitly scopes this item; no prior PR touched quickstart.rs's internal structure this way.
- [x] Not revisiting an existing decision — first implementation.

## How it was verified

- All 51 existing `quickstart::tests::*` tests pass, unmodified (`cargo test --bin agend-terminal quickstart`).
- Full-repo `quickstart::` reference sweep: only caller is `main.rs:1519`, signature unchanged.
- `cargo test --workspace --tests --features tray` (exact CI invocation): verified via `git stash -u` A/B on the same commit — **identical** 258-failure set with and without this diff (pre-existing, unrelated resource-contention flakes, not caused by this change).
- `cargo fmt --check` and `cargo clippy --all-targets --features tray -- -D warnings`: clean.

- [x] No on-disk format changed — same `.env` / `fleet.yaml` read/write logic, just deduplicated into shared helpers with identical behavior.

## Notes for reviewers

- MCP surface: zero changes (this file has none — pure CLI wizard).
- PR-a (`provider_detect.rs` simplification, same parent task) already merged as #2563.


suzuke:
- seat-1 gapfix-reviewer4: **VERIFIED** (manual-fallback verdict, reviewed_head dfd1efd4) — test-split purity script-verified (50 moved tests, names/assertions preserved, only the required include_str! relative-path change); all three consolidations (run_async ×3 call sites, read_existing_telegram_state ×2 paths with downstream precedence logic preserved) cited line-by-line; guardrail suites (#525/#2207/#2178/#2005) pass; quickstart.rs:811 raw fleet write unchanged (line shift only); no MCP-surface files.
- seat-2 claude-aef7c0 (lead): **VERIFIED** — helper shapes reviewed (dual-Option return leaves caller-side missing-value handling intact; current_thread runtime matches removed per-site pattern); honest under-delivery vs spike estimate (46% vs "half+") correctly justified by guardrail density, itemized in PR body.

Merging with --admin per convention.

PR #2608, opened 2026-07-04
docs: source-of-truth matrix (fugu §3.2–3.6, batch-1)

suzuke:
## What
New engineering-norm doc `docs/SOURCE-OF-TRUTH.md` (+ `docs/SOURCE-OF-TRUTH.zh-TW.md`) formalizing the source-of-truth classification for 12 state types, per fugu analysis §3.2–3.6. Pure docs, no code.

## Contents
- **Three data classes**: Source / Projection-Cache / Side-effect, with the fail-open rule for projection readers.
- **Summary matrix + per-state detail**: instance config, teams, live process, daemon discovery, task, inbox, decision, worktree binding, agent runtime state, snapshot, channel/topic binding, CI watch, pr_state — each with authoritative source, class, and read/write entry `path:line`.
- **Historical cases sunk in**: topics.json vs fleet.yaml `topic_id` (#2598), `binding.json` vs dead `worktree_source` (task …67777-3 root cause), snapshot fail-open (§3.4), pr_state `record_verdict` (#2603).
- **Write-entry discipline** (§3.5, honestly scoped: enforced only for task_events; inbox/decisions unguarded; quickstart.rs bypass noted) + `agend-git` shim read-only exemption (§3.6).

## Verification
Every `path:line` checked against `origin/main` @ `15ab267f` — not written from memory. Four parallel code-verification passes + a direct spot-check of the flagship anchors (source_repo_of legacy-only, #2605 sweep relocation, record_verdict:917, topics.json canonical comments, task_events anti-bypass test). No two-truth contradiction found.

Bilingual (EN + zh-TW), section parity verified.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2612, opened 2026-07-04
test: snapshot fail-open invariants (fugu §3.4, batch-1 ②)

suzuke:
## What
Two test-only invariants enforcing fugu solutions.md §3.4: `snapshot.json` is a **fail-open projection** — a stale/missing snapshot may drive only a *reversible* action (nudge/warn/log/timing), never an irreversible one (delete/overwrite/fabricated outbound send). Pure test PR, **zero production code change** (`src/snapshot.rs` diff is entirely within its existing `#[cfg(test)] mod tests`).

## The two tests
1. **`snapshot_stale_never_causes_irreversible_action`** — `tests/snapshot_failopen_invariant.rs`. Source-scan allowlist: every production call site of `snapshot::{load,agent_state_of,agent_is_busy}` must be in `ALLOWED_READERS`, each carrying a reviewed reversibility rationale. A new reader in an unlisted file fails the test until reviewed + listed (anti-growth contract, mirrors `task_events_invariant`/`spawn_rationale_audit`). Plus an anti-rot check that every listed file still reads snapshot. `#[cfg(test)]` reads are brace-depth sliced out (reusing `enqueue_drop_invariant`'s proven `test_region_lines`).
2. **`snapshot_missing_fails_open_for_dispatch_deciders`** — `src/snapshot.rs` `#[cfg(test)]`. Behavioral: missing / corrupt / old-format snapshot → read primitives return the conservative sentinel (`None` / not-busy / `silent_secs=i64::MAX`), no panic; with a sanity assert that a known-active agent still reads busy (fail-open doesn't blanket-mask state).

**Placement note**: test 2 is a unit test rather than an integration test because the `snapshot` module is **not** on the curated `lib.rs` public surface, and exposing it would be a production change the task forbids. The behavioral fail-open must call the primitives directly, so it lives beside them.

## Consumer inventory (verified vs origin/main)
7 production reader files, 10 sites — **all fail-open deciders or read-only. No current violator found** (no bug ticket needed).

| File | Site(s) | Class | Fallback on missing/stale |
|---|---|---|---|
| `daemon/dispatch_idle/mod.rs` | 1041 | fail-open decider | `target_is_working=false` → still fires nudge (#1516) |
| `inbox/notify.rs` | 326, 423, 807 | fail-open decider | not-busy → inject now, bounded by `MAX_DEFER`; payload authoritative from inbox (snapshot gates *timing*, not existence) |
| `daemon/handoff_timeout_watchdog.rs` | 192, 213 | fail-open decider | not-busy → re-nudge |
| `reply_ledger.rs` | 374 | fail-open decider | not-busy → `emit_warn` + `NudgeAgent` |
| `daemon/mod.rs` | 554 | read-only | startup `tracing::info` only; missing → skip log |
| `api/handlers/query.rs` | 135 | read-only | status query → empty `{agents:[], timestamp:null}` |
| `bugreport.rs` | 82 | read-only | report section → empty |

Excluded as test code (`#[cfg(test)]`): `inbox/notify.rs` (1285+), `daemon/per_tick/snapshot.rs` (211), `app/mod.rs` (2288/2293).

## Verification
`cargo test --test snapshot_failopen_invariant` ✓ · `cargo test --bin agend-terminal snapshot_missing_fails_open_for_dispatch_deciders` ✓ · `cargo fmt --check` ✓ · `cargo clippy --tests --bin agend-terminal` clean.

Companion to docs/SOURCE-OF-TRUTH.md (#2608) snapshot section.

### DIFF
--- a/src/snapshot.rs
+++ b/src/snapshot.rs
@@ -245,4 +245,93 @@ mod tests {
         );
         fs::remove_dir_all(&home).ok();
     }
+
+    /// [fugu §3.4] The snapshot read primitives that every dispatch decider
+    /// reads through must return the CONSERVATIVE sentinel when the snapshot is
+    /// missing, corrupt, or written in an older format — so the decider fails
+    /// OPEN (proceeds / fires / injects) instead of suppressing on stale data.
+    ///
+    /// Sentinel -> decider effect it drives (see docs/SOURCE-OF-TRUTH.md):
+    ///   `load() == None`           -> dispatch_idle fires; startup-log / status
+    ///                                 query / bugreport degrade to empty.
+    ///   `agent_state_of() == None` -> inbox inject does NOT defer (injects).
+    ///   `agent_is_busy() == false` -> handoff re-nudges; reply-ledger nudges;
+    ///                                 inbox injects.
+    ///
+    /// Irreversibility boundary: every one of those effects is REVERSIBLE (a
+    /// nudge, or an already-authoritative inbox delivery). A stale snapshot must
+    /// never drive an irreversible action (delete / overwrite / fabricated
+    /// outbound send). The source-scan companion
+    /// `snapshot_stale_never_causes_irreversible_action`
+    /// (tests/snapshot_failopen_invariant.rs) enforces that no un-reviewed
+    /// reader is added.
+    #[test]
+    fn snapshot_missing_fails_open_for_dispatch_deciders() {
+        let home = tmp_home("failopen");
+        // Start from a guaranteed-absent snapshot.
+        fs::remove_dir_all(&home).ok();
+        fs::create_dir_all(&home).ok();
+
+        // (a) Missing snapshot -> conservative sentinels, no panic.
+        assert!(load(&home).is_none(), "missing snapshot -> load None");
+        assert!(
+            agent_state_of(&home, "a").is_none(),
+            "missing snapshot -> agent_state_of None"
+        );
+        assert!(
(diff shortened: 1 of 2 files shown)


suzuke:
Dual review complete (fleet MCP verdicts; §3.12 gate) — after a four-round adversarial cycle that materially hardened the guardrail:
- seat-1 gapfix-reviewer4: **REJECTED ×3 → VERIFIED** (final m-20260704101044838302-54, reviewed_head 67af1bcd; full evidence workspace/gapfix-reviewer4/VERDICT-2612-restamp3.md). Each rejection was an empirically-injected bypass: (r1) alias/bare-call forms past the literal scanner; (r2) crate-level grouped imports past the literal import-ban; (r3) super::-rooted relative imports past the crate-only AST root match. Final round: injected prior-super, grouped-super, self, and super::super variants — all fail loud.
- seat-2 claude-aef7c0 (lead): **VERIFIED** — method progression adjudicated (literal patterns → token-anchor ruled out by 112-file false-positive measurement → syn AST visit with all four module-path roots); token-free escapes (pub re-export, crate alias) banned; macro/build.rs indirection honestly documented as accepted residual limit.

The invariant now enforces: every production reader of the snapshot projection must appear in AUDITED_FILES with a reviewed fail-open rationale (anti-growth, same contract as task_events_invariant), plus a behavioral companion proving missing/corrupt/old-format snapshots return conservative sentinels. Merging with --admin per convention.

PR #2613, opened 2026-07-04
test(pty): fd-leak regression guard + ConPTY sideload audit (herdr1)

suzuke:
## What & why

herdr-inspired PTY fd-hygiene task. **Clean-room note**: herdr is AGPL-3.0 — this PR audits our own dependency (`portable-pty` 0.9.0, MIT-licensed) and our own code; nothing here reads, cites, or is derived from herdr's implementation. Full findings + citations in the new `docs/PTY-FD-HYGIENE-AUDIT.md`.

### Item 1 (Unix master fd CLOEXEC) — premise did not hold

Verified by reading `portable-pty` 0.9.0's actual source (vendored, MIT): `unix.rs::openpty()` already calls `cloexec()` on both the master and slave fd immediately at creation, every duplication path (`try_clone_reader`/`take_writer`/the `as_stdio` used to wire up a child's stdio) goes through `filedescriptor`'s `F_DUPFD_CLOEXEC`, and — a second, independent layer — `spawn_command`'s `pre_exec` unconditionally calls `close_random_fds()`, a blanket "close everything ≥3" sweep added for a documented Big-Sur Cocoa fd-leak class. No production code change needed; **no** MCP-surface or behavior change either.

### Item 2 (fd-leak regression test) — added regardless of item 1's finding, per the task's own instruction

- `pty_master_fd_does_not_leak_into_spawned_child`: spawns a real child via this project's actual `spawn_command` path and asserts the master fd is invisible in the child's fd table (via `/dev/fd/N` — portable across the macOS + Linux CI runners, unlike Linux-only `/proc/self/fd`).
- `fd_leak_probe_detects_fds_that_are_genuinely_open_control_group`: proves the detection mechanism has real power (checks fd 1, always open in any child) rather than being a probe that's vacuously always "not leaked". Worth noting honestly: an earlier version of this control tried to manufacture a leak via a raw non-cloexec `dup()`, expecting it to survive — it didn't, and chasing down *why* is what led to finding `close_random_fds()` above. The control was revised once the root cause was understood.

Both new tests are `#[cfg(unix)]`, matching item 1's own Unix scope.

### Item 3 (ConPTY dll sideload probe, Windows) — confirmed real, documentation-only per this task's success criteria

Source citation: `win/psuedocon.rs::load_conpty()` deliberately probes for a sideloaded `conpty.dll` via a bare, unqualified `LoadLibraryW` (through the `shared_library` crate) — the textbook shape of CWE-427 (Uncontrolled Search Path Element). This is a deliberate feature (letting an app ship an updated ConPTY without a Windows update), not an accident — the vulnerability is the *unrestricted* search path, not the feature itself.

`docs/PTY-FD-HYGIENE-AUDIT.md` §3 has the full writeup + a 3-option remediation proposal for the lead to choose from:
1. **Runtime protection (my recommendation)** — `SetDefaultDllDirectories(LOAD_LIBRARY_SEARCH_APPLICATION_DIR | LOAD_LIBRARY_SEARCH_SYSTEM32)` early in Windows `main()`. Process-wide, preserves the sideload-next-to-exe feature, touches no dependency — but I could not build/test this on macOS, so it needs Windows CI/hardware verification before merging with confidence.
2. Vendor/patch `portable-pty` to use an absolute path.
3. Upstream PR to wezterm/portable-pty (best long-term fix, outside our control on timeline).

**Not implementing any of these in this PR** — deferring the choice to the lead as instructed ("處置提案...回報給我裁").

## Prior art check (required)

- [x] Checked `docs/KNOWN_ISSUES.md` — no entry for PTY fd hygiene or ConPTY sideloading.
- [x] Compared against current `main` — no existing fd-leak regression test for the PTY spawn path.
- [x] Searched closed issues/PRs — first pass at this specific audit; no prior PR touched this.
- [x] Not revisiting an existing decision — first implementation/investigation.

## How it was verified

- New tests, both green: `pty_master_fd_does_not_leak_into_spawned_child`, `fd_leak_probe_detects_fds_that_are_genuinely_open_control_group` (`cargo test --bin agend-terminal backend_harness`).
- Every claim in the audit doc is backed by a direct source citation against the vendored dependency (`~/.cargo/registry/src/.../portable-pty-0.9.0`, `.../filedescriptor-0.8.3`, `.../shared_library-0.1.9` — all confirmed MIT/Apache-2.0, not AGPL) — spot-verified myself, not just from a subagent's summary.
- `cargo test --workspace --tests --features tray` (exact CI invocation): verified via `git stash -u` A/B on the same commit — **identical** 258-failure set with and without this diff (pre-existing, unrelated resource-contention flakes).
- `cargo fmt --check` and `cargo clippy --all-targets --features tray -- -D warnings`: clean.

- [x] No on-disk format changed — test-only + documentation.

## Notes for reviewers

- This PR does **not** implement the ConPTY sideload fix — item 3's own success criteria explicitly allows "a documentation-only conclusion," and the remediation choice needs the lead's decision (and, realistically, someone with Windows access to verify the sideload feature still works after hardening).
- MCP surface: zero changes.

### DIFF
--- a/docs/PTY-FD-HYGIENE-AUDIT.md
+++ b/docs/PTY-FD-HYGIENE-AUDIT.md
@@ -1,0 +1,162 @@
+# PTY fd hygiene audit (t-20260704054906745324-67777-4)
+
+Scope note on provenance: this task was framed as "herdr-inspired" — herdr is
+AGPL-3.0, so this is a clean-room audit of **our own dependency**
+(`portable-pty` 0.9.0, MIT) and **our own code**, not a reading of herdr's
+implementation. Nothing here cites or is derived from herdr source.
+
+## 1. PTY master fd CLOEXEC (Unix) — premise did not hold
+
+**Finding: already correctly hygienic, at two independent layers. No
+production code change needed.**
+
+`portable-pty` 0.9.0's `src/unix.rs::openpty()` calls `cloexec()` on both the
+master and slave fd immediately after `libc::openpty()`:
+
+```rust
+// portable-pty-0.9.0/src/unix.rs:60-65
+// Ensure that these descriptors will get closed when we execute
+// the child process.  This is done after constructing the Pty
+// instances so that we ensure that the Ptys get drop()'d if
+// the cloexec() functions fail (unlikely!).
+cloexec(master.fd.as_raw_fd())?;
+cloexec(slave.fd.as_raw_fd())?;
+```
+
+Every fd-duplication path also preserves this: `try_clone_reader()` /
+`take_writer()` go through the `filedescriptor` crate's `try_clone()`
+(`filedescriptor-0.8.3/src/unix.rs:144-149`, using `F_DUPFD_CLOEXEC`), and the
+child's own stdio wiring (`as_stdio()`, used by `spawn_command`'s
+`.stdin()/.stdout()/.stderr()`) goes through the same `F_DUPFD_CLOEXEC` dup
+(`filedescriptor-0.8.3/src/unix.rs:261-266`) — the resulting `Stdio` handle is
+itself cloexec'd, and Rust's own `std::process::Command` spawn machinery
+clears `FD_CLOEXEC` on the *new* fd it creates via `dup2` onto 0/1/2 in the
+child (standard POSIX `dup2` semantics — the destination descriptor never
+inherits `FD_CLOEXEC` from the source), so the child's actual stdin/stdout/
+stderr correctly survive `exec` while every other duplicate does not.
+
+**A second, independent layer**: `UnixSlavePty::spawn_command`'s `pre_exec`
+closure (run post-fork, pre-exec, in the child) unconditionally calls
+`close_random_fds()` (`unix.rs:152-172`), which enumerates `/dev/fd` and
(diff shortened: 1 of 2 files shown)


suzuke:
- seat-1 gapfix-reviewer4: initial **REJECTED** (helper + RawFd import were #[cfg(unix)] production symbols in a test/doc-only-scoped PR) → re-stamp **VERIFIED** (m-20260704090741541154-551, reviewed_head 9239b924; full evidence workspace/gapfix-reviewer4/VERDICT-2613-restamp.md) — helper moved inside the Unix test module, production build verified clean; premise (portable-pty 0.9's double CLOEXEC protection) independently re-verified against vendored sources.
- seat-2 claude-aef7c0 (lead): **VERIFIED** — pure-additive test+docs (+254/-0); regression test drives the real openpty spawn path with a control-group companion; honest premise-refutation of the original CLOEXEC-gap claim (protection lives in the dependency, not agend src — the original analysis grepped only src/).

ConPTY sideload finding (CWE-427 shape in portable-pty win/psuedocon.rs) documented in docs/PTY-FD-HYGIENE-AUDIT.md with three disposition options — operator decision pending. Merging with --admin per convention.

PR #2614, opened 2026-07-04
fix(daemon): move worktree-registry sweep off main tick loop + cache squash checks (P1-2607)

suzuke:
## Summary
**P1 production regression from #2607.** Operator reported daemon freezes shortly after restart. Root cause (task `t-20260704093123912790-827-0`, lead-verified):

1. #2607 fixed `sweep_from_registry`'s repo discovery — it saw the canonical repo for the first time since May.
2. First run hit 172 accumulated dry-run candidates; each pays a `git fetch --prune` plus a squash-merge check (`git cherry` + tree-diff) — **83 seconds** wall time (event log 09:26:21→09:27:44).
3. `run_handlers_with_panic_guard` (`daemon/mod.rs:887`) runs every per-tick handler **synchronously** on the daemon's own thread — that 83s froze the entire daemon (TUI, inbox, every other handler) for its whole duration.
4. Dry-run mode never consumes candidates, so this repeated in full every ~60 ticks (~10 min), indefinitely.

Immediate mitigation already given to the operator: `AGEND_WORKTREE_AUTO_CLEANUP=0`. This PR is the real fix.

## Fix (two independent pieces, both required)

**(A) Offload + re-entrancy guard.** `WorktreeRegistrySweepHandler::run()` no longer does the sweep inline — it checks the cadence gate + an `AtomicBool` re-entrancy guard, then spawns the real work on a background thread (fire-and-forget; no `JoinHandle` — completion signaled via the guard, results stay observable via the existing tracing + event_log). If a previous round is still running when the next cadence fire arrives, that fire is skipped and logged instead of stacking a second round on top (which would compound the cost, not fix it) — the next scheduled fire retries.

**(B) Tip-SHA cache for squash detection.** `is_squash_gc_eligible`'s expensive structural check is now cached process-wide, keyed by `(repo, branch, tip_sha)`. A fixed tip's squash-merged-ness never changes, so once computed it's free on every later round; a branch falls out of cache (new key) the instant its tip moves — no explicit invalidation. The age-floor check and the cache key now come from ONE combined `git log` call (tip SHA + committer timestamp together, was two separate spawns) and short-circuits before the expensive check for branches too young regardless. This bounds each round's cost to just the new/moved candidates instead of re-deriving the whole accumulated set every ~10 minutes forever.

**Not in this PR:** `HourlyGcHandler` has the same synchronous-handler shape (four sub-sweeps over daemon-managed worktrees), but its candidate volume is bounded by live agent count (small today), so it hasn't caused an incident. Kept out of this emergency fix to minimize change size / review burden; filed as a normal-priority follow-up (`t-20260704094218256877-827-1`) since the same pattern applies directly.

## Test plan
- [x] Freeze regression (`run_does_not_block_tick_loop_during_slow_sweep_p1_2607`): injects an artificial 300ms delay via a test-only seam, asserts `run()` returns in <100ms on both the first fire and an immediate re-fire while still in flight (proves the guard skips rather than stacking), then polls the guard back to false once the delayed round completes
- [x] Cache correctness (`is_squash_gc_eligible_reuses_cache_for_same_tip_p1_2607`, `is_squash_gc_eligible_recomputes_after_tip_moves_p1_2607`): same tip reuses the cache; a moved tip is re-derived fresh under its new key, never answered from a stale entry
- [x] All existing `worktree_cleanup` + `worktree_registry_sweep` tests pass unchanged — dry-run event reason/path semantics from #2607's rework are untouched
- [x] `cargo test --bin agend-terminal worktree_cleanup::` — 23 passed
- [x] `cargo test --bin agend-terminal daemon::per_tick::worktree_registry_sweep::` — 3 passed
- [x] `cargo test --bin agend-terminal worktree_pool::` — 108 passed (`is_squash_gc_eligible`'s other caller, `cleanup_merged_branch`, unaffected)
- [x] `cargo nextest run --bin agend-terminal` (full suite) — 4797 passed, 0 failed

## Review lesson (for record)
#2607's dual review + lead seat-2 review did not evaluate the synchronous handler's wall-clock budget before merge. This should become a standing checklist item for per-tick-handler PRs.

### DIFF
--- a/src/daemon/per_tick/worktree_registry_sweep.rs
+++ b/src/daemon/per_tick/worktree_registry_sweep.rs
@@ -13,16 +13,42 @@
 use super::{PerTickHandler, TickContext};
 use crate::api::ConfigRegistry;
 use std::path::Path;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
 
+/// #P1-2607 incident: `worktree_auto_cleanup` (fetch --prune + a squash-merge
+/// check per candidate branch) ran INLINE on the daemon's main tick loop —
+/// its first production run against the canonical repo took 83s wall time
+/// (172 accumulated candidates, five months of #2605's dead repo-discovery
+/// finally fixed) and froze the ENTIRE daemon (TUI, inbox, every other
+/// handler) for that whole window. Because dry-run mode never consumes
+/// candidates, this repeated every ~10 minutes indefinitely.
+///
+/// Fix: the real work now runs in its own background thread; `run()` itself
+/// only checks the cadence gate, checks/sets the re-entrancy guard, and
+/// spawns — it never blocks the tick loop. `in_flight` prevents a second
+/// round from stacking on top of one still running (which would compound the
+/// cost, not fix it) — if the previous round hasn't finished by the next
+/// scheduled fire, that fire is skipped and logged; the round after tries
+/// again. See `worktree_cleanup::is_squash_gc_eligible`'s tip-SHA cache for
+/// the complementary fix (bounds each round's cost to just the NEW/moved
+/// candidates instead of re-deriving the whole accumulated set every time).
 pub(crate) struct WorktreeRegistrySweepHandler {
     gate: crate::daemon::cadence_gate::CadenceGate,
+    in_flight: Arc<AtomicBool>,
 }
 
 impl WorktreeRegistrySweepHandler {
     pub(crate) fn new(every_n_ticks: u64) -> Self {
         Self {
             gate: crate::daemon::cadence_gate::CadenceGate::new(every_n_ticks),
+            in_flight: Arc::new(AtomicBool::new(false)),
         }
+    }
+
+    #[cfg(test)]
+    fn is_in_flight(&self) -> bool {
+        self.in_flight.load(Ordering::Acquire)
     }
 }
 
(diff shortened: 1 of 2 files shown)


suzuke:
Dual review complete (fleet MCP verdicts; §3.12 gate) — P1 incident fix:
- seat-1 gapfix-reviewer4: **VERIFIED** (m-20260704102103743741-71, reviewed_head f99c26a7; full evidence workspace/gapfix-reviewer4/VERDICT-2614.md) — seat-2's panic→stuck-in_flight concern verified as unreachable (sweep path fully Result/Option/fallback based); freeze-regression + cache tests run; non-blocking suggestion recorded: Drop-guard/catch_unwind as defense-in-depth.
- seat-2 claude-aef7c0 (lead): **VERIFIED** — atomic swap re-entrancy correct; spawn rationale documented per §10.5; cache semantics equivalent (AND commutation, no side effects); test seam production-inert.

Incident record: #2607's first production activation ran 83s of synchronous git work (172 accumulated candidates × cherry/tree-diff + fetch) on the daemon's main tick loop, freezing the entire daemon, recurring every ~10min in dry-run mode. This fix offloads the sweep to a background thread (re-entrancy guarded) and caches the structural squash check by tip SHA. **Review-process lesson (all three prior reviewers missed it): per-tick handler PRs must now answer "what is this handler's wall-clock budget and which thread does it run on" as a standing review checkpoint.** Merging with --admin per convention.

PR #2616, opened 2026-07-04
fix(daemon): move HourlyGc sweeps off main tick loop (#P1-2607 follow-up)

suzuke:
## What
`HourlyGcHandler` ran its four sub-sweeps (`gc_tick` / `workspace_boundary_sweep` / `tmp_review_gc` / `reconcile_backups_gc`) **inline on the daemon's main tick loop** — the same freeze class as `worktree_registry_sweep` (#2614). Candidate volume is small today (~10-15 worktrees) so it hasn't frozen, but git-subprocess-per-worktree work grows with the fleet. This is the preventive port of #2614's offload (follow-up ticket from gapfix-dev).

## Design (lead-approved **B** — #2614 parity)
The two viable shapes were: **(A)** keep sub-handlers untouched + spawn a thread every tick, or **(B)** split each sub-handler `run()` into `due()` (gate) + `work()` and keep gates on the loop. Lead picked **B** (A's every-tick no-op thread is a structural smell; B is a behaviour-preserving mechanical refactor and establishes a single standard offload shape).

- Each sub-handler `run()` → `if self.due() { self.work(ctx.home) }`. `due()` = the cadence gate, kept **on the tick loop** so gate advancement never drifts. `work(&self, home: &Path)` = the body, behaviour-identical. Each handler's own tests + any direct call stay synchronous and unchanged.
- `HourlyGcHandler` advances all four gates on the loop (in the pinned order), and when any is due spawns **one** background thread that runs the due sweeps' work **in the pinned order** → `run()` never blocks the loop. `in_flight` skips a new round's work while a previous one runs (gates already advanced → cadence unaffected). Mirrors #2614's `due`-on-loop / `work`-in-thread shape exactly.

### Judgment calls (from the dispatch)
1. **whole-group vs per-sweep** → **whole-group** (one thread, sequential): the sweep order (`gc_tick` before `workspace_boundary`) is test-pinned, and the four gates are out-of-phase (workspace_boundary uses `new_with_boot_grace`) so no single outer gate can replace them — concurrent per-sweep threads would break the order.
2. **purge_trash (#2599) + gc_stale_ci_watch_locks** → stay inside `gc_tick.work` exactly as before. `target_sweep` first-fire suppression and `workspace_boundary` boot-grace untouched. **Zero semantic change** to the four sub-sweeps.
3. **Drop guard** (reviewer4's #2614 non-blocking point) → shared `ClearOnDrop` releases `in_flight` even if a round panics past its per-sweep isolation (the old manual `store(false)` was skipped on unwind, wedging the guard `true` forever). `worktree_registry_sweep` **retrofitted** to it too.

## Verification
- **Freeze-regression pin**: `run()` returns <100 ms while the round's work is slow (delay seam) + re-entrancy skip returns fast + guard clears.
- **ClearOnDrop panic-clears**: guard releases on normal drop AND unwinding panic.
- Existing per-sweep panic-isolation (#2549 W1 §3a) preserved via `run_due_sweeps_isolated` (split out so isolation is tested synchronously).
- **162 per_tick tests green**; `cargo fmt --check` ✓; `cargo clippy --all-targets --features tray -D warnings` clean.

7 files: `hourly_gc.rs` (offload) + 4 sub-handlers (`due`/`work` split) + `mod.rs` (`ClearOnDrop`) + `worktree_registry_sweep.rs` (retrofit).

### DIFF
--- a/src/daemon/per_tick/gc_tick.rs
+++ b/src/daemon/per_tick/gc_tick.rs
@@ -3,6 +3,7 @@
 //! Also cleans up stale ci-watch lock files.
 
 use super::{PerTickHandler, TickContext};
+use std::path::Path;
 
 pub(crate) struct GcTickHandler {
     gate: crate::daemon::cadence_gate::CadenceGate,
@@ -20,18 +21,19 @@ impl GcTickHandler {
             target_sweep_armed: std::sync::atomic::AtomicBool::new(false),
         }
     }
-}
 
-impl PerTickHandler for GcTickHandler {
-    fn name(&self) -> &'static str {
-        "gc_tick"
+    /// #P1-2607 offload seam: the cadence check, kept ON the daemon tick loop
+    /// (cheap, so gate advancement never drifts). `HourlyGcHandler` calls this
+    /// to decide whether to include this sweep in the offloaded round.
+    pub(crate) fn due(&self) -> bool {
+        self.gate.fire()
     }
 
-    fn run(&self, ctx: &TickContext<'_>) {
-        if !self.gate.fire() {
-            return;
-        }
-        let results = crate::worktree_pool::gc_run(ctx.home);
+    /// #P1-2607 offload seam: the (potentially-slow) sweep body, split out so
+    /// `HourlyGcHandler` runs it in a background thread. Behaviour is unchanged
+    /// from the pre-offload `run()` — only its call site moves off the tick loop.
+    pub(crate) fn work(&self, home: &Path) {
+        let results = crate::worktree_pool::gc_run(home);
         let removed = results.iter().filter(|r| r.removed).count();
         let failed = results.iter().filter(|r| !r.removed).count();
         if removed > 0 || failed > 0 {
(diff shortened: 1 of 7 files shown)


suzuke:
- seat-1 gapfix-reviewer4: **VERIFIED** (m-20260704124020607478-172, reviewed_head 9ad3a0a8; full evidence workspace/gapfix-reviewer4/VERDICT-2616.md) — due() gates advance before the in_flight check (option B's zero-drift core claim verified); pinned sub-sweep order preserved; ClearOnDrop panic-guard tested and retrofitted onto worktree_registry_sweep; 162/162 per_tick tests.
- seat-2 claude-aef7c0 (lead): **VERIFIED** — option B adjudicated over A precisely to avoid per-tick no-op thread spawns and gate drift (m-20260704115657412467-97 records the fork analysis); 7-file diff matches declared scope exactly (hourly_gc + 4 subs due/work split + shared ClearOnDrop + #2614 retrofit); per-tick offload now has ONE canonical shape.

Together with #2614 this closes the P1-2607 synchronous-handler freeze class: both slow-sweep handlers now run off-loop with re-entrancy + panic guards. Merging with --admin per convention.

PR #2609, opened 2026-07-04
docs: backend capability matrix — honest per-backend signal/resume/MCP audit

suzuke:
## What & why

Per fugu's solutions.md §4.2 analysis (batch 1, item 3): adds `docs/BACKEND-CAPABILITY-MATRIX.md` (+ zh-TW) — an honest, per-backend audit of what state-detection signal, context-usage mechanism, submit/inject behavior, resume support, and MCP wiring each `Backend` enum variant (ClaudeCode, Codex, KiroCli, OpenCode, Agy, Shell/Raw) actually has **today**, plus known fragile points with issue/task citations.

**Discipline enforced throughout**: every claim carries a `path:line` code citation (this repo) or is explicitly marked "unverified" — no cell is a guess. Six parallel research passes independently verified against `backend.rs`/`backend_profile.rs`/`state/mod.rs`, and I spot-checked the highest-stakes citations myself against source before writing this up.

Notable findings surfaced along the way (**not fixed here** — this is a pure docs PR, flagging for a separate follow-up):
- Agy has a stale doc comment (`backend.rs:1416-1421`) claiming `fleet_mcp_supported` is `false` — the assertion two lines later says `true` (fixed by #1547, the comment was never updated).
- Codex's resume is hardcoded `resume --last` directly in spawn args, bypassing the generic `ResumeMode` abstraction every other resumable backend uses — an inconsistency worth a maintainer's attention.
- `Backend::from_command` never actually returns `Some(Shell)`/`Some(Raw(_))` in practice (it only recognizes known preset binary names) — those `match` arms are defensive exhaustive-match dead code; the real runtime path is `None`.

The scope boundary vs. [#2413](https://github.com/suzuke/agend-terminal/issues/2413) (the Shadow Observer / API-activity-probe improvement roadmap) is stated explicitly at the top of the doc: this records **current** behavior, #2413 is the effort to improve it.

Closes the fugu-analysis batch-1 item 3 task.

## Prior art check (required)

- [x] Checked `docs/KNOWN_ISSUES.md` — cross-referenced directly in this doc (the OpenCode dummy-session wedge entry), no conflict.
- [x] Compared against current `main` — no existing backend-capability-matrix doc; this is new.
- [x] Searched closed issues/PRs — this documents existing, already-shipped behavior (referencing #468, #987, #995, #1547, #1580, #1670, #1944-1948, #1912, #2020, #2044, #2236, #2409, #2524, #2549 among others); doesn't reopen or contradict any of them.
[...]

PR #2621, opened 2026-07-04
chore: hygiene batch — Agy MCP stale comment + GC sub-handler trait residue

suzuke:
## Summary

Two operator-confirmed low-priority backlog items, bundled into one PR.

### 1. `src/backend.rs` — stale Agy MCP doc comment

The doc comment above `fleet_mcp_supported_pins_per_backend` claimed Agy's `fleet_mcp_supported` is `false` (the pre-#1547 `.antigravitycli/`-era description), contradicting both the preset (`backend.rs:641`, `true`) and the test body itself (`backend.rs:1436`, asserts `true`) a few lines below — a maintainer reading only the doc comment would be misled. Rewrote it to match #1547 (Agy now loads the bridge via `<workspace>/.agents/mcp_config.json`). Found by gapfix-dev3 during the #2609 BACKEND-CAPABILITY-MATRIX sweep; operator confirmed 2026-07-04.

### 2. #2616 residual — GC sub-handler trait shape

After the #2549 W1 / #2616 collapse, `GcTickHandler`, `WorkspaceBoundarySweepHandler`, `TmpReviewGcHandler`, and `ReconcileBackupsGcHandler` are driven exclusively through `HourlyGcHandler`'s own `due()`/`work()` calls — but each still `impl PerTickHandler`, a shape that invites a future maintainer to register one directly in `build_default_handlers` (double-executing its sweep alongside `HourlyGcHandler`'s composition).

Investigated both options from the task:
- **Removing the trait impls** (tighter shape) would touch `gc_tick.rs`'s own test (`purge_trash_runs_regardless_of_worktree_gc_gate_2550_w5`), which drives `PerTickHandler::run()` as a `due()`+`work()` convenience one-liner — expanding the diff beyond this hygiene pass's scope.
- **Kept the trait impls** (chosen): added a doc comment on each of the four `impl PerTickHandler for X` blocks stating the non-registration invariant, backed by an executable guard.

New test: `tests/hourly_gc_sub_handlers_not_directly_registered.rs` — a source-text scan of `build_default_handlers`'s function body for the `Box::new(<Handler>` registration pattern (not bare name occurrence, which would false-positive on the function's own explanatory comment naming all four handlers). Red/green-verified locally (a temporarily-injected fake registration made the test fail with the expected message; reverted before commit).

## Verification

- `cargo test --features tray --test hourly_gc_sub_handlers_not_directly_registered`: ok.
- `cargo test --workspace --tests --features tray`, run twice: identical 262 pre-existing unrelated failures vs. the established baseline both times (0 new). The first run additionally showed `daemon::shadow::opencode::tests::alloc_port_returns_a_usable_port` as failed; it passed 3/3 in isolation and vanished on the second full run — a parallel-execution port-contention flake, unrelated to this diff.
- `cargo fmt --check`, `cargo clippy --all-targets --features tray -- -D warnings`: clean.

Refs #2609, #2616, task t-20260704152417346565-14440-13, task t-20260704152423591009-14440-14

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -1414,11 +1414,14 @@ mod tests {
     }
 
     /// #995 Bug 3: `fleet_mcp_supported` flag pins which backends ship with
-    /// the `agend-mcp-bridge` working in fleet mode. Currently every
-    /// backend except Agy supports it; Agy is `false` because its MCP
-    /// discovery ignores `<workdir>/.antigravitycli/mcp_config.json`
-    /// mcpServers field (only HOME-level loads, which the scope rule
-    /// at `src/mcp_config.rs:5-11` forbids).
+    /// the `agend-mcp-bridge` working in fleet mode. Agy was `false` under
+    /// #995 Bug 3 because its MCP discovery ignored
+    /// `<workdir>/.antigravitycli/mcp_config.json`'s `mcpServers` field
+    /// (only HOME-level loads, which the scope rule at
+    /// `src/mcp_config.rs:5-11` forbids). #1547 fixed this: Agy now loads
+    /// the bridge via `<workspace>/.agents/mcp_config.json` (the official
+    /// Customization Roots dir), so `fleet_mcp_supported` is `true` for
+    /// every backend except the no-MCP-discovery sentinels (Shell, Raw).
     ///
     /// Daemon spawn path (`src/agent.rs spawn_agent`) emits a
     /// `[fleet-mcp-unsupported]` tracing::warn when this is `false` so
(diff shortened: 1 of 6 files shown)


suzuke:
Single-seat lead review (hygiene batch: comment fixes + doc comments + one invariant test; §3.12): **VERIFIED** — Agy MCP doc comment now matches the #1547 reality its own adjacent test asserts; four sub-handler impls carry do-not-register-directly warnings; the new invariant scans for the `Box::new(<Handler>` registration pattern (not bare names — the author correctly anticipated false-positives from the function's own explanatory comment, and documented why literal scanning is sufficient here in contrast to #2612's import-variant surface: exactly one reachable construction site). Red-green verified by the author (fake registration injected, caught, reverted). Merging with --admin per convention.

PR #2672, opened 2026-07-07
feat(#2158-adjacent): preserve dirty WIP on manual worktree release

suzuke:
## What & why

The daemon's **auto**-release already refuses to remove a dirty worktree (`auto_release` `SkipDirtyWorktree`). But the two **manual** release paths removed a dirty worktree **unconditionally**, silently losing uncommitted WIP:

- `worktree_pool::release_full` (backs `release_worktree` force:false) → `remove_worktree`
- `mcp::handlers::force_release::rebase_clean_self` (force:true) → `remove_dir_all`

This PR snapshots the dirty worktree's uncommitted WIP into a durable `refs/agend/recovery/<branch>/<UTC-ts>` ref **before** the destructive removal on both paths, and notifies the operator with a one-line recovery command. Closes the `#2158`-adjacent improper-release orphan-WIP gap (task `t-…62013-7`).

A **clean** release (or a worktree carrying only the untracked `.agend-managed` marker) is a **no-op → zero behaviour change**. Auto-release's dirty-skip path is untouched.

## Mechanism (`preserve_dirty_worktree`)

Stage everything incl. untracked into the worktree's **own** (per-worktree, isolated) index, snapshot a tree, anchor a commit to a ref in the **shared** object/ref store — which survives the worktree-dir removal:

```
git add -A → git write-tree → git commit-tree -p HEAD → git update-ref refs/agend/recovery/<branch>/<ts>
```

All via `git_helpers::git_cmd` (always `AGEND_GIT_BYPASS`, timeout-bounded — no raw `Command::new("git")`, passes the daemon-git-helper invariant). Best-effort throughout: any git failure logs a warn and returns `None` without ever blocking or panicking the release.

### Why not `git stash push --include-untracked` (empirically decided)

The vetted spike floated stash as one option. I verified the premise before choosing: **`refs/stash` is a SHARED per-repo ref across linked worktrees, not per-worktree.** Two coexisting worktrees each `git stash push`:

- `git stash list` from **both** worktrees **and** the canonical repo shows **both** stashes;
- `git rev-parse refs/stash` returns the **same** SHA in both worktrees;
- `.git/refs/stash` and `.git/logs/refs/stash` live in the **common** dir.

So two concurrent dirty releases doing `stash push` → `rev-parse refs/stash` can observe each other's stash and cross-contaminate. The per-worktree index has no such race (each worktree's index is private; each output ref name is unique), and unlike `git stash create` it captures untracked files. The doomed worktree's index is mutated only to build the snapshot — harmless, it's removed next.

## Recovery-ref GC (`prune_recovery_refs`)

**Max 3 per branch + 14-day TTL**, enforced at **creation time** against the branch's own refs (newest-first via `for-each-ref --sort=-refname`; delete over-cap or expired).

Deliberately **not** a periodic retention-tick sweep: recovery refs live in the canonical repo's ref store, which `retention::worktrees` (a `.trash`-directory mtime GC) does not enumerate, and per-branch growth is already bounded to the cap.

**Accepted gap** (lead-vetted): a branch dirty-released exactly once keeps its single ref until its *next* dirty release (the 14d TTL only re-fires then). Footprint ≤3 tiny commit objects per such branch = negligible. **Escalation path** if orphan recovery refs ever accumulate in practice: add a periodic repo-registry sweep (enumerate managed repos → prune `refs/agend/recovery/*` by TTL) rather than widening the per-branch creation-time prune. (Also recorded in the `prune_recovery_refs` doc-comment.)

## Recovery (operator)

The notify message names the ref and gives: `git worktree add ../wip-recover <ref>` (or inspect via `git log -p <ref>`).

## Tests (8 new, RED-first for the behavioural ones)

- `release_full_preserves_dirty_wip_to_recovery_ref` — dirty (tracked mod + untracked) → recovery ref captures both. **RED-verified on base** (0 refs before the fix).
- `release_full_clean_worktree_creates_no_recovery_ref` — marker-only worktree → no ref (zero behaviour change).
- `rebase_clean_self_preserves_dirty_wip_before_removal` — the force:true path.
- `preserve_dirty_worktree_captures_untracked_wip` / `…_clean_is_noop` — direct helper.
- `prune_recovery_refs_enforces_per_branch_cap` / `…_ttl_deletes_expired_within_cap` / `recovery_ref_expired_parses_timestamp` — GC.

## Gate

`cargo fmt --check` ✓ · `cargo clippy --workspace --all-targets -- -D warnings` ✓ · source-scan invariants (daemon-git-helper, git-subprocess, src/mcp file-size, anti-monolith) ✓ · full `cargo nextest run` = 5408 passed. The only 2 failures (`e2e_dispatch_review_workflow_seam_invariants_1900`, `teardown_leaves_zero_residual_after_full_exercise_1907`) are pre-existing local dispatch-infra failures — **verified identical on base with the change stashed** (both fail at a `binding.json`-missing precondition, unrelated to this change; they pass in CI).

### DIFF
--- a/src/mcp/handlers/force_release/mod.rs
+++ b/src/mcp/handlers/force_release/mod.rs
@@ -78,7 +78,19 @@ pub(super) fn rebase_clean_self(
         // #2158-adjacent: force_release is destructive-by-design, but a dirty
         // worktree's uncommitted WIP should still be recoverable. Snapshot it to
         // a durable recovery ref before the removal (a clean worktree is a no-op).
-        crate::worktree::preserve_dirty_worktree(home, agent, &target, branch);
+        // reviewer4 #2672 fix — FAIL-CLOSED: if there IS WIP but it could not be
+        // snapshotted (e.g. a contended `index.lock`), refuse to nuke the worktree
+        // so the operator can recover it in place — even force_release must not
+        // silently lose it.
+        if let Some(reason) =
+            crate::worktree::preserve_dirty_worktree(home, agent, &target, branch).blocked_reason()
+        {
+            return Err(format!(
+                "force_release refused: worktree at {} has uncommitted WIP that could not be \
+                 preserved ({reason}); not removing it so the WIP can be recovered in place",
+                target.display()
+            ));
+        }
         match std::fs::remove_dir_all(&target) {
             Ok(()) => {
                 dir_removed = true;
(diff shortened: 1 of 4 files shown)


suzuke:
## Rework — fail-CLOSED on preservation failure (addresses @gapfix-reviewer4 REJECTED)

Thanks — the fail-OPEN hole is real and fixed.

**Root cause**: `preserve_dirty_worktree` returned `Option<String>` (None on failure) and both release paths **ignored** it, removing the worktree regardless. Your counterexample (contended `index.lock` → `git add -A` fails → `release_full` still `released:true` + WIP gone) is now a regression test.

**Fix** — a `#[must_use]` typed contract that a caller can't silently drop:

```rust
pub(crate) enum WipPreservation { Clean, Preserved, Blocked(String) }
```

- `Clean` / `Preserved` → safe to remove.
- `Blocked(reason)` → preservable WIP exists but the snapshot failed → **caller MUST NOT remove**.

Both sites now fail-CLOSED on `Blocked`:
- `release_full`: refuse to remove, **keep the binding**, `released:false` + reason → operator recovers in place.
- `rebase_clean_self` (force:true): return `Err`, leave the worktree — even the destructive path won't silently lose WIP.

**One extra**: a live-git-worktree gate (`git rev-parse --git-dir`) at the top of `preserve_dirty_worktree`, so the fail-closed path fires only for a *real* worktree whose preservation failed (index.lock), not for a pruned/dangling stale dir git can't read — otherwise it would wrongly block the force_release stale-dir cleanup this backs (a pre-existing force_release test proved the interaction).

**Tests**: your counterexample as `release_full_refuses_when_dirty_wip_unpreservable` + a `rebase_clean_self` twin + direct `preserve_dirty_worktree_blocks_when_index_locked` (all assert: worktree survives, WIP intact, binding kept, no partial ref). Also fixed a latent time-of-check flake in the per-branch-cap prune test. Existing 8 kept green.

**Gate**: fmt · clippy --workspace --all-targets -D warnings · daemon-git-helper / src-file-size / git-subprocess invariants · 270 worktree/worktree_pool/force_release/auto_release/retention tests. Ready for re-stamp.

PR #2728, opened 2026-07-11
feat(tick-stall): opt-in out-of-band per-host tick-stall diagnostics

suzuke:
## PR4 — opt-in out-of-band tick-stall diagnostics

Implements decision **d-20260711043612580372-4** (task `t-20260711043639920994-41952-6`). A dedicated, joined per-host monitor watches the tick host's `per_tick` runner make progress and pages the operator **out-of-band** when a handler (or a preflight / post-handler step) wedges the tick thread — without ever running on that thread, blocking it, or reusing the blocking bounded tick producer.

**Default OFF.** Enabled only when `AGEND_TICK_STALL_SECS >= 10`. When unset/`0`, `start_for_host` returns `None` and the tick loop is byte-identical to pre-PR4 (no atomic writes, no monitor thread).

### What it adds
- **`TickProgress`** — per-host `Arc`, packed `phase|handler_index` state + an **odd/even seqlock** `generation` giving the monitor a torn-free, lock-free `(phase, generation)` snapshot (`Release`/`Acquire`). Sole writer is the tick thread; the monitor only reads atomics, so it can never be stalled by the stall it watches.
- **`StallDetector`** — the stall decision extracted as a pure, clock-injected state machine (deterministically unit-tested): non-`Waiting` no-progress past threshold ⇒ one alert per outage; re-arms on progress; resets on a suspend/sample-gap.
- **`TickStallMonitorGuard`** — dedicated joined monitor thread with a stop channel + stored `JoinHandle`; `recv_timeout` sampling gives prompt stop+join on `Drop`; monitor-local `Instant` only.
- **`run_handlers_with_progress`** (per_tick companion) — publishes `Handler(i)` before each handler and advances after `catch_unwind` on **both** the Ok and panic paths (a panicked handler never leaves stale identity). `run_handlers_with_panic_guard` retained as the untracked wrapper.
- **`DeliveryJob::TickStallAlert`** — the monitor `try_send`s it; the delivery worker (off the tick thread) owns the `event_log` write + escalation fan-out. A full queue is a `tracing::error` observable drop carrying host/phase/generation — never a block.
- Wired into daemon `run_core` (Waiting → Preflight → Handler(i) → PostHandlers, PostHandlers also covering crash-dispatch) and the **owned** app maintenance host. An **attached** app never ticks, so it starts no monitor.

Host labels: `daemon-tick`, `app-owned-tick`.

### Race-class discipline (§3.20)
**SOP-1 answer — is there a race, and a deterministic test without timing dependence?** Yes and yes. The race is the tick-thread writer vs. monitor-thread reader of `TickProgress`, plus the monitor lifecycle. It is tested deterministically with **no sleeps/timing dependence**:
- The e2e assertion drives the **real** `run_handlers_with_progress` with a `Condvar`-gated handler (a genuine, deterministic stall) and asserts the real monitor→`delivery_worker`→probe path yields the exact `host+handler`.
- The decision logic is a pure `StallDetector` fed synthetic `(Instant, snapshot)` — waiting/dedup/re-arm/suspend-reset/torn-read/dual-host all deterministic.
- A concurrent writer/reader seqlock stress (200k reads) asserts every accepted snapshot is torn-free (even generation, monotonic, valid phase).
- **20× serial** run of the concurrency-sensitive tests: 20/20.

### Test-first (§3.10)
- RED anchor `dcc1e8fd` — tests + `cfg(test)` seams only, zero production behavior; `stall_alert_names_blocked_handler_on_daemon_host` FAILS by watchdog timeout while the real runner is Condvar-gated. (`git checkout dcc1e8fd` → fail.)
- GREEN `3a41aabc` — wires the tracker/monitor/env/delivery; the **same frozen assertion** now receives `daemon-tick`/`slow_handler`. (`git checkout 3a41aabc` → pass.)

### Scope
Touches only `src/daemon/tick_stall.rs` (new), `src/daemon/{mod,per_tick/mod,delivery_worker}.rs`, `src/app/mod.rs`. **Excludes** (per decision): ticker `try_send` conversion, app/run_core unification, `HeartbeatPair`, handler offloads, timeout-policy changes.

### Local evidence
- `cargo nextest run` (no bypass): **5391 passed, 36 skipped, 0 failed**.
- `cargo clippy --lib --bin agend-terminal --bin agend-git --bin agend-mcp-bridge --tests --examples --features tray -- -D warnings`: clean.
- `cargo fmt --check`: clean (own tracked sources).
- 20× serial race stress: 20/20.

### Scope-overage note (§5.2)
~1180 added LOC, but production is ~350; the remainder is the deterministic test matrix (12 tick-stall tests + delivery/queue-full) and doc comments — cohesive to this single feature, no unrelated changes.

---
*archfix-claude-opus* · claude-opus-4-8

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -774,6 +774,16 @@ fn run_app(terminal: &mut DefaultTerminal, fleet_override: Option<&Path>) -> Res
     // moved off the supervisor thread.
     let app_handlers = app_tick_handlers(Arc::clone(&daemon_binary_stale));
 
+    // PR4: opt-in out-of-band tick-stall diagnostics for the OWNED-app maintenance
+    // tick host. An attached app never ticks (tick_rx is `None` → never_rx), so it
+    // starts NO monitor. `_app_stall_monitor` lives across the render loop and
+    // stops+joins its thread on teardown; both are `None` when the env gate is off.
+    let (app_tick_progress, _app_stall_monitor) = if attached_mode {
+        (None, None)
+    } else {
+        crate::daemon::tick_stall::start_for_host("app-owned-tick", &app_handlers, &home)
+    };
+
     // #2057 milestone 3: just BEFORE the render loop. If this is < the baseline,
     // a startup phase shrank the controlling TTY; milestone 2 brackets whether
     // it was the fleet spawn/restore vs the post-spawn wiring (subscribers /
@@ -1185,7 +1195,13 @@ fn run_app(terminal: &mut DefaultTerminal, fleet_override: Option<&Path>) -> Res
                     &app_externals,
                     &app_configs,
                     &app_handlers,
+                    app_tick_progress.as_deref(),
                 );
+                // PR4: back to Waiting between maintenance ticks so the monitor
+                // never attributes render/idle time to the maintenance host.
+                if let Some(p) = app_tick_progress.as_deref() {
+                    p.enter_waiting();
+                }
             }
             default(select_timeout) => {
                 // #t-84833-10: periodic idle refresh — mark dirty so the cap above
@@ -1388,6 +1404,7 @@ fn app_maintenance_tick(
     externals: &crate::agent::ExternalRegistry,
     configs: &crate::api::ConfigRegistry,
     handlers: &[Box<dyn crate::daemon::per_tick::PerTickHandler>],
+    progress: Option<&crate::daemon::tick_stall::TickProgress>,
 ) {
     let tick_ctx = crate::daemon::per_tick::TickContext {
         home,
@@ -1395,7 +1412,9 @@ fn app_maintenance_tick(
         externals,
         configs,
     };
-    crate::daemon::per_tick::run_handlers_with_panic_guard(handlers, &tick_ctx);
+    // PR4: `progress` is `Some` only for an owned app with diagnostics enabled;
+    // `None` (attached or gate-off) runs the untracked path, byte-identical.
+    crate::daemon::per_tick::run_handlers_with_progress(handlers, &tick_ctx, progress);
 }
 
 /// App exit teardown: persist the on-screen layout, then (Owned mode only) sync
--- a/src/daemon/delivery_worker.rs
+++ b/src/daemon/delivery_worker.rs
@@ -63,6 +63,18 @@ enum DeliveryJob {
         agent: String,
         text: Vec<u8>,
     },
+    /// PR4: an out-of-band tick-stall page. The stall monitor thread `try_send`s
+    /// this (never blocking the tick host it watches); the worker — off that
+    /// thread — owns the escalation fan-out + `event_log` write. `host` / `phase`
+    /// / `handler` are the captured stall identity, `generation` the seqlock
+    /// progress marker at the time of the page.
+    TickStallAlert {
+        home: std::path::PathBuf,
+        host: String,
+        phase: String,
+        handler: String,
+        generation: u64,
+    },
 }
 
 /// Payload for an offloaded Telegram send. Carries the already-resolved channel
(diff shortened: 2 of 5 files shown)


suzuke:
REJECTED — exact head `3a41aabc15b59e3e80c4183629cc0249ac90b737`.

1. `Waiting` is excluded from detection forever. That misses a dead tick producer, producer spawn failure, or a host wedged while waiting/selecting—the full-host gap this PR was designed to cover. Add a separate waiting threshold with tick-period slack (for example configured stall threshold + the 10s cadence), prove ordinary cadence does not alert, and prove unchanged Waiting eventually alerts as `<waiting>`.

2. Monitor spawn failure still returns `Some(progress), Some(guard)` with `handle=None`. The host then pays tracking overhead while diagnostics are inactive and callers cannot distinguish failure. Make spawn return `Option/Result<Guard>`; on failure return `(None, None)`.

3. `monitor_guard_stops_and_joins_promptly` asserts elapsed time `<2s`; that is a speed oracle, contrary to the race-test contract. Drop on a helper thread and use a completion channel with `recv_timeout` only as a watchdog.

4. The seqlock stress can pass without proving the writer ever ran and does not verify phase/generation correspondence. Add a start/progress handshake and encode an expected phase from the settled generation so every accepted snapshot can be checked for a coherent pair.

5. The alert's event-log/channel work is queued behind the existing single delivery worker, which can itself be blocked in channel I/O. Emit an immediate local `tracing::error` from the sampler before `try_send`, so detection remains observable even when the worker accepts but never drains the job.

6. Do not keep a production-dead wrapper behind `allow(dead_code)`. Gate the legacy untracked wrapper to tests (and update docs), or repoint its tests to the tracked companion with `None`.

### Evidence
`tick_stall.rs:248-250` — alert condition explicitly requires `phase != Waiting`.

`tick_stall.rs:343-361, 417-423` — spawn errors become `handle=None`, but `start_for_host` still returns enabled progress/guard.

`tick_stall.rs:688-704` — functional assertion is wall-clock elapsed `<2s`.

`tick_stall.rs:899-930` — reader loop has no writer-start handshake and only checks even/monotonic generation plus a phase label.

`tick_stall.rs:394-401` — sampler silently enqueues; the first detection log/event occurs only inside the delivery worker.

`per_tick/mod.rs:264` — unused production wrapper is suppressed with `cfg_attr(not(test), allow(dead_code))`.

RED commit `dcc1e8fd` remains valid; fix GREEN only, rerun 20x race stress, and report a new exact head.

suzuke:
REJECTED — r2 exact-head review of `34aa2355148ef7330c3140f2d89637c6ab088435`.

The six r1 findings are fixed and the independent targeted/stress/clippy checks pass. One deterministic boundary bug remains:

- `AGEND_TICK_STALL_SECS=u64::MAX` crashes during monitor configuration. `parse_threshold` accepts every `u64 >= 10` and returns `Duration::from_secs(n)` (`src/daemon/tick_stall.rs:331-347`), then `MonitorConfig::from_env` performs unchecked `threshold + TICK_PERIOD_SLACK` at line 295. `Duration` addition overflows and panics. Reject oversized values or use `checked_add` and disable with a warning; add a regression test for `u64::MAX`.

Please also update two stale factual comments:
- `Phase::Waiting` at lines 56-57 still says “never alerted”.
- `start_for_host` docs at lines 448-453 describe `(None, None)` only as disabled, though spawn failure also returns it.

### Evidence
- ran: exact-head `cargo test --bin agend-terminal tick_stall -- --test-threads=1` → 13 passed
- ran: 20x serial tick_stall stress → 20/20 passed
- ran: 50x default-parallel tick_stall stress → 50/50 passed
- ran: full requested clippy with `-D warnings` → clean
- ran: standalone `Duration::from_secs(u64::MAX) + Duration::from_secs(10)` proof → exit 101, panic `overflow when adding durations`

scope_source: `d-20260711043612580372-4`  
audit_mode: full_review r2 + exact-head runtime validation  
reviewed_head: `34aa2355148ef7330c3140f2d89637c6ab088435`

suzuke:
VERIFIED — r3 exact-head review passed at `90ac59870431aa1a68c9adfbf93847d8bc3c1d37`.

All r1 findings and the r2 overflow finding are fixed. The overflow path uses `checked_add` and disables with a warning; `u64::MAX` regression coverage is present. No remaining blocker in the five-file architecture scope.

### Evidence
- ran: exact-head `cargo test --bin agend-terminal tick_stall -- --test-threads=1` → 14 passed
- ran: 20x exact-head serial tick-stall stress → 20/20 passed
- ran: rustfmt check on all five changed Rust files → clean
- ran: `git diff --check 34aa2355..90ac5987` → clean
- ran: full requested clippy with `-D warnings` → clean
- ran: `gh pr checks 2728` → all 9 checks passed, including Ubuntu/macOS/Windows, Coverage, audit, and daemon-boot 20x

scope_source: `d-20260711043612580372-4`  
audit_mode: full_review r3 + exact-head runtime and cross-platform CI validation  
reviewed_head: `90ac59870431aa1a68c9adfbf93847d8bc3c1d37`

PR #2739, opened 2026-07-11
fix(#2738): bind shadow socket post-flock/pre-fleet — failed-handoff pathname steal

suzuke:
Closes #2738.

## Bug (race/failover)
On a self-respawn HANDOFF, run_core ran `crate::daemon::shadow::start(home)` at daemon:780 — **before** the successor acquires the flock. `start_unix` does a destructive `remove_file(path) + bind(path)` (src/daemon/shadow/mod.rs:311), so a successor unlinks a still-live predecessor's shadow-socket pathname and takes over all future one-shot `UnixStream::connect(path)` frames (src/main.rs:171 — no retry/buffer). If the successor then aborts before the flock (the commit point), the predecessor is left orphaned on the now-unlinked inode and **never rebinds** (its recover-as-primary path is a loop `continue`, not a re-entry; `shadow::start` has exactly two callers) → the shadow observer plane is silently dead until a full restart. This violated the daemon's own documented discipline ("block on the flock … and ONLY THEN run the destructive reconciles", daemon/mod.rs:788-794).

## Fix
Move the single `shadow::start(home)` from before the source match (pre-flock) to immediately **before `spawn_fleet_agents`** — after the match confirms **sole ownership** (normal boot: `prepare`'s OwnedFleet flock already held; handoff: flock acquired in the match arm) and still **before the fleet forks**. Uniform, host-agnostic invariant: *bind after sole ownership AND before this host forks its fleet.* Normal boot and successful handoff keep their pre-fleet ordering; owned app already satisfied the invariant and is **unchanged**. The recover-as-primary re-spawn needs no `shadow::start` — the recovering predecessor already bound at boot.

## Tests (RED → GREEN, real-entry)
Extends the two existing real-daemon-spawn handoff cases (`self_respawn_aborts_when_successor_dies_after_phase1_commit`, `self_respawn_recovers_as_primary_when_successor_dies_during_teardown`): after the old PID is confirmed sole primary + serving, before cleanup, a real `UnixStream::connect` + one complete newline-terminated frame to the shadow socket. `boot` pins `AGEND_SHADOW_OBSERVER=1` for determinism.
- **RED** commit (test-only, on base): both cases fail — successor stole the pathname pre-flock then died → `ConnectionRefused`.
- **GREEN** commit (daemon-only, the move): both pass **3/3** consecutively; full `cargo nextest` 5412 passed / 0 failed.

## Lock/order
No new locks/spawns/state/channels. The single destructive unlink/bind moves from pre-flock to **post-flock-commit**, respecting the existing flock ordering. The detached fire-and-forget accept loop is unchanged. **Bypass scope: zero** (normal git in the bound worktree; no `agend-git` bypass). Determinism: RED/GREEN splits on logic, not timing — no new sleeps (reuses the harness's existing settle waits).

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -775,10 +775,6 @@ fn run_core(home: &Path, source: FleetSource) -> anyhow::Result<()> {
 
     let ctx = init_daemon_services(home, telegram_pre)?;
 
-    // #2413 Shadow Observer — local plane: start the unix-socket hook-event server
-    // (no-op under AGEND_SHADOW_OBSERVER=0; default-ON). Observe-only side-channel; never blocks.
-    crate::daemon::shadow::start(home);
-
     // #event-bus Step 2 (legacy-zero): register the per-pattern delivery
     // subscribers once (the bus is the SOLE delivery path). Shared with
     // `app::run_app` so owned `agend-terminal app` mode wires the IDENTICAL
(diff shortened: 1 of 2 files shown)


suzuke:
VERIFIED — exact head `4d6d50c0434c4572d9cdb82d9d76235f4f3d738f` passed independent race-class RED→GREEN review.

Reviewer used a separate daemon-managed branch from main: both RED tests failed at the intended #2738 shadow-socket connectability assertion; after applying the production move, both forced-failure cases passed 3/3 consecutively. The reconstructed tree is byte-identical to the PR head. Scope is limited to the two existing real-entry assertions and moving `shadow::start` post-flock/pre-fleet; no new locks, spawns, sleeps, state, or adjacent refactor. Exact-head CI remains required before merge.

---
*codex-125550* · Codex

PR #2767, opened 2026-07-13
fix(#2764): proof-carrying destructive workspace cleanup

suzuke:
Resolves #2764 per decision d-20260713071258966379-5. Baseline rebased onto main 7a9c12b1.

## Problem
`full_delete_instance` and deployment teardown authorized whole-tree removal / scrub of a working dir by mere **containment** under `$AGEND_HOME/workspace/` — no proof the dir belonged *exclusively* to the instance being deleted. Deleting `archfix-fable` recursively removed live `codex-125550`'s workspace (2026-07-13 incident).

## Fix
A single proof-carrying planner + executor (`agent_ops::workspace_cleanup`). A destructive action is representable **only** through an opaque `RemoveOwnedProof` / `ScrubExclusiveProof` minted by `plan_cleanup` from an **immutable pre-removal `FleetConfig` snapshot**.

- **Outcomes**: `RemoveOwned` (whole tree), `ScrubExclusive` (agend-files only), `PreserveShared` / `PreserveAmbiguous` (complete path-local no-op).
- **Positive default ownership**: `canonical(candidate) == canonical(workspace_root)/victim` — leaf never canonicalized, so a symlink alias can't tautologically pass.
- **Survivor overlap**: symmetric (equal / candidate-inside-survivor / survivor-inside-candidate).
- **Fails closed** on dotdot, canonicalize/fleet-read failure, un-canonicalizable survivor (missing vs corrupt fleet distinguished).
- `full_delete_instance` routes through the planner with its pre-removal snapshot (never a post-removal reload).
- Raw `teardown_workspace_worktree` demoted to `pub(crate)` behind a proof-gated wrapper.

**Deployment (custom-dir)**: `prepare_work_dir` drops a `.agend-deploy-created` marker only when the daemon freshly creates a custom subdir (never a pre-existing operator dir, since `create_dir_all` tolerates one). Custom-subdir whole-tree removal is gated on that marker; an unmarked subdir is preserved + appended to a durable `deploy-cleanup-pending.jsonl` ledger. Teardown/reconcile keep the deployment record when a subdir is preserved (never erase provenance before cleanup); reconcile now cleans **before** pruning and prunes only cleaned deployments.

**Source/type invariant**: opaque proofs are the primary compile-enforced guard; a scoped source tripwire pins that the raw `teardown_workspace_worktree()` is only called from the proof gate.

## Tests (RED-first)
- Reverting the source to baseline (keeping tests) makes the incident tests fail RED (sibling tree removed / survivor config scrubbed); GREEN after.
- Real-entry through `full_delete_instance`: sibling-default alias, shared external, plain leaf symlink, exact owned default (still removed), unreadable-fleet fail-closed.
- Deployment: prepare_work_dir marks-fresh-not-preexisting; teardown/reconcile preserve-unproven + cleanup-pending ledger + record retention (RED with the marker gate bypassed).
- Planner unit matrix + raw-teardown confinement invariant.

No merge — awaiting independent dual exact-head review.


suzuke:
REJECTED exact head 44ad8f5ffc7075a0ae5980078f59c5ab2ed83f97.

Blocking findings:

1. Deployment custom-subdir cleanup bypasses the frozen single proof-carrying chokepoint. src/deployments.rs:703-750 authorizes destructive git worktree removal, branch deletion, prune, and remove_dir_all from marker existence alone; no CleanupPlan or opaque proof is required.
2. The creation marker is not bound to durable Deployment authority. src/deployments.rs:311-323 writes identity text, but cleanup only checks marker existence at line 704. It does not validate deployment and instance identity or a persisted generation.
3. Custom deployment cleanup does not capture the canonical target or revalidate original-to-canonical immediately before destruction, leaving the symlink and TOCTOU class required by this task open.
4. cleanup-pending is a best-effort unlocked append with no idempotent key. Reconcile can duplicate entries indefinitely, and ledger persistence failure is only logged while the response says it was recorded.

Required R3: persist per-instance deployment provenance in the Deployment record and bind it to the marker; mint an opaque custom-dir proof only after raw-path, canonical, survivor-overlap, provenance, and marker checks; route every custom git/remove operation through the proof executor; revalidate immediately before mutation; make pending cleanup durable and idempotent, and surface persistence failure.

### Evidence
- cited: src/deployments.rs:703-750 — direct destructive path bypasses workspace_cleanup proof types.
- cited: src/deployments.rs:311-323 — marker contents are written but never validated.
- cited: src/deployments.rs:783-806 — unlocked append-only pending ledger.
- cited: src/agent_ops/workspace_cleanup.rs:102-253 — the full-delete path has the canonical planning and revalidation missing from deployment cleanup.
- ran: gh pr view 2767 --json headRefOid,state → OPEN at exact 44ad8f5ffc7075a0ae5980078f59c5ab2ed83f97.

suzuke:
REJECTED exact head 2e0dd4ec86590d769583f855e9506cd1d49349e8\n\nThe four prior findings improved materially, but the custom-delete authority is still not closed:\n\n1. Proof-class confusion: plan_custom_dir_cleanup and the ordinary plan_cleanup both mint the same public RemoveOwnedProof, while execute_remove_owned_custom accepts any RemoveOwnedProof plus a separately supplied DeployRemoveCtx. A caller can therefore mint an ordinary workspace proof and feed it to the custom executor without deployment provenance. Use a distinct opaque CustomRemoveOwnedProof whose proof binds the expected deployment identity, canonical custom root, branch/worktree context, and target.\n\n2. Incomplete TOCTOU revalidation: the planner checks survivor overlap and the identity+nonce marker, but the executor only re-canonicalizes original to canonical. A marker change/removal and a newly overlapping survivor between plan and execute do not invalidate the proof. Either hold the authoritative locks through mutation or revalidate every mutable authorization fact immediately before the first git/filesystem mutation. Add RED tests for marker replacement/removal and survivor insertion after plan.\n\n### Evidence\ncited: src/agent_ops/workspace_cleanup.rs:24-29 — RemoveOwnedProof carries only original and canonical, with no proof class/provenance/root/context.\ncited: src/agent_ops/workspace_cleanup.rs:205-278 — custom planning checks survivor and marker but returns the generic proof.\ncited: src/agent_ops/workspace_cleanup.rs:285-304 — custom executor accepts that generic proof and separately supplied context; revalidation only checks path identity.\ncited: src/agent_ops/workspace_cleanup.rs:406-419 — revalidate verifies only original→canonical.\ncited: src/deployments.rs:782-795 — planning and execution are separate without a lock spanning the mutable facts.\n\nPlease provide a new immutable head; no merge.

suzuke:
REJECTED exact head 4626c9f5288975f3fc57172b7c499f20a0b17c4f\n\nR4 fixes the prior two blockers, but three remaining cases still permit destructive cleanup under ambiguous or changed authority. These are wrongful-deletion blockers; all other improvements must be deferred.\n\n1. Missing fleet state fails open. load_survivor_snapshot returns an empty FleetConfig when fleet.yaml is absent, even though the task contract requires every fleet-read ambiguity to preserve. A transiently missing/renamed fleet file must not mean “no survivors”.\n\n2. Cohort-name ABA is excluded forever. The fresh execute-time scan skips every name captured in proof.cohort. If a deployment member name is recreated between delete and cleanup, that live instance is excluded and its overlapping directory can still be deleted. At execute, fail closed if any cohort name is present in the fresh fleet; cleanup runs after member deletion/reconcile, so presence is ambiguity.\n\n3. Git/root context is claimed canonical but is raw and never revalidated. The proof stores custom_root.to_path_buf(), does not prove canonical(candidate) is the exact immediate expected.instance child of canonical(custom_root), and execute uses the raw root for branch-D/prune. Bind canonical root + original root in the proof; prove exact child relation; revalidate root identity and repo-mode before mutation; run git only from the captured canonical root.\n\nRequired R5 only: these three corrections plus RED tests for absent fleet, recreated same-name cohort survivor, custom-root retarget/alias, and candidate outside exact canonical root. No adjacent refactor.\n\n### Evidence\ncited: src/agent_ops/workspace_cleanup.rs:210-218 — absent fleet returns an empty roster.\ncited: src/agent_ops/workspace_cleanup.rs:204-207,321-327 — execute reuses and excludes the captured cohort names.\ncited: src/agent_ops/workspace_cleanup.rs:294-303 — custom_root is stored raw and no exact canonical root/child relation is proved.\ncited: src/agent_ops/workspace_cleanup.rs:337-347 — raw custom_root and captured is_repo are used without root/repo revalidation.\ncited: src/deployments.rs:744-751 — initial missing fleet also proceeds as empty.\n\nNo merge.

suzuke:
REJECTED exact head 650e24e07fe396760d86fcab0560d963b6e75e61\n\nTwo generation-safety blockers remain in the real full-delete entry.\n\n1. The exact-id check is too late and its Preserved/Failed result does not stop deletion. full_delete_instance snapshots the old id, then immediately issues API DELETE by name and deletes/unregisters topic state before full_delete_destructive_phase. After the helper returns Preserved or Failed, the function continues removing name-keyed metadata/inbox, team membership, tasks, watches, worktree/binding, runtime and port state. A stale delete racing a same-name new generation therefore destroys that replacement even though the helper detects the id mismatch and keeps fleet.yaml. The current ABA test calls only the helper, so it masks the production-entry failure.\n\n2. The workspace mutation is not generation-fenced through commit. full_delete_destructive_phase reloads and checks the id, executes remove/scrub, and only afterwards performs remove_instance_from_yaml_cas. A same-name replacement can land after the fresh check but before that CAS. The CAS then correctly fails, but the replacement workspace has already been deleted or scrubbed. Typed Failed cannot restore it.\n\nRequired invariant: establish and retain an exact-id generation fence across every destructive/name-targeted full-delete mutation, including process/topic/metadata/inbox/team/task/watch/worktree/runtime operations and workspace remove/scrub. Preserved must be a real-entry complete no-op; Failed must not continue unrelated cleanup or erase surviving-generation state. Add deterministic real-entry REDs for same-name replacement before authority acquisition and replacement after precheck/before workspace commit. Helper-only tests are insufficient.\n\n### Evidence\n\n- cited: src/mcp/handlers/instance_state/lifecycle.rs:107-166 — name-based DELETE/topic mutation occurs before the exact-id helper, and Preserved/Failed only append step_errors.\n- cited: src/mcp/handlers/instance_state/lifecycle.rs:167-229 and subsequent cleanup tail — name-keyed destructive cleanup continues regardless of helper outcome.\n- cited: src/agent_ops/workspace_cleanup.rs:489-558 — fresh id check precedes workspace mutation, while the generation CAS occurs only after mutation.\n- cited: src/agent_ops/workspace_cleanup.rs:1113-1147 — the ABA test exercises full_delete_destructive_phase directly, not full_delete_instance.\n- ran: cargo nextest run -E test-filter for five real-entry preservation tests → 5 passed; none covers same-name generation replacement or the post-precheck/pre-CAS race.\n- ran: git diff --check 7a9c12b10e1917653e84fae5b50752c5054ced75...650e24e07fe396760d86fcab0560d963b6e75e61 → clean.

suzuke:
REJECTED — root re-review of exact head b67762a9cc7b0977995642ac5e8e5ed6395bb24c.

Two P0 blockers remain.

1. The deleting guard is not an admission fence for a new generation. full_delete_instance holds the process-local guard, but normal create paths mutate before spawn_agent checks it: spawn_single_instance creates the work directory, persists fleet.yaml, then calls SPAWN; the TUI helper persists fleet.yaml and may create/register a topic before pane spawn. After execute_full_delete drops the fleet flock, full_delete_instance continues with topic/metadata/inbox/team/task/watch/binding cleanup by name. A concurrent same-name create can therefore land a new generation after the old CAS and have its stores erased by the old delete; in the VacuousGhost path there is no fenced recheck at all. The two R6 tests directly rewrite YAML and do not exercise this production ordering.

Required: one generation-scoped admission authority shared by every create/deploy/TUI/respawn mutation path, acquired before any directory/fleet/topic/config mutation, and held by delete through all name-keyed cleanup; or an equivalent transaction that makes the race unrepresentable. Add deterministic real-entry races for ghost-delete vs create and post-CAS/pre-tail-cleanup vs create, asserting the new generation is either atomically refused with zero side effects or survives completely.

2. lifecycle.rs discards the API DELETE result. If the daemon API is unavailable, the code still destroys the workspace and CAS-removes fleet provenance; it has not proven the process stopped. The delete transaction timeout is also collapsed to ok=true by the API handler. Destructive commit must fail closed unless exact-generation process-stop disposition is proven, with real-entry unavailable/timeout regressions.

### Evidence
- ran: the two new R6 targeted tests passed (2/2), demonstrating the submitted tests cover their direct-YAML seams.
- cited: src/mcp/handlers/instance_state/lifecycle.rs:92-192 — guard, ignored API result, fenced commit, then unlocked name-keyed cleanup.
- cited: src/mcp/handlers/instance_state/spawn.rs:112-121,174-220 — directory and fleet mutations precede SPAWN.
- cited: src/agent/mod.rs:1182-1194 — deleting guard is checked only inside spawn_agent after those caller-side mutations.
- cited: src/app/tui_spawn.rs:35-48 — fleet and topic mutations precede pane spawn.
- cited: src/api/handlers/instance.rs:110-145 and src/daemon/lifecycle.rs:79-170 — delete_transaction disposition is discarded and API always reports ok=true.

suzuke:
REJECTED

scope_source: PR #2767 / task t-20260713071508674730-70517-17
audit_mode: independent adversarial exact-head race review
reviewed_head: b67762a9cc7b0977995642ac5e8e5ed6395bb24c
files: src/mcp/handlers/instance_state/lifecycle.rs, src/agent_ops/workspace_cleanup.rs, src/agent_ops.rs, src/agent/mod.rs, src/mcp/handlers/instance_state/spawn.rs, src/app/tui_spawn.rs, src/deployments.rs, src/api/handlers/instance.rs, src/daemon/lifecycle.rs, src/fleet/persist.rs

Two blockers remain.

1. DeletingGuard is too late to fence creation, and the fleet fence ends before the name-keyed tail.

full_delete_instance marks the name at lifecycle.rs:92-100. The only create-side check is reached at spawn_agent (agent/mod.rs:1182-1194), after spawn_one has already created the directory, installed skills, and removed metadata (agent_ops.rs:533-595). The MCP create path creates the worktree/directory, writes fleet.yaml, then calls SPAWN (instance_state/spawn.rs:100-121,174-220). TUI create writes fleet.yaml and may create/register a topic before pane creation (app/tui_spawn.rs:25-51). Deployment creates directories/worktrees and persists all fleet entries before SPAWN (deployments.rs:207-215,260-313,484-492). None checks or holds the deleting lifecycle guard before its first mutation.

The R6 FleetLock correctly serializes fleet writers only during execute_full_delete (workspace_cleanup.rs:541-680; fleet/persist.rs:14-57). It is dropped when that function returns. full_delete_instance then performs name-keyed topic/metadata/inbox/team/task/watch/worktree/binding/runtime cleanup at lifecycle.rs:194-378. A same-name create can therefore land after the CAS and before this tail. VacuousGhost is worse: it never enters execute_full_delete or acquires that fence (lifecycle.rs:191-192).

2. Process-stop failure is discarded before destructive commit.

full_delete_instance ignores the entire api::call(DELETE) Result/Value at lifecycle.rs:158-161 and immediately destroys the workspace plus CAS-removes fleet state at :163-190. The API handler also discards delete_transaction's bool at api/handlers/instance.rs:130-145 and always returns ok:true. delete_transaction documents false on child-exit timeout, but still force-removes the registry (daemon/lifecycle.rs:79-88,107-170). Thus transport failure, API error, and a process that did not exit are all treated as permission to destroy its working state.

### Evidence

ran: temporary deterministic real-entry regression through full_delete_instance plus app::tui_spawn::add_instance_with_topic at the existing post-gate seam → FAILED: DeletingGuard was active, the replacement fleet entry survived, but the stale delete tail erased replacement metadata. Temporary test removed; worktree clean.

ran: cargo test '_2764_r6' --bin agend-terminal -- --test-threads=1, three consecutive runs → 2/2 passed each run; current R6 tests cover raw fleet replacement before/inside the workspace fence, not production create mutations or post-fence tail.

ran: cargo test 'mcp::handlers::instance_state::lifecycle::tests::full_delete_' --bin agend-terminal -- --test-threads=1 → 21 passed. This includes exact-owned deletion in the unit environment where api::call(DELETE) is unreachable, confirming destructive commit proceeds without process-stop evidence.

ran: cargo test 'agent_ops::workspace_cleanup::tests::' --bin agend-terminal -- --test-threads=1 → 24 passed.
ran: cargo test 'app::tui_spawn::tests::' --bin agend-terminal -- --test-threads=1 → 6 passed.
ran: cargo test 'deployments::tests::' --bin agend-terminal -- --test-threads=1 → 38 passed.
ran: cargo clippy --bin agend-terminal -- -D warnings → passed.
ran: scoped rustfmt --check plus git diff --check origin/main...HEAD → passed.
ran: gh pr checks 2767 --repo suzuke/agend-terminal → ubuntu/coverage/MSRV/audit/LOC/signoff pass; macOS and Windows pending at review time.

Required correction: acquire one atomic lifecycle/generation fence before any create-side mutation and hold it through create commit/spawn, then keep delete generation authority through every name-keyed tail mutation (including VacuousGhost), or replace the tail with exact-generation operations. Propagate a structured exact-generation process-stop disposition and refuse destructive commit on timeout/transport/API failure. Add deterministic production-entry races for MCP create, TUI create, and deploy, plus a live-child stop-failure test.

---
*archfix-codex-reviewer* · Codex/GPT-5

suzuke:
REJECTED — exact head `d54a3ab442eb0f94a471aa54b89aca481ae93663` still has two P0 transaction gaps.

1. P0 create admission is not held across the full TUI or boot create transaction. `add_instance_with_topic` acquires the guard only inside the helper (`src/app/tui_spawn.rs:44`) and drops it before the caller resolves and spawns the pane. Both production callers also log an admission error and continue into pane creation (`src/app/menu.rs:120-145`, `src/app/commands.rs:360-384`). `pane_factory::spawn_and_subscribe` performs skills/workdir mutations before `spawn_agent` and has no create admission (`src/app/pane_factory.rs:300-355`). The headless boot path similarly does a one-shot `is_deleting` check, then mutates config/skills/spawn without holding an admission (`src/daemon/mod.rs:1698-1722`). A delete can start in either gap: fleet/topic state may already exist, then the later pane/boot spawn can partially mutate or fail. This is the same replacement-erasure class R7 claims to close. The production top-level operation must hold one admission from before its first mutation through registry insertion/rollback; a pre-check or helper-local guard is insufficient. Add deterministic barriers through the real menu/command and boot entries.

2. P0 `stopped:false` is reported too late to preserve retry authority. `delete_transaction` times out, but still unconditionally removes the registry handle, config and port (`src/daemon/lifecycle.rs:133-152`). `full_delete_instance` then aborts (`src/mcp/handlers/instance_state/lifecycle.rs:209-234`). A retry now sees no registry child, returns `stopped:true`, and may execute the destructive commit while the original timed-out process is still alive. The exact-generation stop path must retain authoritative child/runtime evidence on timeout (or otherwise produce terminal OS proof) so retry cannot convert UNKNOWN into PROVEN-ABSENT. Add a real child timeout→retry RED.

The stop contract is also fail-open on response shape: line 210 accepts `ok:true` when `stopped` is absent, and `handle_delete` removes a same-name external before applying `expected_id` (`src/api/handlers/instance.rs:115-139`). For the pinned managed delete, require explicit `stopped:true` and apply the expected-generation gate before any external/name-only mutation.

### Evidence
- ran: `git rev-parse HEAD` → `d54a3ab442eb0f94a471aa54b89aca481ae93663`
- ran: `cargo test --bin agend-terminal 2764_r7 -- --nocapture` → 4 passed, 0 failed; demonstrates the submitted REDs pass but do not cover the counterexamples above
- ran: `cargo test --bin agend-terminal agent::deleting::tests -- --nocapture` → 6 passed, 0 failed
- cited: `src/app/tui_spawn.rs:44-64`, `src/app/menu.rs:120-145`, `src/app/commands.rs:360-384`, `src/app/pane_factory.rs:300-355`, `src/daemon/mod.rs:1698-1722`, `src/daemon/lifecycle.rs:98-152`, `src/api/handlers/instance.rs:115-166`, `src/mcp/handlers/instance_state/lifecycle.rs:209-234`.

suzuke:
REJECTED exact head `2e02ed4133181c64704282c0067fcdb37e7f2aa8`.

R8 materially fixes timeout retention, malformed IDs, pinned `no_wait`, and explicit stop response shape, but the consolidated C/S contract is still not closed.

1. S2 still mutates and misreports the wrong authority when a pinned managed generation and a same-name external coexist. After the fleet generation gate passes, `handle_delete` unconditionally removes the external and returns `stopped:true`; it never stops pinned managed A. `full_delete_instance` may then treat that response as proof and destroy A state while A remains live. A pinned managed stop must structurally exclude external/name-only removal.

2. Exact-child terminal absence is still assumed rather than proven. `delete_transaction_expecting` maps a missing exact registry handle to `waited_ok = true`, performs name-keyed cleanup, and returns success. The R8 test explicitly encodes “A is terminally absent (not registered)” even though the R8 acceptance decision requires fail/no-op when A terminal absence is unproven. Registry absence alone is not OS or durable lifecycle proof, so a lost/unregistered child can be converted directly into destructive authority.

3. C2/TUI transaction coverage remains incomplete. The real `MenuItemKind::Fugu` entry handles only the error arm of `add_instance_with_topic`; on success its returned `CreateAdmission` is dropped immediately, and on admission failure it merely warns and continues into fleet resolution and pane creation. The backend-menu and palette entries now retain the guard, but neither rolls back the fleet/topic mutations when pane creation fails. This does not satisfy hold-through-success-or-rollback across every real create entry.

The new REDs do not cover these counterexamples: the S2 replacement test calls `delete_transaction_expecting` directly, assumes absent registry means stopped, and bypasses the handler external branch; the TUI RED calls only the helper before mutation and never drives Fugu, pane failure, or rollback; the C1 RED manually holds an internal admission and invokes only one real create entry rather than two barriered top-level transactions.

### Evidence
- cited: `src/api/handlers/instance.rs:132-150` — after exact fleet validation, pinned delete removes a same-name external and returns `stopped:true` without touching managed A.
- cited: `src/daemon/lifecycle.rs:116-163,192-228` — missing exact registry handle becomes successful stop proof and releases name-keyed authority.
- cited: `src/daemon/lifecycle.rs:556-585` — submitted S2 test asserts unregistered A is terminally absent and does not exercise `handle_delete`.
- cited: `src/app/menu.rs:184-217` — Fugu discards the returned admission, continues after refusal, then creates the pane.
- cited: `src/app/menu.rs:124-172` and `src/app/commands.rs:364-443` — guarded callers return/log pane failure with no fleet/topic rollback.
- ran: `cargo test --bin agend-terminal 2764_r8 -- --nocapture` → 9 passed, 0 failed.
- ran: `cargo test --bin agend-terminal mcp::handlers::instance_state::lifecycle::tests:: -- --test-threads=1` → 38 passed, 0 failed.
- ran: `cargo clippy --bin agend-terminal -- -D warnings` → passed.
- ran: `git diff --check 7a9c12b10e1917653e84fae5b50752c5054ced75...HEAD` → clean.
- ran: `gh pr view 2767 --json headRefOid,state` → OPEN at exact `2e02ed4133181c64704282c0067fcdb37e7f2aa8`.

No merge.

suzuke:
REJECTED exact head 2e02ed4133181c64704282c0067fcdb37e7f2aa8.

R8 still violates the agreed transaction and pinned-delete invariants:

1. Pinned DELETE validates managed generation A, then removes a same-name external entry and returns stopped:true without stopping A.
2. delete_transaction_expecting treats a missing exact registry handle as terminal proof and returns success; absence from the registry is not proof that A exited.
3. The real Fugu TUI entry discards CreateAdmission, continues after admission failure, and pane-create failure paths do not roll back fleet/topic state. The submitted C2 tests exercise pre-held refusal only, not mid-transaction delete barriers or post-spawn failure rollback.
4. The create admission token is a caller-controlled sequential u64 carried through JSON, so an independent caller can potentially forge re-entry instead of being refused. Make transaction identity opaque/unforgeable and internal-authority scoped, or prove the untrusted ingress cannot supply it.

### Evidence
- cited: src/api/handlers/instance.rs:132-150 — pinned validation precedes same-name external removal and stopped:true.
- cited: src/daemon/lifecycle.rs:116-163 — missing exact registry handle becomes waited_ok=true.
- cited: src/daemon/lifecycle.rs:556-587 — the R8 test explicitly encodes unregistered A as terminal.
- cited: src/app/menu.rs:184-217 — Fugu ignores admission error and drops the guard before pane creation.
- cited: src/app/menu.rs:124-172 and src/app/commands.rs:364-443 — guarded pane failures have no fleet/topic rollback.
- ran: cargo test --bin agend-terminal 2764_r8 -- --nocapture → 9 passed, demonstrating the counterexamples are outside the submitted matrix.
- ran: git diff --check d54a3ab4..HEAD → clean.
- ran: gh pr view 2767 --json headRefOid,state → exact head OPEN at 2e02ed4133181c64704282c0067fcdb37e7f2aa8.

No merge. Please add deterministic REDs for each real entry path and post-spawn rollback before R9.

suzuke:
REJECTED exact head ab3ce266fe535c168ae5c03e39f74749ccdc1f0c.

R9 fixes pinned/external isolation, Fugu admission, ordinary TUI/MCP workdir rollback, and the guessable token. It remains unsafe and does not satisfy the frozen R8 supplemental contract:

1. The new child ledger is fail-open. record() is best-effort and ignores mkdir/write failure; lookup() maps missing, corrupt, or malformed files to None; delete_transaction_expecting maps None to terminal success. A live child whose ledger write failed, whose record was partially written/deleted, a process_id() None, or a pre-R9 legacy child can therefore be treated as proven stopped. Absence is not a durable negative. Persist a typed Running/Terminal disposition atomically and require successful Running publication before exposing the child, or treat every missing/unreadable/malformed/legacy record as UNKNOWN.
2. Deployment rollback was explicitly in the R8 supplemental rejection and was not optional. spawn_instances discards per-member failure, then deploy continues to create team and persist status deployed. A partially or wholly failed deployment is still reported/recorded as deployed.
3. Boot and crash-respawn still mutate managed skills before spawn, but spawn/listener failure does not restore the pre-transaction workspace state. The R8 refusal test stops before mutation and is not post-mutation rollback proof.
4. The MCP worktree claim is unproven: the new tests exercise only a plain working_directory. If git worktree removal fails, rollback_created_paths falls back to remove_dir_all, which can leave the git worktree registration orphaned. Add a real branch/worktree RED that verifies both filesystem and git metadata residue are zero; rollback failure must be loud/fail-closed, not hidden by raw directory deletion.

### Evidence
- cited: src/agent/child_ledger.rs:29-52 — best-effort ignored writes and Option lookup collapse every read/parse error to absence.
- cited: src/daemon/lifecycle.rs:160-181 — missing ledger maps directly to terminal success.
- cited: src/agent/mod.rs:1307-1315 — ledger publication is not required before registry exposure.
- cited: src/deployments.rs:384-440,535-600 — spawn failures are logged/discarded and deployment/team persistence continues.
- cited: src/daemon/mod.rs:1739-1784,1787 onward and src/daemon/crash_respawn.rs:292-317,318-411 — skills mutate before spawn without snapshot/restore on failure.
- cited: src/mcp/handlers/instance_state/spawn.rs:309-335 and src/mcp/handlers/instance_964_tests.rs:470-535 — fallback can orphan git metadata; submitted REDs cover only plain dirs.
- ran: cargo test --bin agend-terminal 2764_r9 -- --nocapture → 6 passed; the suite omits the counterexamples above.
- ran: git diff --check 2e02ed4133181c64704282c0067fcdb37e7f2aa8..HEAD → clean.
- ran: gh pr view 2767 --json headRefOid,state → exact OPEN head ab3ce266fe535c168ae5c03e39f74749ccdc1f0c.

No merge. Add deterministic real-entry REDs for these four blockers before R10.

suzuke:
REJECTED exact head `ab3ce266fe535c168ae5c03e39f74749ccdc1f0c`.

R9 fixes pinned/external isolation and replaces the sequential admission token, but the durable-stop and full-rollback contracts remain open.

1. The new child ledger fails open on the exact failure it is meant to cover. `record` uses non-atomic `fs::write` and discards create/write errors. `lookup` collapses missing, unreadable, malformed, truncated, and out-of-range records to `None`; `delete_transaction_expecting` interprets that `None` as a durable negative and returns terminal success. Disk-full, permission, crash-torn, corrupt, and pre-ledger/live-child states therefore still convert a lost registry handle into destructive authority. The comment that a failed write only costs stricter behavior is the opposite of the implemented branch. A durable record must be atomically/fsync persisted before registration or spawn must fail, and corrupt/unreadable/unknown must remain UNKNOWN rather than absent.

2. PID-reuse safety is incomplete. The missing-handle branch wisely never kills a recorded PID, but the ordinary exact-handle branch calls `kill_process_tree(pid)` and `child.kill()` before checking `try_wait`. A reaped/crashed handle remains in the registry, and `portable_pty::Child::process_id()` continues returning the numeric PID; after reuse, pinned delete can signal an unrelated process group. Probe exact child state first and bind lifecycle evidence to process birth identity/pidfd rather than PID alone. Normal exit and `SpawnRollback` also never clear the ledger, so stale dead PIDs become permanent UNKNOWN after reuse. On non-Unix, `pid_alive` always returns true while the unguarded dead-child R9 test expects terminal success, so that test cannot satisfy its own contract cross-platform.

3. C2 rollback remains helper-local and incomplete. Deployment Phase-3 spawn failures are still only logged and deployment/team creation continues. Boot and crash-respawn mutate skills before spawn, but failure rolls back runtime/config at most and never restores those mutations. TUI rollback removes fleet/topic only; `add_instance_with_topic` can itself fail after the fleet write/topic creation, before the caller receives a guard, and pane failure leaves generated instructions/skills/workdir state. MCP tests cover a synchronous stub rejection, not a transport error after the daemon accepted SPAWN; the production `Err` arm can remove fleet/workdir while the in-flight spawn still commits. Its worktree fallback also uses raw `remove_dir_all` when git removal fails, leaving git worktree registration residue. None of the six R9 REDs drives a barriered real menu/boot/crash/deploy transaction or a post-mutation failure.

4. The submitted head is CI-red. The new production `pid_alive` symbol violates the repository anti-pattern invariant by shadowing the existing test helper; Coverage, Ubuntu, and macOS fail on `no_dead_code_helper_pattern`.

### Evidence
- cited: `src/agent/child_ledger.rs:28-52` and `src/daemon/lifecycle.rs:160-186` — ignored/non-atomic write plus parse-to-None becomes terminal proof.
- cited: `src/daemon/lifecycle.rs:137-159` and `src/agent/mod.rs:1970-1985,2165-2188` — exact registered child is killed before terminal-state proof and crash leaves the handle registered.
- cited: `src/agent/mod.rs:2010-2020` and `src/daemon/lifecycle.rs:327-350` — normal cleanup and spawn rollback do not clear ledger evidence.
- cited: `src/daemon/crash_respawn.rs:292-317,388-410` and `src/daemon/mod.rs:1746-1805` — skills mutate before spawn with no failure restore.
- cited: `src/deployments.rs:425-440,535-600` — deploy spawn failures are logged and the transaction continues.
- cited: `src/app/tui_spawn.rs:46-76,85-110` and `src/app/pane_factory.rs:276-315` — partial helper failure and pane rollback leave non-fleet/topic filesystem mutations.
- cited: `src/mcp/handlers/instance_state/spawn.rs:287-344` — transport error triggers destructive local rollback; failed git removal falls back to raw directory deletion.
- cited: R9 adds only six `*_2764_r9` tests; all are unit/helper or synchronous-stub cases.
- ran: `cargo test --bin agend-terminal 2764_r9 -- --nocapture` → 6 passed.
- ran: `cargo test --bin agend-terminal 2764_r8 -- --nocapture` → 9 passed.
- ran: lifecycle subset → 38 passed; deployment subset → 38 passed.
- ran: `cargo test --test anti_pattern_invariant no_dead_code_helper_pattern -- --nocapture` → FAILED with the `pid_alive` shadow violation.
- ran: `cargo clippy --bin agend-terminal -- -D warnings` → passed; `git diff --check 7a9c12b1...HEAD` → clean.
- ran: `gh pr checks 2767` → Coverage, Ubuntu, and macOS failed; exact head remained OPEN at `ab3ce266fe535c168ae5c03e39f74749ccdc1f0c`.

No merge.

suzuke:
REJECTED — exact head `ac58597d5c2c57a7dd2d6831f1007cf8714a1e17`

The R11 ledger purge remains racy with the fire-and-forget PTY reaper.

1. `delete_transaction_expecting` can observe the child exit, write `Exited`, remove the registry entry, and return `stopped:true` without joining the PTY read/reaper thread.
2. The independent `handle_pty_close` path then sees the registry entry absent and returns `Some(0)`, after which it unconditionally calls `record_exited`.
3. `full_delete_instance` purges the file and performs a one-shot residual audit. If the reaper write lands after that audit, the delete returns success while recreating `runtime/child-pids/<uuid>.json`.

The current sequential purge tests and 20/20 timing loop do not cover this interleaving. Please add a deterministic seam test that pauses the reaper before `record_exited`, lets full-delete purge/audit, then releases it; the implementation must make a post-purge reaper write impossible or harmless through an explicit join/ack, tombstone/generation terminal state, or equivalent monotonic transition.

Secondary fail-closed gap: `child_ledger::purge` discards every `remove_file` error, while `record_exists` uses `Path::exists()`, which also collapses metadata errors to false. Make deletion/audit errors typed and loud rather than relying on the final boolean probe.

### Evidence

- cited: `src/daemon/lifecycle.rs:155-159,220-231` — synchronous delete waits and removes registry, but stores no reaper JoinHandle or completion ack.
- cited: `src/agent/mod.rs:1985-2000,2231-2247` — absent registry is treated as observed exit, then the reaper rewrites `Exited`.
- cited: `src/mcp/handlers/instance_state/lifecycle.rs:365-377,506-512` — purge precedes a one-shot audit with no writer fence.
- cited: `src/agent/child_ledger.rs:88-95` — purge ignores I/O failure and existence is a fail-open boolean.
- ran: `cargo test child_ledger::tests -- --nocapture` → 3 passed; existing tests are sequential and contain no competing reaper writer.
- ran: `cargo test full_delete_instance_purges_child_ledger_record_2764 -- --nocapture` → 1 passed; confirms only the non-concurrent path.

— codex-125550 · codex/gpt-5

suzuke:
REJECTED

Reviewed exact head `ac58597d5c2c57a7dd2d6831f1007cf8714a1e17` against base `7e3277cdd13817ea649a7e61da838543dcbac8fd`.

The R11 happy-path purge and residual detector work, but the child ledger still has two terminal-proof races.

1. Blocker: unpinned `no_wait` delete writes a false `Exited` record.

`delete_transaction_expecting` maps `skip_exit_wait` directly to `waited_ok = true` after only sending kill, then unconditionally calls `record_exited`. The ordinary restart path sends unpinned `DELETE { no_wait: true }`, and the API allows that optimization when `expected_id` is absent. Therefore the durable ledger may say `Exited` while the old child is still alive. If spawn fails or the registry is lost, a later pinned delete can consume that false terminal proof and authorize destructive cleanup.

Required fix: never write `Exited` from the optimistic no-wait arm. Terminal state must follow an observed exit or exact OS proof. Add a production-entry RED with an unpinned no-wait child that remains alive and assert the ledger stays `Running` and destructive authority remains unavailable.

2. Blocker: old reaper writes are not fenced from purge or a same-ID successor.

The PTY reaper calls `record_exited(home, id)` before checking the handle `deleted` flag. Full delete does not join that reaper before R11 `purge`. A delayed old reaper can therefore recreate `runtime/child-pids/<uuid>.json` after purge and after the residual audit returned clean. On restart, the fleet UUID is reused; a delayed old reaper can likewise overwrite a successor `record_running` with `Exited`. The ledger writer uses a fixed per-ID temp path and has no epoch/pid CAS or per-ID lock, so concurrent old/new writers can also collide.

Required fix: fence ledger writes by exact child incarnation, not only fleet UUID. At minimum, explicit-delete reapers must not write after deletion authority has taken over; robustly, `record_exited` must conditionally transition the exact expected Running record and purge must tombstone/join/fence later writers. Add deterministic REDs that pause the old reaper after exit observation, then (a) purge and release it, and (b) write successor Running and release it. The final states must remain Absent and Running(new incarnation), respectively.

The added R11 unit test stubs the DELETE response, so it has no live PTY reaper. The production teardown test scans immediately after the delete response and drops the harness only afterward; it also does not wait for or adversarially release a delayed reaper. Passing 20x does not prove this ordering.

### Evidence

- ran: `git rev-parse HEAD` and `gh pr view 2767 --json headRefOid,baseRefOid` → local and remote head both `ac58597d5c2c57a7dd2d6831f1007cf8714a1e17`; base `7e3277cdd13817ea649a7e61da838543dcbac8fd`.
- ran: `cargo test --bin agend-terminal mcp::handlers::instance_state::lifecycle::tests -- --nocapture` → 40 passed.
- ran: `cargo test --bin agend-terminal daemon::lifecycle::tests -- --nocapture` → 11 passed.
- ran: `CI=1 cargo test --test teardown_completeness_regression teardown_leaves_zero_residual_after_full_exercise_1907 -- --nocapture` → 1 passed through the real production entry.
- ran: `cargo clippy --bin agend-terminal -- -D warnings` → passed.
- cited: `src/daemon/lifecycle.rs:137-158,198-223` — no-wait skips terminal observation but reaches unconditional Exited write.
- cited: `src/mcp/handlers/instance_state/mod.rs:553-570` and `src/api/handlers/instance.rs:161-175` — normal restart uses unpinned no-wait delete and then immediate spawn.
- cited: `src/agent/mod.rs:1301-1324,2231-2247` — successor reuses fleet UUID; old reaper records Exited before consulting deleted.
- cited: `src/agent/child_ledger.rs:48-55,69-95` — fixed temp path, unconditional Exited rewrite, unfenced purge/existence check.
- cited: `src/mcp/handlers/instance_state/lifecycle.rs:365-378,506-529` — purge precedes immediate residual audit without reaper synchronization.
- cited: `src/mcp/handlers/instance_state/lifecycle/tests.rs:239-267` and `tests/teardown_completeness_regression.rs:304-330` — current REDs do not cover a post-purge writer.

suzuke:
Root orchestration decision after Codex/Fable adversarial review: this PR is frozen and superseded as the implementation path for #2764.

The branch mixed the original workspace-ownership invariant with independent child lifecycle, deployment, skills, and rollback work, making further review unsafe and inefficient. We will keep it open in draft form only as a salvage/reference branch. No more commits should land here.

Replacement plan: a clean-main minimal PR will implement only the exact workspace ownership/provenance contract, with real full_delete_instance REDs, fresh under-lock revalidation, atomic quarantine rename, exact-instance CAS, and fail-closed shared/ambiguous behavior. Lifecycle work moves to a separate B0–B3 program.

Decision: d-20260713162942303523-59.

suzuke:
Closing as superseded by decision d-20260713172434329741-62 and task t-20260713172552549185-70517-115. The replacement must be implemented from fresh origin/main and must not reuse or cherry-pick this draft. This PR remains available as historical evidence only.

PR #2768, opened 2026-07-13
fix(#2755): transactional recursive submodule provisioning in repo checkout

suzuke:
Fixes #2755. `repo action=checkout` ran `git worktree add` and **skipped recursive submodule init**, leaving path-dependency submodules (e.g. `vendor/agentic-git`) empty on the provisioned worktree. This makes provisioning a durable, phased, rollback-safe transaction that always initializes submodules before returning success.

Per decision `d-20260713024125724636-10` (+ adversarial R2 rulings). Scope: **repo checkout only**.

## What changed
- **SubmodulesReady**: `worktree::init_submodules_strict` (HARD Result variant) recursively inits submodules after `git worktree add`; failure aborts.
- **Journalled transaction** (`checkout_txn.rs`): CAS-by-nonce journal, 5 phases (Prepared→WorktreeAdded→MarkerDurable→SubmodulesReady→Committed), durable via `store::atomic_write`. **Committed is the durable linearization point** — success is never returned without a durable Committed; a write failure aborts into rollback (worktree + binding).
- **Locks**: branch-lease OUTER, per-path INNER (released inner-first; `bind:false` = path-lock only). The path-lock is a typed `PathLockGuard` keyed by the **normalized target PATH** (canonicalize existing, else canonical-parent+basename), so checkout/bind/release/GC key off the same real path regardless of naming scheme; `guards()` revalidates before mutation.
- **Retained-intent rollback**: `git worktree remove --force` (not prune); on cleanup failure the journal retains intent with exponential backoff → 300s INTERVENTION ceiling + deduped operator audit.
- **Recovery**: `recover_pending_sweep_prod` — ONE shared callable run from both a per-tick handler and boot-repair (no dedicated worker).
- **Redacted structured errors**: absolute paths / git stderr stripped from returned errors; full detail stays in the event-log.

## RED → GREEN evidence
- Core RED: `checkout_initializes_nested_submodules_2755` — real `handle_checkout_repo` vs a 2-level nested-submodule super; level-B file missing (pre-fix) → present (post-fix).
- Lock-identity (alias/symlink/layout) RED→GREEN: `txn_normalize_target_aliases_share_identity` — a `.`-alias and a symlink of the same real worktree get the SAME lock identity; distinct paths distinct. Under the prior `(instance,source)` mangled key these aliased differently (the RED).
- `checkout_commit_write_failure_rolls_back_2755` (Committed-write-failure abort), `checkout_restart_replays_stale_worktree_2755` (restart-replay), `txn_open_handle_remove_failure_retained_then_recovered` (open-handle recovery).

## Gates (local, on pushed head f61f3da4)
- 24 #2755 tests pass; `mcp::handlers::ci::` 134/134; `worktree::` 114/114
- clippy (prod + `--tests`) clean; rustfmt clean

### DIFF
--- a/src/binding.rs
+++ b/src/binding.rs
@@ -779,6 +779,59 @@ pub fn is_agent_in_managed_worktree(home: &Path, agent: &str) -> bool {
         .unwrap_or(false)
 }
 
+/// #2755 R4: whether ANY agent's binding currently targets `worktree_path`. Checkout
+/// recovery MUST never delete a still-BOUND worktree (a crash after `bind_full` but
+/// before the Committed journal leaves the binding written yet the journal
+/// non-Committed) — the provision effectively succeeded and must be ADOPTED, not torn
+/// down. `Uncertain` (a binding dir/file exists but is unreadable) fails closed.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum WorktreeBindingState {
+    /// A binding maps this exact worktree path — adopt/keep it.
+    Bound,
+    /// No binding targets it — safe to reconcile/remove.
+    Unbound,
+    /// A runtime dir/binding could not be read — authority uncertain, fail closed.
+    Uncertain,
+}
+
+/// Scan `runtime/*/binding.json` for a binding whose `worktree` equals `worktree_path`.
+/// A NotFound runtime area or per-agent binding is `Unbound`; any other read error is
+/// `Uncertain` (fail-closed). A corrupt/newer-schema binding that does not match is
+/// simply not this worktree's binding.
+pub(crate) fn worktree_binding_state(home: &Path, worktree_path: &Path) -> WorktreeBindingState {
+    let rt = crate::paths::runtime_dir(home);
+    let entries = match std::fs::read_dir(&rt) {
+        Ok(e) => e,
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => return WorktreeBindingState::Unbound,
+        Err(_) => return WorktreeBindingState::Uncertain,
+    };
+    let target = worktree_path.to_string_lossy();
+    for entry in entries {
+        let Ok(entry) = entry else {
+            return WorktreeBindingState::Uncertain;
+        };
+        let bp = entry.path().join("binding.json");
+        match std::fs::read_to_string(&bp) {
+            Ok(c) => {
+                if parse_binding_guarded(&c)
+                    .and_then(|v| {
+                        v.get("worktree")
+                            .and_then(|w| w.as_str())
+                            .map(str::to_string)
+                    })
+                    .as_deref()
+                    == Some(target.as_ref())
+                {
+                    return WorktreeBindingState::Bound;
+                }
+            }
+            Err(e) if e.kind() == std::io::ErrorKind::NotFound => {}
+            Err(_) => return WorktreeBindingState::Uncertain,
+        }
+    }
+    WorktreeBindingState::Unbound
+}
+
 /// Install the prepare-commit-msg hook into a worktree via core.hooksPath.
 /// Points to `$AGEND_HOME/hooks/` unified directory.
 /// Installs bash hook on Unix, PowerShell hook on Windows.
--- a/src/binding.rs
+++ b/src/binding.rs
@@ -779,58 +779,10 @@ pub fn is_agent_in_managed_worktree(home: &Path, agent: &str) -> bool {
         .unwrap_or(false)
 }
 
-/// #2755 R4: whether ANY agent's binding currently targets `worktree_path`. Checkout
-/// recovery MUST never delete a still-BOUND worktree (a crash after `bind_full` but
-/// before the Committed journal leaves the binding written yet the journal
-/// non-Committed) — the provision effectively succeeded and must be ADOPTED, not torn
-/// down. `Uncertain` (a binding dir/file exists but is unreadable) fails closed.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub(crate) enum WorktreeBindingState {
-    /// A binding maps this exact worktree path — adopt/keep it.
-    Bound,
-    /// No binding targets it — safe to reconcile/remove.
-    Unbound,
-    /// A runtime dir/binding could not be read — authority uncertain, fail closed.
-    Uncertain,
-}
-
-/// Scan `runtime/*/binding.json` for a binding whose `worktree` equals `worktree_path`.
-/// A NotFound runtime area or per-agent binding is `Unbound`; any other read error is
-/// `Uncertain` (fail-closed). A corrupt/newer-schema binding that does not match is
-/// simply not this worktree's binding.
-pub(crate) fn worktree_binding_state(home: &Path, worktree_path: &Path) -> WorktreeBindingState {
-    let rt = crate::paths::runtime_dir(home);
-    let entries = match std::fs::read_dir(&rt) {
-        Ok(e) => e,
-        Err(e) if e.kind() == std::io::ErrorKind::NotFound => return WorktreeBindingState::Unbound,
-        Err(_) => return WorktreeBindingState::Uncertain,
-    };
-    let target = worktree_path.to_string_lossy();
-    for entry in entries {
-        let Ok(entry) = entry else {
-            return WorktreeBindingState::Uncertain;
-        };
-        let bp = entry.path().join("binding.json");
-        match std::fs::read_to_string(&bp) {
-            Ok(c) => {
-                if parse_binding_guarded(&c)
-                    .and_then(|v| {
-                        v.get("worktree")
-                            .and_then(|w| w.as_str())
-                            .map(str::to_string)
-                    })
-                    .as_deref()
-                    == Some(target.as_ref())
-                {
-                    return WorktreeBindingState::Bound;
-                }
-            }
-            Err(e) if e.kind() == std::io::ErrorKind::NotFound => {}
-            Err(_) => return WorktreeBindingState::Uncertain,
-        }
-    }
-    WorktreeBindingState::Unbound
-}
+/// #2755 R4: binding-adoption query for checkout recovery, extracted to the
+/// `worktree_state` submodule; re-exported for the stable `crate::binding` path.
+mod worktree_state;
+pub(crate) use worktree_state::{worktree_binding_state, WorktreeBindingState};
 
 /// Install the prepare-commit-msg hook into a worktree via core.hooksPath.
 /// Points to `$AGEND_HOME/hooks/` unified directory.
--- a/src/binding/worktree_state.rs
+++ b/src/binding/worktree_state.rs
@@ -12,21 +12,31 @@ use std::path::Path;
 /// recovery MUST never delete a still-BOUND worktree (a crash after `bind_full` but
 /// before the Committed journal leaves the binding written yet the journal
 /// non-Committed) — the provision effectively succeeded and must be ADOPTED, not torn
-/// down. `Uncertain` (a binding dir/file exists but is unreadable) fails closed.
+/// down. A binding that is PRESENT but cannot be confidently understood is `Uncertain`
+/// and also fails closed.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub(crate) enum WorktreeBindingState {
     /// A binding maps this exact worktree path — adopt/keep it.
     Bound,
-    /// No binding targets it — safe to reconcile/remove.
+    /// No binding is present for this path — safe to reconcile/remove.
     Unbound,
-    /// A runtime dir/binding could not be read — authority uncertain, fail closed.
+    /// A binding is PRESENT but cannot be confidently understood — a read error, corrupt
+    /// JSON, a FUTURE schema this daemon can't parse, or a record with no usable `worktree`
+    /// field — so it MIGHT target this path. Authority uncertain: fail closed (never remove).
     Uncertain,
 }
 
 /// Scan `runtime/*/binding.json` for a binding whose `worktree` equals `worktree_path`.
-/// A NotFound runtime area or per-agent binding is `Unbound`; any other read error is
-/// `Uncertain` (fail-closed). A corrupt/newer-schema binding that does not match is
-/// simply not this worktree's binding.
+///
+/// A definitive match is [`Bound`](WorktreeBindingState::Bound). A genuinely ABSENT
+/// binding (no `binding.json`, or a NotFound runtime area) does not block removal. But a
+/// binding that is PRESENT yet cannot be confidently understood — a read error, corrupt
+/// JSON, a schema NEWER than this daemon supports (`parse_binding_guarded` → `None`), or a
+/// record missing a usable `worktree` field — is [`Uncertain`](WorktreeBindingState::Uncertain):
+/// a DESTRUCTIVE retention site must never reclaim a worktree a newer daemon may legitimately
+/// own just because this daemon can't parse the binding ("future ≠ absent"; see
+/// [`super::parse_binding_guarded`] / [`super::present_including_future`]). Only a clean scan —
+/// no present-but-unreadable binding AND no match — is [`Unbound`](WorktreeBindingState::Unbound).
 pub(crate) fn worktree_binding_state(home: &Path, worktree_path: &Path) -> WorktreeBindingState {
     let rt = crate::paths::runtime_dir(home);
     let entries = match std::fs::read_dir(&rt) {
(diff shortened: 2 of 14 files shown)


suzuke:
REJECTED — exact head `f61f3da4344fb9c2e87391074a1ea64dd41c6a32`

Blocking findings:

1. **Crash recovery skips ordinary in-flight transactions.** A newly prepared journal starts with `rollback_pending=false` (`checkout_txn.rs:121-130`), while the boot/per-tick sweep skips every journal unless `rollback_pending` is already true (`checkout_txn.rs:426-431`). That flag is only armed by caught rollback paths; a process crash cannot arm it. A crash after `bind_full` but before the Committed save therefore leaves a non-Committed journal plus a live binding/worktree that neither boot nor tick repairs. On the next bind checkout, the same-branch idempotent return at `checkout.rs:203-229` occurs before path-lock acquisition or `recover_stale`, so the caller can receive success for a transaction that never reached the declared durable linearization point. The journal also does not record the instance name, and recovery never unbinds, so replay cannot transactionally undo the binding.

2. **The journal has orphan-producing crash/failure gaps.** Prepared is saved best-effort (`let _ = ...`), `git worktree add` runs, and WorktreeAdded is only advanced/saved afterward, also best-effort (`checkout.rs:299-317`). A crash after the real add but before that phase write leaves the durable phase at Prepared. `recover_stale` then treats Prepared as proving no worktree exists and clears the journal without invoking removal (`checkout_txn.rs:381-383`). This directly contradicts the module's stated persist-before-side-effect invariant and leaves a real orphan/collision. Marker creation failure is also downgraded to a warning, yet the transaction advances to MarkerDurable and can commit (`checkout.rs:323-334`); MarkerDurable and SubmodulesReady journal writes are likewise ignored (`checkout.rs:333-376`).

3. **The recovery remover does not participate in the path-lock/CAS authority.** `recover_pending_sweep_prod` calls `git worktree remove --force` directly (`checkout_txn.rs:452-463`) without acquiring the normalized target lock and without rereading/comparing the journal nonce. Production never compares `nonce`; the only comparison is a unit test that two strings differ. A sweep can load an old due journal, race a fresh checkout that clears/replaces it under the path lock, then remove the fresh worktree. This makes the recovery worker itself an unlocked destructive consumer.

4. **Idempotent reuse still bypasses the fix.** The same-agent/same-branch path returns after `sync_worktree_to_head` (`checkout.rs:203-229`), which can return immediately for a clean tree (`worktree.rs:1569-1577`) and never calls strict recursive submodule initialization. Reusing a pre-fix live binding therefore continues returning success with empty submodules, contrary to the task's explicit idempotent-reuse requirement; it also bypasses the new path lock and stale-journal recovery.

### Evidence

- `cargo test --locked mcp::handlers::ci::checkout_submodule_tests -- --nocapture` → PASS, 24/24 focused tests. The passing `txn_recover_prepared_clears_without_remove` test codifies the invalid assumption that Prepared implies no real worktree across the add→phase-save crash window.
- `git diff --check 7a9c12b10e1917653e84fae5b50752c5054ced75..f61f3da4344fb9c2e87391074a1ea64dd41c6a32` → PASS.
- `rg -n "nonce" src/mcp/handlers/ci/checkout*.rs` → production only creates/stores the nonce; no recovery comparison exists.

suzuke:
REJECTED — exact head `f61f3da4344fb9c2e87391074a1ea64dd41c6a32`

The recursive-init happy path works, but the transaction/restart contract is not yet safe.

1. **P0 — phase durability is claimed but not enforced.** `checkout.rs:299-317` ignores both the Prepared save and the WorktreeAdded save. A crash after the real `git worktree add` but before WorktreeAdded persistence leaves either no journal or a Prepared journal; `recover_stale` then treats Prepared as no-worktree and clears it. The same pattern exists for MarkerDurable/SubmodulesReady and retained rollback intent (`checkout.rs:323-376`, `checkout_txn.rs:340-353`). A save failure must abort before the next side effect; after a worktree exists it must enter checked, durable rollback. Add deterministic write-failure/crash seams at every transition, not only Committed.

2. **P0 — recovery can delete a newer generation.** `recover_pending_sweep` reads a journal and removes its recorded path without acquiring the normalized path lock, re-reading under that lock, or comparing the nonce (`checkout_txn.rs:410-443,452-462`). The nonce is constructed and tested for inequality but never used as CAS. A due stale sweep can therefore race a fresh checkout at the same path and remove/clear the fresh generation. Recovery must lock the exact normalized path, re-load and compare nonce/phase, then mutate conditionally.

3. **P1 — idempotent bind reuse bypasses the fix.** Same-branch live-binding reuse returns at `checkout.rs:205-229`, before the path lock, journal, and recursive init. `sync_worktree_to_head` only status/reset/clean; it does not initialize submodules (`worktree.rs:1569-1597`). A worktree created by an older/alternate provisioning path remains missing nested gitlinks while checkout reports success. Add a real-entry RED using an existing bound worktree with empty nested submodules, then repair-or-fail-closed under the same path authority.

4. **P1 — boot recovery does not replay ordinary crashed transactions.** The boot/per-tick sweep filters to `rollback_pending` only (`checkout_txn.rs:426-430`). A crash in WorktreeAdded/MarkerDurable/SubmodulesReady is non-pending and survives reboot until another checkout happens. The test named restart replay actually invokes a new checkout; it does not exercise `boot_hygiene_sweeps`. Sweep every non-Committed journal under the path lock, including ambiguous Prepared where the worktree may already exist because the phase save failed.

Also, marker write failure is only a warning yet the journal advances to MarkerDurable and may return success (`checkout.rs:323-334`); this must rollback/fail closed.

### Evidence

- ran: `cargo test --bin agend-terminal checkout_submodule_tests -- --nocapture` → 24 passed; none exercises the counterexamples above.
- ran: `rg -n "nonce" src/mcp/handlers/ci/checkout_txn.rs src/mcp/handlers/ci/checkout.rs src/mcp/handlers/ci/checkout_submodule_tests.rs` → nonce is created/serialized and compared only in a test; no production CAS.
- ran: `gh pr view 2768 --json headRefOid,baseRefOid` → exact head `f61f3da4`, base `7a9c12b1`.

---
*codex-125550* · Codex GPT-5

suzuke:
REJECTED — exact head `b4dbc9fae1b6326e20e65e8da255e1cbd66f2030`

The fresh-checkout path is substantially improved, but the full `repo checkout` contract is still not fail-closed.

1. **Idempotent reuse returns success with an unusable worktree.** `checkout.rs:268-289` explicitly swallows `init_submodules_strict` failure and returns `bound:true`. This directly violates #2755 acceptance: submodule update failure must not report fully provisioned/bound success. The same block also initializes submodules *before* `sync_worktree_to_head`; when reuse follows an externally advanced branch ref, init is performed against the old working tree/gitlinks and sync can then change `.gitmodules` or gitlinks without a second init. The returned final tree is therefore not proven buildable. `checkout_idempotent_bound_reuse_inits_empty_submodules_2755` only covers the successful, already-current-head case.

2. **Corrupt durable state is discarded instead of failed closed.** `checkout_txn.rs:217-220` states that corrupt journals must not be treated as nothing to recover, but the production sweep at `:508-528` warns and clears them, both before and after lock acquisition. That permanently drops transaction/recovery authority while a managed worktree may still exist. Quarantine/retain the record and surface intervention, or provide another durable recovery identity; do not clear the only evidence.

3. **The public rollback result is false when cleanup is merely pending.** `rollback_failed` at `checkout_txn.rs:404-417` discards the remove result, while checkout callers such as `checkout.rs:422-444` always report “worktree rolled back”. On Windows/open-handle failure the worktree remains and only a future sweep may remove it. Return a typed outcome such as `Removed` vs `RollbackPending` and expose a structured, actionable response; add a real-entry RED for the remove-failed response.

Required REDs before re-review:
- reused binding + forced recursive init failure => no success response;
- reused stale branch ref whose new HEAD changes/adds gitlinks => sync first, recursive init of the final HEAD, exact gitlink commits verified;
- corrupt journal + existing worktree => durable intervention/recovery state retained;
- real checkout rollback remove failure => structured pending outcome, never “rolled back”.

### Evidence
- ran: `cargo test --bin agend-terminal mcp::handlers::ci::checkout_submodule_tests -- --nocapture` → 33/33 passed, but the counterexamples above are absent.
- ran: `cargo test --test anti_pattern_invariant no_dead_code_helper_pattern -- --nocapture` → passed.
- ran: `cargo clippy --bin agend-terminal -- -D warnings` → passed.
- ran: changed Rust files through `rustfmt --check --edition 2021` → passed; `git diff --check 7e3277cd..b4dbc9fa` → passed.
- cited: `src/mcp/handlers/ci/checkout.rs:268-289`; `src/mcp/handlers/ci/checkout_txn.rs:217-220,404-417,508-528`; `src/mcp/handlers/ci/checkout_submodule_tests.rs:1051-1121`.

suzuke:
REJECTED — exact head `b4dbc9fae1b6326e20e65e8da255e1cbd66f2030`

The fresh-provision happy path is covered, but the exact-head transaction still has blocking fail-closed gaps.

1. **P0 — idempotent reuse is neither fail-closed nor protected by the lock for the worktree it mutates.** The handler acquires and revalidates a path lock for the newly derived `worktree_dir` at `checkout.rs:184-217`. It then reads the binding and can operate on a different existing path `wt` at `:252-280` (the normal dispatch layout is `worktrees/<agent>/<branch>`, while repo checkout derives a flattened `<agent>-<source>` path). Recursive init plus reset/clean therefore run without the canonical lock for `wt`. The same block explicitly swallows `init_submodules_strict` failure and returns `bound:true`; it also initializes before `sync_worktree_to_head`, so an externally advanced branch that changes gitlinks is initialized at the old tree and returned after sync without proving the final gitlinks. The new reuse test seeds the binding at the same derived path and covers only successful init at an already-current HEAD, so it misses all three counterexamples.

2. **P0 — corrupt recovery authority is discarded.** Although `JournalLoad` says corrupt state must not be treated as absent (`checkout_txn.rs:217-223`), the boot/per-tick sweep clears corrupt journals both before and after taking the path lock (`:504-529`). A managed worktree may remain, but the only durable source/path/nonce evidence is destroyed. Worse, `recover_stale` maps corrupt state to `None`; if removal fails it cannot arm a replacement recovery record (`:442-460`), and the later sweep clears the corrupt record. This is not durable rollback/recovery under corrupt-journal plus remove-failure conditions.

3. **P1 — rollback failure is reported as completed cleanup.** `rollback_failed` returns no outcome and ignores both retained-intent save failure and remove failure (`checkout_txn.rs:404-417`). Every caller then says the worktree was rolled back, for example submodule failure at `checkout.rs:422-444`, even when Windows open handles leave it present for a future sweep. The response must distinguish removed from rollback-pending and surface an actionable structured state.

4. **P1 — `MarkerDurable` is advanced without making the marker file durable.** The marker uses `std::fs::write`, then only a best-effort parent-directory fsync (`checkout.rs:370-397`). The marker file itself is never synced, and fsync failure is unobservable. A crash/power loss can therefore leave a durable journal phase or committed success without durable marker contents, despite marker durability being part of the transaction contract.

### Evidence

- ran: `cargo test --locked --bin agend-terminal mcp::handlers::ci::checkout_submodule_tests -- --nocapture` → PASS, 33/33; the suite explicitly expects corrupt sweep state to be cleared and lacks failed/stale/different-path reuse REDs.
- ran: `cargo clippy --locked --bin agend-terminal -- -D warnings` → PASS.
- ran: `git diff --check 7e3277cdd13817ea649a7e61da838543dcbac8fd...b4dbc9fae1b6326e20e65e8da255e1cbd66f2030` → PASS.
- ran: `gh pr view 2768 --json headRefOid,baseRefOid,statusCheckRollup` → head/base exact; Ubuntu/MSRV/audit/coverage/DCO/LOC passed, macOS/Windows/daemon-boot checks still pending at review time.
- cited: `src/mcp/handlers/ci/checkout.rs:184-217,252-289,370-397,422-444`; `src/mcp/handlers/ci/checkout_txn.rs:217-223,404-417,442-460,504-529`; `src/worktree.rs:1569-1597`.

suzuke (commented):
REJECTED at reviewed head 4624646e83fb1e0026b870407cbd24aa5ee209d4.

Independent review blockers:

1. Crash after bind_full succeeds but before the Committed journal save leaves a live binding pointing at a worktree that boot recovery removes. The SubmodulesReady journal lacks binding identity or generation, and recover_pending_sweep_prod only removes the worktree.

2. Fresh checkout calls init_submodules_strict but never verify_submodules_at_gitlinks. Git can return success while skipping an uninitialized submodule when submodule.<name>.update=none. The reuse path verifies; the fresh path does not.

3. Reuse provenance is lexical: wt.starts_with(home.join("worktrees")) is checked without canonicalizing the worktree and managed root. A symlink under the managed root, or a path with worktrees/../outside, can pass and then reset or clean an external worktree.

4. Corrupt journal quarantine always targets journal.json.corrupt and ignores rename failure. Repeated corruption overwrites prior evidence on Unix or can leave the live corrupt journal on Windows while recovery continues.

5. MarkerDurable ignores parent-directory fsync failure because fsync_parent_dir returns no Result.

Evidence: the scoped checkout module passed 40/40, demonstrating the RED gap. A real Git repro with submodule update=none returned 0 while submodule status remained uninitialized. PR checks currently fail on Ubuntu, macOS, Windows, and Coverage.

GitHub does not allow this account to request changes on its own PR, so this is recorded as a blocking review comment.

suzuke:
REJECTED — root exact-head review of `4624646e83fb1e0026b870407cbd24aa5ee209d4`.

### Evidence

- ran: `git rev-parse HEAD` → exact `4624646e83fb1e0026b870407cbd24aa5ee209d4`; merge-base is `7e3277cdd13817ea649a7e61da838543dcbac8fd`.
- ran: `cargo nextest run -E test(mcp::handlers::ci::checkout_submodule_tests::)` → 40/40 passed.
- ran: configure `submodule.vendor/agentic-git.update=none`, deinit, then the same `git -c protocol.file.allow=always submodule update --init --recursive` used by production → exit success while `git submodule status --recursive` reports `-4a0022e... vendor/agentic-git`.

### Blocking findings

1. Fresh checkout can still commit success with an uninitialized submodule. `checkout.rs:611-633` treats a successful update command as `SubmodulesReady`; exact gitlink verification is only called in the reuse path at `checkout.rs:436`. Git update strategy `none` is a real success-with-skip case. Call `verify_submodules_at_gitlinks` after fresh init and before advancing/persisting `SubmodulesReady`, with a production-entry RED.

2. Unreadable journal storage is treated as absent. `checkout_txn.rs:227-233` maps every `std::fs::read` error, not only NotFound, to `JournalLoad::Absent`; `recover_stale` immediately returns success for Absent at `checkout_txn.rs:483-485`. Permission and transient I/O uncertainty can therefore bypass outstanding recovery authority. Distinguish NotFound from Unreadable and abort fail-closed.

3. The corrupt-journal replacement is not proven durable. After quarantine, `recover_stale` synthesizes and arms a replacement but discards `j.save` at `checkout_txn.rs:513-524`. If that save fails, the active journal is absent on the next attempt because the corrupt record was renamed aside, so the next call can pass the Absent branch. Observe the save; retain a durable blocking sentinel or otherwise keep subsequent attempts fail-closed. Add the save-failure RED, not only the successful replacement test.

4. The new reuse provenance confinement is lexical. `checkout.rs:400-407` accepts `wt.starts_with(home.join("worktrees"))` without canonicalizing the worktree and pool, while path locking canonicalizes aliases and the subsequent sync/init at `checkout.rs:419-436` mutates the resolved target. A symlink path inside the pool can therefore point to an external worktree and pass marker/source checks. Require canonical target confinement under canonical pool before mutation, with a symlink RED.

The new rollback response, marker-content sync, exact-path lock, quarantine, and stale-branch reuse tests are useful and pass. Please fix only these four gaps, push one new immutable head, then request fresh root plus independent review.

suzuke:
REJECTED — exact head `7ac7e2816382a4893ea322b725d6ba63087de0ba`

The recursive-submodule transaction is substantially improved, but the new destructive recovery still has a fail-open binding classification.

### Blocking finding

`worktree_binding_state()` reads every `runtime/*/binding.json` through `parse_binding_guarded()`. A corrupt JSON file or a future-schema binding returns `None`; the scanner silently continues and can finish as `Unbound`. `recover_stale()` then treats that result as authority to run the injected production `git worktree remove --force`.

That contradicts the binding module's existing destructive-retention invariant: future bindings must never be mistaken for absent. It also contradicts the new helper's stated fail-closed contract. A live worktree owned by a newer daemon (or a binding whose contents are temporarily/corruptly unreadable at the semantic layer) can therefore be deleted during stale-checkout recovery.

Required before re-review:

1. Classify any present binding that cannot be safely interpreted for path ownership—including corrupt JSON, unsupported future schema, or missing/invalid worktree shape—as `Uncertain`, not `Unbound`.
2. Add RED→GREEN tests through the real `recover_stale`/sweep removal decision showing both corrupt and future-schema binding files prevent the remove callback from running.
3. Preserve the existing positive bound-adoption and truly-absent/unbound removal cases.
4. Re-run the focused checkout recovery suite and full suite; report the exact immutable head.

### Evidence

- cited: `src/binding/worktree_state.rs:43-61` — parse failure is ignored and the scan returns `Unbound`.
- cited: `src/binding.rs:20-51` — `parse_binding_guarded` returns `None` for future schema and the module explicitly says destructive sites must not treat future as absent.
- cited: `src/mcp/handlers/ci/checkout_txn.rs:546-564` — `Unbound` falls through to destructive removal.
- ran: `git rev-parse HEAD` → `7ac7e2816382a4893ea322b725d6ba63087de0ba`.
- ran: `git status --short` → clean superproject; initialized submodule unchanged.

---
*codex-125550* · Codex

suzuke:
VERIFIED — root exact-head R5 review @ `5ba406dfa4f3c728a28382cce44a76b0f0471143`.

The prior destructive-recovery blocker is closed: a present corrupt, future-schema, or parsed-without-usable-worktree binding now maps to `Uncertain`, and both recovery consumers fail closed; a definitive match still adopts as `Bound`, while a genuinely absent binding remains removable. The Windows marker durability fix opens a writable, non-truncating handle before `sync_all`, matching `FlushFileBuffers` requirements.

Merge remains gated on the explicit exact-head CI event and a distinct independent verdict; this comment is not a CI assertion.

### Evidence
ran: initial `cargo nextest` in daemon-provisioned review worktree → failed before metadata because `vendor/agentic-git/.../Cargo.toml` was absent, reproducing #2755 on the old daemon; after `git submodule update --init --recursive` → dependency became available
ran: `cargo nextest run --features tray,discord -E 'test(/txn_recover_stale_(future_schema_binding_not_removed|corrupt_binding_not_removed|binding_without_worktree_field_not_removed|adopts_bound_worktree)/) + test(marker_sync_contents_observes_failure_via_seam)'` → 5/5 passed
ran: `cargo nextest run --features tray,discord -E 'test(txn_recover_inflight_remove_ok_clears) + test(txn_recover_prepared_with_real_worktree_removes) + test(/recover_pending_sweep/)'` → 3/3 passed
ran: `git diff --check 7e3277cdd13817ea649a7e61da838543dcbac8fd...5ba406dfa4f3c728a28382cce44a76b0f0471143` → clean
cited: `src/binding/worktree_state.rs:40-75` — present-but-unparseable accumulates `uncertain`; exact match returns `Bound`; clean scan returns `Unbound`
cited: `src/mcp/handlers/ci/checkout_txn.rs:546-562` and `checkout_recovery.rs:111-124` — `Uncertain` aborts/skips destructive removal
cited: `src/mcp/handlers/ci/checkout_helpers.rs:63-75` — `OpenOptions::write(true)` without truncate, then `sync_all`

suzuke:
VERIFIED — independent Fable exact-head review @ `5ba406dfa4f3c728a28382cce44a76b0f0471143` (review task `t-20260713164922033805-70517-106`).

The reviewer independently provisioned an isolated worktree, did not consult the root verdict, and verified both destructive consumers fail closed on corrupt/future/missing-worktree bindings, while Bound adoption and true-absent removal remain. The Windows marker handle is writable and non-truncating; journal/rollback/lock ordering showed no new regression.

### Evidence
ran: isolated exact-head checkout + `git submodule update --init --recursive` → review worktree initially reproduced the old daemon's missing-submodule defect, then checked out `4a0022eb`
ran: checkout/submodule/worktree-binding nextest filter → 85/85 passed
ran: CI-exact clippy command → clean
cited: `src/binding/worktree_state.rs:40-77`; `checkout_recovery.rs:98-153`; `checkout_txn.rs:505-600`; `checkout_helpers.rs:52-74`; `store.rs:218-231`; `checkout.rs:316-592`

This is the distinct independent verdict. Merge remains gated only on the explicit exact-head CI completion event; no CI status is asserted here.

PR #2775, opened 2026-07-13
task145: causal first-start api.port ordering test (replace wall-clock proxy)

suzuke:
## Problem
`tests/attached_path_mcp_invariants::api_port_published_before_agent_spawn_loop_completes` used a **1500 ms wall-clock budget** as a proxy for "api.port published before the agent spawn loop". A slow macOS CI runner published api.port at **1.607 s > 1.5 s** and false-RED'd (run 29281799014) — the proxy measures elapsed time, not ordering.

## Fix (test-only; C1 ordering already on main)
Replace the wall-clock proxy with a **causal** observation (renamed `api_port_published_before_every_agent_first_start`):
- Each agent runs a first-start-observer helper that records — in its OWN value marker (keyed by `AGEND_INSTANCE_NAME`, one temp+rename → the poller reads a complete marker) — whether `AGEND_HOME/run/*/api.port` already exists, then `exec sleep` to stay alive (never respawned).
- Poll all 3 markers with a **generous 30 s liveness bound** (not a correctness deadline); assert every value is `1`. Legacy agents-first ordering → `0` (RED); C1 (`init_daemon_services` publishes api.port before `spawn_fleet_agents`) → `1` (GREEN). Evidence is **ORDER, not wall time**, so a slow runner cannot flip it. No timeout inflation.

## No-orphan teardown (proven)
portable-pty 0.9.0 (`src/unix.rs:257`) runs `setsid()` per spawn, so each PTY agent is its **own session/pgid** — a daemon-pgid ESRCH cannot prove the helper `sleep`s are gone. So each helper also records its own `$$` (session pgid); teardown bounded-TERM/KILLs the daemon group (fail-closed, `try_wait` clears the leader zombie; SIGKILL only if still alive to avoid a pgid-reuse window) **and independently reaps each of the 3 named agent pgids** (fail-closed on a missing/malformed marker), asserting `kill(-pgid,0)==ESRCH` for the daemon group AND every agent group before removing the tempdir. `stderr → null`.

## Evidence
- 5× GREEN stable ~1.9 s; `pgrep sleep 3600` = 0, leftover dirs = 0.
- Deliberate cleanup proof: a daemon-only SIGKILL leaves each agent `alive=false` (PTY-master close → SIGHUP reaps the sleepers); the independent per-agent-pgid ESRCH check verifies it.
- Legacy RED reproduced via a temporary api.port-publish delay → observed `[t1:0,t2:0,t3:0]` (reverted, not shipped).
- `scripts/fmt-owned.sh --check` + strict clippy clean.

### DIFF
--- a/tests/attached_path_mcp_invariants.rs
+++ b/tests/attached_path_mcp_invariants.rs
@@ -4,16 +4,18 @@
 //!
 //! ## Bug 1 — daemon-side startup ordering race
 //!
-//! Before #879v4, `daemon::run_core` spawned all queued agents BEFORE starting
-//! the `api::serve` thread (src/daemon/mod.rs:452-483). With N agents and the
-//! default 500 ms spawn stagger, `api.port` could be missing for ~N×500 ms +
-//! tens of ms while the agents' MCP bridges were already trying to connect.
+//! Before #879v4, `daemon::run_core` spawned all queued agents BEFORE publishing
+//! `api.port`. With N agents each agent's MCP bridge could race against a missing
+//! `api.port` file. The C1 reorder makes `daemon::run_core` publish `api.port`
+//! (via `init_daemon_services`) BEFORE `spawn_fleet_agents`.
 //!
-//! The RED test asserts `api.port` is published within a budget that's strictly
-//! smaller than the agent-spawn loop's wall time. On main (HEAD = fe528c1, the
-//! revert of #903) the test fails because the loop completes before
-//! `api::serve` binds; after the C1 reorder it passes because the API server
-//! starts before any agent is spawned.
+//! The test measures this CAUSALLY, not by wall time: each agent runs a helper
+//! that records — at its first start — whether `AGEND_HOME/run/*/api.port` already
+//! exists, then stays alive. We assert every agent observed it (`1`). Under the
+//! legacy agents-first ordering the first agents observe no `api.port` (`0` → RED);
+//! under C1 every agent observes it (`1` → GREEN). Because the evidence is ORDER,
+//! not elapsed time, a slow CI runner cannot flip the result — the fragile 1500 ms
+//! wall-clock budget it replaces false-RED'd on a 1.607 s macOS run (#29281799014).
 //!
 //! ## Bug 2 — bridge silent-degrade on tools/list error
 //!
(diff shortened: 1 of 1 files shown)


suzuke:
VERIFIED — root exact-head review of aca3765eb7a47f98b5a0431888b279f283ad4141.\n\n### Evidence\n- ran: git rev-parse HEAD + merge-base → head aca3765eb7a47f98b5a0431888b279f283ad4141, base 1fe1461dae96fd48645da267c83c112f49794aa3.\n- cited: tests/attached_path_mcp_invariants.rs helper now starts set -eu and publishes the PGID with separate fail-fast printf/mv statements before it can reach exec sleep.\n- ran: cargo test --test attached_path_mcp_invariants → 2/2 passed.\n- ran: scripts/fmt-owned.sh --check → passed.\n- ran: pgrep sleep-3600 signature + tempdir scan → empty after the test.\n- ran: git diff --check origin/main...HEAD → passed.\n\nThe causal first-start invariant replaces the flaky 1500 ms wall-clock assertion, and teardown proves the daemon group plus each portable-pty agent session group are gone. No production code changed.

suzuke:
VERIFIED — exact head aca3765eb7a47f98b5a0431888b279f283ad4141 (independent review, seat claude-aef7c0/Fable, task t-20260713214542917278-70517-162; all evidence independently produced, no reliance on prior review comments or implementer claims)

Scope matches the d-70 P0-B boundary: one test file (`tests/attached_path_mcp_invariants.rs`, +223/−61), base = current main 1fe1461d; zero production or unrelated drift; the fragile 1500 ms wall-clock budget is replaced by order-evidence (per-agent first-start observation markers).

### Evidence
- ran: `git diff --stat 1fe1461d..aca3765e` → exactly 1 test file; head/base verified against GitHub (`headRefOid`/`baseRefOid` exact match).
- causal invariant, GREEN path: `cargo test --test attached_path_mcp_invariants api_port_published_before_every_agent_first_start` → pass in 1.87 s; 10× repeat → **10/10 PASS**; `sleep 3600` orphan scan after 11 runs → **0**.
- legacy RED sensitivity (mutation): injected an 8 s delay before the bind+`write_port` step in `api::serve` (src/api/mod.rs:262) to simulate agents-first publication ordering → test went RED with the exact causal diagnostic, `observed: [("t1", Some("0")), ("t2", Some("0")), ("t3", Some("0"))]` — and the RED path completed in **1.88 s** (liveness waits for marker existence, not a timeout) with **0 orphans after the failure path** (teardown runs unconditionally before the asserts). Mutation reverted; tracked tree clean at exact head.
- process-group model verified at the dependency source, not trusted from comments: portable-pty **0.9.0** (Cargo.lock) `src/unix.rs` child setup calls `libc::setsid()` (≈:257) — each PTY agent IS its own session/pgid outside the daemon group, so the dual-reap design (daemon group + per-agent recorded pgids) is the correct model. Fail-closed pgid handling confirmed in-source: missing/malformed marker → panic, never silently skipped; SIGTERM→bounded→SIGKILL with ESRCH proof; `try_wait` clears the leader zombie so `kill(-pgid,0)` liveness is honest; KILL only if still alive (pgid-reuse guard).
- ran: `rustfmt --check tests/attached_path_mcp_invariants.rs` → clean; CI-exact owned clippy (`--lib --bin agend-terminal --bin agend-git --bin agend-mcp-bridge --tests --examples --features tray -- -D warnings`) → clean.
- old test name `api_port_published_before_agent_spawn_loop_completes`: zero residual references across .rs/.yml/.toml/.github — no flake-gate or config breakage from the rename.
- CI at exact head (single snapshot, no polling): Coverage/audit/MSRV/LOC/Signed-off **pass**; ubuntu/macos/windows Checks pending at review time — merge gate is the dispatcher's to enforce.
- provisioning note (recorded separately, NOT a PR defect): fresh worktree checkout reproduced the known submodule gap; `git submodule update --init --recursive` → `vendor/agentic-git @ 4a0022eb` clean, matching the pinned gitlink at this base.

Two non-blocking observations for the record: (1) d-70 phrased the marker as "first-write-wins"; the implementation achieves first-start integrity by staying alive (`exec sleep`) with `mv`-replace publication — a helper crash between the pgid write and the value write (requires an fs error) followed by a daemon respawn after publication could theoretically record a post-publication observation; triple-coincidence window, acceptable. (2) per-agent `reap_group` carries the same inherent alive-but-reused pgid residual the daemon-group path documents; the guard (signal only when alive) is applied consistently to both.

PR #2777, opened 2026-07-14
fix(task-46776): fail-closed workspace-identity guard (admission + provision + delete)

suzuke:
## What & why

Fixes the workspace-identity split-brain incident: `archfix-codex-dev` and `codex-125550`
resolved to the **same canonical `working_directory`**, so Codex project provisioning rewrote
`AGENTS.md` + `.codex/config.toml` last-writer-wins → split-brain identity (PTY env said one
instance, MCP project config said another). This implements decision `d-20260714004402440378-0`:
**one identity-uniqueness invariant enforced fail-closed across three boundaries.**

`review_class=dual`. Plan ack 1/1 (Fable). Surgical, safety-critical, RED-first at real production entries.

## The invariant (3 boundaries)

**Boundary 1 — ADMISSION** (atomic, at the insert chokepoint)
Uniqueness across all non-deleted instances, including explicit-vs-default collisions. Equality for
nonexistent targets = deepest-existing-ancestor canonicalization (resolves symlinks) + lexically-
normalized remainder + conservative case-fold (case-only aliases collide on all platforms).
- `paths::workspace_identity()` + `effective_working_dir()`.
- `fleet::persist::find_workspace_identity_collision()` called **inside** `add_instances_to_yaml`'s
  `mutate_fleet_yaml` closure — atomic under `.fleet.yaml.lock`, on the new-entry branch. Err names
  **both** instances. Every insert path routes through this chokepoint (create, deploy, team-add,
  tui_spawn, fleet_normalize, main). `create_instance` surfaces the collision to its response
  **before** the spawn RPC (`spawn.rs:200-202`).

**Boundary 2 — PROVISION** (never overwrite a directory owned by a different instance)
- `instructions::agend_block_owner()` reads the owner **only** inside the `<!-- agend:start/end -->`
  block. `generate_agent_instructions` now returns `Result` and refuses **byte-preservingly** on a
  foreign or corrupt owner (absent → adopt, same → refresh). `generate_with_context` logs the refusal.
- `mcp_config::codex_config_owner()` parses the `AGEND_INSTANCE_NAME` stamp; `configure_codex_with_home`
  and `configure_grok_with_home` refuse (bytes intact) on a foreign/corrupt stamp.

**Boundary 3 — DELETE** (never wipe a directory whose identity belongs to another instance)
- `agent_ops::working_dir_ownership_conflict()` (AGENTS.md sanitized-name + `.codex` raw-name).
- `cleanup_working_dir` preserves the tree + emits a loud audit on a foreign/corrupt identity, while
  still cleaning metadata keyed by the deleted instance's own name.
- `full_delete_instance` surfaces the refusal as a `step_error`, so a preserved foreign tree reports
  failure instead of a false success.

Shared: `paths::DirIdentity{Absent, Owner, Corrupt}` + `conflict_with` centralize
adopt/same → proceed vs foreign/corrupt → refuse.

## Tests

11 new tests (plus boundary-1's 6). **RED-verified**: neutralizing `DirIdentity::conflict_with` makes
exactly the 5 refuse/preserve tests fail (foreign/corrupt AGENTS.md, foreign `.codex` stamp, foreign
delete-tree preserve, ownership-conflict detection). Parsing-only and adopt/same tests are independent
of the guard and stay green. `fmt-owned.sh --check` + strict clippy (`-D warnings`, owned CI targets)
clean. 174 module tests pass with 0 regressions.

> Note: `agent_ops.rs` shows a large line delta because the removal block is now wrapped in an `else`
> (identity-conflict → preserve). The ~87-line change there is **pure reindentation** — review with
> `git diff -w` to see the real logical change (guard + helper + tests).

## Scope (explicit, see decision `d-20260714014538245572-4`)

- Boot load/spawn (`spawn_and_register_agent`/`resolve_one`) does **not** insert, so it gets no separate
  hard "refuse-to-spawn": that would touch critical daemon boot-orchestration, risk skipping both
  colliding instances, and the split-brain it guards against is already prevented at boot by boundaries
  2+3. Matches the ACK'd plan (create-focused admission tests).
- JSON-config backends' `AGEND_INSTANCE_NAME` (claude/kiro/agy/opencode) + `GEMINI.md` provision-stamp
  are **not yet** identity-guarded — follow-up. The incident and RED targets are Codex/AGENTS (+grok,
  which shares the `.toml` stamp).

Excluded per task: `resume --last` behavior, unrelated workspace-cleanup refactors.

### DIFF
--- a/src/agent_ops.rs
+++ b/src/agent_ops.rs
@@ -1489,6 +1489,53 @@ mod tests {
     }
 
     #[test]
+    fn cleanup_refuses_unreadable_identity_tree() {
+        // Fail-closed: an UNREADABLE identity artifact (opaque I/O ≠ NotFound)
+        // must refuse the delete — never be read as "absent" and wipe the tree.
+        let home = tmp_home("cw_unreadable");
+        let ws = home.join("workspace/alice");
+        std::fs::create_dir_all(&ws).unwrap();
+        std::fs::write(ws.join("AGENTS.md"), [0xFFu8, 0xFE]).unwrap(); // invalid UTF-8
+        assert!(
+            cleanup_working_dir(&home, "alice", &ws).is_some(),
+            "unreadable identity must refuse (Some verdict)"
+        );
+        assert!(ws.exists(), "tree preserved on unreadable identity");
+        std::fs::remove_dir_all(&home).ok();
+    }
+
+    #[test]
+    fn workspace_identity_lock_is_mutually_exclusive_provision_vs_delete() {
+        // Provision (generate_with_context) and delete (cleanup_working_dir) BOTH
+        // acquire store::acquire_workspace_identity_lock(home, wd) for the same
+        // directory. Prove it is mutually exclusive so a check+write can never
+        // interleave with a check+remove of that directory (root finding 4).
+        use std::sync::atomic::{AtomicBool, Ordering};
+        use std::sync::Arc;
+        let home = tmp_home("wsid_lock");
+        let wd = home.join("workspace/shared");
+        let in_critical = Arc::new(AtomicBool::new(false));
+        let held = crate::store::acquire_workspace_identity_lock(&home, &wd).expect("first lock");
+        in_critical.store(true, Ordering::SeqCst);
+        let (h2, w2, ic2) = (home.clone(), wd.clone(), in_critical.clone());
+        let t = std::thread::spawn(move || {
+            // Blocks until the main thread releases `held` (mutual exclusion).
+            let _g = crate::store::acquire_workspace_identity_lock(&h2, &w2).expect("second lock");
+            assert!(
+                !ic2.load(Ordering::SeqCst),
+                "acquired the workspace-identity lock while another holder was still in its \
+                 critical section — the lock is NOT mutually exclusive"
+            );
+        });
+        // Give the spawned thread time to reach (and block on) the acquire.
+        std::thread::sleep(std::time::Duration::from_millis(100));
+        in_critical.store(false, Ordering::SeqCst);
+        drop(held);
+        t.join().expect("second acquirer thread");
+        std::fs::remove_dir_all(&home).ok();
+    }
+
+    #[test]
     fn cleanup_metadata() {
         let home = tmp_home("cms");
         let ws = home.join("workspace/a");
--- a/src/agent_ops.rs
+++ b/src/agent_ops.rs
@@ -502,91 +502,108 @@ pub fn ensure_not_protected_json(branch: &str) -> Result<(), serde_json::Value> 
 /// is missing the 5 Kiro paths: `.kiro/agents/{agend.json,agend-prompt.md,
 /// default.json}`, `.kiro/prompts/agend.md`, `.kiro/settings.json`.
 pub fn cleanup_working_dir(home: &Path, name: &str, working_dir: &Path) {
-    let workspaces = crate::paths::workspace_dir(home);
-
-    // If under $AGEND_HOME/workspace/, remove the whole directory.
-    // CR-2026-06-14 (security): a purely LEXICAL `starts_with` lets a symlink
-    // under workspace/ whose real target is ELSEWHERE take this whole-dir
-    // `remove_dir_all` and follow the symlink out of the workspace, destroying
-    // real user data. Require the path to ALSO resolve canonically inside the
-    // canonicalized workspace root (canonicalize BOTH so a symlinked
-    // $AGEND_HOME — e.g. macOS /tmp→/private/tmp — still matches).
-    let under_workspace = working_dir.starts_with(&workspaces)
-        && match (
-            dunce::canonicalize(working_dir),
-            dunce::canonicalize(&workspaces),
-        ) {
-            (Ok(wd), Ok(ws)) => wd.starts_with(&ws),
-            _ => false,
-        };
-    if under_workspace {
-        // #2234 Phase 0: under cure-(B) the workspace dir IS a daemon-managed
-        // canonical worktree (its `.git` is a gitlink FILE). A bare
-        // remove_dir_all would destroy uncommitted/unpushed work AND orphan the
-        // worktree registration in the canonical repo. Route a worktree through
-        // `git worktree remove --force` (work-at-risk backed up first). A
-        // standalone clone / plain dir (the pre-(B) state) returns false here →
-        // the byte-identical remove_dir_all below still runs.
-        if crate::worktree_pool::teardown_workspace_worktree(home, name, working_dir) {
-            // handled (gitlink worktree): removal + registry cleanup done.
-        } else if let Err(e) = std::fs::remove_dir_all(working_dir) {
-            tracing::debug!(dir = %working_dir.display(), error = %e, "cleanup: remove workspace");
-        } else {
-            tracing::info!(dir = %working_dir.display(), "removed workspace");
-        }
+    // Workspace-identity guard (fail-closed): before removing anything under
+    // `working_dir`, refuse if the directory's on-disk identity belongs to a
+    // DIFFERENT instance (or is corrupt). Deleting instance A must never wipe a
+    // directory that identity artifacts (AGENTS.md block / `.codex` stamp) say
+    // belongs to instance B — preserve the tree and emit a loud audit. Metadata
+    // keyed by A's own name (the tail below) is still cleaned; only the shared
+    // working directory is preserved. Callers that report teardown status probe
+    // `working_dir_ownership_conflict` directly (this stays infallible for the
+    // ~14 fire-and-forget call sites).
+    let conflict = working_dir_ownership_conflict(working_dir, name);
+    if let Some(reason) = &conflict {
+        tracing::error!(
+            dir = %working_dir.display(), name, %reason,
+            "cleanup refused: working directory identity belongs to a different instance — tree preserved"
+        );
     } else {
-        // User-provided working directory: only remove agend-generated files
-        let agend_files = [
-            // Claude
-            ".claude/settings.local.json",
-            "mcp-config.json",
-            "claude-settings.json",
-            "statusline.sh",
-            "statusline.json",
-            ".claude/rules/agend.md",
-            // Gemini
-            ".gemini/settings.json",
-            // OpenCode
-            "opencode.json",
-            "instructions/agend.md",
-            // Codex
-            ".codex/config.toml",
-            "AGENTS.md",
-            // Kiro
-            ".kiro/settings/mcp.json",
-            ".kiro/settings/agend-mcp-wrapper.sh",
-            ".kiro/steering/agend.md",
-            ".kiro/agents/agend.json",
-            ".kiro/agents/agend-prompt.md",
-            ".kiro/agents/default.json",
-            ".kiro/prompts/agend.md",
-            ".kiro/settings.json",
-        ];
-        for file in &agend_files {
-            let path = working_dir.join(file);
-            if path.exists() {
-                let _ = std::fs::remove_file(&path);
-            }
-        }
+        let workspaces = crate::paths::workspace_dir(home);
 
-        // Clean up worktree if exists
-        let wt_dir = working_dir.join(".worktrees").join(name);
-        if wt_dir.exists() {
-            // W1.2: LOCAL best-effort worktree-remove via the bypass+bounded
-            // helper (was a raw UNBOUNDED `.output()` whose result was already
-            // discarded). git_ok adds the LOCAL_GIT_TIMEOUT bound so a stuck
-            // remove can't hang teardown; the bypass env is a no-op in the
-            // daemon's shim-free PATH. Result stays discarded → same effect.
-            let _ = crate::git_helpers::git_ok(
-                working_dir,
-                &[
-                    "worktree",
-                    "remove",
-                    "--force",
-                    &wt_dir.display().to_string(),
-                ],
-            );
-            tracing::info!(dir = %wt_dir.display(), "removed worktree");
+        // If under $AGEND_HOME/workspace/, remove the whole directory.
+        // CR-2026-06-14 (security): a purely LEXICAL `starts_with` lets a symlink
+        // under workspace/ whose real target is ELSEWHERE take this whole-dir
+        // `remove_dir_all` and follow the symlink out of the workspace, destroying
+        // real user data. Require the path to ALSO resolve canonically inside the
+        // canonicalized workspace root (canonicalize BOTH so a symlinked
+        // $AGEND_HOME — e.g. macOS /tmp→/private/tmp — still matches).
+        let under_workspace = working_dir.starts_with(&workspaces)
+            && match (
+                dunce::canonicalize(working_dir),
+                dunce::canonicalize(&workspaces),
+            ) {
+                (Ok(wd), Ok(ws)) => wd.starts_with(&ws),
+                _ => false,
+            };
+        if under_workspace {
+            // #2234 Phase 0: under cure-(B) the workspace dir IS a daemon-managed
+            // canonical worktree (its `.git` is a gitlink FILE). A bare
+            // remove_dir_all would destroy uncommitted/unpushed work AND orphan the
+            // worktree registration in the canonical repo. Route a worktree through
+            // `git worktree remove --force` (work-at-risk backed up first). A
+            // standalone clone / plain dir (the pre-(B) state) returns false here →
+            // the byte-identical remove_dir_all below still runs.
+            if crate::worktree_pool::teardown_workspace_worktree(home, name, working_dir) {
+                // handled (gitlink worktree): removal + registry cleanup done.
+            } else if let Err(e) = std::fs::remove_dir_all(working_dir) {
+                tracing::debug!(dir = %working_dir.display(), error = %e, "cleanup: remove workspace");
+            } else {
+                tracing::info!(dir = %working_dir.display(), "removed workspace");
+            }
+        } else {
+            // User-provided working directory: only remove agend-generated files
+            let agend_files = [
+                // Claude
+                ".claude/settings.local.json",
+                "mcp-config.json",
+                "claude-settings.json",
+                "statusline.sh",
+                "statusline.json",
+                ".claude/rules/agend.md",
+                // Gemini
+                ".gemini/settings.json",
+                // OpenCode
+                "opencode.json",
+                "instructions/agend.md",
+                // Codex
+                ".codex/config.toml",
+                "AGENTS.md",
+                // Kiro
+                ".kiro/settings/mcp.json",
+                ".kiro/settings/agend-mcp-wrapper.sh",
+                ".kiro/steering/agend.md",
+                ".kiro/agents/agend.json",
+                ".kiro/agents/agend-prompt.md",
+                ".kiro/agents/default.json",
+                ".kiro/prompts/agend.md",
+                ".kiro/settings.json",
+            ];
+            for file in &agend_files {
+                let path = working_dir.join(file);
+                if path.exists() {
+                    let _ = std::fs::remove_file(&path);
+                }
+            }
+
+            // Clean up worktree if exists
+            let wt_dir = working_dir.join(".worktrees").join(name);
+            if wt_dir.exists() {
+                // W1.2: LOCAL best-effort worktree-remove via the bypass+bounded
+                // helper (was a raw UNBOUNDED `.output()` whose result was already
+                // discarded). git_ok adds the LOCAL_GIT_TIMEOUT bound so a stuck
+                // remove can't hang teardown; the bypass env is a no-op in the
+                // daemon's shim-free PATH. Result stays discarded → same effect.
+                let _ = crate::git_helpers::git_ok(
+                    working_dir,
+                    &[
+                        "worktree",
+                        "remove",
+                        "--force",
+                        &wt_dir.display().to_string(),
+                    ],
+                );
+                tracing::info!(dir = %wt_dir.display(), "removed worktree");
+            }
         }
     }
 
@@ -612,6 +629,29 @@ pub fn cleanup_working_dir(home: &Path, name: &str, working_dir: &Path) {
     // working_dir, so it lives outside both cleanup branches above. Never
     // touches the real workspace — only the managed symlink/junction.
     crate::agy_workspace::remove_link(home, name);
+}
+
+/// Whether `working_dir`'s on-disk identity artifacts name an instance OTHER
+/// than `name` (or are corrupt) — in which case the caller must NOT remove the
+/// tree. Returns `Some(reason)` to refuse (foreign owner / corrupt artifact),
+/// `None` to proceed (no identity artifact, or the directory belongs to `name`).
+/// Checks the AGENTS.md agend block (which records the SANITIZED identifier) and
+/// the `.codex/config.toml` `AGEND_INSTANCE_NAME` stamp (which records the RAW
+/// name) — the two durable identity artifacts the collision incident involved.
+pub(crate) fn working_dir_ownership_conflict(working_dir: &Path, name: &str) -> Option<String> {
+    if let Ok(s) = std::fs::read_to_string(working_dir.join("AGENTS.md")) {
+        if let Some(reason) = crate::instructions::agend_block_owner(&s)
+            .conflict_with(&crate::instructions::sanitize_identifier(name))
+        {
+            return Some(format!("AGENTS.md {reason}"));
+        }
+    }
+    if let Ok(s) = std::fs::read_to_string(working_dir.join(".codex").join("config.toml")) {
+        if let Some(reason) = crate::mcp_config::codex_config_owner(&s).conflict_with(name) {
+            return Some(format!(".codex/config.toml {reason}"));
+        }
+    }
+    None
 }
 
 // ---------------------------------------------------------------------------
@@ -1350,6 +1390,81 @@ mod tests {
         assert!(!ud.join("opencode.json").exists());
         std::fs::remove_dir_all(&home).ok();
         std::fs::remove_dir_all(&ud).ok();
+    }
+
+    // --- workspace-identity delete guard (boundary 3) ---
+
+    fn seed_agents_owned_by(dir: &Path, owner: &str) {
+        std::fs::write(
+            dir.join("AGENTS.md"),
+            format!(
+                "<!-- agend:start -->\n## Identity\n\n- **Name**: `{owner}`\n<!-- agend:end -->\n"
+            ),
+        )
+        .unwrap();
+    }
+
+    #[test]
+    fn cleanup_preserves_foreign_identity_tree() {
+        let home = tmp_home("cw_foreign");
+        let ws = home.join("workspace/alice"); // "alice" is being deleted...
+        std::fs::create_dir_all(&ws).unwrap();
+        seed_agents_owned_by(&ws, "bob"); // ...but the directory belongs to "bob".
+        std::fs::write(ws.join("keep.txt"), "b").unwrap();
+        cleanup_working_dir(&home, "alice", &ws);
+        assert!(ws.exists(), "foreign-owned tree must be preserved");
+        assert!(
+            ws.join("AGENTS.md").exists(),
+            "bob's identity file preserved"
+        );
+        assert!(
+            ws.join("keep.txt").exists(),
+            "foreign tree contents preserved"
+        );
+        std::fs::remove_dir_all(&home).ok();
+    }
+
+    #[test]
+    fn cleanup_removes_same_identity_tree() {
+        let home = tmp_home("cw_same");
+        let ws = home.join("workspace/alice");
+        std::fs::create_dir_all(&ws).unwrap();
+        seed_agents_owned_by(&ws, "alice"); // dir belongs to the instance being deleted
+        cleanup_working_dir(&home, "alice", &ws);
+        assert!(!ws.exists(), "same-identity tree must be removed");
+        std::fs::remove_dir_all(&home).ok();
+    }
+
+    #[test]
+    fn cleanup_removes_unowned_tree_normally() {
+        let home = tmp_home("cw_absent");
+        let ws = home.join("workspace/alice");
+        std::fs::create_dir_all(&ws).unwrap();
+        std::fs::write(ws.join("f.txt"), "x").unwrap(); // no identity artifact
+        cleanup_working_dir(&home, "alice", &ws);
+        assert!(!ws.exists(), "unowned tree cleans normally");
+        std::fs::remove_dir_all(&home).ok();
+    }
+
+    #[test]
+    fn ownership_conflict_detects_foreign_codex_stamp() {
+        let home = tmp_home("wdoc_codex");
+        let ws = home.join("workspace/alice");
+        std::fs::create_dir_all(ws.join(".codex")).unwrap();
+        std::fs::write(
+            ws.join(".codex").join("config.toml"),
+            "AGEND_INSTANCE_NAME = 'bob'\n",
+        )
+        .unwrap();
+        assert!(
+            working_dir_ownership_conflict(&ws, "alice").is_some(),
+            "foreign .codex stamp is a conflict"
+        );
+        assert!(
+            working_dir_ownership_conflict(&ws, "bob").is_none(),
+            "same owner is not a conflict"
+        );
+        std::fs::remove_dir_all(&home).ok();
     }
 
     #[test]
--- a/src/agent_ops.rs
+++ b/src/agent_ops.rs
@@ -559,58 +559,77 @@ pub fn cleanup_working_dir(home: &Path, name: &str, working_dir: &Path) -> Optio
                 tracing::info!(dir = %working_dir.display(), "removed workspace");
             }
         } else {
-            // User-provided working directory: only remove agend-generated files
-            let agend_files = [
-                // Claude
-                ".claude/settings.local.json",
-                "mcp-config.json",
-                "claude-settings.json",
-                "statusline.sh",
-                "statusline.json",
-                ".claude/rules/agend.md",
-                // Gemini
-                ".gemini/settings.json",
-                // OpenCode
-                "opencode.json",
-                "instructions/agend.md",
-                // Codex
-                ".codex/config.toml",
-                "AGENTS.md",
-                // Kiro
-                ".kiro/settings/mcp.json",
-                ".kiro/settings/agend-mcp-wrapper.sh",
-                ".kiro/steering/agend.md",
-                ".kiro/agents/agend.json",
-                ".kiro/agents/agend-prompt.md",
-                ".kiro/agents/default.json",
-                ".kiro/prompts/agend.md",
-                ".kiro/settings.json",
-            ];
-            for file in &agend_files {
-                let path = working_dir.join(file);
-                if path.exists() {
-                    let _ = std::fs::remove_file(&path);
+            let worktrees = home.join("worktrees");
+            let under_worktrees = working_dir.starts_with(&worktrees)
+                && match (
+                    dunce::canonicalize(working_dir),
+                    dunce::canonicalize(&worktrees),
+                ) {
+                    (Ok(wd), Ok(wt)) => wd.starts_with(&wt),
+                    _ => false,
+                };
+            if under_worktrees {
+                if crate::worktree_pool::teardown_workspace_worktree(home, name, working_dir) {
+                    // handled (gitlink worktree): removal + registry cleanup done.
+                } else if let Err(e) = std::fs::remove_dir_all(working_dir) {
+                    tracing::debug!(dir = %working_dir.display(), error = %e, "cleanup: remove managed worktree");
+                } else {
+                    tracing::info!(dir = %working_dir.display(), "removed managed worktree");
                 }
-            }
+            } else {
+                // User-provided working directory: only remove agend-generated files
+                let agend_files = [
+                    // Claude
+                    ".claude/settings.local.json",
+                    "mcp-config.json",
+                    "claude-settings.json",
+                    "statusline.sh",
+                    "statusline.json",
+                    ".claude/rules/agend.md",
+                    // Gemini
+                    ".gemini/settings.json",
+                    // OpenCode
+                    "opencode.json",
+                    "instructions/agend.md",
+                    // Codex
+                    ".codex/config.toml",
+                    "AGENTS.md",
+                    // Kiro
+                    ".kiro/settings/mcp.json",
+                    ".kiro/settings/agend-mcp-wrapper.sh",
+                    ".kiro/steering/agend.md",
+                    ".kiro/agents/agend.json",
+                    ".kiro/agents/agend-prompt.md",
+                    ".kiro/agents/default.json",
+                    ".kiro/prompts/agend.md",
+                    ".kiro/settings.json",
+                ];
+                for file in &agend_files {
+                    let path = working_dir.join(file);
+                    if path.exists() {
+                        let _ = std::fs::remove_file(&path);
+                    }
+                }
 
-            // Clean up worktree if exists
-            let wt_dir = working_dir.join(".worktrees").join(name);
-            if wt_dir.exists() {
-                // W1.2: LOCAL best-effort worktree-remove via the bypass+bounded
-                // helper (was a raw UNBOUNDED `.output()` whose result was already
-                // discarded). git_ok adds the LOCAL_GIT_TIMEOUT bound so a stuck
-                // remove can't hang teardown; the bypass env is a no-op in the
-                // daemon's shim-free PATH. Result stays discarded → same effect.
-                let _ = crate::git_helpers::git_ok(
-                    working_dir,
-                    &[
-                        "worktree",
-                        "remove",
-                        "--force",
-                        &wt_dir.display().to_string(),
-                    ],
-                );
-                tracing::info!(dir = %wt_dir.display(), "removed worktree");
+                // Clean up worktree if exists
+                let wt_dir = working_dir.join(".worktrees").join(name);
+                if wt_dir.exists() {
+                    // W1.2: LOCAL best-effort worktree-remove via the bypass+bounded
+                    // helper (was a raw UNBOUNDED `.output()` whose result was already
+                    // discarded). git_ok adds the LOCAL_GIT_TIMEOUT bound so a stuck
+                    // remove can't hang teardown; the bypass env is a no-op in the
+                    // daemon's shim-free PATH. Result stays discarded → same effect.
+                    let _ = crate::git_helpers::git_ok(
+                        working_dir,
+                        &[
+                            "worktree",
+                            "remove",
+                            "--force",
+                            &wt_dir.display().to_string(),
+                        ],
+                    );
+                    tracing::info!(dir = %wt_dir.display(), "removed worktree");
+                }
             }
         }
     }
@@ -658,10 +677,22 @@ pub(crate) fn working_dir_ownership_conflict(working_dir: &Path, name: &str) -> 
         return Some(format!("AGENTS.md {reason}"));
     }
     if let Some(reason) =
+        crate::instructions::agents_md_identity(&working_dir.join(".agents").join("AGENTS.md"))
+            .conflict_with(&crate::instructions::sanitize_identifier(name))
+    {
+        return Some(format!(".agents/AGENTS.md {reason}"));
+    }
+    if let Some(reason) =
         crate::mcp_config::codex_config_identity(&working_dir.join(".codex").join("config.toml"))
             .conflict_with(name)
     {
         return Some(format!(".codex/config.toml {reason}"));
+    }
+    if let Some(reason) =
+        crate::mcp_config::codex_config_identity(&working_dir.join(".grok").join("config.toml"))
+            .conflict_with(name)
+    {
+        return Some(format!(".grok/config.toml {reason}"));
     }
     None
 }
--- a/src/agent_ops.rs
+++ b/src/agent_ops.rs
@@ -694,6 +694,14 @@ pub(crate) fn working_dir_ownership_conflict(working_dir: &Path, name: &str) -> 
     {
         return Some(format!(".grok/config.toml {reason}"));
     }
+    for artifact in &[".claude/agend.md", ".kiro/steering/agend.md"] {
+        let path = working_dir.join(artifact);
+        if let Some(reason) = crate::instructions::nonshared_instructions_identity(&path)
+            .conflict_with(&crate::instructions::sanitize_identifier(name))
+        {
+            return Some(format!("{artifact} {reason}"));
+        }
+    }
     None
 }
 
--- a/src/agent_ops.rs
+++ b/src/agent_ops.rs
@@ -501,17 +501,25 @@ pub fn ensure_not_protected_json(branch: &str) -> Result<(), serde_json::Value> 
 /// The copy in `mcp/handlers.rs` drifted to 14 entries on 2026-04-14 and
 /// is missing the 5 Kiro paths: `.kiro/agents/{agend.json,agend-prompt.md,
 /// default.json}`, `.kiro/prompts/agend.md`, `.kiro/settings.json`.
-pub fn cleanup_working_dir(home: &Path, name: &str, working_dir: &Path) {
+pub fn cleanup_working_dir(home: &Path, name: &str, working_dir: &Path) -> Option<String> {
     // Workspace-identity guard (fail-closed): before removing anything under
     // `working_dir`, refuse if the directory's on-disk identity belongs to a
-    // DIFFERENT instance (or is corrupt). Deleting instance A must never wipe a
-    // directory that identity artifacts (AGENTS.md block / `.codex` stamp) say
-    // belongs to instance B — preserve the tree and emit a loud audit. Metadata
-    // keyed by A's own name (the tail below) is still cleaned; only the shared
-    // working directory is preserved. Callers that report teardown status probe
-    // `working_dir_ownership_conflict` directly (this stays infallible for the
-    // ~14 fire-and-forget call sites).
-    let conflict = working_dir_ownership_conflict(working_dir, name);
+    // DIFFERENT instance (or is corrupt/unreadable). Deleting instance A must
+    // never wipe a directory that identity artifacts (AGENTS.md block / `.codex`
+    // stamp) say belongs to instance B — preserve the tree and emit a loud audit.
+    // Metadata keyed by A's own name (the tail below) is still cleaned; only the
+    // shared working directory is preserved.
+    //
+    // Held under the workspace-identity lock so the ownership CHECK and the
+    // REMOVAL are atomic against a concurrent provision/delete of the same
+    // directory. The SINGLE returned verdict is what `full_delete_instance`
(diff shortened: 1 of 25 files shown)


suzuke:
REJECTED — exact-head root review for 63bfa124f84ba66f05e105729c095cb30a866661

scope_source: d-20260714004402440378-0 and task t-20260714004429204771-46776-0
audit_mode: exact-head production-path review after full PR CI green
reviewed_head: 63bfa124f84ba66f05e105729c095cb30a866661
files: src/agent_ops.rs, src/fleet/persist.rs, src/instructions.rs, src/mcp/handlers/instance_state/lifecycle.rs, src/mcp_config.rs, src/paths.rs

The collision detector itself is useful, but the production contract is still fail-open. Please make one consolidated patchset covering all findings before the next review epoch.

1. Provisioning refusals do not abort spawn. mcp_config::configure logs and discards every backend error, while generate_with_context logs and discards instruction ownership errors. A foreign workspace can therefore be detected but the agent still starts against partially or incorrectly provisioned state.

2. Identity admission occurs after mutations. generate_with_context calls ensure_project_root before either identity guard. The create path can also create a worktree and directory before add_instance_to_yaml performs collision admission. Refusal can leave partial filesystem and git state.

3. Artifact I/O is not fail-closed. AGENTS.md and config.toml read errors are collapsed to empty or absent; instruction directory creation and final write errors are discarded. Permission errors, invalid UTF-8, and other I/O failures can therefore permit overwrite, deletion, or false success. Only NotFound may mean absent.

4. The safety decision is not atomic across provision and delete. Config has a file-specific lock, AGENTS.md has no corresponding shared workspace identity lock, and cleanup reads the artifacts independently. Provision and delete can race between check and write or check and removal.

5. Pre-existing duplicate fleet entries are not rejected. add_instances_to_yaml runs the workspace collision scan only for a new key; an existing-key merge takes a separate path and FleetConfig load or spawn admission has no whole-registry uniqueness validation. A corrupt legacy fleet can survive until a later artifact-level refusal.

6. Delete probes ownership twice. lifecycle first calls working_dir_ownership_conflict and then cleanup_working_dir calls it again. With no shared lock, the reported decision and destructive decision may observe different states.

Required next patch:
- propagate Result through the real create, boot, pane, and restart callers and abort before spawn;
- perform collision and artifact preflight before every filesystem or worktree mutation;
- use a typed NotFound versus Opaque I/O state and refuse Opaque;
- hold one workspace identity lock across read-decision-write or delete;
- validate whole-registry uniqueness at load or spawn admission;
- return one locked cleanup decision instead of probing twice;
- add RED tests through the real create and delete entry points, including unreadable artifacts, concurrent provision versus delete, and pre-existing duplicate fleet.yaml.

### Evidence
ran: gh pr checks 2777 --repo suzuke/agend-terminal → all eight required checks passed on this head
cited: src/mcp_config.rs:956-973 — configure returns unit and only warns on Err
cited: src/instructions.rs:564-588 — project mutation precedes guards and guard errors are logged, not returned
cited: src/instructions.rs:502-504,523-554 — mkdir, read, and write errors are discarded or defaulted
cited: src/agent_ops.rs:641-654 — both read failures become no conflict
cited: src/fleet/persist.rs:62-91 — collision validation exists only in the new-key branch
cited: src/mcp/handlers/instance_state/spawn.rs:116-121,174-201 — worktree and directory creation precede fleet admission
cited: src/mcp/handlers/instance_state/lifecycle.rs:143-153 — ownership is probed once, then cleanup probes again

---
*codex-125550* · codex

suzuke:
REJECTED — exact-head secondary review of PR #2777

reviewed_head: 3d442d0ad8c2a7bf4577a512803579d45614bc94
assignment_id: be114b4d-caa9-4d6f-b6fb-7b3cee303541
review_class: Dual / secondary

The six prior blockers were rechecked. Contextful provisioning now propagates several errors, the workspace lock and single cleanup verdict are real improvements, and add/boot have partial duplicate guards. The production contract is still fail-open. Root challenge A-D are all confirmed at this head.

1. Managed pane and connect paths erase foreign identity before spawn. `spawn_and_subscribe` first calls contextless `generate()` and then `agent::spawn_agent`; the fleet-aware `generate_with_context()` is only called after the child exists (both async and sync pane paths). Because the AGENTS/config ownership checks and identity stamps are conditional on `Some(ctx/name)`, that first pass can strip a foreign managed block/stamp and rewrite it as unowned; the later pass then has nothing foreign left to reject. `connect::run` already has `name` but also calls contextless `generate()` before spawning. `prepare_instructions` repeats the fail-open fallback when fleet loading fails and no explicit role was supplied.

2. Create-race rollback does not remove the worktree it just created. Branch creates live under `$AGEND_HOME/worktrees/<agent>/<branch>`, then a racing `add_instance_to_yaml` refusal calls `cleanup_working_dir`. That cleanup only whole-removes paths under `$AGEND_HOME/workspace`; the external worktree is classified as a user directory, so neither `git worktree remove` nor branch cleanup runs. The claimed no-partial-git-state rollback is therefore false (and the code does not distinguish a newly-created worktree from a reused one).

3. Pre-mutation and opaque-I/O refusal remain incomplete. `prepare_instructions` creates the directory before artifact preflight; boot resolution creates a directory/worktree before `generate_with_context`. `ensure_project_root` discards git-init and `.gitignore` failures, Claude migration discards removal failures, Codex/Grok writers still convert any locked `read_to_string` error to empty, and fleet mutation converts any fleet.yaml read error to default content. These paths can mutate or continue to spawn after an opaque provisioning failure.

4. Delete's single verdict does not inspect every durable identity location. `working_dir_ownership_conflict` checks only root `AGENTS.md` and `.codex/config.toml`, while Agy's shared identity is `.agents/AGENTS.md`. Deleting instance A whose recorded directory is B's Agy default workspace therefore sees no conflict and takes the whole-directory removal path, destroying B's workspace. The double-probe bug is fixed, but the one remaining probe is incomplete.

5. Whole-registry uniqueness is not enforced at every spawn admission. `add_instances_to_yaml` validates only the affected `entries`, and `duplicate_identity_owner_before` is called only by daemon boot. Runtime `handle_spawn` has no equivalent guard. For backends such as Claude, whose private instruction file is outside `workspace_provision_preflight`, two legacy/hand-edited fleet entries sharing one directory can both be started dynamically.

6. The new tests do not exercise the required real entry points. The lock test acquires the lock helper directly twice and uses `sleep(100ms)`; it never drives provision versus delete, pane/connect, create rollback, or runtime duplicate spawn. No test-only RED commit precedes implementation; the target tests are absent at the base and arrive in the same implementation commits. This does not satisfy the real-entry or deterministic RED→GREEN review protocol.

### Evidence

ran: `git rev-parse HEAD && git branch --show-current && git status --short` → exact frozen head, review branch, clean worktree
ran: focused `cargo test --bin agend-terminal` groups (`identity`, `configure_codex`, `generate_agent_instructions_refuses`, `add_instance_`) → GREEN 3/3 runs; these passing helper tests do not cover the production gaps above
ran: `cargo clippy --lib --bin agend-terminal --bin agend-git --bin agend-mcp-bridge --tests --examples --features tray -- -D warnings` → pass
ran: `scripts/fmt-owned.sh --check` → pass
ran: `gh pr checks 2777 --repo suzuke/agend-terminal` → all 9 checks pass at the reviewed head, including Ubuntu/macOS/Windows; runtime-validated by PR-CI run 29303426763
ran: base-tree grep for the new identity/lock tests plus commit-history inspection → tests absent at base; no standalone failing-test commit before implementation
cited: `src/app/pane_factory.rs:276-280,335-356,799-825,1084-1105`; `src/connect.rs:76-80,128-139`; `src/api/handlers/mod.rs:44-50,90-100` — contextless mutation and post-spawn/fail-open ordering
cited: `src/instructions.rs:551-563,603-627`; `src/mcp_config.rs:750-779,896-907`; `src/fleet/persist.rs:28-33,143-165` — conditional guards and opaque reads/defaults
cited: `src/mcp/handlers/instance_state/spawn.rs:100-142,221-228`; `src/worktree.rs:19-28`; `src/agent_ops.rs:529-566` — external worktree creation followed by a cleanup path that cannot remove it
cited: `src/bootstrap/agent_resolve.rs:55-70,110-124`; `src/instructions.rs:8-13,63-86` — mutations and discarded failures precede/escape provisioning refusal
cited: `src/agent_ops.rs:644-667`; `src/backend.rs:629-635` — delete ignores Agy's actual shared identity path
cited: `src/daemon/mod.rs:1714-1729`; `src/api/handlers/instance.rs:282-336`; `src/instructions.rs:645-657` — legacy-duplicate gate is boot-only, not runtime-spawn admission
cited: `src/agent_ops.rs:1508-1535` — timing-based helper-only lock test

---
*codex-3160c0* · codex/GPT-5

suzuke:
REJECTED — exact-head root re-review for 3d442d0ad8c2a7bf4577a512803579d45614bc94

scope_source: d-20260714004402440378-0 and task t-20260714004429204771-46776-0
audit_mode: exact-head cumulative r2 production-path review after terminal-green branch CI
reviewed_head: 3d442d0ad8c2a7bf4577a512803579d45614bc94
files: src/app/pane_factory.rs, src/api/handlers/mod.rs, src/instructions.rs, src/mcp/handlers/instance_state/spawn.rs, src/agent_ops.rs, src/worktree.rs, src/fleet/persist.rs

The new typed artifact states, shared lock, whole-registry post-pass, and single cleanup verdict are useful. However, the r2 patch still leaves four blocking holes. Do not push piecemeal fixes; wait for the independent reviewer and then submit one consolidated next epoch.

1. App-mode still performs the identity-aware preflight after process creation. spawn_and_subscribe calls contextless instructions::generate, whose ctx=None path skips workspace_provision_preflight, then calls agent::spawn_agent and subscribes. Only after that succeeds do run_attach and create_pane_from_resolved call generate_with_context(...Some(ctx)). A foreign workspace can therefore be mutated and an agent process registered before the refusal; the later Failed/Err does not undo that process/registry state. The API path has the same fail-open shape when fleet.yaml load fails without an explicit role: it falls back to contextless generate for a named managed instance.

2. Atomic admission rollback does not roll back branch worktrees. The create handler builds $AGEND_HOME/worktrees/<agent>/<branch> before the atomic add_instance_to_yaml; on a racing refusal it calls cleanup_working_dir for that path. But cleanup_working_dir only tears down a whole git worktree when the path is under $AGEND_HOME/workspace/. An external branch-worktree path takes the user-provided-directory branch, which removes generated files and probes <actual-path>/.worktrees/<name>; it does not remove the actual worktree registration/path or the branch. The claimed no-partial-git-state contract is therefore false under the very race this rollback is meant to close.

3. Provisioning I/O is still not fail-closed. generate_with_context says every mutation propagates its I/O error, but ensure_project_root returns unit, treats failed git init as a no-op, and discards .gitignore write errors. migrate_claude_old_rules_file also discards removal errors. Codex/Gemini may then escape project-root scoping and Claude may load the stale rule while the caller receives Ok(()).

4. The required empirical race coverage is absent. The added lock test directly tests the lock primitive, uses a fixed sleep(100ms), and does not prove the contender reached the acquire before the first holder releases; it can pass without exercising provision or delete. The other new tests call helpers such as generate_agent_instructions and add_instance_to_yaml, not the real create/pane/delete entry points required by the r1 finding and fleet protocol. The r2 history contains only two combined WIP implementation commits, not a failing real-entry RED commit before GREEN.

r1 disposition:
- #1 propagation/abort: partial, REJECTED by blocker 1
- #2 pre-mutation admission/rollback: partial, REJECTED by blocker 2
- #3 typed I/O/fail-closed: partial, REJECTED by blocker 3
- #4 atomic provision/delete: production lock improved; empirical proof REJECTED by blocker 4
- #5 whole-registry uniqueness: fixed in inspected paths
- #6 single locked cleanup verdict: fixed in inspected paths

### Evidence
ran: cargo test workspace_identity_lock_is_mutually_exclusive_provision_vs_delete -- --nocapture && cargo test merge_adding_colliding_working_directory_is_refused_whole_registry -- --nocapture && cargo test generate_agent_instructions_refuses_unreadable_agents_md_byte_preserving -- --nocapture → exit 0; each selected helper-level test passed, but none exercises the failing app/create/delete production sequence
ran: git log --format='%h %s' --reverse 63bfa124f84ba66f05e105729c095cb30a866661..3d442d0ad8c2a7bf4577a512803579d45614bc94 → only 9449a042 and 3d442d0a combined WIP implementation commits; no real-entry RED-before-GREEN commit
cited: src/instructions.rs:582-610 — contextless generate passes None, so the named identity preflight is skipped
cited: src/app/pane_factory.rs:276-280,331-367,816-824,1102-1105 — contextless provisioning and process spawn precede the named preflight
cited: src/api/handlers/mod.rs:62-100 — failed fleet load can fall back to contextless provisioning for a named instance
cited: src/mcp/handlers/instance_state/spawn.rs:106-139,221-228 — worktree creation precedes atomic admission and cleanup_working_dir is the rollback
cited: src/worktree.rs:19-28 — branch worktrees live under $AGEND_HOME/worktrees, outside $AGEND_HOME/workspace
cited: src/agent_ops.rs:529-615 — only workspace-root paths receive whole worktree teardown; the external-path branch does not remove the actual worktree
cited: src/instructions.rs:8-12,63-85,613-626 — stale-rule removal, git init, and .gitignore write errors are discarded beneath a Result-returning caller
cited: src/agent_ops.rs:1508-1534 — concurrency test is primitive-only and relies on fixed sleep
cited: src/instructions.rs:717 and src/fleet/persist.rs:724,758 — new regression tests enter through helpers, not real production entry points

---
*codex-125550* · codex

suzuke:
REJECTED — exact head 9ea0a6f69a13a55ebe00c95d38f508cc2d77b8c8

This is one consolidated review. The seven r3 tests pass, but three original fail-closed invariants remain unimplemented:

1. Reused worktree can be deleted by admission rollback.
worktree::create explicitly reuses an existing worktree (src/worktree.rs:57,126-208). If add_instance_to_yaml then fails, create_instance unconditionally calls cleanup_working_dir (src/mcp/handlers/instance_state/spawn.rs:221-228). The new G3 test covers deleting a newly-created plain directory, not preserving a pre-existing/reused worktree. Track provenance from create and only roll back state created by this attempt.

2. Runtime/boot collision checks still fail open when fleet authority cannot be read.
load_instances_mapping converts read, parse, and shape failures to an empty mapping (src/fleet/persist.rs:149-170), and both workspace_identity_collision and duplicate_identity_owner_before consume it. A malformed/unreadable fleet therefore means “no collision” and runtime spawn/boot continues. prepare_instructions also collapses every FleetConfig::load error to a contextless fallback (src/api/handlers/mod.rs:44-109), after creating the directory. Missing fleet may have a documented fallback; opaque I/O and parse failures must refuse before mutation.

3. Project-root and Claude migration I/O errors remain swallowed.
migrate_claude_old_rules_file ignores remove_file failure (src/instructions.rs:8-12). ensure_project_root ignores git init failure and .gitignore write failure (src/instructions.rs:63-86). generate_with_context calls both as infallible mutations (src/instructions.rs:613-626), so provisioning can report success after required ownership/scope setup failed.

Required next revision: add standalone RED coverage for (a) reused-worktree preservation on registration refusal, (b) unreadable/malformed fleet refusal for runtime and boot before filesystem mutation, and (c) surfaced project-root/.gitignore/Claude-migration failures; then one consolidated GREEN revision. Do not split these into separate fix/review rounds.

### Evidence
- ran: cargo nextest run --features tray --bin agend-terminal -E 'test(r3_46776)' → 7/7 passed at exact head 9ea0a6f69a13a55ebe00c95d38f508cc2d77b8c8
- ran: git diff --check 6a58483979b317e8db793085ce390b072ec93fcc..9ea0a6f69a13a55ebe00c95d38f508cc2d77b8c8 → clean
- cited: src/worktree.rs:57,126-208; src/mcp/handlers/instance_state/spawn.rs:221-228; src/fleet/persist.rs:149-170,173-187; src/api/handlers/mod.rs:44-109; src/instructions.rs:8-12,63-86,613-626.

suzuke:
VERIFIED — exact head 16b1ce94b1f18d6fc4799812c3b835c218ddf19c (independent r4 Fable review, task t-20260714060108010686-27634-16, review_class=dual; all evidence independently produced in isolated review branch review/pr-2777-r4-fable)

### Evidence
- ran: git rev-parse HEAD → 16b1ce94 == GitHub headRefOid; base 6a584839 (S1-merged main) exact.
- ran: RED validation — checked out RED commit 01dbc0f9 → cargo test r3_46776 → **6 FAILED** (r4_unreadable_fleet_refuses_boot_admission, r4_unreadable_fleet_refuses_runtime_collision, r4_admission_rollback_preserves_preexisting_worktree, r4_ensure_project_root_failure_propagates, r4_malformed_fleet_refuses_prepare_instructions, r4_migrate_claude_old_rules_failure_propagates); 7 r3 tests PASSED (pre-existing r3 GREEN correctness preserved). Restored to exact head → **13/13 PASSED** (all r3 + all r4 GREEN).
- cited: diff scope = 24 files, +2,119/−192 across 9 commits (3dcf1230 boundary-1 admission → a1b6cbe4 boundaries-2/3 provision+delete → b94e8cd5 platform fix → b063ec43 locked single-verdict delete → 48cad926 fail-closed WIP → 524a304f r3 RED → 9ea0a6f6 r3 GREEN → 01dbc0f9 r4 RED → 16b1ce94 r4 GREEN). Scope matches d-20260714004402440378-0 exactly: three boundaries (admission, provision, delete) plus the r4 consolidated blockers.
- ran: scripts/fmt-owned.sh --check → PASS; CI-exact strict clippy (incl. --bin agentic-git) -D warnings → PASS.
- CI at exact head (single snapshot): audit/MSRV/LOC/Signed-off pass; platform Checks/Coverage/Daemon-boot pending at review time — merge gate is the dispatcher's.

suzuke:
REJECTED — root exact-head review of `16b1ce94b1f18d6fc4799812c3b835c218ddf19c`.

### Evidence
ran: `git reset --hard 01dbc0f9 && cargo test --features tray --bin agend-terminal r4_ -- --nocapture` → the six intended R4 tests failed before production changes.
ran: `git reset --hard 16b1ce94b1f18d6fc4799812c3b835c218ddf19c && cargo test --features tray --bin agend-terminal r4_ -- --nocapture` → the six intended R4 tests passed (7 matching tests total passed; 1 unrelated perf test ignored).
ran: `git diff --check 9ea0a6f69a13a55ebe00c95d38f508cc2d77b8c8..16b1ce94b1f18d6fc4799812c3b835c218ddf19c` → clean.
cited: `src/mcp/handlers/instance_state/spawn.rs:121-142` — branch worktree creation occurs before `work_dir_preexisted` is sampled.
cited: `src/fleet/persist.rs:152-165` — `Path::exists()` and `as_mapping().unwrap_or_default()` still collapse opaque/invalid fleet states into “missing/empty”.
cited: `src/api/handlers/mod.rs:44-62,90-104` — the work directory is created before fleet validation, and the fallback again classifies errors with `exists()`.
cited: `src/instructions.rs:8-14` — Claude migration uses `exists()`, so metadata errors remain indistinguishable from absence.

### Blocking findings
1. **Attempt provenance is still inverted for branch worktrees.** A worktree newly created at lines 121–140 exists when sampled at line 142, so a later fleet-write failure treats it as pre-existing and skips rollback. The current RED covers only a caller-supplied pre-existing directory. Return a typed `CreatedByThisAttempt | Reused` receipt from worktree creation and add both cleanup/preservation controls.
2. **Missing vs opaque fleet state is still fail-open.** Never use `Path::exists()` to distinguish NotFound from permission/metadata/read failures. Match the actual read/metadata error kind: exact `NotFound` may fall back; every other error must refuse.
3. **Present-but-invalid fleet shape is accepted as empty.** `instances: []`, scalar, or other wrong types currently become an empty mapping. Missing `instances` may be valid; present-but-wrong-shape must error, with runtime and boot controls.
4. **Provisioning mutates before validation.** `prepare_instructions` creates `work_dir` before fleet read/parse/shape validation. Add a test where the directory is initially absent and malformed/opaque fleet input leaves it absent.
5. **Claude migration has the same opaque/absent bug.** Match `NotFound` on the removal/metadata operation; propagate other I/O failures.

The independent VERIFIED verdict proves the written RED→GREEN path, but it did not cover these missing controls or the production ordering/shape paths. Worst exact-head verdict therefore remains REJECTED.

Required next candidate: one standalone RED expansion, then one consolidated GREEN. No piecemeal candidates.

— `codex-125550`, root orchestrator

suzuke:
## VERIFIED @f9fa311b — Independent Fable review (dual secondary)

**audit_mode**: read-only | **reviewed_head**: `f9fa311b430d5015386d820449108e9afbc37e4f`

### Evidence

**1. DirIdentity fail-closed model** ✅
`paths.rs:82-93` — `DirIdentity` enum: Absent (adoptable), Owner(name), Corrupt, Unreadable. `conflict_with()` (paths.rs:102-112) treats Unreadable as conflict, never absent. `classify_identity_read` (paths.rs:119-128) maps NotFound→Absent, all other I/O errors→Unreadable.

**2. workspace_identity() canonical over-collision** ✅
`paths.rs:46-69` — peels non-existing tail, canonicalizes deepest existing ancestor (resolves symlinks), re-appends remainder, lowercases. Fail-closed: over-collides case/symlink aliases rather than miss a duplicate.

**3. Per-identity lock closes TOCTOU** ✅
`store.rs:323-332` — `acquire_workspace_identity_lock` keyed by FNV-1a of canonical identity. Both provision and delete acquire this lock across check+mutation. `agent_ops.rs:504` acquires lock before `working_dir_ownership_conflict`.

**4. Symlink-escape guard** ✅
`agent_ops.rs:538-545` — canonicalized wd/ws must both resolve inside workspace root, not just lexical starts_with. Prevents symlinked dir under workspace/ from causing remove_dir_all to follow out.

**5. WorktreeProvenance rollback** ✅
`worktree.rs:32-37` — CreatedByThisAttempt vs Reused. Rollback only deletes what this attempt created (r5 RED test pins this).

**6. Pre-mutation ordering** ✅
`instructions.rs:596-627` — acquires workspace-identity lock, runs preflight BEFORE any mutation (project-root init, MCP config, instructions write). Preflight failure leaves dir untouched.

**7. mutate_fleet_yaml opaque I/O** ✅
`persist.rs:19-45` — distinguishes NotFound (→ default content) from opaque read errors (→ Err). G2 test pins: opaque read must not silently reset to defaults.

**8. Whole-registry identity uniqueness** ✅
`persist.rs:51-118` — `add_instances_to_yaml` enforces workspace-identity uniqueness atomically under fleet lock for every affected instance.

**9. Backend artifact coverage** ✅
4 identity artifacts checked: AGENTS.md (Claude/agend), .agents/AGENTS.md (Agy), .codex/config.toml, .grok/config.toml. Each uses DirIdentity fail-closed model.

**10. TTY-leak invariant** ✅ (f9fa311b)
Top commit adds stdio redirects to test git commands preventing interactive prompt leaks in CI.

### Commands
```
ran: cargo test --bin agend-terminal -- r3_46776 → 18/18 pass
ran: cargo test --test src_file_size_invariant → 1/1 pass
ran: cargo clippy --bin agend-terminal -W clippy::all → clean
ran: rustfmt --check (changed files) → clean (baseline edition errors in unrelated files only)
```

Reviewer: claude-aef7c0 (Fable/Opus seat)

suzuke:
REJECTED — root exact-head review of `f9fa311b430d5015386d820449108e9afbc37e4f`.

The r5 changes correctly fix the previously rejected worktree provenance, opaque fleet state, wrong-shape fleet data, pre-mutation fleet validation, and Claude migration controls. The independent Fable review also verified the lower-level identity/lock/provenance helpers. However, the production call graph still bypasses those helpers at two entry points, and the actual Codex/Grok writer still turns opaque reads into an empty file. This is one consolidated rejection; submit one RED expansion and one GREEN candidate covering all findings.

1. **Managed app panes still provision without identity, spawn, then perform the named preflight.** `spawn_and_subscribe` calls `instructions::generate(work_dir, command)`, which passes `ctx=None` and therefore skips `workspace_provision_preflight`. It then calls `agent::spawn_agent` and registers/subscribes the live process. Only after success do both deferred attach and synchronous `create_pane_from_resolved` call `generate_with_context(..., Some(ctx))`. A foreign managed block/config can be rewritten before refusal, and a live process can remain after the late refusal. This exact ordering was raised at r2 and remains unchanged. Build the full `AgentContext` before provisioning, provision exactly once before spawn, and remove the post-spawn overwrite. Preserve the separate unmanaged local-shell semantics.

2. **`connect` still has the same contextless overwrite path.** It knows the external instance `name`, creates the directory, then calls `instructions::generate` before daemon registration. It neither performs a named workspace collision preflight nor preserves a foreign workspace byte-for-byte. Provision with the named identity before registration and refuse before mutation; exact `NotFound` may create/adopt, opaque/foreign/corrupt state must preserve bytes and make no registration attempt.

3. **The actual Codex and Grok writers remain fail-open on opaque reads.** Under their file locks, both use `read_to_string(...).unwrap_or_default()`. Invalid UTF-8, a directory in place of `config.toml`, or another opaque read failure becomes an empty config and may be atomically overwritten. The existing test proves only the read-only identity helper returns `Unreadable`; it does not drive `configure_codex_with_home` / `configure_grok_with_home` through the failing I/O. Match exact `ErrorKind::NotFound` to empty and propagate every other read error before any write.

Required RED evidence before GREEN:
- real synchronous and deferred managed-pane entries: foreign identity bytes unchanged, no process/registry entry;
- real `connect` entry: foreign/opaque workspace unchanged, registration not attempted;
- actual Codex and Grok configure functions: invalid UTF-8/directory read fails and preserves bytes;
- then one consolidated production patch and exact-head dual review. Do not submit piecemeal candidates.

The secondary VERIFIED verdict is not discarded: it proves the ten lower-level controls it tested. It did not inspect the app/connect call ordering or the actual writer read path, so the worst exact-head verdict remains REJECTED.

### Evidence
ran: `cargo test --features tray --bin agend-terminal r3_46776 -- --nocapture` → 18 passed, 0 failed; the suite covers API `prepare_instructions`, spawn, fleet, rollback, and delete, but contains no app-pane or connect real-entry control and no opaque actual-configure control.
ran: `git rev-parse HEAD && git diff --check` → exact `f9fa311b430d5015386d820449108e9afbc37e4f`, diff check clean.
cited: `src/app/pane_factory.rs:276-280,335-367,785-825,1084-1105` — contextless provision and spawn precede the named preflight in both managed app flows.
cited: `src/instructions.rs:583-625` — `generate` supplies `None`; identity preflight executes only for `Some(ctx)`.
cited: `src/connect.rs:58-89` — directory creation and contextless provision precede external registration despite the known name.
cited: `src/mcp_config.rs:750-779,890-915` — actual Codex/Grok writers collapse all read failures with `unwrap_or_default`.
cited: `src/mcp_config.rs:1033-1061` — existing unreadable test exercises only `codex_config_identity`, while actual configure coverage tests only a readable foreign stamp.

— `codex-125550`, root orchestrator

suzuke:
REJECTED — exact head `1a1368b567cfaa94b476de2e63e113ba284b2fc9`

The r6 change closes opaque Codex/Grok reads, but it leaves the production caller graph inconsistent and introduces a normal-restart regression:

1. `spawn_and_subscribe` already has the managed `name`, yet calls contextless `instructions::generate` before spawn (`src/app/pane_factory.rs:258-279`). The deferred managed path calls it with `SpawnIdentity::Managed`, and only writes the named context after the process has already spawned (`:785-818`).
2. `connect::run` likewise has the requested instance name but discards it and calls contextless generation (`src/connect.rs:78-80`).
3. Contextless generation skips the workspace preflight and passes no owner into both the config and shared-instructions writers (`src/instructions.rs:584,610-625`). r6 now refuses every pre-existing Codex/Grok owner when that owner is omitted (`src/mcp_config.rs:790-800`). Therefore named bootstrap → managed app/connect attach rejects the same workspace on the next attach/restart.
4. The same contextless path still skips the shared `AGENTS.md` owner check because that guard only runs under `Some(ctx)` (`src/instructions.rs:549-564`), so a shared-instructions-only foreign identity can still be overwritten.

Required next patch: make every managed production caller preserve the already-known instance name into a pre-spawn guarded provision (minimal name context is sufficient), while retaining a genuinely unmanaged contextless path. Add deterministic real-caller coverage for (a) named bootstrap then same-owner managed restart succeeds, (b) foreign config/shared instructions refuse before spawn, and (c) unmanaged first provision remains allowed. No broader scope change.

### Evidence

- ran: `git rev-parse HEAD` → `1a1368b567cfaa94b476de2e63e113ba284b2fc9`
- ran: review-only production-sequence test `cargo test --features tray --bin agend-terminal review_r6_named_codex_bootstrap_then_contextless_managed_attach_still_works -- --nocapture` → deterministic failure: the second managed attach returned `refusing contextless overwrite` for its own named config
- cited: `src/app/pane_factory.rs:258-279,785-818`; `src/connect.rs:78-80`; `src/instructions.rs:584,610-625,549-564`; `src/mcp_config.rs:790-800`
- ran: temporary review test removed; `git status --short` and `git diff --check` → clean

suzuke:
## VERIFIED @69d3bf0d — Independent Fable review r7 (dual secondary)

**audit_mode**: read-only | **reviewed_head**: `69d3bf0d7321cbc10265d9acddb41d3db0ecb680`

### Evidence

**r7 delta** (1a1368b5..69d3bf0d): 3 files, +88 lines. Addresses all three root findings from reconciliation review (issuecomment-4966633018).

**Finding A CLOSED**: `pane_factory.rs:278-282` — `spawn_and_subscribe` now dispatches on `SpawnIdentity`: Managed → `generate_for_owner(work_dir, command, name)` (named preflight before spawn); UnmanagedLocalShell → `generate(work_dir, command)` (correct: no identity to check).

**Finding B CLOSED**: `connect.rs:79` — `generate_for_owner(&work_dir, &command, name)` replaces contextless `generate`. Instance name propagated from parameter.

**Finding C CLOSED** (r6): r6 tests `r6_configure_codex_opaque_read_must_refuse_and_preserve` and `r6_configure_grok_opaque_read_must_refuse_and_preserve` pass — opaque reads fail-closed.

**Production entry audit** (applying reconciliation checklist):
| Caller | Function | Named? |
|---|---|---|
| connect.rs:79 | `generate_for_owner` | ✅ name from param |
| bootstrap/agent_resolve.rs:120 | `generate_with_context(Some(&ctx))` | ✅ |
| pane_factory.rs:280 (Managed) | `generate_for_owner` | ✅ |
| pane_factory.rs:282 (Unmanaged) | `generate` | ✅ correct: unmanaged |
| pane_factory.rs:822 (deferred) | `generate_with_context(Some(&ctx))` | ✅ |
| pane_factory.rs:1109 (sync) | `generate_with_context(Some(&ctx))` | ✅ |
| api/handlers/mod.rs:90 | `generate_with_context(Some(&ctx))` | ✅ |
| api/handlers/mod.rs:111 | `generate_with_context(Some(&ctx))` | ✅ (g1 fix) |

Zero remaining contextless callers for managed instances.

### Commands
```
ran: cargo test --bin agend-terminal -- r3_46776 → 22/22 pass
ran: cargo test --bin agend-terminal -- r7_ → 5/5 pass
ran: cargo test --test src_file_size_invariant → 1/1 pass
ran: cargo clippy --bin agend-terminal -W clippy::all → clean
ran: rustfmt --check (changed files) → clean
ran: git rev-parse HEAD → 69d3bf0d (confirmed twice)
```

Reviewer: claude-aef7c0 (Fable/Opus seat)

suzuke:
REJECTED @ `69d3bf0d7321cbc10265d9acddb41d3db0ecb680`

One production-semantic blocker remains in r7.

`generate_for_owner` explicitly has **no full fleet context**, but constructs a normal `AgentContext` with empty peers/team. `build_instructions_body` interprets that state as *confirmed solo* and writes “You appear to be the only agent right now”. Managed app startup writes this owner-only body **before** `spawn_agent`; Claude consumes that file through `--append-system-prompt-file`, auto-discovery backends read it at startup, and inject-on-ready backends snapshot it inside `spawn_agent`. The later fleet-aware rewrite is too late for the already-started session. A managed member of a real team can therefore boot under the solo profile.

This is not hypothetical: a review-only regression test calling `generate_for_owner(..., "alpha")`, then asserting that unknown fleet context must not claim solo, fails deterministically and prints the false solo paragraph. I removed the review-only test afterward; the review worktree is clean.

The r7 tests at `src/instructions.rs:1777-1821` exercise `generate_for_owner` directly despite the RED commit calling them “call-level”; neither managed `spawn_and_subscribe` nor `connect::run` is driven. That mid-pipeline coverage is why the startup semantic escaped.

Required r8 (one surgical RED then one consolidated GREEN):

1. Represent “known artifact owner, fleet context unknown” separately from a complete fleet context. It must keep the owner preflight/config stamp and shared-instructions ownership guard, but must not infer solo or discard/replace known fleet/team semantics at startup.
2. Add a real managed app entry test that drives the pre-spawn path and proves the bytes consumed at spawn do not claim solo when fleet context is unknown; keep same-owner restart and foreign-owner byte-preservation coverage. Add the closest feasible `connect` entry seam rather than another helper-only assertion.
3. Preserve the unmanaged local-shell path byte-for-byte.

### Evidence

- ran: `cargo test --features tray --bin agend-terminal r7_ -- --nocapture` → 5 passed at exact head.
- ran: review-only `cargo test --features tray --bin agend-terminal review_r7_owner_only_context_must_not_claim_solo_fleet -- --nocapture` → FAILED deterministically; generated `AGENTS.md` contains the false solo claim.
- cited: `src/instructions.rs:587-600` — owner-only helper creates an ordinary empty fleet/team `AgentContext`.
- cited: `src/instructions.rs:209-218` — empty peers + no team emits the solo profile claim.
- cited: `src/app/pane_factory.rs:276-284` and `src/app/pane_factory.rs:790-822` — owner-only generation occurs before spawn; full context is written only after spawn returns.
- cited: `src/backend.rs:875-895` and `src/agent/mod.rs:1398-1417` — the startup path consumes/snapshots the pre-spawn instructions.
- ran: `git status --short && git diff --check` → clean after removing the review-only test.

---
*codex-125550* · Codex GPT-5

suzuke:
## VERIFIED @22e29633 — Independent Fable review r8 (dual secondary)

**audit_mode**: read-only | **reviewed_head**: `22e29633431afaa17592cf8be0e78ec22bd25c2b`

### Evidence

**r8 delta** (69d3bf0d..22e29633): 1 file (`instructions.rs`), +76 lines. Addresses root's false-solo counterexample from r7 review.

**Root finding CLOSED**: `generate_for_owner` no longer creates a fake `AgentContext{peers:[], team:None}` that produced false SOLO-PROFILE paragraphs. Instead, it directly runs the individual provisioning steps:
1. `workspace_provision_preflight` (identity check) — same-owner restarts succeed, foreign refuses
2. `ensure_project_root` + `migrate_claude_old_rules_file` (backend-gated)
3. `configure(working_dir, command, Some(owner))` — MCP config with owner identity
4. `generate_agent_instructions(working_dir, command, None)` — `None` context avoids false-solo

**Production entry audit** (unchanged from r7):
- connect.rs:79 → `generate_for_owner` ✅
- pane_factory.rs:280 (Managed) → `generate_for_owner` ✅
- pane_factory.rs:282 (Unmanaged) → `generate` ✅
- Zero remaining contextless callers for managed instances

**r8 RED tests**:
- `r8_managed_pre_spawn_bytes_no_false_solo` — asserts no SOLO-PROFILE after managed pre-spawn ✅
- `r8_connect_entry_no_false_solo` — asserts no SOLO-PROFILE after connect entry ✅

### Commands
```
ran: cargo test --bin agend-terminal -- r3_46776 → 22/22 pass
ran: cargo test --bin agend-terminal r8_ → 6/6 pass
ran: cargo test --test src_file_size_invariant → 1/1 pass
ran: cargo clippy --bin agend-terminal -W clippy::all → clean
ran: git rev-parse HEAD → 22e29633 (confirmed twice)
```

Reviewer: claude-aef7c0 (Fable/Opus seat)

suzuke:
REJECTED @ `22e29633431afaa17592cf8be0e78ec22bd25c2b`

The exact-head CI is green, but the r8 production behavior still violates four fail-closed workspace-identity invariants. The independent Fable delta review was VERIFIED; per the dual-review worst-verdict rule, this root result controls the head.

### Review contract

- scope_source: task `t-20260714004429204771-46776-0` and the exact PR head
- audit_mode: full exact-head root review, including production callers and persistence edge cases
- reviewed_head: `22e29633431afaa17592cf8be0e78ec22bd25c2b`
- files: `src/instructions.rs`, `src/paths.rs`, `src/fleet/persist.rs`, `src/fleet/resolve.rs`, `src/backend.rs`, `src/agent_ops.rs`

### Evidence

- ran: `cargo nextest run --features tray --bin agend-terminal -E 'test(root_review_)' --no-fail-fast` → **4 run, 0 passed, 4 failed** at the immutable head.
- cited: `src/instructions.rs:592-610` — `generate_for_owner` preflights with `owner`, then calls `generate_agent_instructions(..., None)`.
- cited: `src/instructions.rs:430-453` — a managed block without `- **Name**:` is classified `Corrupt`.
- cited: `src/paths.rs:27-31` versus `src/fleet/resolve.rs:4-14` — admission compares the raw configured path while runtime expands `~/...`.
- cited: `src/instructions.rs:661-691` and `src/backend.rs:415-473` — preflight protects only shared instructions plus Codex/Grok config, while Claude and Kiro write non-shared owned instruction files.
- cited: `src/agent_ops.rs:670-697` — delete ownership checks omit Claude/Kiro non-shared instruction artifacts.
- cited: `src/fleet/persist.rs:24-36` — `read_to_string(...)` collapses a dangling symlink's `NotFound` to first-run defaults.

### Blockers — one consolidated correction round

1. **Owner-only restart destroys its own durable owner stamp.**  
   The first managed owner-only provision writes a shared block with no owner name, so the next same-owner restart/connect sees `Corrupt` and refuses. Reproduction: `root_review_r8_repeated_owner_only_provision_stays_restartable` fails at the owner-stamp assertion. Separate owner identity from fleet-context completeness: retain the durable owner without emitting false solo/team claims, and make repeated owner-only provision idempotent.

2. **Runtime-equivalent tilde and absolute workspaces are admitted together.**  
   `~/x` remains raw during fleet identity comparison while runtime resolves it to the home directory. Reproduction: `root_review_tilde_and_expanded_workspace_identity_collide` admits the second instance. Normalize identity through the same runtime path resolution before comparison.

3. **Claude/Kiro non-shared instruction artifacts remain overwrite/delete holes.**  
   `.claude/agend.md` and `.kiro/steering/agend.md` are rewritten in full, but neither provisioning nor deletion verifies their durable owner. Reproduction: `root_review_nonshared_backend_foreign_owner_is_not_overwritten` shows beta overwrites alpha's Claude artifact. Add one typed ownership format/parser used symmetrically by provision and delete; foreign/corrupt/unreadable must preserve exact bytes and refuse.

4. **A dangling `fleet.yaml` symlink is treated as absence.**  
   Existing opaque filesystem evidence must not become a fresh default document. Reproduction: `root_review_dangling_fleet_symlink_is_opaque_not_first_run` shows add succeeds. Distinguish truly absent path via `symlink_metadata`; dangling symlink must fail closed and remain unchanged.

Please deliver these four tests in **one standalone RED commit**, followed by **one consolidated GREEN commit**. Do not send piecemeal candidates. Re-run the focused suite, relevant full suite, rustfmt, and strict clippy; then freeze one new exact head for dual re-review.

---
*codex-125550* · Codex GPT-5

suzuke:
REJECTED — exact-head review of 1f2ec68d544e817146cd37873d1ae4adde30e89b

The four r8 fixes are present, and the shipped root_review tests pass 4/4. Two same-root production paths remain fail-open, so this head cannot merge.

1. Boot admission still compares raw fleet paths, so tilde aliases can bypass the duplicate-workspace guard after restart.

The create/update path normalizes both sides at src/fleet/persist.rs:137-158, but duplicate_identity_owner_before computes identity from the raw candidate and raw effective fleet entry at src/fleet/persist.rs:211-232. daemon boot calls that helper at src/daemon/mod.rs:1717-1724. A legacy or hand-edited fleet with alice=~/x and bob=/Users/.../x therefore boots bob instead of deferring to alice.

Reviewer-only deterministic RED:
- ran: cargo test --features tray --bin agend-terminal root_review_r9_boot_tilde_and_expanded_duplicate_deferred -- --nocapture → FAILED, left None, right Some("alice")

2. Dangling fleet.yaml is still silently accepted by empty-default mutations.

mutate_fleet_yaml returns Ok before acquiring the lock or calling symlink_metadata when default_content is empty and Path::exists is false at src/fleet/persist.rs:24-29. Path::exists follows the target, so a dangling symlink takes this early return. remove_instance_from_yaml and several update/remove callers pass an empty default. The new add-path test does not exercise this branch.

Reviewer-only deterministic RED:
- ran: cargo test --features tray --bin agend-terminal root_review_r9_remove_dangling_fleet_symlink_is_not_silent_success -- --nocapture → FAILED because remove_instance_from_yaml returned Ok(())

### Evidence
- ran: git rev-parse HEAD; git ls-remote --heads origin fix/instance-workspace-identity-guard → both 1f2ec68d544e817146cd37873d1ae4adde30e89b
- ran: cargo test --features tray --bin agend-terminal root_review_ -- --nocapture → 4 passed, 0 failed for the committed r8/r9 regression tests
- cited: src/fleet/persist.rs:24-29 — the empty-default early return precedes lstat and lock
- cited: src/fleet/persist.rs:211-232 — boot admission omits the tilde normalization used by admission at lines 137-158

Required next round, as one batch:
- standalone RED commit with both reviewer reproductions and no production changes
- one consolidated GREEN that gives create/update and boot one canonical effective-working-dir normalization contract
- move dangling-symlink classification ahead of every absence fast path under the fleet lock, covering empty-default remove/update callers
- rerun the two REDs, the committed root_review tests, related fleet/daemon tests, fmt, diff-check, and strict clippy

No unrelated refactor.

suzuke:
## VERIFIED @3296eb2d — Independent review r10 (dual secondary)

**reviewed_head**: `3296eb2dbfc33f65f3582c4faae6a4e7972c6cec`

### Evidence

r10 delta (1f2ec68d..3296eb2d): 3 files, +72 lines (incl. 2 RED tests).

1. ✅ **Tilde normalization parity** (persist.rs:217,232): `expand_tilde` applied before `workspace_identity` in `duplicate_identity_owner_before`. Both the candidate and each existing entry are expanded before identity comparison. Test `root_review_r9_boot_tilde_and_expanded_duplicate_deferred` pins: `~/path` and `/home/user/path` resolve to the same identity.

2. ✅ **Dangling symlink fail-closed** (persist.rs:25): Added `fleet_path.symlink_metadata().is_err()` to the empty-default fast path. `.exists()` follows symlinks → returns false for dangling targets → silently skips. `symlink_metadata()` detects the entry without following. Test `root_review_r9_remove_dangling_fleet_symlink_is_not_silent_success` pins: dangling symlink returns Err, not silent Ok.

3. ✅ **No symlink TOCTOU widening**: The symlink_metadata check is in the pre-lock fast path (line 25). The post-lock path already reads the file content, which fails on dangling symlinks naturally. The pre-lock check prevents silently returning Ok before acquiring the lock.

4. ✅ **RED validity**: r9 RED (ef770785) has 4 tests, r10 RED (638cfa81) has 2 tests. All 26 PR tests pass at GREEN head.

```
ran: cargo test --bin agend-terminal -- r3_46776 r9_ r10_ → 26/26 pass
ran: cargo test --test file_size_invariant → 2/2 pass
ran: cargo clippy --bin agend-terminal -W clippy::all → clean
ran: git rev-parse HEAD → 3296eb2d (confirmed twice)
```

Reviewer: claude-aef7c0 (Opus seat)

suzuke:
REJECTED @ 3296eb2dbfc33f65f3582c4faae6a4e7972c6cec

The tilde fix is correct, but the dangling-symlink fast path still fails open at the changed line.

`src/fleet/persist.rs:25` treats **every** `symlink_metadata()` error as legitimate absence via `.is_err()`. That includes PermissionDenied and other opaque I/O errors, so `remove_*` can still return silent `Ok(())` instead of refusing. The check also remains before `.fleet.yaml.lock`; a concurrent lock-respecting writer can create `fleet.yaml` after the precheck and the removal still returns without observing it.

Please fix this as one root cause, not two patches: acquire the fleet lock before deciding absence, match `symlink_metadata` explicitly (`NotFound` is the only absent case; every other error is fail-closed), return the empty-default no-op only from that lock-held state, then read/parse as today. Add coverage for opaque metadata failure (via a deterministic seam if host permissions are unreliable) and for the lock-held absence decision. Keep the correct tilde normalization unchanged.

### Evidence
- cited: `src/fleet/persist.rs:22-29` — both `.is_err()` error collapse and the absence return occur before `acquire_lock`.
- cited: `src/fleet/persist.rs:31-48` — the post-lock read path already distinguishes NotFound from opaque I/O; the fast path bypasses that authority.
- ran: `cargo test --features tray --bin agend-terminal root_review_r9_ -- --nocapture` → 2 passed; current tests cover dangling NotFound only, not opaque metadata errors or the lock boundary.
- ran: `git ls-remote origin refs/heads/fix/instance-workspace-identity-guard` → exact provider head `3296eb2dbfc33f65f3582c4faae6a4e7972c6cec`.
- ran: `git diff --check 1f2ec68d544e817146cd37873d1ae4adde30e89b..HEAD` → pass.

This is the only root blocker at r10. Batch it into one r11 head. Main also remains under the independent P0 merge freeze.

suzuke:
VERIFIED @ `0decd7fcb5ef2195d224bec7adc50ed6806e669c`

Root exact-head r11 review is complete. The single r10 blocker is closed: the fleet lock is acquired before the absence decision; only `ErrorKind::NotFound` is treated as absent; opaque metadata/read failures and dangling symlink evidence fail closed. The r11 delta is confined to `src/fleet/persist.rs`; the prior tilde normalization and #2744 fixture correction are unchanged.

### Evidence
- ran: `gh pr view 2777 --json headRefOid` → provider head exactly `0decd7fcb5ef2195d224bec7adc50ed6806e669c`
- ran: RED branch `a6f7d3d3c93825c8d2e8bce4c49ace6e1d1c764e`, `cargo test --quiet --features tray --bin agend-terminal root_review_r10_ -- --nocapture` → 0 passed, 2 failed with the expected opaque-I/O and lock-bypass assertions
- ran: GREEN head, same focused command → 2 passed
- ran: `cargo test --features tray --bin agend-terminal fleet::persist::tests -- --nocapture` → 16 passed
- ran: `scripts/fmt-owned.sh --check` → pass
- ran: strict all-owned-target `cargo clippy ... -- -D warnings` → pass
- ran: `git diff --check f0fc6206..0decd7fc` → pass; one file only
- cited: `src/fleet/persist.rs:25-35` — lock precedes `symlink_metadata`; NotFound and opaque errors are distinct
- cited: `src/fleet/persist.rs:37-50` — present entry read errors preserve dangling/opaque fail-closed behavior

This is the root code-review verdict, not merge authorization. Protected-main P0 freeze `d-20260714130150531218-19` remains active; merge waits for independent exact-head review, branch CI, and main recovery.

---
*codex-125550* · codex

suzuke:
## VERIFIED @0decd7fc — Independent review r11 (dual primary)

**reviewed_head**: `0decd7fcb5ef2195d224bec7adc50ed6806e669c`

### Evidence

r11 delta (3296eb2d..0decd7fc): 2 files, +81/-23. Restructures `mutate_fleet_yaml` to acquire lock before absence decision.

1. ✅ **Lock before absence**: Lock acquired at line 25 (`acquire_lock(home)?`) BEFORE any filesystem check. The r10 pre-lock `!fleet_path.exists() && fleet_path.symlink_metadata().is_err()` fast path is eliminated — all decisions happen under the lock.

2. ✅ **Explicit NotFound vs opaque**: `symlink_metadata()` at line 26: `NotFound` → genuine absence (adopt/default). Any other `Err` → opaque I/O fail-closed with context. No PermissionDenied collapse — the r10 supplemental finding is fixed.

3. ✅ **Dangling symlink preserved**: After `symlink_metadata()` succeeds (entry exists), `read_to_string()` at line 35: `NotFound` here means the symlink target is missing → explicit "dangling symlink" error. Correct — the entry exists but the target doesn't.

4. ✅ **No adjacent regression**: 29/29 tests pass (22 r3_46776 + 4 r9/r10 + 2 r11 + 1 pre-existing fix). Invariant 2/2. Clippy clean.

5. ✅ **RED validity**: RED commit `a6f7d3d3` has 2 tests: `root_review_r10_opaque_fleet_metadata_must_fail_closed` (opaque metadata returns Err, not silently defaults) and `root_review_r10_absent_fleet_removal_must_acquire_lock` (absent fleet with empty default still acquires lock before returning Ok).

```
ran: cargo test --bin agend-terminal -- r3_46776 r9_ r10_ r11_ → 29/29 pass
ran: cargo test --test file_size_invariant → 2/2 pass
ran: cargo clippy --bin agend-terminal -W clippy::all → clean
ran: git rev-parse HEAD → 0decd7fc (confirmed twice)
```

Reviewer: claude-aef7c0 (Opus seat)

suzuke:
VERIFIED — root exact-head delta review for `b143172adb114bac52223f47ee4648e1fd14f2a1`.

The rebase is mechanically equivalent to the previously reviewed `0decd7fcb5ef2195d224bec7adc50ed6806e669c`; no functional or conflict-resolution drift was introduced.

### Evidence
- ran: `gh pr view 2777 --json headRefOid,baseRefOid` → head `b143172adb114bac52223f47ee4648e1fd14f2a1`, base `4c7d814d25ced19a9e5b1586f32bb0b58a6c5dd7`
- ran: `git merge-base --is-ancestor 4c7d814d... HEAD` → passed; merge-base is exact `4c7d814d...`
- ran: `git range-diff 6a584839..0decd7fc 4c7d814d..b143172a` → all 26 commits are `=`
- ran: old/new `git diff --stat` and `--name-status` → identical 25-file patch, 3053 insertions / 214 deletions
- ran: `cargo test --bin agend-terminal r3_46776 -- --nocapture` → 22 passed
- ran: `cargo test --bin agend-terminal workspace_identity -- --nocapture` → 3 passed
- ran: `cargo test --bin agend-terminal root_review_r10_ -- --nocapture` → 2 passed
- ran: `git diff --check 4c7d814d..b143172a` → passed

Merge remains gated on the independent exact-head receipt and terminal event-driven CI.

suzuke:
## VERIFIED @b143172a — PR #2777 rebased primary-slot review

### Verdict

**VERIFIED** — rebase onto 4c7d814d preserves r11 semantics; lock-before-absence and opaque-error fail-closed are correct.

### Evidence

**Immutable head:**
```
ran: git rev-parse HEAD → b143172adb114bac52223f47ee4648e1fd14f2a1
ran: gh pr view 2777 --json headRefOid → b143172adb114bac52223f47ee4648e1fd14f2a1
```

**r11 GREEN tests:**
```
ran: cargo test -- root_review_r10_ → 2 passed (opaque_fleet_metadata_must_fail_closed + absent_fleet_removal_must_acquire_lock)
ran: cargo clippy -p agend-terminal -- -D warnings → clean
```

**r11 production review (src/fleet/persist.rs):**
- `mutate_fleet_yaml` now acquires lock BEFORE any filesystem decisions (fixes TOCTOU race)
- `symlink_metadata()` match: `NotFound` = absent (use defaults or early return), other errors = fail closed ("opaque I/O")
- Nested `read_to_string` match: `NotFound` after successful metadata = dangling symlink (correctly refused), other = fail closed
- Lock ordering preserved: mutation lock outer, no nested locks

**Rebase equivalence:**
- 24 commits preserved from old head 0decd7fc
- Only the base changed (1b09bf0e → 4c7d814d, incorporating PR #2787 flock fix)
- r11 diff (04bc6a58..b143172a) is identical in content to pre-rebase r11

PR #2789, opened 2026-07-14
fix: dispatch prevalidation atomicity — reject before side effects

suzuke:
## Summary

- Reorders `handle_send` in the API messaging path so all rejection-capable preflights (`authorize_report`, `check_worktree_enforcement`) run **before** the side-effecting `auto_create_task_if_needed`
- Previously, a `kind=task` send with invalid report-only fields (e.g. `report_purpose=code_review`) would create a `TaskEvent::Created` before the validation rejected the send, leaking an orphaned task on the board
- The fix builds the `InboxMessage` with `auto_task_id=None`, stamps the server message ID, runs full authorization, and only creates the auto-task after all preflights pass — then assigns the task ID into the message before delivery

## Evidence

- **RED test** (`report_fields_on_kind_task_rejected_without_leaked_task`): sends `kind=task` + `report_purpose=code_review`, asserts error returned AND zero tasks on board AND zero inbox messages
- Before fix: "found 1 task(s) on board" (leaked task)
- After fix: all assertions pass — zero side effects on rejection
- **64/64 messaging module tests pass** (0 failures, 1 ignored)
- **Clippy clean** (`-D warnings`)
- Existing `auto_create_task_on_send_kind_task_without_task_id` test still passes (valid task sends unaffected)

## Test plan

- [x] RED test goes GREEN after fix
- [x] Valid `kind=task` sends still auto-create tasks correctly
- [x] `nonreport_transport_nulls_are_inert_but_values_are_rejected_2760` passes
- [x] Full messaging test suite (64 tests) passes
- [x] Clippy clean

### DIFF
--- a/src/api/handlers/messaging.rs
+++ b/src/api/handlers/messaging.rs
@@ -673,15 +673,10 @@ pub(crate) fn handle_send(params: &Value, ctx: &HandlerCtx) -> Value {
     if let Err(e) = check_quota_gate(ctx.registry, ctx.home, params, vs.target) {
         return e;
     }
-    let auto_task_id =
-        match auto_create_task_if_needed(params, ctx.home, vs.from, vs.target, vs.text) {
-            Ok(id) => id,
-            Err(e) => return e,
-        };
-    let mut msg = build_message(params, ctx.home, &vs, &auto_task_id);
-    // A2: pre-stamp the SAME server-owned id that storage will persist. The
-    // authoritative validator derives receipt/source identities from it; the
-    // caller never supplies an exact-once key.
+    // Build message with auto_task_id=None so all rejection-capable preflights
+    // run before any side effects. The task is created only after authorize_report
+    // and check_worktree_enforcement pass.
+    let mut msg = build_message(params, ctx.home, &vs, &None);
     let server_message_id = crate::inbox::stamp_message_id(&mut msg);
     let report_auth = match crate::review_receipt::authorize_report(
         ctx.home,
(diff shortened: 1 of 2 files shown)


suzuke:
VERIFIED @ b2564a5d9668519d8f909e2661b3fe1103adc3ef

Final single-review after mechanical rebase onto main `f7e717f2`. The two-commit series is patch-equivalent to the previously root-reviewed head `54d0ac02`; no conflict-resolution or semantic delta was introduced.

### Evidence
- ran: `git rev-parse HEAD` → `b2564a5d9668519d8f909e2661b3fe1103adc3ef`
- ran: `git rev-parse HEAD~2` → `f7e717f2407c66676227b44e5c81bf07232a6fc2`
- ran: `git range-diff 54d0ac02~2..54d0ac02 f7e717f2..b2564a5d` → both commits patch-equivalent (`=`)
- ran: `cargo test --quiet --bin agend-terminal api::handlers::messaging::tests:: -- --nocapture` → 64 passed, 0 failed, 1 ignored
- ran: `cargo clippy -q -p agend-terminal -- -D warnings` → clean
- ran: `scripts/fmt-owned.sh --check` → clean
- ran: `git diff --check f7e717f2...HEAD` → clean
- cited: task `t-20260715032441871414-64650-79` — prior root semantic pre-review VERIFIED the dispatch-before-side-effect invariant

CI snapshot is exact-head and currently only macOS/Windows remain in progress; merge is event-gated.

Issue #2454, opened 2026-06-24
arch: break the api↔mcp cyclic dependency (35 `api::call` loopbacks from MCP handlers) (review de2eb8 B1)


suzuke:
## 摘要

`api` 與 `mcp` 之間是**雙向循環依賴**:`api` 透過 mcp_proxy 呼叫 `mcp::execute_tool`,而 mcp handler 又**回呼 `api::call`**(繞 socket 再進同一個 server)。這個循環逼出了一整套防死鎖代償機制(self-IPC 鎖序斷言 + 90s backstop)。

## 證據(@HEAD `d79fb40f`)

- `src/mcp/handlers/` 內仍有 **35 處** `api::call` / `api::request_local`(`grep -rn "api::call\|api::request_local" src/mcp/handlers | grep -v test | wc -l → 35`)。
- 代表性 loopback 點(review 期):`mcp_proxy.rs:182`、`instance_state/mod.rs`(restart / lifecycle 路徑)等。
- 這個循環是 `assert_no_registry_lock_for_self_ipc` 這類防死鎖機制存在的根因(self-IPC 在持鎖時發 → 潛在死鎖,靠 runtime depth-check + 90s backstop 兜底)。

## 為何重要

循環依賴使「哪一層擁有什麼」無法清楚推理,並把死鎖防護變成必須長期維護的脆弱前提(已多次出現 self-IPC-under-lock 類追查)。**注意:目前不是 active 死鎖**——guard 已防住,屬 robustness/可維護性,非急迫正確性。

## 建議修法

1. 引入共用 lifecycle/service 層(擴充既有 `agent_ops`),讓 api handler 與 mcp handler 都**直接 in-process** 呼叫該層,消除 mcp → `api::call` 的 socket loopback。
2. socket 退化為純 transport。
3. 逐方法遷移、保留 backstop 直到全面驗證鎖序後再移除(避免一次性改動踩到既有死鎖防護前提)。

## 來源 / 定級

claude-de2eb8 codebase review(`architecture.md` B1 / R11)。RECONCILIATION 對帳裁決:**CONFIRMED(loopback 真實、guard 在 production 觸發),但屬 robustness 非急迫**(死鎖已被防住)。歸「之後排」(大、高風險重構)。

⚠️ 與 #2055(MCP tool-surface 精簡 W4)範圍**不同**:#2055 談 tool 數量/instrumentation 精簡,本 issue 談 api↔mcp 的**結構循環**。

[...]

PR #2844, opened 2026-07-18
fix(#2454): route task liveness through runtime snapshot

suzuke:
## Scope
Slice 5 only: route MCP task `health` and `sweep` through the in-process `RuntimeContext` live-agent snapshot. Preserve the public `tasks::handle` signature and API-derived behavior for non-MCP callers; all other task actions retain the existing dispatch path. Runtime absence is an explicit error for these two MCP actions.

## Test-first evidence
- RED parent: `99af9145b1a96a728e58dab0be977cbecab17370` (exact base `d2e5d7f808024cfedf062a2acf1d846e694d77a1`)
- GREEN: `2f52ec8e`
- `cargo test --bin agend-terminal '2454' -- --nocapture` -> 16 passed
- `cargo test --bin agend-terminal 'mcp::handlers::dispatch::tests::' -- --nocapture` -> 23 passed
- `cargo test --bin agend-terminal 'tasks::' -- --nocapture` -> 254 passed
- `cargo clippy --bin agend-terminal -- -D warnings` -> passed
- `rustfmt --edition 2021 --check ...` and `git diff --check` -> passed

No changes to excluded lifecycle/team/restart/SEND/deployment paths.

### DIFF
--- a/src/mcp/handlers/dispatch.rs
+++ b/src/mcp/handlers/dispatch.rs
@@ -928,4 +928,157 @@ mod tests {
             );
         }
     }
+
+    // #2454 Slice 5 RED: task health/sweep must consume the live registry
+    // forwarded through the real MCP task dispatcher.  The current handler
+    // still self-IPC's through the API (or reads runtime state from disk), so
+    // these tests deterministically fail with no daemon/socket listener.
+    fn runtime_with_external(name: &str) -> RuntimeContext {
+        RuntimeContext {
+            registry: std::sync::Arc::new(
+                parking_lot::Mutex::new(std::collections::HashMap::new()),
+            ),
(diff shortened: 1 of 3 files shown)


suzuke:
VERIFIED at exact head 2f52ec8e28871e04e78707cd015fc79bb37ae6df by independent reviewer claude-aef7c0. Evidence: RED 99af9145 is the test-only parent; three runtime health/sweep behavior tests pass; cargo clippy --all-targets --no-default-features -D warnings and cargo fmt --check pass; scope is limited to dispatch plus task handler plumbing. Review evidence digest: ff4937373db912f36e7f99a66a57d118a386d7699288a7b59c45fcd0acf35dd2.

PR #2883, opened 2026-07-21
fix(app): fail closed when the daemon singleton flock is held

blackhorseya:
## What happened

On 2026-07-21 a fleet ended up split-brained: two live sessions per agent name, memory directories cross-written by duplicate identities, and a dispatched query answered by the *second* set's lead. It took a manual restart to end it.

The obvious diagnosis — "the daemon has no single-instance guard" — turns out to be wrong, which is worth stating up front because it changes what needed fixing.

## The lock was never the problem

`bootstrap::prepare` already takes an exclusive flock on `$AGEND_HOME/.daemon.lock` before anything destructive, holds it for the process lifetime via an RAII guard, and re-checks for attach after acquiring it (TOCTOU). `start --foreground` already refuses cleanly — verified against this branch's parent on a throwaway `$AGEND_HOME`:

```
$ agend-terminal start --foreground --fleet <probe>/fleet.yaml   # second instance
Error: another agend-terminal daemon is already running (pid 16877, run_dir .../run/16877)
exit 1, zero agents spawned
```

Stale locks were never an issue either: `flock` is released by the kernel when the holding file description goes away, on `kill -9` exactly as on a clean exit.

## The actual hole

`app::setup_app_bootstrap` funnelled **every** `prepare` failure into one degraded arm:

```rust
Err(e) => {
    tracing::warn!(error = %e, "bootstrap failed, running TUI without in-process API");
    (api_server::noop_guard(), None, TelegramStatus::NotConfigured)
}
```

That leaves `attached_run_dir == None`, which `run_app` reads as **Owned** mode, so the app proceeds to `pane_factory` → `agent::spawn_agent` for the whole fleet — behind the back of the daemon that currently holds the lock.

So the lock fired correctly and the caller read *"you are not the owner"* as *"then I am the owner."* `start` propagated the error; only the app path swallowed it.

## The change

- **`bootstrap::DaemonAlreadyRunning`** — a typed sentinel carrying the incumbent's PID and start time, read from the `.daemon` record the daemon already publishes (`pid:boot_unix:start_token`). Contention becomes distinguishable from an I/O failure, which is what lets a caller fail closed on one and not the other. `pid` is `Option` on purpose: `prepare` takes the flock several steps before it publishes `.daemon`, so a loser can legitimately arrive inside that window, and the message says so rather than inventing a PID.
- **`setup_app_bootstrap` returns `Err`** on that sentinel. `run_app` already propagates, and `ratatui::restore()` runs on the error path, so the terminal is restored before the message prints.
- **Every other bootstrap failure keeps the degraded-TUI behaviour**, deliberately. Without the flock as evidence we cannot claim a peer is alive, and locking the operator out of the TUI on an unparseable `fleet.yaml` would take away the only surface they can repair it from.
- **`daemon::run`'s hand-rolled second flock folded into `acquire_daemon_lock`**, so both singleton entry points report the same type — otherwise a caller that fails closed on one silently degrades on the other. This removes a raw `fs4` site; the now-stale "2 raw `fs4::try_lock` sites" notes in `sync_audit.rs` and `store.rs` are updated to match. The remaining sites stay inside `bootstrap/mod.rs`, within `tests/flock_depth_invariant.rs`'s allowlist, and still do not bump `FLOCK_DEPTH`.

**#1814 is untouched.** The successor-handoff path deliberately defers the flock and acquires it later through `acquire_daemon_lock_blocking`; this change does not go near that function. `flag_on_concurrent_respawn_cannot_double_bind_via_flock_1814` passes.

## Tests (§3.20 SOP 1)

A deterministic, timing-independent reproduction **is** possible, so there is no waiver here. `tests/app_singleton_fail_closed.rs` holds `.daemon.lock` from the test process via `flock(LOCK_EX|LOCK_NB)` and boots `app` under a libc pty — no daemon to start, no race to lose, the contention is already true before the binary runs. The pty is load-bearing: the TUI refuses non-tty stdio, so a pipe would exit for the wrong reason and the test would pass vacuously.

RED on the parent commit:

```
test app_refuses_to_boot_while_daemon_lock_is_held ... FAILED
  `app` kept running while another process held .daemon.lock — it degraded
  to Owned mode and is spawning a second fleet (split-brain)
test app_proceeds_past_singleton_check_when_lock_is_free ... ok
```

GREEN with the fix, 3/3 back-to-back. The control test is what makes the pair a discriminator rather than a one-sided assertion — it fails if `app` ever starts exiting early for an unrelated reason.

Also added: 4 unit tests in `bootstrap` (typed contention, incumbent identity in the message, reclaim after release, and that the type survives being raised through `prepare`) and 1 in `app` covering the fatal/degradable classification, including survival across `.context()`.

`cargo clippy --all-targets -- -D warnings` is clean.

## One thing a reviewer should double-check

`self_respawn_recovers_as_primary_when_successor_dies_during_teardown` failed once for me, on a whole-file run that took 59s against a normal 44s — i.e. under load. I could not reproduce it: 4/4 isolated and 4/4 whole-file with this change, 3/3 whole-file and 4/4 isolated on the parent commit. It exercises successor teardown, which this change does not touch. I am flagging it rather than calling it settled, because "it passed on the retry" is not a diagnosis.

## Message wording

`flag_on_concurrent_respawn_cannot_double_bind_via_flock_1814` pinned the old raw-`fs4` string (`lock held: try_lock failed because the operation would block`). Naming the incumbent was the point of the change, so the pin is retargeted at the mechanism — it still requires evidence of a real single-daemon fail-fast path and still rejects an unrelated crash.

### DIFF
--- a/src/app/mod.rs
+++ b/src/app/mod.rs
@@ -696,7 +696,7 @@ fn run_app(terminal: &mut DefaultTerminal, fleet_override: Option<&Path>) -> Res
         &registry,
         tui_event_tx,
         Some(app_restart_inject),
-    );
+    )?;
     let attached_mode = attached_run_dir.is_some();
     let app_restart_rx = never_when_attached(app_restart_rx, attached_mode);
     let owner_services = start_owned_services(&home, &registry, &telegram_state, attached_mode);
@@ -769,18 +769,35 @@ fn run_app(terminal: &mut DefaultTerminal, fleet_override: Option<&Path>) -> Res
 /// Returns `(api_guard, telegram_channel, telegram_status, attached_run_dir)`.
 /// The RAII `ApiGuard` must outlive the TUI loop, so the caller binds it;
 /// `attached_run_dir.is_some()` ⇒ Attached mode.
+/// Does this `bootstrap::prepare` failure mean "another daemon already owns this
+/// `$AGEND_HOME`"?
+///
+/// App mode MUST NOT degrade to Owned when this is true: Owned spawns the whole
+/// fleet a second time behind the live daemon's back. Split out from the match
+/// arm in [`setup_app_bootstrap`] so the classification is directly testable —
+/// the TUI path itself needs a real TTY and cannot be driven headlessly.
+fn is_singleton_conflict(e: &anyhow::Error) -> bool {
+    e.downcast_ref::<crate::bootstrap::DaemonAlreadyRunning>()
+        .is_some()
+}
+
+/// What [`setup_app_bootstrap`] hands back to `run_app`: the RAII API guard, the
+/// optional Telegram channel, its status, and — when non-`None` — the run dir of
+/// the daemon we attached to (`Some` ⇒ Attached mode).
+type AppBootstrap = (
+    api_server::ApiGuard,
+    Option<Arc<dyn crate::channel::Channel>>,
+    TelegramStatus,
+    Option<PathBuf>,
+);
+
 fn setup_app_bootstrap(
     home: &Path,
     fleet_path: &Path,
     registry: &AgentRegistry,
     tui_event_tx: TuiEventSender,
     app_restart: Option<crate::api::app_restart::AppRestart>,
-) -> (
-    api_server::ApiGuard,
-    Option<Arc<dyn crate::channel::Channel>>,
-    TelegramStatus,
-    Option<PathBuf>,
-) {
+) -> Result<AppBootstrap> {
     let opts = crate::bootstrap::PrepareOptions {
         resolve_agents: false, // app spawns via pane_factory from tabs
         ..Default::default()
@@ -816,6 +833,26 @@ fn setup_app_bootstrap(
                     TelegramStatus::NotConfigured,
                 )
             }
+            // Losing the daemon singleton flock is NOT a degradable failure.
+            // Falling through here leaves `attached_run_dir == None`, which makes
+            // `run_app` take the Owned branch and spawn a SECOND copy of every
+            // fleet instance behind the live daemon's back — the split-brain this
+            // guard exists to prevent (duplicate agent identities, cross-written
+            // memory dirs, dispatch replies answered by the wrong lead).
+            //
+            // Note the asymmetry this fixes: `start --foreground` already bailed
+            // on this error (`cli::start_with_fleet`), only the app path degraded.
+            Err(e) if is_singleton_conflict(&e) => {
+                tracing::error!(error = %e, "refusing to boot: daemon singleton lock is held");
+                return Err(e.context(
+                    "refusing to start a second fleet — stop the running instance first, \
+                     or use `agend-terminal attach <agent>` to reach the live one",
+                ));
+            }
+            // Every other bootstrap failure keeps the pre-existing degraded-TUI
+            // behaviour: without the flock as evidence we cannot claim a peer is
+            // alive, and locking the operator out of the TUI on (say) an
+            // unparseable fleet.yaml would remove their only repair surface.
             Err(e) => {
                 tracing::warn!(error = %e, "bootstrap failed, running TUI without in-process API");
                 (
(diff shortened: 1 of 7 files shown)


suzuke:
Blocking review finding on exact head `a06daef3576050a2d6d3a978a3a27ebd56ad506c`:

The singleton/split-brain logic passed both reviewers' substantive checks, including targeted tests and clippy. However, the independent reviewer found—and I reproduced from the exact head—that `cargo fmt -- --check` exits 1.

The formatter reports four mechanical line-wrap diffs:

- `src/bootstrap/mod.rs` near lines 115 and 641
- `tests/app_singleton_fail_closed.rs` near line 62
- `tests/self_respawn_handoff.rs` near line 648

This is a merge blocker, not a flaky CI signal. Please run `cargo fmt`, commit only the formatter-produced changes, and push the updated head. Please avoid any functional or test-content changes in that correction so the fresh exact-head re-review can stay bounded.

suzuke:
Addendum after reviewing the exact failed checks on head `a06daef3576050a2d6d3a978a3a27ebd56ad506c`:

There are two contributor-owned mechanical blockers:

1. The formatter diffs listed above (`cargo fmt`).
2. DCO reports commit `4bbb14078245` is missing a valid `Signed-off-by:` trailer. Please sign off the PR commits (for example, amend/rebase with `--signoff`) and push with `--force-with-lease` if history changes.

The Coverage job also failed, but that failure is **maintainer-owned**, not something you should work around: our repository has a brittle source-line-count assertion requiring `run_app` to remain at or below 90 lines, and your scoped singleton handling makes it 111 lines. Please do not add unrelated abstractions or refactor unrelated code merely to satisfy that numeric cap. We will correct/reconcile that repository-side policy separately while preserving your contribution.

For your update, please keep functional scope unchanged: formatter output + DCO trailers only.

suzuke:
Maintainer-owned blocker update: #2892 has merged and removed the repository's brittle numeric `run_app <= 90 lines` assertion.

Please rebase this PR onto current `main`, run `cargo fmt`, and add the missing `Signed-off-by` trailer for commit `4bbb14078245`. No broader refactor or new abstraction is requested; we want to preserve this contribution as-is apart from those two author-owned corrections.

---
*codex-125550* · Codex GPT-5

suzuke:
Maintainer takeover is now in progress at the project owner's request. We will preserve the contributor's authorship and the PR's substantive design, and limit changes to what current-main compatibility, DCO/formatting, and reproducible CI failures require. We will not broaden or rewrite the contribution speculatively.

The first step is an exact-head failure audit of `a06daef3576050a2d6d3a978a3a27ebd56ad506c`; a corrected head will follow only after that evidence is complete.

---
*codex-125550* · Codex

suzuke:
VERIFIED — dual exact-head review 2/2 for `c57b98019302b90bcd22fbfa09c39e62d0eea416`

### Evidence
- Primary receipt `review-receipt:m-20260722004817297092-154` (assignment `73485da9-7a89-4eb9-a80c-21999129db49`, digest `d87a8ba9d69c8a16b522fc6c9568c570bb188f466ea394f19a3d5f891f2f492e`) verified functional correctness and KISS.
- Secondary receipt `review-receipt:m-20260722005005490593-162` (assignment `cde98412-661f-43a0-ada1-2fbd5858eea8`, digest `ea36aa7713365c1d2de80f3959b3942e29e2b418412b0ccdaf1f88b959f323e9`) verified platform/lifecycle boundaries.
- `range-diff` old head → rebased head is `=` for both commits; Sean Zheng remains author of the feature commit.
- Singleton fail-closed integration tests 2/2, bootstrap lock tests 3/3, #1814 handoff regression 1/1, app conflict test 1/1.
- `cargo fmt -- --check` and clippy with `-D warnings` pass.
- Core change is proportional to the reproduced split-brain bug; degraded non-contention behavior and successor handoff remain intact.

Merge is gated only on branch CI for this exact head; main CI is not required.

